PR #8, opened 2015-06-21
validate root file against cert store

endophage:
As part of bootstrapping a TUF client, validate the downloaded root.json against our cert store. Per conversation with @NathanMcCauley the logic is as follows:

1. Parse root.json
2. Read Key IDs associated with the root role.
3. Read keys (from the root.json) associated with the root role key IDs.
4. Validate keys from root.json against local cert pool.
5. Using only keys that validated against cert pool, validate signatures on the root.json, requiring a threshold of 1.

### DIFF
--- a/cmd/notary/keys.go
+++ b/cmd/notary/keys.go
@@ -66,7 +66,7 @@ func keysRemove(cmd *cobra.Command, args []string) {
 	//TODO (diogo): Validate Global Unique Name. We probably want to reject 1 char GUNs.
 	gunOrID := args[0]
 
-	// Try to retreive the ID from the CA store.
+	// Try to retrieve the ID from the CA store.
 	cert, err := caStore.GetCertificateBykID(gunOrID)
 	if err == nil {
 		fmt.Printf("Removing: ")
@@ -74,6 +74,20 @@ func keysRemove(cmd *cobra.Command, args []string) {
 
 		// If the ID is found, remove it.
 		err = caStore.RemoveCert(cert)
+		if err != nil {
+			fatalf("failed to remove certificate from KeyStore")
+		}
+		return
+	}
+
+	// Try to retrieve the ID from the Certificate store.
+	cert, err = certificateStore.GetCertificateBykID(gunOrID)
+	if err == nil {
+		fmt.Printf("Removing: ")
+		printCert(cert)
+
+		// If the ID is found, remove it.
+		err = certificateStore.RemoveCert(cert)
 		if err != nil {
 			fatalf("failed to remove certificate from KeyStore")
 		}
@@ -140,10 +154,16 @@ func keysTrust(cmd *cobra.Command, args []string) {
 		fatalf("aborting action.")
 	}
 
-	err = caStore.AddCert(cert)
+	err = nil
+	if cert.IsCA {
+		err = caStore.AddCert(cert)
+	} else {
+		err = certificateStore.AddCert(cert)
+	}
 	if err != nil {
 		fatalf("error adding certificate from file: %v", err)
 	}
+
 	fmt.Printf("Adding: ")
 	printCert(cert)
 
@@ -155,9 +175,16 @@ func keysList(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
-	fmt.Println("# Trusted Certificates:")
+	fmt.Println("# Trusted CAs:")
 	trustedCAs := caStore.GetCertificates()
 	for _, c := range trustedCAs {
+		printCert(c)
+	}
+
+	fmt.Println("")
+	fmt.Println("# Trusted Certificates:")
+	trustedCerts := certificateStore.GetCertificates()
+	for _, c := range trustedCerts {
 		printCert(c)
 	}
 
@@ -185,7 +212,7 @@ func keysGenerate(cmd *cobra.Command, args []string) {
 		fatalf("could not generate key: %v", err)
 	}
 
-	caStore.AddCert(cert)
+	certificateStore.AddCert(cert)
 	fingerprint := trustmanager.FingerprintCert(cert)
 	fmt.Println("Generated new keypair with ID: ", string(fingerprint))
 }
--- a/cmd/notary/keys.go
+++ b/cmd/notary/keys.go
@@ -67,7 +67,7 @@ func keysRemove(cmd *cobra.Command, args []string) {
 	gunOrID := args[0]
 
 	// Try to retrieve the ID from the CA store.
-	cert, err := caStore.GetCertificateBykID(gunOrID)
+	cert, err := caStore.GetCertificateByFingerprint(gunOrID)
 	if err == nil {
 		fmt.Printf("Removing: ")
 		printCert(cert)
@@ -81,7 +81,7 @@ func keysRemove(cmd *cobra.Command, args []string) {
 	}
 
 	// Try to retrieve the ID from the Certificate store.
-	cert, err = certificateStore.GetCertificateBykID(gunOrID)
+	cert, err = certificateStore.GetCertificateByFingerprint(gunOrID)
 	if err == nil {
 		fmt.Printf("Removing: ")
 		printCert(cert)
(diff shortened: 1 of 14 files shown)


cmd/notary/tuf.go
```diff
@@ -461,6 +464,77 @@ func bootstrapClient(remote store.RemoteStore, repo *tuf.TufRepo, kdb *keys.KeyD
...
+		// Check to see if this leafCertificate has a chain to one of the Root CAs
+		// of our CA Store.
+		certList := []*x509.Certificate{leafCert}
+		err = trustmanager.Verify(caStore, gun, certList)
+		if err == nil {
+			certs[kID] = rootSigned.Keys[kID]
+		}
+	}
+	_, err = signed.VerifyRoot(root, 0, certs, 1)
+	if err != nil {
```
endophage: `return err`

cmd/notary/tuf.go
```diff
@@ -461,6 +464,77 @@ func bootstrapClient(remote store.RemoteStore, repo *tuf.TufRepo, kdb *keys.KeyD
...
+	for _, kID := range rootSigned.Roles["root"].KeyIDs {
+		// TODO(dlaw): currently assuming only one cert contained in
+		// public key entry. Need to fix when we want to pass in chains.
+		k, _ := pem.Decode([]byte(rootSigned.Keys["kid"].Public()))
+		decodedCerts, err := x509.ParseCertificates(k.Bytes)
+		if err != nil {
+			continue
+		}
+
+		// TODO(diogo): Assuming that first certificate is the leaf-cert. Need to
```
endophage: We should document this assumption somewhere. As we control format we can standardize on always putting the leaf cert first.

cmd/notary/tuf.go
```diff
@@ -461,6 +464,77 @@ func bootstrapClient(remote store.RemoteStore, repo *tuf.TufRepo, kdb *keys.KeyD
...
+	rootSigned := &data.Root{}
+	err := json.Unmarshal(root.Signed, rootSigned)
+	if err != nil {
+		return err
+	}
+	certs := make(map[string]*data.PublicKey)
+	for _, kID := range rootSigned.Roles["root"].KeyIDs {
+		// TODO(dlaw): currently assuming only one cert contained in
+		// public key entry. Need to fix when we want to pass in chains.
+		k, _ := pem.Decode([]byte(rootSigned.Keys["kid"].Public()))
```
endophage: I think I hacked this in while we were prepping for the demo. We should check the error response from Decode.

cmd/notary/tuf.go
```diff
@@ -461,6 +464,77 @@ func bootstrapClient(remote store.RemoteStore, repo *tuf.TufRepo, kdb *keys.KeyD
...
+		decodedCerts, err := x509.ParseCertificates(k.Bytes)
+		if err != nil {
+			continue
+		}
+
+		// TODO(diogo): Assuming that first certificate is the leaf-cert. Need to
+		// iterate over all decodedCerts and find a non-CA one (should be the last).
+		leafCert := decodedCerts[0]
+		leafID := string(trustmanager.FingerprintCert(leafCert))
```
endophage: Took me a minute to work out what was going on here. We should add a new key type to indicate a cert chain. Maybe something like `pem:chain:rsa` (aiming for the type string to indicate order of operations, we start with a pem, after decoding we get a chain, and each item in the chain is used for rsa, or whatever cipher we choose).

endophage: Also, IDs are hex strings right? If we've done a correct hex encoding, the `string()` case shouldn't be necessary.

cmd/notary/tuf.go
diogomonica: I Intend on fixing this. I think depending on ordering for something to work is crappy programming.

cmd/notary/tuf.go
diogomonica: `Decode` doesn't return `error`: https://golang.org/pkg/encoding/pem/#Decode

cmd/notary/tuf.go
diogomonica: My thought was that someone could give me malicious JSON data where the `kID` does not actually match the contents of the PEM certificate inside of Public. This way, by having me computing the `leafID` from `leafCert`, it ensure that I always do a `certificateStore.GetCertificateByKID()` with the right `kID`. This said, if everything goes correctly, `kID == leafID`.

We're going to need tons of tests around this once it's refactored into libnotary.

I added a commit that changes `trustmanager.FingerprintCert` to return type `string`. I also renamed `kID` as `fingerprint` everywhere in notary.

cmd/notary/tuf.go
endophage: Ah, nevermind then. Saw the `_` and assumed I was skipping a check when I put that in. Not familiar enough with the pem lib.

cmd/notary/tuf.go
endophage: No worries about recomputing, that completely makes sense. This also though, breaks down possibly multiple certs in the TUF key into their individual IDs. Was that not the purpose here?

cmd/notary/keys.go
```diff
@@ -66,8 +66,8 @@ func keysRemove(cmd *cobra.Command, args []string) {
 	//TODO (diogo): Validate Global Unique Name. We probably want to reject 1 char GUNs.
 	gunOrID := args[0]
 
-	// Try to retreive the ID from the CA store.
-	cert, err := caStore.GetCertificateBykID(gunOrID)
+	// Try to retrieve the ID from the CA store.
+	cert, err := caStore.GetCertificateByFingerprint(gunOrID)
```
NathanMcCauley: Why `gunOrID`? Because we can't tell from cmdline which type it is?

NathanMcCauley: Wish we had a way to optimistically parse and figure it out earlier in the method.

diogomonica: Since GUNs can really be anything, there really isn't a good way.

trustmanager/x509store_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+	// Create a store and add the CA root
+	store := NewX509MemStore()
+	err = store.AddCert(rootCA)
+	if err != nil {
+		t.Fatalf("failed to load certificate from file: %v", err)
+	}
+
+	// Get our certList with Leaf Cert and Intermediate
+	certList := []*x509.Certificate{intermediateCA, intermediateCA}
```
NathanMcCauley: Why have the intermediate twice?

diogomonica: The point of the test was not having a leaf. Since the method requires at least 2 certs, or it errors out, I put two intermediateCA certs in it.

trustmanager/x509store.go
```diff
@@ -33,3 +37,62 @@ type ValidatorFunc func(cert *x509.Certificate) bool
...
+	}
+
+	// We exited the loop with no leaf certificates
+	if leafCert == nil {
+		return errors.New("no leaf certificates found")
+	}
+
+	// We have one leaf certificate and at least one intermediate. Lets add this
+	// Cert Pool as the Intermediates list on our VerifyOptions
+	opts.Intermediates = intPool
```
NathanMcCauley: Should we harden the logic here by checking that `intPool` isn't empty? Right now the only reason we know that is based on the fact the we checked earlier that certList had two entries.

trustmanager/x509store.go
```diff
@@ -33,3 +37,62 @@ type ValidatorFunc func(cert *x509.Certificate) bool
...
+
+	// Finally, let's call Verify on our leafCert with our fully configured options
+	chains, err := leafCert.Verify(opts)
+	if len(chains) == 0 || err != nil {
+		return fmt.Errorf("certificate validation failed not verify: %v", err)
```
NathanMcCauley: double negative 'failed not verify'

maybe "certificate validation failed. Not verified: "

trustmanager/x509store.go
diogomonica: Well, the code either adds to a cert pool or not based on IsCA. Since we error out if we have less than two certs, and we error out if two leaf-certs are found, this means that intPool can never be empty (AddCert doesn't return any error.

I'm fine with being more defensive though.

NathanMcCauley:
LGTM after nits.

PR #11, opened 2015-06-21
Auth mechanisms

endophage:
adding htpasswd and silly auth to support the same set of mechanisms as distribution


server/server.go
```diff
@@ -96,11 +96,13 @@ func run(ctx context.Context, addr, tlsCertFile, tlsKeyFile string, trust signed
 	}
 	tlsLsnr := tls.NewListener(lsnr, tlsConfig)
 
-	var ac auth.AccessController = nil
-	//ac, err := auth.GetAccessController("token", map[string]interface{}{})
-	//if err != nil {
-	//	return err
-	//}
+	if authMethod == "" {
+		authMethod = "silly"
```
stevvooe: @diogomonica Silly auth is just for testing functionality of the authorization stack. Adding a test here and defaulting to is not the right thing thing to do.

diogomonica: Thank you.

endophage:
@diogomonica any tests on the authentication lib should live in docker/distribution. We shouldn't be writing tests for another package's code in here.

dmcgowan:
needs rebase

client/client_test.go
```diff
@@ -270,80 +260,6 @@ func TestAddListTarget(t *testing.T) {
 	assert.True(t, newFileFound, "second changelist file not found")
 
 	changelistDir.Close()
-
```
NathanMcCauley: Why does this test get removed?

endophage: It shouldn't have been. Git did a weird merge on this file and I just realized I pulled this file from my forked master rather than docker/master. Fixing now.

dmcgowan:
I will be testing this today, also another rebase is needed

stevvooe:
LGTM

utils/http.go
```diff
@@ -55,6 +55,18 @@ func (root *rootHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	ctx = context.WithValue(ctx, "http.request", r)
 
+	if root.auth != nil {
+		var err error
+		access := buildAccessRecords(vars["imageName"], root.actions...)
+		if ctx, err = root.auth.Authorized(ctx, access...); err != nil {
+			if err, ok := err.(auth.Challenge); ok {
+				err.ServeHTTP(w, r)
```
dmcgowan: The auth challenge sets the `WWW-Authenticate` header, but it does not set the status code. Currently it returns a 200 with the header.

endophage: Returning http.StatusUnauthorized now.

dmcgowan:
LGTM

PR #21, opened 2015-06-24
adding exec command that pipes into shell directly

progrium:
Allows you to stop doing this:
```
curl example.com/install.sh | notary verify example.com/install v1 | sh
```
And start doing this:
```
curl example.com/install.sh | notary exec example.com/install v1
```

Signed-off-by: Jeff Lindsay <progrium@gmail.com>

### DIFF
--- a/cmd/notary/tuf.go
+++ b/cmd/notary/tuf.go
@@ -5,10 +5,13 @@ import (
 	"crypto/sha256"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"path"
 	"path/filepath"
+	"syscall"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/endophage/gotuf"
@@ -69,7 +72,14 @@ var cmdVerify = &cobra.Command{
 	Use:   "verify [ GUN ] <target>",
 	Short: "verifies if the content is included in the trusted collection",
 	Long:  "verifies if the data passed in STDIN is included in the trusted collection identified by the Global Unique Name.",
-	Run:   verify,
+	Run:   verifyCmd,
+}
+
+var cmdExec = &cobra.Command{
+	Use:   "exec [ GUN ] <target>",
+	Short: "verifies content in trusted collection and pipes into shell",
+	Long:  "verifies if the data passed in STDIN is verified and then it is effectively piped into SHELL if set or /bin/sh.",
+	Run:   execCmd,
 }
 
 func tufAdd(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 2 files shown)


cmd/notary/tuf.go
```diff
@@ -348,7 +358,50 @@ func tufRemove(cmd *cobra.Command, args []string) {
...
+	}
+	shCmd := exec.Command(sh)
+	shCmd.Stdout = os.Stdout
+	shCmd.Stderr = os.Stderr
+	shStdin, err := shCmd.StdinPipe()
+	if err != nil {
+		logrus.Error("unable to create stdin pipe:", err)
+		return
+	}
+	go verify(cmd, args, shStdin)
```
somethingnew2-0: Doesn't this cause a race condition where a command could run before it is verified?

cmd/notary/tuf.go
```diff
@@ -398,7 +451,8 @@ func verify(cmd *cobra.Command, args []string) {
 		logrus.Error("notary: data not present in the trusted collection.")
 		os.Exit(1)
 	} else {
-		_, _ = os.Stdout.Write(payload)
+		_, _ = output.Write(payload)
+		output.Close()
```
endophage: `defer output.Close()` is more idiomatic and ensures it will always execute, even if the write panics.

endophage:
@diogomonica @NathanMcCauley one more set of eyes please.

cmd/notary/tuf.go
progrium: fair enough.

endophage: Sorry, I'm having a rough week. That defer should also be moved to the top of the function. Obviously it's never going to get executed if the write panics and it's the next line of code. Also, postcondition on the function should be output buffer is closed, in the current implementation, the output buffer may or may not be closed depending on whether there's an error or not. (it'll get cleaned up on exit anyway, but better explicit than implicit)

progrium: Yep, sounds right. I'll fix with any other suggestions from review.

NathanMcCauley:
@progrium thanks for the PR, the code looks really clean!

My fundamental question is this: why is this better than piping to `sh`? Notary is expected to be providing trust over arbitrary content, not just executable content, so its not clear to me why `notary` should directly support executing content.

Is `exec` expected to get smarter so that it can figure out that anything other than a shell script has been downloaded? For example, how will `exec` work if you pull down an ELF binary? A jar?

progrium:
It's not "better" but it's still quite a common case that's made shorter and debatably cleaner. It also adds the affordance for it in a more discoverable way. 

Though you have certainly inspired an interesting path for it. It could indeed get smarter. Files with `#!` get piped to shell, other detectable headers can be handled appropriately. Maybe. Interesting to think about, though not necessary. I would actually say that should be something another tool might do that uses notary as a library in the process. 

Regardless, I think even if it continues as nothing smarter than this, it would be preferred by many over a second pipe into `sh`. I don't think it reduces the idea that it's more general than a tool to verify shell scripts, but I think it adds a bit of flair that will encourage more people to use it.

diogomonica:
@progrium thanks for the contribution. We have tons of stuff to do here, and we would love to keep you engaged in the project. We're working on some documentation/blog post about notary and where we would like to go with it.

As far as `exec` goes, I think this functionality shouldn't be in notary. Notary should exclusively do trust operations:
- Create/Remove/Operate on trusted collections
- Create/Remove/Operate on keys
- Provide an easy way to verify if a specific content is valid or not 

The reason why verify receives from stdin is just to make it easy for people to verify the validity of arbitrary data. Stuff like `cat file | notary 2&>  /dev/null  && command` or the noted example `curl ... | notary | sh` are just cool examples of this, but ultimately `notary verify GUN file` should be the way people verify the validity of content.

We could replicate exec easily by doing:

```sh
nexec() {
 notary verify "$1" "$2" | sh
}
```
And then:
```sh
➜  demo $  cat install.sh | nexec diogomonica.com/scripts v1
Installing...
```

IMHO would love to keep notary's functionality tight. But just my $0.02.

progrium:
If there was a library form of notary for doing verifications, it would probably just be a single method, right? `Verify(data []byte, gun, file string) bool`?

endophage:
We need to create a library form to better integrate it with docker/docker. If you wanted to take a crack at that it would be awesome! We're on a reasonably tight schedule for it though and I'm looking to have that PR in by Monday. I was planning to do it after I add timestamping support to the server, which I should be PR-ing this evening or tomorrow.

progrium:
I'm in no rush, so I'll stay out of your way. But good to hear. 

Anyway, I'd close this but maybe we can leave it open for further comments until y'all get tired of seeing it?

diogomonica:
Cleaning up PRs. I don't think we want this for now.

PR #23, opened 2015-07-01
Removed all references of GUN from filestore

diogomonica:
@endophage removed GUN references for the FileStore as you requested.

The FileStore still does operations on directories (which might be sightly counterintuitive), but at least now it's explicit that you're removing a directory, and should pass a param accordingly.

This also bails if you pass a file, instead of a directory.


endophage:
LGTM. Just renaming things.

PR #24, opened 2015-07-01
Adding support for encrypted files to filestore

diogomonica:
@NathanMcCauley @endophage adds support for encrypted files in filestore. The main goal is to encrypt root keys in a future PR.

### DIFF
--- a/trustmanager/filestore.go
+++ b/trustmanager/filestore.go
@@ -17,11 +17,13 @@ const private os.FileMode = 0700
 const encryptedExt string = "enc"
 
 const (
-	SaltSize = 32
-	KeySize  = 32
-	ScryptN  = 32768
-	ScryptR  = 8
-	ScryptP  = 1
+	AESBlocksize = 16
+	SaltSize     = 32
+	KeySize      = 32
+	NonceSize    = 12
+	ScryptN      = 32768
+	ScryptR      = 8
+	ScryptP      = 1
 )
 
 // FileStore is the interface for all FileStores
@@ -75,8 +77,8 @@ func (f *fileStore) Add(name string, data []byte) error {
 	return ioutil.WriteFile(filePath, data, f.perms)
 }
 
-// AddEncrypted writes encrypted data to a file with a given name, given a key
-func (f *fileStore) AddEncrypted(name string, data []byte, passphrase string) error {
+// AddEncrypted writes encrypted data to a file with a given name, given a passphrase used for key derivation
+func (f *fileStore) AddEncrypted(name string, plaintext []byte, passphrase string) error {
 	filePath := f.genEncryptedFilePath(name)
 	createDirectory(filepath.Dir(filePath), f.perms)
 
@@ -106,18 +108,23 @@ func (f *fileStore) AddEncrypted(name string, data []byte, passphrase string) er
 	}
 
 	// Generate random nonce for GCM
-	nonce := make([]byte, gcm.NonceSize())
+	fmt.Println(gcm.NonceSize())
+	nonce := make([]byte, NonceSize)
 	_, err = rand.Read(nonce)
 	if err != nil {
 		return err
 	}
 
-	// Encypts the data using the nonce
-	cipherText := gcm.Seal(nonce, nonce, data, nil)
+	// Append the nonce to the finalBytes so we can use it on decrypt
+	finalBytes := append(salt, nonce...)
 
-	// Append the salt to the outData so we can use it on decrypt
-	outData := append(salt, cipherText...)
-	return ioutil.WriteFile(filePath, outData, f.perms)
+	// Encypts the plaintext using the nonce
+	cipherText := gcm.Seal(nil, nonce, plaintext, nil)
+
+	// Append the ciphertext to the finalBytes
+	finalBytes = append(finalBytes, cipherText...)
+
+	return ioutil.WriteFile(filePath, finalBytes, f.perms)
 }
 
 // Remove removes a file identified by a name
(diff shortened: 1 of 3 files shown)


trustmanager/filestore.go
```diff
@@ -97,6 +150,52 @@ func (f *fileStore) GetData(name string) ([]byte, error) {
...
+		return nil, err
+	}
+
+	// If the data doesn't have at least this size, it doesn't have any data.
+	if len(data) <= SaltSize+24+4 {
+		return nil, fmt.Errorf("Error while decrypting, not enough data in: %s", filePath)
+	}
+
+	// Get the salt from the first SaltSize bytes in data
+	salt := make([]byte, SaltSize)
```
endophage: just use slices

```go
salt := data[:SaltSize]
data = data[SaltSize:]
```

Can do the same for the nonce at 187:188, assuming the above change is made it becomes:

```go
nonce := data[:gcm.NonceSize()]
data = data[gcm.NonceSize():]
```

Then `gcm.Open` just takes `data` with no additional slicing. You can ignore the reassignments to data but should definitely just make use of slicing to assign salt and nonce.

trustmanager/filestore.go
```diff
@@ -61,7 +69,52 @@ func (f *fileStore) Add(name string, data []byte) error {
...
+	// Derive a key from the passphrase
+	// First generate random salt for scrypt
+	salt := make([]byte, SaltSize)
+	_, err := rand.Read(salt)
+	if err != nil {
+		return err
+	}
+
+	// With the salt, generate key derived from passphrase
+	derivedKey, err := scrypt.Key([]byte(passphrase), salt, 16384, 8, 1, KeySize)
```
endophage: Should those integer constants be declared at the top of the file for clarity? They're reused below.

trustmanager/filestore.go
```diff
@@ -61,7 +75,52 @@ func (f *fileStore) Add(name string, data []byte) error {
 	return ioutil.WriteFile(filePath, data, f.perms)
 }
 
-// Remove removes a file identified by name
+// AddEncrypted writes encrypted data to a file with a given name, given a key
```
NathanMcCauley: s/given a key/given a passphrase used for key derivation/

trustmanager/filestore.go
```diff
@@ -61,7 +75,52 @@ func (f *fileStore) Add(name string, data []byte) error {
...
+func (f *fileStore) AddEncrypted(name string, data []byte, passphrase string) error {
+	filePath := f.genEncryptedFilePath(name)
+	createDirectory(filepath.Dir(filePath), f.perms)
```
NathanMcCauley: Can `createDirectory` error?

trustmanager/filestore.go
```diff
@@ -61,7 +75,52 @@ func (f *fileStore) Add(name string, data []byte) error {
...
+
+	// Generate random nonce for GCM
+	nonce := make([]byte, gcm.NonceSize())
+	_, err = rand.Read(nonce)
+	if err != nil {
+		return err
+	}
+
+	// Encypts the data using the nonce
+	cipherText := gcm.Seal(nonce, nonce, data, nil)
```
NathanMcCauley: Why is `nonce` being passed for the `dst` argument?

trustmanager/filestore.go
```diff
@@ -61,7 +75,52 @@ func (f *fileStore) Add(name string, data []byte) error {
 	return ioutil.WriteFile(filePath, data, f.perms)
 }
 
-// Remove removes a file identified by name
+// AddEncrypted writes encrypted data to a file with a given name, given a key
+func (f *fileStore) AddEncrypted(name string, data []byte, passphrase string) error {
```
NathanMcCauley: Could we change the `data` arg to be `plaintext`?

NathanMcCauley: Its a bit confusing b/c the GCM lib's notion of `data` is the associated data for AEAD mode.

trustmanager/filestore.go
```diff
@@ -97,6 +156,52 @@ func (f *fileStore) GetData(name string) ([]byte, error) {
...
+// GetEncrypted decrypts and returns data given a file name
+func (f *fileStore) GetEncrypted(name string, passphrase string) ([]byte, error) {
+	filePath := f.genEncryptedFilePath(name)
+	data, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		return nil, err
+	}
+
+	// If the data doesn't have at least this size, it doesn't have any data.
+	if len(data) <= SaltSize+24+4 {
```
NathanMcCauley: What are these magic numbers? They probably deserve to have named constants.

trustmanager/filestore.go
diogomonica: No, it's best effort. If it can't do it, the file writing will fail.

trustmanager/filestore_test.go
```diff
@@ -50,7 +50,99 @@ func TestAddFile(t *testing.T) {
 	}
 }
 
-func TestRemoveFile(t *testing.T) {
+func TestAddAndGetEncrypted(t *testing.T) {
```
NathanMcCauley: Could you add a test that does a bitflip on the on-disk file and make sure that decryption fails due to authentication failure?

trustmanager/filestore.go
NathanMcCauley: Ah, looks like its to append the ciphertext on to the nonce.

NathanMcCauley:
Is this the right approach? Should we instead be using encrypted PEMs? It seems like golang supports those natively (https://golang.org/pkg/crypto/x509/#DecryptPEMBlock)?

trustmanager/filestore_test.go
diogomonica: I have a check that uses the wrong key to decrypt. But I can do a bit-flip.

diogomonica:
PEM encryption supports no authenticated encryption modes: 

```go
const (
        _ PEMCipher = iota
        PEMCipherDES
        PEMCipher3DES
        PEMCipherAES128
        PEMCipherAES192
        PEMCipherAES256
)
```

Since the filestore makes this transparent, I don't see a big issue, but for compatibility we might want PEM encrypted.

trustmanager/filestore_test.go
NathanMcCauley: wrong key is different. since we're using a counter mode bit flips allow for changing the ciphertext and getting a predictable change in the decrypted plaintext -- that is, bit flip in ciphertext leads to exact same bit flipped in the decrypted plaintext

trustmanager/filestore.go
```diff
@@ -97,6 +162,53 @@ func (f *fileStore) GetData(name string) ([]byte, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	gcm, err := cipher.NewGCM(c)
+	if err != nil {
+		return nil, err
+	}
+
+	gcm.Overhead()
```
NathanMcCauley: Unused?

trustmanager/filestore.go
```diff
@@ -97,6 +162,53 @@ func (f *fileStore) GetData(name string) ([]byte, error) {
 	return data, nil
 }
 
+// GetEncrypted decrypts and returns data given a file name
+func (f *fileStore) GetEncrypted(name string, passphrase string) ([]byte, error) {
```
NathanMcCauley: s/GetEncrypted/GetDecrypted/

thaJeztah:
@NathanMcCauley could you add a short comment why this was closed?

thaJeztah:
Ah, I think this is replaced by; https://github.com/docker/notary/pull/27, correct?

diogomonica:
@thaJeztah yeah, sorry about that. We're implementing this functionality in #27 instead to make it multi-platform and a more generic way of encrypting private keys.

PR #25, opened 2015-07-01
adding timestamping and some general cleanup

endophage:
Added timestamping, a memory store (for testing), tests, and a new dependency on stretchr/testify (only for the tests), which provides nice assertions.

Also some cleanup: removed the graceful restart stuff for simplicity and switched configuration over to viper for consistency.


cmd/notary-server/main.go
```diff
@@ -14,13 +14,14 @@ import (
 	"github.com/Sirupsen/logrus"
 	_ "github.com/docker/distribution/registry/auth/token"
 	"github.com/endophage/gotuf/signed"
+	//	"github.com/garyburd/redigo/redis"
```
NathanMcCauley: cleanup?

server/server_test.go
```diff
@@ -4,33 +4,33 @@ import (
...
-	"time"
 
 	"github.com/endophage/gotuf/signed"
 	"golang.org/x/net/context"
-
-	"github.com/docker/notary/config"
 )
 
-func TestRunBadCerts(t *testing.T) {
+// Per production deployments, we have to make running with
```
NathanMcCauley: Just delete this if its no longer needed.

server/storage/database.go
```diff
@@ -78,3 +87,28 @@ func (db *MySQLStorage) Delete(gun string) error {
...
+		return "", nil, err
+	}
+
+	return cipher, public, err
+}
+func (db *MySQLStorage) SetTimestampKey(gun, cipher string, public []byte) error {
+	stmt := "INSERT INTO `timestamp_keys` (`gun`, `cipher`, `public`) VALUES (?,?,?);"
+	_, err := db.Exec(stmt, gun, cipher, public)
+	if err, ok := err.(*mysql.MySQLError); ok {
+		if err.Number == 1022 { // duplicate key error
```
NathanMcCauley: Is there a set of consts available from the mysql package we're using?

server/timestamp/timestamp.go
```diff
@@ -0,0 +1,114 @@
...
+import (
+	"encoding/json"
+	"github.com/endophage/gotuf/data"
+	"github.com/endophage/gotuf/signed"
+	"time"
+
+	"github.com/docker/notary/server/storage"
+)
+
+func GetTimestampKey(gun string, store storage.MetaStore, crypto signed.CryptoService) (*data.TUFKey, error) {
```
NathanMcCauley: There has to be a way to reduce nesting in this method.

server/timestamp/timestamp.go
```diff
@@ -0,0 +1,114 @@
+package timestamp
```
NathanMcCauley: All methods in this file need comments.

server/timestamp/timestamp.go
```diff
@@ -0,0 +1,114 @@
...
+				}
+				return nil, err
+			}
+			return &key.TUFKey, nil
+		}
+	}
+	return data.NewTUFKey(cipher, string(public), ""), nil
+}
+
+func GetTimestamp(gun string, store storage.MetaStore, signer *signed.Signer) ([]byte, error) {
```
NathanMcCauley: s/GetTimestamp/GetOrCreateTimestamp/

server/timestamp/timestamp.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		if _, ok := err.(*storage.ErrNotFound); !ok {
+			// If we received an ErrNotFound, we're going to
+			// generate the first timestamp, any other error
+			// should be returned here
+			return nil, err
+		}
+	}
+	var ts *data.SignedTimestamp
+	if d != nil {
```
NathanMcCauley: Why is this variable called `d`?

server/timestamp/timestamp.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+		if !timestampExpired(ts) {
+			return d, nil
+		}
+	}
+	sgnd, version, err := createTimestamp(gun, ts, store, signer)
+	if err != nil {
+		return nil, err
+	}
+	out, err := json.Marshal(sgnd)
```
NathanMcCauley: Why do we have to do so much marshalling manually? I wish either gotuf or the stores took care of this and we just sent them types.

endophage: I'm slowly working on cleaning up a lot of this stuff. Latest push fixes all the casting from []byte to string and back for the key values. Still working out how to make some of the interfaces more intelligent to abstract away the json pieces.

server/storage/database.go
endophage: Unfortunately not. I had to go dig this error out of the MySQL docs. The library takes the raw output from MySQL and parses it into its error object which includes the `Number` field.

server/timestamp/timestamp.go
endophage: Done :-)

cmd/notary/tuf.go
```diff
@@ -302,10 +299,10 @@ func tufPublish(cmd *cobra.Command, args []string) {
 	if err != nil {
 		fatalf(err.Error())
 	}
-	timestamp, err := filestore.GetMeta("timestamp", 0)
-	if err != nil {
-		fatalf(err.Error())
-	}
+	//timestamp, err := filestore.GetMeta("timestamp", 0)
```
NathanMcCauley: Can we just remove the commented out code?

endophage: Yeah sorry. I habitually leave it in for my own reference for a few commits, but I'm aware it clutters things too. Pulling it now

cmd/notary-server/config.json
```diff
@@ -11,5 +11,8 @@
 	},
 	"logging": {
 		"level": 5
+	},
+	"storage": {
+		"db_url": "dockercondemo:dockercondemo@tcp(notarymysql:3306)/dockercondemo"
```
diogomonica: Is this a dev-config? if so, shouldn't the file reflect it? Also, we should start deleting mentions to dockercon.

cmd/notary/tuf.go
```diff
@@ -503,15 +481,16 @@ func bootstrapRepo(gun string, repo *tuf.TufRepo) store.MetadataStore {
...
-	err = json.Unmarshal(timestampJSON, timestamp)
-	if err != nil {
-		fatalf(err.Error())
-	}
-	repo.SetTimestamp(timestamp)
 	return filestore
 }
+
+// Use this to initialize remote HTTPStores from the config settings
+func getRemoteStore(gun string) (store.RemoteStore, error) {
```
diogomonica: This should probably take a transport and baseURL instead of assuming HTTP and going to notary:4443

server/handlers/default.go
```diff
@@ -145,3 +146,84 @@ func DeleteHandler(ctx context.Context, w http.ResponseWriter, r *http.Request)
...
+	}
+
+	vars := mux.Vars(r)
+	gun := vars["imageName"]
+
+	out, err := timestamp.GetOrCreateTimestamp(gun, store, signer)
+	if err != nil {
+		return &errors.HTTPError{
+			HTTPStatus: http.StatusInternalServerError,
+			Code:       9999,
```
diogomonica: Code 9999?

server/server.go
```diff
@@ -5,132 +5,78 @@ import (
...
 		return err
 	}
-	lsnr, err := net.ListenTCP("tcp", tcpAddr)
+
+	keypair, err := tls.LoadX509KeyPair(tlsCertFile, tlsKeyFile)
 	if err != nil {
-		return err
+		// must be able to run without certs. In prod, users may
+		// want load balancer to terminate TLS
+		logrus.Errorf("[Notary Server] Error loading TLS keys %s", err)
```
diogomonica: Not sure what the current state is, but I think the last time we discussed this we said we had to support non TLS clients. :(

server/storage/interface.go
```diff
@@ -0,0 +1,9 @@
+package storage
+
+type MetaStore interface {
+	UpdateCurrent(gun, role string, version int, data []byte) error
```
diogomonica: Why not just `Update` and `Get` ?

server/timestamp/timestamp.go
```diff
@@ -0,0 +1,133 @@
...
+	cipher, public, err := store.GetTimestampKey(gun)
+	if err != nil {
+		if _, ok := err.(*storage.ErrNoKey); ok {
+			key, err := crypto.Create("timestamp")
+			if err != nil {
+				return nil, err
+			}
+			err = store.SetTimestampKey(gun, key.Cipher(), []byte(key.Public()))
+			if err != nil {
+				if _, ok := err.(*storage.ErrTimestampKeyExists); ok {
```
diogomonica: 5 nested ifs.

diogomonica:
LGTM

server/timestamp/timestamp.go
diogomonica: david, diogo, derek? choose one :)

cmd/notary-server/config.json
endophage: This is the only config we have for mysql for now. I'm leaving it there so that when I replace it I have a reminder of the mysql connection url.

It should probably also be replicated into the dev config file.

cmd/notary/tuf.go
endophage: Other than the GUN, I'm expecting the rest of the values to come from configuration. I can rename the function to be more specific about it being an HTTPStore but the internals of the function will be updated to pull values from viper.

server/handlers/default.go
endophage: Placeholder error code. One of my cleanup tasks is to create a set of error codes.

server/server.go
endophage: Yep. If the user doesn't provide certs in the configuration it just prints the error (in case they did want TLS but configured wrong), and continues to run without TLS.

server/storage/interface.go
endophage: Eventually we will want functions that operate on old version (OK, maybe there won't be an Update for old versions, but there will be a Get). The ___Current versions are shortcuts to interact with the most recent version and they are relevant now.

server/timestamp/timestamp.go
endophage: Got it down to 3 with some reorganization.

diogomonica:
@endophage needs rebase to be merged

server/handlers/default.go
```diff
@@ -145,3 +146,84 @@ func DeleteHandler(ctx context.Context, w http.ResponseWriter, r *http.Request)
...
+	}
+
+	logrus.Debug("Writing data")
+	w.Write(out)
+	return nil
+}
+
+func GetTimestampKeyHandler(ctx context.Context, w http.ResponseWriter, r *http.Request) *errors.HTTPError {
+	s := ctx.Value("metaStore")
+	store, ok := s.(storage.MetaStore)
```
NathanMcCauley: I really dislike this pattern, I wish there was a better way. For now it'll have to do.

NathanMcCauley:
LGTM

server/handlers/default.go
endophage: I am of the same opinion, but it seems to be how the context package was designed to be used :-(

PR #27, opened 2015-07-04
Added a keyfilestore with encrypted PEM support

diogomonica:
@NathanMcCauley @endophage @dmcgowan 

Implementation of a keyfilestore that supports encrypted files. Nathan convinced me that this is the right way of storing encrypted keys.

This will be the basis for the encrypted storage of root keys, and I will add support for bundling soon (for export)

### DIFF
--- a/cmd/notary/main.go
+++ b/cmd/notary/main.go
@@ -25,7 +25,7 @@ const tufDir string = configPath + "tuf/"
 
 var caStore trustmanager.X509Store
 var certificateStore trustmanager.X509Store
-var privKeyStore trustmanager.FileStore
+var privKeyStore *trustmanager.KeyFileStore
 
 var rawOutput bool
 
@@ -91,7 +91,7 @@ func init() {
 		fatalf("could not create X509FileStore: %v", err)
 	}
 
-	privKeyStore, err = trustmanager.NewPrivateFileStore(finalPrivDir, "key")
+	privKeyStore, err = trustmanager.NewKeyFileStore(finalPrivDir)
 	if err != nil {
 		fatalf("could not create FileStore: %v", err)
 	}
--- a/cmd/notary/main.go
+++ b/cmd/notary/main.go
@@ -25,7 +25,7 @@ const tufDir string = configPath + "tuf/"
 
 var caStore trustmanager.X509Store
 var certificateStore trustmanager.X509Store
-var privKeyStore *trustmanager.KeyFileStore
+var privKeyStore trustmanager.EncryptedFileStore
 
 var rawOutput bool
 
--- a/trustmanager/keyfilestore.go
+++ b/trustmanager/keyfilestore.go
@@ -13,7 +13,7 @@ type KeyFileStore struct {
 
 // NewKeyFileStore returns a new KeyFileStore creating a private directory to
 // hold the keys.
-func NewKeyFileStore(baseDir string) (EncryptedFileStore, error) {
+func NewKeyFileStore(baseDir string) (*KeyFileStore, error) {
 	fileStore, err := NewFileStore(baseDir, keyExtension)
 	if err != nil {
 		return nil, err
--- a/trustmanager/keyfilestore.go
+++ b/trustmanager/keyfilestore.go
@@ -1,0 +1,99 @@
+package trustmanager
+
+import (
+	"crypto"
+	"errors"
+	"fmt"
+)
+
+const (
+	keyExtension = "key"
+)
+
+// KeyFileStore persists and manages private keys on disk
+type KeyFileStore struct {
+	fingerprintMap map[string]string
+	fileStore      FileStore
+}
+
+// NewKeyFileStore returns a new KeyFileStore.
+func NewKeyFileStore(directory string) (*KeyFileStore, error) {
+	fileStore, err := NewPrivateFileStore(directory, keyExtension)
+	if err != nil {
+		return nil, err
+	}
+
+	return &KeyFileStore{
+		fingerprintMap: make(map[string]string),
+		fileStore:      fileStore,
+	}, nil
+}
+
+// Add stores both the PrivateKey bytes in a file
+func (s *KeyFileStore) Add(fileName string, privKey crypto.PrivateKey) error {
+	if privKey == nil {
+		return errors.New("adding nil key to keyFileStore")
+	}
+
+	pemKey, err := KeyToPEM(privKey)
+	if err != nil {
+		return err
+	}
+
+	return s.fileStore.Add(fileName, pemKey)
+}
+
+// Get returns a PrivateKey given a filename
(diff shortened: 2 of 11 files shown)


cmd/notary/main.go
```diff
@@ -25,7 +25,7 @@ const tufDir string = configPath + "tuf/"
 
 var caStore trustmanager.X509Store
 var certificateStore trustmanager.X509Store
-var privKeyStore trustmanager.FileStore
+var privKeyStore *trustmanager.KeyFileStore
```
endophage: This should probably be an interface type with KeyFileStore implementing the interface. It will make it more flexible if we want to add other types of key store in the future.

endophage:
After adding some kind of "KeyStore" interface, I'd break up the KeyFileStore into 2 separate types. One, a basic FileStore without encryption, then an "EncryptedStore" that can be wrapped around any KeyStore (hopefully you can infer all the ...'s):

```go
type KeyStore interface {
    Get(...) ...
    Add(...) ...
    Remove(...) ...
    ...
}

// implements KeyStore
type FileStore struct {
    ...
}

// implements KeyStore 
type EncryptedStore struct {
    KeyStore // embeds any other key store to add encryption
    passphrase string // if there's something better to use here, obviously do it
}

func (es EncryptedStore) Add(fileName string, privKey crypto.PrivateKey) ... {
    if privKey == nil {
        return errors.New("adding nil key to keyFileStore")
    }

    encryptedKey, err := KeyToEncryptedPEM(privKey, es.passphrase)
    if err != nil {
        return err
    }

    fmt.Println(string(encryptedKey))
    return s.fileStore.Add(fileName, encryptedKey)
}

// you can infer EncryptedStore.Get
```

The encrypted store could now be wrapped around any other store (various remote stores might be desirable in the future).

endophage:
I'd also try and have the stores operate on `[]byte` types with a higher level wrapper to deal with (de)serializing of formats (PEM in our case).

trustmanager/keyfilestore.go
```diff
@@ -0,0 +1,99 @@
...
+		return nil, err
+	}
+
+	return &KeyFileStore{
+		fingerprintMap: make(map[string]string),
+		fileStore:      fileStore,
+	}, nil
+}
+
+// Add stores both the PrivateKey bytes in a file
```
NathanMcCauley: s/both//

trustmanager/x509utils.go
```diff
@@ -52,11 +56,41 @@ func ToPEM(cert *x509.Certificate) []byte {
 	return pemCert
 }
 
-// TeyToPEM is an utility function returns a PEM encoded Key
-func KeyToPEM(keyBytes []byte) []byte {
-	keyPEMBytes := pem.EncodeToMemory(&pem.Block{Type: "RSA PRIVATE KEY", Bytes: keyBytes})
+// KeyToPEM is an utility function returns a PEM encoded Key
```
NathanMcCauley: KeyToPEM returns a PEM encoded key from a crypto.PrivateKey

trustmanager/x509utils.go
```diff
@@ -52,11 +56,41 @@ func ToPEM(cert *x509.Certificate) []byte {
...
+	rsaKey, ok := key.(*rsa.PrivateKey)
+	if !ok {
+		return nil, errors.New("only RSA keys are currently supported")
+	}
+
+	keyBytes := x509.MarshalPKCS1PrivateKey(rsaKey)
+	return pem.EncodeToMemory(&pem.Block{Type: "RSA PRIVATE KEY", Bytes: keyBytes}), nil
+}
+
+// KeyToEncryptedPEM is an utility function returns a PEM encoded Key
```
NathanMcCauley: KeyToPEM returns a encrypted PEM encoded key from a crypto.PrivateKey

trustmanager/x509utils.go
```diff
@@ -83,6 +117,7 @@ func loadCertFromPEM(pemBytes []byte) (*x509.Certificate, error) {
 	return nil, errors.New("no certificates found in PEM data")
 }
 
+// FingerprintCert returns a TUF compliant fingerprint for a X509 Certificate
```
NathanMcCauley: What does it mean to be 'TUF compliant'?

trustmanager/x509utils.go
```diff
@@ -125,3 +160,58 @@ func LoadCertFromFile(filename string) (*x509.Certificate, error) {
...
+	key, err := ParseRawPrivateKey(pemBytes)
+	if err != nil {
+		return nil, err
+	}
+	return key, nil
+}
+
+// ParseRawPrivateKey returns a private key from a PEM encoded private key. It
+// only supports RSA (PKCS#1).
+func ParseRawPrivateKey(pemBytes []byte) (crypto.PrivateKey, error) {
```
NathanMcCauley: s/Raw/Pem/

trustmanager/x509utils.go
```diff
@@ -125,3 +160,58 @@ func LoadCertFromFile(filename string) (*x509.Certificate, error) {
...
+	case "RSA PRIVATE KEY":
+		return x509.ParsePKCS1PrivateKey(block.Bytes)
+	default:
+		return nil, fmt.Errorf("unsupported key type %q", block.Type)
+	}
+}
+
+// ParseRawEncryptedPrivateKey returns a private key from a PEM encrypted private key. It
+// only supports RSA (PKCS#1).
+func ParseRawEncryptedPrivateKey(pemBytes []byte, passphrase string) (crypto.PrivateKey, error) {
```
NathanMcCauley: s/Raw/Pem/

trustmanager/x509utils.go
diogomonica: It means using the TUF json format for fingerprinting keys.

trustmanager/keyfilestore.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+// KeyFileStore persists and manages private keys on disk
+type KeyFileStore struct {
+	FileStore
+}
+
+// NewKeyFileStore returns a new KeyFileStore creating a private directory to
+// hold the keys.
+func NewKeyFileStore(baseDir string) (EncryptedFileStore, error) {
```
endophage: As a general rule, NewXxx functions should return type Xxx, not the interface it implements. In the case that the interface lives in a different package it saves a dependency. It also takes greater advantage of Go's pseudo-duck typing when dealing with interfaces.

endophage:
LGTM

(one question, is the reason the EncryptedFileStore doesn't have the passphrase as a field because a user may be asked for multiple different password during a single command execution?)

diogomonica:
@endophage sorry, I didn't answer to that before. Yes, you might want to have different root keys having different passphrases (and still manage all root keys within a single keyfilestore).

NathanMcCauley:
LGTM

PR #28, opened 2015-07-05
changelist implementation + integration

endophage:
Adds changelists. An explicit `init` should no longer be called. On `publish` the repo is initialized if necessary, the changelist is applied, and the updates are pushed to the server.

Please provide review but bear in mind the following TODOs:
- Add an enum for Change.Action values
- Add a Changelist implementation that writes a file for each change.

### DIFF
--- a/client/changelist/change.go
+++ b/client/changelist/change.go
@@ -1,0 +1,46 @@
+package changelist
+
+// tufChange represents a change to a TUF repo
+type tufChange struct {
+	// Abbreviated because Go doesn't permit a field and method of the same name
+	Actn       int    `json:"action"`
+	Role       string `json:"role"`
+	ChangeType string `json:"type"`
+	ChangePath string `json:"path"`
+	Data       []byte `json:"data"`
+}
+
+// NewTufChange initializes a tufChange object
+func NewTufChange(action int, role, changeType, changePath string, content []byte) *tufChange {
+	return &tufChange{
+		Actn:       action,
+		Role:       role,
+		ChangeType: changeType,
+		ChangePath: changePath,
+		Data:       content,
+	}
+}
+
+// Action return c.Actn
+func (c tufChange) Action() int {
+	return c.Actn
+}
+
+// Scope returns c.Role
+func (c tufChange) Scope() string {
+	return c.Role
+}
+
+// Type returns c.ChangeType
+func (c tufChange) Type() string {
+	return c.ChangeType
+}
+
+// Path return c.ChangePath
+func (c tufChange) Path() string {
+	return c.ChangePath
+}
+
+func (c tufChange) Content() []byte {
+	return c.Data
+}
--- a/client/changelist/files_changelist.go
+++ b/client/changelist/files_changelist.go
@@ -1,0 +1,105 @@
+package changelist
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"sort"
+	"time"
+
+	"code.google.com/p/go-uuid/uuid"
+	"github.com/Sirupsen/logrus"
+)
+
+type fileChangelist struct {
+	dir string
+}
+
+func NewFileChangelist(dir string) (*fileChangelist, error) {
+	logrus.Debug("Making dir path: ", dir)
+	err := os.MkdirAll(dir, 0700)
+	if err != nil {
+		return nil, err
+	}
+	return &fileChangelist{dir: dir}, nil
+}
+
+func (cl fileChangelist) List() []Change {
+	changes := make([]Change, 0)
+	dir, err := os.Open(cl.dir)
+	if err != nil {
+		return changes
+	}
+	fileInfos, err := dir.Readdir(0)
+	if err != nil {
+		return changes
+	}
+	sort.Sort(fileChanges(fileInfos))
+	for _, f := range fileInfos {
+		if f.IsDir() {
+			continue
+		}
+		raw, err := ioutil.ReadFile(path.Join(cl.dir, f.Name()))
+		if err != nil {
+			// TODO(david): How should we handle this?
+			fmt.Println(err.Error())
+			continue
+		}
+		c := &tufChange{}
+		err = json.Unmarshal(raw, c)
+		if err != nil {
+			// TODO(david): How should we handle this?
+			fmt.Println(err.Error())
+			continue
+		}
+		changes = append(changes, c)
+	}
+	return changes
+}
+
+func (cl fileChangelist) Add(c Change) error {
+	cJson, err := json.Marshal(c)
+	if err != nil {
+		return err
+	}
+	filename := fmt.Sprintf("%020d_%s.change", time.Now().UnixNano(), uuid.New())
+	return ioutil.WriteFile(path.Join(cl.dir, filename), cJson, 0644)
+}
+
+func (cl fileChangelist) Clear(archive string) error {
+	dir, err := os.Open(cl.dir)
+	if err != nil {
+		return err
+	}
+	files, err := dir.Readdir(0)
+	if err != nil {
+		return err
+	}
+	for _, f := range files {
+		os.Remove(path.Join(cl.dir, f.Name()))
+	}
+	return nil
+}
+
+func (cl fileChangelist) Close() error {
+	// Nothing to do here
+	return nil
+}
+
+type fileChanges []os.FileInfo
+
+func (cs fileChanges) Len() int {
+	return len(cs)
+}
+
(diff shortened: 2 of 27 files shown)


client/changelist/change.go
```diff
@@ -0,0 +1,73 @@
+package changelist
+
+import (
+	"encoding/json"
+
+	"github.com/endophage/gotuf/data"
+)
+
+type tufChange struct {
+	Actn       string `json:"action"`
```
diogomonica: Why not just Action?

endophage: Can't have a field and method with the same name on a struct.

client/changelist/change.go
```diff
@@ -0,0 +1,73 @@
...
+		Actn:       action,
+		Role:       role,
+		ChangeType: changeType,
+		ChangePath: changePath,
+		Data:       content,
+	}
+}
+
+func NewTufTargetChange(action, target string, hash map[string]data.HexBytes, size int64, custom map[string]interface{}) (*tufChange, error) {
+	jsonCustom, err := json.Marshal(custom)
```
diogomonica: Why are we Marshaling and unmarshaling? Maybe add a comment?

client/changelist/change.go
```diff
@@ -0,0 +1,73 @@
...
+
+type tufChange struct {
+	Actn       string `json:"action"`
+	Role       string `json:"role"`
+	ChangeType string `json:"type"`
+	ChangePath string `json:"path"`
+	Data       []byte `json:"data"`
+}
+
+func NewTufChange(action, role, changeType, changePath string, content []byte) *tufChange {
```
diogomonica: Needs comment.

client/changelist/change.go
```diff
@@ -0,0 +1,73 @@
...
+	return &tufChange{
+		Actn:       action,
+		Role:       role,
+		ChangeType: changeType,
+		ChangePath: changePath,
+		Data:       content,
+	}
+}
+
+func NewTufTargetChange(action, target string, hash map[string]data.HexBytes, size int64, custom map[string]interface{}) (*tufChange, error) {
```
diogomonica: Needs comment. Not sure what it's going on here with the custom. Is it a best effort addition of extra params?

endophage: Every TUF target has a custom field that can contain any information another application wants to put in there. It's defined in the TUF spec and the only way to deal with it in Go is with a `map[string]interface{}`

client/changelist/change.go
endophage: Will do. FYI the Unmarshal is a `json.RawMessage.UnmarshalJSON` which is a glorified cast. `json.RawMessage` is defined as `type RawMessage []byte`

dmcgowan: This is more than a glorified case, it is a copy `// UnmarshalJSON sets *m to a copy of data.`. Why not just actually cast here, this is perfectly acceptable `json.RawMessage(jsonCustom)`

client/changelist/change.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	customRaw := json.RawMessage{}
+	err = customRaw.UnmarshalJSON(jsonCustom)
+	if err != nil {
+		return nil, err
+	}
+	content := &data.FileMeta{
+		Length: size,
+		Hashes: hash,
+		Custom: &customRaw,
```
dmcgowan: I know this is an upstream definition, but since you are the author I will ask here. Why does use a pointer?

client/changelist/change.go
endophage: That's fair, it does copy. I've just had some weird behaviour with direct casts to json.RawMessage, although I realize there shouldn't be any issues with that. I can switch it over.

client/changelist/change.go
endophage: Upstream as you say. I don't have a good reason, but the original flynn go-tuf implementation would let you add many targets with a single, shared piece of custom data in one operation so that may have been the original reason.

endophage: I just removed NewTufTargetChange instead. I didn't end up using it in the command due to how we generate the FileMeta object.

client/changelist/files_changelist.go
```diff
@@ -0,0 +1,103 @@
...
+		return changes
+	}
+	sort.Sort(fileChanges(fileInfos))
+	for _, f := range fileInfos {
+		if f.IsDir() {
+			continue
+		}
+		raw, err := ioutil.ReadFile(path.Join(cl.dir, f.Name()))
+		if err != nil {
+			// TODO(david): How should we handle this?
```
NathanMcCauley: I think we need to fail hard if theres a corrupted/unreadable file in this dir. Otherwise we risk losing data.

client/changelist/files_changelist.go
```diff
@@ -0,0 +1,103 @@
...
+			fmt.Println(err.Error())
+			continue
+		}
+		c := &tufChange{}
+		err = json.Unmarshal(raw, c)
+		if err != nil {
+			// TODO(david): How should we handle this?
```
NathanMcCauley: same comment as above

client/changelist/files_changelist.go
```diff
@@ -0,0 +1,103 @@
...
+		return nil, err
+	}
+	return &fileChangelist{dir: dir}, nil
+}
+
+func (cl fileChangelist) List() []Change {
+	changes := make([]Change, 0)
+	dir, err := os.Open(cl.dir)
+	if err != nil {
+		return changes
```
NathanMcCauley: Shouldn't this return an error?

cmd/notary/tuf.go
```diff
@@ -324,20 +350,33 @@ func tufRemove(cmd *cobra.Command, args []string) {
 	}
 	gun := args[0]
 	targetName := args[1]
-	kdb := keys.NewDB()
-	signer := signed.NewSigner(NewCryptoService(gun))
-	repo := tuf.NewTufRepo(kdb, signer)
+	//	kdb := keys.NewDB()
```
NathanMcCauley: remove?

cmd/notary/tuf.go
```diff
@@ -324,20 +350,33 @@ func tufRemove(cmd *cobra.Command, args []string) {
...
-
-	err := repo.RemoveTargets("targets", targetName)
+	clPath := path.Join(viper.GetString("tufDir"), gun, "changelist")
+	cl, err := changelist.NewFileChangelist(clPath)
 	if err != nil {
 		fatalf(err.Error())
 	}
 
-	saveRepo(repo, filestore)
+	c := changelist.NewTufChange("remove", "targets", "target", targetName, nil)
```
NathanMcCauley: function now takes an `int`

client/changelist/change.go
```diff
@@ -0,0 +1,46 @@
+package changelist
+
+// tufChange represents a change to a TUF repo
+type tufChange struct {
+	// Abbreviated because Go doesn't permit a field and method of the same name
+	Actn       int    `json:"action"`
+	Role       string `json:"role"`
```
NathanMcCauley: Any reason that `Role` and `ChangeType` aren't enums as well?

client/changelist/files_changelist.go
endophage: If the directory doesn't exist, I felt returning an empty list was appropriate. No dir, and empty dir, are functionally the same thing for us.

client/changelist/change.go
endophage: delegations mean roles have some structural rules but can be almost anything (either they're a base role, or they're prefixed with "targets/"). ChangeType could also be an enum, I'll add that.

client/changelist/files_changelist.go
NathanMcCauley: disagree, specifying an empty directory probably indicates there is nothing available.

a nonexistent directory seems more like a misconfiguration

client/changelist/files_changelist.go
NathanMcCauley: json unmarshalling failure almost certainly means corruption, right?

client/changelist/files_changelist.go
endophage: Agreed, just wanted input before I went that route.

client/changelist/files_changelist.go
endophage: All the directory paths, both for reading and writing, are generated from a combination of user input and static configuration. If a user is messing with the configuration they're going to get weird behaviour. 

Assuming user's aren't messing with configuration (which they shouldn't unless they really know what they're doing), a non-existent directory indicates no changes are pending, which is functionally identical to an empty directory.

dmcgowan:
Work in progress is no longer using this branch

PR #29, opened 2015-07-06
WIP: client interface

diogomonica:
DO NOT MERGE, THIS IS A WIP.

Attempting to refactor notary cli into libnotary and changing root generation to be independent from repository generation.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"crypto/x509"
 	"encoding/json"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"io/ioutil"
@@ -16,10 +17,12 @@ import (
 
 	"github.com/docker/notary/trustmanager"
 	"github.com/endophage/gotuf"
+	tufclient "github.com/endophage/gotuf/client"
 	"github.com/endophage/gotuf/data"
 	"github.com/endophage/gotuf/keys"
 	"github.com/endophage/gotuf/signed"
 	"github.com/endophage/gotuf/store"
+
 	"github.com/spf13/viper"
 )
 
@@ -40,6 +43,18 @@ type Client interface {
 	GetRepository(gun string, baseURL string, transport http.RoundTripper) (Repository, error)
 }
 
+// Repository is the interface that represents a Notary Repository
+type Repository interface {
+	Update() error
+	Initialize(key *data.PublicKey) error
+
+	AddTarget(target *Target) error
+	ListTargets() ([]*Target, error)
+	GetTargetByName(name string) (*Target, error)
+
+	Publish() error
+}
+
 type NotaryClient struct {
 	configFile       string
 	caStore          trustmanager.X509Store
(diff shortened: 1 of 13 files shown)


client/client.go
```diff
@@ -0,0 +1,421 @@
...
+	signer       *signed.Signer
+	tufRepo      *tuf.TufRepo
+	fileStore    store.MetadataStore
+	privKeyStore trustmanager.FileStore
+}
+
+// Target represents a simplified version of the data TUF operates on.
+type Target struct {
+	Name   string
+	Hashes data.Hashes
```
dmcgowan: This still exposes a TUF data type on the published interface `data.Hashes`. Ideally we could integrate a docker package such as `distribution/digest`.

endophage: A receiver of a Target object shouldn't need to know anything about TUF though right? `data.Hashes` is just a `map[string][]byte` with some convenience functions defined on the `[]byte`

client/client.go
```diff
@@ -0,0 +1,421 @@
...
+
+// Update forces TUF to download the remote timestamps and verify if there are
+// any remote changes.
+func (r *NotaryRepository) Update() error {
+	return nil
+}
+
+// Initialize creates a new repository by using rootKey as the root Key for the
+// TUF repository.
+func (r *NotaryRepository) Initialize(rootKey *data.PublicKey) error {
```
endophage: If we're aiming to reduce the dependencies of calling code on tuf, this should probably take `rootKeyID string`

client/client.go
```diff
@@ -0,0 +1,421 @@
...
+	ListTargets() ([]*Target, error)
+	GetTargetByName(name string) (*Target, error)
+
+	Publish() error
+}
+
+// NewClient is a helper method that returns a new notary Client, given a config
+// file. It makes the assumption that the base directory for the config file will
+// be the place where trust information is being cached locally.
+func NewClient(configFile string) (*NotaryClient, error) {
```
endophage: This shouldn't be doing config parsing. Pass in the config values themselves. Both our commands and docker/docker already do config parsing.

client/client.go
```diff
@@ -0,0 +1,421 @@
...
+
+// ErrRepositoryNotExist gets returned when trying to make an action over a repository
+/// that doesn't exist
+var ErrRepositoryNotExist = errors.New("repository does not exist")
+
+// Client is the interface that defines the Notary Client type
+type Client interface {
+	ListPrivateKeys() []data.PrivateKey
+	GenRootKey(passphrase string) (*data.PublicKey, error)
+	GetRepository(gun string, baseURL string, transport http.RoundTripper) (Repository, error)
```
endophage: Why use a RoundTripper? Does it afford some benefit to us over just doing an `http.Get`?

diogomonica: @dmcgowan this one is yours.

client/client.go
diogomonica: Agreed. I'm tired. Changing it.

client/client.go
diogomonica: And then create a NewRole with the string directly for root?

client/client.go
```diff
@@ -0,0 +1,576 @@
...
+	privKeyStore, err := trustmanager.NewKeyFileStore(filepath.Join(c.baseDir, privDir))
+	if err != nil {
+		return nil, err
+	}
+
+	signer := signed.NewSigner(NewCryptoService(gun, privKeyStore))
+
+	return &NotaryRepository{Gun: gun,
+		baseURL:          baseURL,
+		tufRepoPath:      filepath.Join(c.baseDir, tufDir),
```
endophage: `filepath.Join(c.baseDir, tufDir, gun)`

client/client.go
```diff
@@ -0,0 +1,576 @@
...
+	GenRootKey(passphrase string) (*data.PublicKey, error)
+	GetRepository(gun string, baseURL string, transport http.RoundTripper) (Repository, error)
+}
+
+// Repository is the interface that represents a Notary Repository
+type Repository interface {
+	Update() error
+	Initialize(key *data.PublicKey) error
+
+	AddTarget(target *Target) error
```
dmcgowan: Target does not need to be mutated nor null, we don't need a pointer here

dmcgowan:
Work is progress is currently being done in another branch

PR #31, opened 2015-07-07
Fix compilation error

dmcgowan:
Rufus needs to be updated and tests updated to convert the algorithm to string.


diogomonica:
No longer required.

PR #33, opened 2015-07-08
Test improvements

diogomonica:
@aaronlehmann

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -66,7 +66,7 @@ type NotaryClient struct {
 	baseDir          string
 	caStore          trustmanager.X509Store
 	certificateStore trustmanager.X509Store
-	rootKeyStore     trustmanager.KeyFileStore
+	rootKeyStore     *trustmanager.KeyFileStore
 }
 
 type NotaryRepository struct {
@@ -77,7 +77,7 @@ type NotaryRepository struct {
 	signer           *signed.Signer
 	tufRepo          *tuf.TufRepo
 	fileStore        store.MetadataStore
-	privKeyStore     trustmanager.KeyFileStore
+	privKeyStore     *trustmanager.KeyFileStore
 	caStore          trustmanager.X509Store
 	certificateStore trustmanager.X509Store
 }
@@ -589,13 +589,14 @@ func (c *NotaryClient) GetRepository(gun string, baseURL string, transport http.
 		return nil, err
 	}
 
-	signer := signed.NewSigner(NewCryptoService(gun, *privKeyStore))
+	signer := signed.NewSigner(NewCryptoService(gun, privKeyStore))
 
 	return &NotaryRepository{Gun: gun,
 		baseURL:          baseURL,
 		tufRepoPath:      filepath.Join(c.baseDir, tufDir, gun),
 		transport:        transport,
 		signer:           signer,
+		privKeyStore:     privKeyStore,
 		caStore:          c.caStore,
 		certificateStore: c.certificateStore}, nil
 }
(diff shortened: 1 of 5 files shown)


client/client.go
```diff
@@ -619,13 +620,14 @@ func (c *NotaryClient) InitRepository(gun string, baseURL string, transport http
...
 
-	signer := signed.NewSigner(NewCryptoService(gun, *privKeyStore))
+	signer := signed.NewSigner(NewCryptoService(gun, privKeyStore))
 
 	nRepo := &NotaryRepository{Gun: gun,
 		baseURL:          baseURL,
 		tufRepoPath:      filepath.Join(c.baseDir, tufDir, gun),
 		transport:        transport,
 		signer:           signer,
+		privKeyStore:     privKeyStore,
```
diogomonica: Not currently being used for anything except tests, correct?

client/client_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// Look for keys in root_keys
+	// There should be a file named after the key ID of the root key we
+	// passed in.
+	rootKeyFilename := rootKey.ID() + ".key"
+	if _, err := os.Stat(filepath.Join(tempBaseDir, "private", "root_keys", rootKeyFilename)); err != nil {
+		t.Fatal("missing root key")
+	}
+
+	// Also expect a symlink from the key ID of the certificate key to this
```
diogomonica: This comment doesn't seem to be testing what it says. We're checking if we're automatically adding the generated certificate for the repository to our trusted list of certificates here.

client/client_test.go
```diff
@@ -0,0 +1,116 @@
...
+	// root key
+	certificates := client.certificateStore.GetCertificates()
+	if len(certificates) != 1 {
+		t.Fatalf("unexpected number of certificates (%d)", len(certificates))
+	}
+
+	certID := data.NewPublicKey("RSA", trustmanager.CertToPEM(certificates[0])).ID()
```
diogomonica: You could use `trustmanager.FingerprintCert(certificates[0])`. It returns a string with the TUF compliant ID.

client/client_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+	actualDest, err := os.Readlink(filepath.Join(tempBaseDir, "private", "root_keys", certID+".key"))
```
diogomonica: Maybe refactor `".key"` into `expectedExt` in the beginning of the test?

client/client.go
aaronlehmann: Correct. I think it's worth having here. I can add a comment saying only tests use it right now, if you want.

client/client_test.go
```diff
@@ -0,0 +1,116 @@
...
+		t.Fatal("missing trusted certificate")
+	}
+
+	// Check that tuf metadata files were created
+	expectedTUFMetadataFiles := []string{
+		filepath.Join("tuf", gun, "metadata", "root.json"),
+		filepath.Join("tuf", gun, "metadata", "snapshot.json"),
+		filepath.Join("tuf", gun, "metadata", "targets.json"),
+	}
+	for _, filename := range expectedTUFMetadataFiles {
```
diogomonica: Maybe we can do some JSON parsing and try to figure out if signatures were added and the root key, snapshots key and targets IDs exist in the file? 

One problem that I've seen before is the fact that tuf silently ignores when it can't find a specific key, and that translates to lack of signatures in the file.

client/client_test.go
aaronlehmann: I might have expressed it badly, but what this comment is talking about is making sure that there's a symlink inside root_keys with the correct name and target. I don't think it has to do with the trusted certificate list - that's slightly farther down when we stat a .crt file inside trusted_certificates.

diogomonica:
Small nits LGTM, feel free to merge it after they get fixed.

client/client_test.go
aaronlehmann: One of the files has a ".crt" extension, so I'm not sure it works to factor this out. But I can replace the literal strings with constants if you think it's worth it.

client/client_test.go
aaronlehmann: Great idea. Added some stuff in this area - please tell me if it's what you had in mind.

client/client_test.go
aaronlehmann: Made this change. Thanks for the suggestion.

PR #35, opened 2015-07-09
cleanup dockerfile

jfrazelle:
I am OCD about having nice dockerfiles

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,12 +2,12 @@ FROM golang
 
 COPY . /go/src/github.com/docker/notary
 
-RUN chmod 777 /tmp/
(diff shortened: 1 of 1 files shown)


jfrazelle:
also why the chmod of /tmp

Dockerfile
```diff
@@ -4,7 +4,9 @@ COPY . /go/src/github.com/docker/notary
 
 RUN chmod 777 /tmp/
```
NathanMcCauley: Is this still needed? Did we just need it for the demo?

dmcgowan:
Please fix tests...jk
LGTM

PR #40, opened 2015-07-10
Remove circle old go version

dmcgowan:
Remove remaining references to old go version


dmcgowan:
Circle green, merging

PR #41, opened 2015-07-10
Deploy notary on staging.

RichardScothern:
Create a stage-config files without SSL keys as SSL is terminated
at the ELB.

Add a service name for correct routing from https://notary-stage.docker.io

### DIFF
--- a/cmd/notary-server/stage-config.json
+++ b/cmd/notary-server/stage-config.json
@@ -1,0 +1,13 @@
+{
+	"server": {
+		"addr": ":4443"
+	},
+	"trust_service": {
+		"type": "local",
+		"hostname": "",
+		"port": ""
+	},
+	"logging": {
+		"level": 5
+	}
+}
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,8 +3,10 @@ notary:
   links:
    - notarymysql
   ports:
-   - "8080:8080"
-   - "4443:4443"
+   - "8081"
+   - "4443"
+  environment:
+    SERVICE_NAME: notary
 #rufus:
 #  volumes:
 #   - /dev/bus/usb/003/010:/dev/bus/usb/002/010


Dockerfile
```diff
@@ -10,4 +12,5 @@ EXPOSE 4443
 
 WORKDIR /go/src/github.com/docker/notary
 
-CMD [ "notary-server", "-config", "cmd/notary-server/dev-config.json" ]
+
+CMD [ "notary-server", "-config", "cmd/notary-server/stage-config.json" ]
```
NathanMcCauley: Does this imply dockerfile per environment? Is that how the rest of the apps in our infra do it?

docker-compose.yml
```diff
@@ -14,3 +16,4 @@ notarymysql:
   build: ./notarymysql/
   ports:
     - "3306:3306"
+
```
NathanMcCauley: remove

Dockerfile
thaJeztah: Perhaps

```Dockerfile
CMD [ "--config", "cmd/notary-server/stage-config.json" ]
```

That way the config can be overridden from the command-line, e.g.

    docker run notary-server -config foobar.json

thaJeztah: (unrelated) Just noticed that notary uses a single dash for non-shorthand options, i.e. `-options`, not `--options`? This is different from Docker itself, which uses `--` for non-shorthand options.

RichardScothern: There currently isn't a way to pass runtime environment variables through docker-compose to the Dockerfile. (https://github.com/docker/compose/issues/1377).  I'll talk to infra.

@thaJeztah  '-config' is a byproduct of the command line library used by notary.

cmd/notary-server/stage-config.json
```diff
@@ -0,0 +1,13 @@
+{
+	"server": {
+		"addr": ":4443"
+	},
+	"trust_service": {
+		"type": "local",
```
diogomonica: This should probably not be local and should be pointed to notary-signer.

docker-compose.yml
```diff
@@ -3,8 +3,10 @@ notary:
   links:
    - notarymysql
   ports:
-   - "8080:8080"
-   - "4443:4443"
+   - "8081"
```
diogomonica: Is this documented somewhere on the wiki? Which ports are owned by whom?

PR #43, opened 2015-07-11
Use logrus instead of fmt.Println

dmcgowan:
Do not use fmt.Println in client, shows up in UX of importers


NathanMcCauley:
LGTM

PR #44, opened 2015-07-11
Atomic updates of metadata.

endophage:
Provides an additional handler to allow clients to update any set of TUF metadata atomically

Most tests target primarily error cases and translation from sql errors to our own. With the mock, testing of happy paths is somewhat moot as they can be forced to work.


server/storage/database.go
```diff
@@ -63,6 +57,32 @@ func (db *MySQLStorage) UpdateCurrent(gun, role string, version int, data []byte
 	return nil
 }
 
+// UpdateMany atomically updates many TUF records in a single transaction
+func (db *MySQLStorage) UpdateMany(gun string, updates []MetaUpdate) error {
+	insertStmt := "INSERT INTO `tuf_files` (`gun`, `role`, `version`, `data`) VALUES (?,?,?,?) WHERE (SELECT count(*) FROM `tuf_files` WHERE `gun`=? AND `role`=? AND `version`>=?) = 0;"
```
NathanMcCauley: Why do we need this `WHERE` clause? Shouldn't this just be a DB uniqueness constraint on (gun, role, version)?

endophage: If a version gets skipped for one reason or another, the uniqueness constraint alone would allow an "old" version to get inserted, which would then never be served out.

server/handlers/default.go
```diff
@@ -35,6 +37,76 @@ func MainHandler(ctx context.Context, w http.ResponseWriter, r *http.Request) *e
...
+	defer r.Body.Close()
+	s := ctx.Value("metaStore")
+	if s == nil {
+		return &errors.HTTPError{
+			HTTPStatus: http.StatusInternalServerError,
+			Code:       9999,
+			Err:        fmt.Errorf("Version store is nil"),
+		}
+	}
+	store, ok := s.(storage.MetaStore)
```
NathanMcCauley: Why do we care about the store's type here?

endophage: Because we're going to call methods on it that are defined on the MetaStore interface. I could move this down to just above line 99, but it's much cleaner to keep this generic setup here at the top.

server/handlers/default.go
```diff
@@ -35,6 +37,76 @@ func MainHandler(ctx context.Context, w http.ResponseWriter, r *http.Request) *e
...
+		}
+	}
+	var updates []storage.MetaUpdate
+	for {
+		part, err := reader.NextPart()
+		if err == io.EOF {
+			break
+		}
+		role := part.FileName()
+		if role == "" {
```
NathanMcCauley: Do we expect to see empty multiparts?

endophage: Not from our own code, but absolutely from people out in the wild trying probing the API.

server/storage/database.go
```diff
@@ -36,26 +37,19 @@ func NewMySQLStorage(db *sql.DB) *MySQLStorage {
...
-	var exists int
-	err := row.Scan(&exists)
-	if err != nil {
-		return err
-	}
-	if exists != 0 {
-		return &ErrOldVersion{}
-	}
+func (db *MySQLStorage) UpdateCurrent(gun string, update MetaUpdate) error {
+	insertStmt := "INSERT INTO `tuf_files` (`gun`, `role`, `version`, `data`) VALUES (?,?,?,?) WHERE (SELECT count(*) FROM `tuf_files` WHERE `gun`=? AND `role`=? AND `version`>=?) = 0"
```
NathanMcCauley: Same comment as above -- isn't the `SELECT` unnecessary?

endophage: Same commend as above. It prevents old versions that didn't already exist being interleaved.

server/storage/database_test.go
```diff
@@ -0,0 +1,278 @@
...
+	"database/sql"
+	"testing"
+
+	"github.com/DATA-DOG/go-sqlmock"
+	"github.com/go-sql-driver/mysql"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMySQLUpdateCurrent(t *testing.T) {
+	db, err := sqlmock.New()
```
NathanMcCauley: Oooh mocking! +1

endophage: It's not the best mocking library, but allows me to simulate errors. Would like to work out a good way to run the happy paths against a real database so I can independently query the DB to ensure the data was inserted correctly

server/handlers/default.go
```diff
@@ -35,6 +37,76 @@ func MainHandler(ctx context.Context, w http.ResponseWriter, r *http.Request) *e
...
+	store, ok := s.(storage.MetaStore)
+	if !ok {
+		return &errors.HTTPError{
+			HTTPStatus: http.StatusInternalServerError,
+			Code:       9999,
+			Err:        fmt.Errorf("Version store not configured"),
+		}
+	}
+	vars := mux.Vars(r)
+	gun := vars["imageName"]
```
NathanMcCauley: Should we sanity check that gun here matches the gun in root.json's cert?

server/handlers/default.go
NathanMcCauley: If they're probing we should fail the request, no?

server/handlers/default.go
endophage: Hmmm, might be a good idea. Should probably do that for the single file update too if it's a root.json being updated. We can add that check in both places for polish after freeze.

server/handlers/default.go
endophage: Certainly, but not in an undefined way. This will fail very gracefully. I'm a subscriber to the idea "if you ask me to do something but there's nothing to do, I'm going to tell you it's done" (i.e. return a 200)

server/storage/database.go
NathanMcCauley: ah interesting, that's a good check

server/handlers/default.go
NathanMcCauley: i see, we have to do a cast b/c we don't know the type coming out of the ctx?

server/handlers/default.go
NathanMcCauley: i mean, if they're probing and generate a malformed request. would love to see this code be pretty strict in what it accepts

server/handlers/default.go
NathanMcCauley: ok

server/handlers/default.go
NathanMcCauley: its gonna cause us pain if our API returns success for non-sense requests :(

NathanMcCauley:
LGTM

server/handlers/default.go
endophage: It's a very academic question: is the null object malformed? (fyi Tony Hoare who invented the idea of a "null reference" thinks it was a bad idea, so you're in good company) We can add an error for the empty request if you think this behaviour will be a pain.

NathanMcCauley: as written it won't just accept empty requests, but empty multiparts interspersed with correct ones. image some client code that correctly reads three files but fails to read one and just inserts something empty. it'll succeed quietly and update the other items. i'd make it fail the request.

endophage: Ah good example. Yep, that case would definitely be contrary to the intention of this handler. Will fix.

endophage: Added an error if role is empty. Also attempt to trim ".json" suffix from filename. Finally, check that the role is valid using `data.ValidRole`. I debated requiring filenames to be passed with .json, or as the naked role name, now either way will work, and we confirm that the resulting role is one we should accept.

server/handlers/default.go
endophage: Ah yeah, `Context.Value` return the `interface{}` type.

PR #45, opened 2015-07-12
Adding ECDSA support

diogomonica:
Added support for ECDSA keys.


client/cli_crypto_service.go
```diff
@@ -10,62 +11,77 @@ import (
...
+		keyName := filepath.Join(ccs.gun, keyid)
 
 		var privKey *data.PrivateKey
 		var err error
 		var method string
 
-		// Read PrivateKey from file
+		// Read PrivateKey from file.
+		// TODO(diogo): This assumes both that only root keys are encrypted and
+		// that encrypted keys are always X509 encoded
```
aaronlehmann: I do think we'll want to change this eventually - I can imagine use cases where the root keys will not have a passphrase (i.e. non-interactive use). I think it would be a good idea to add a flag to the `genericCryptoService` structure to use instead of using the passphrase for deciding whether it's a root key or not.

client/cli_crypto_service.go
```diff
@@ -95,9 +120,8 @@ func (ccs *CryptoService) Sign(keyIDs []string, payload []byte) ([]data.Signatur
 	return signatures, nil
 }
 
-func sign(privKey *data.PrivateKey, hash crypto.Hash, hashed []byte) ([]byte, error) {
-	// TODO(diogo): Implement support for ECDSA.
-	if strings.ToLower(privKey.Cipher()) != "rsa" {
+func rsaSign(privKey *data.PrivateKey, hash crypto.Hash, hashed []byte) ([]byte, error) {
+	if strings.ToUpper(privKey.Cipher()) != "RSA" {
```
aaronlehmann: I wonder if it would make sense to have `Cipher` do the `ToUpper` so we don't have the risk of forgetting to transform to uppercase for a string comparison.

I also think it would be a good idea to make the algorithms string constants we can refer to (something like `signer.AlgoRSA`) to guard against incorrect character case or typos, but this is more of a cosmetic thing. It would also help prevent confusion between cryptographic algorithms and signing methods.

client/client_test.go
```diff
@@ -18,6 +18,8 @@ import (
...
...RwO4U6eVY7V9GpWiv6gOJTHXYFxU/aL0Ty3eQRxwy9tyVRo8EJz5pRex+e6ws1M+jLOviYqW4VocxQ8dZYd+zBvQfWmRfah7XXJ/HPUx2I05zrmR7VbGX6Bu4g5w3KnyIO61gfyQNKF2bm2Q3yblfupx3URvX0bl180R/+QN2Aslr4zxULFE6b+qJqBydrztq+AAP3WmskRxGa6irFnKxkspJqUpQN1mFselj6iQrzAcwkRPoCw0RwCCMq1/OOYvQtgxTJcO4zDVlbw54PvnxPZtcCWw7fO8oZ2Fvo2SDo75CDOATOGaT4Y9iqECgYEAzWZSpFbN9ZHmvq1lJQg//jFAyjsXRNn/nSvyLQILXltz6EHatImnXo3v+SivG91tfzBI1GfDvGUGaJpvKHoomB+qmhd8KIQhO5MBdAKZMf9fZqZofOPTD9xRXECCwdi+XqHBmL+l1OWz+O9Bh+Qobs2as/hQVgHaoXhQpE0NkTcCgYEA/Tjf6JBGl1+WxQDoGZDJrXoejzG9OFW19RjMdmPrg3t4fnbDtqTpZtCzXxPTCSeMrvplKbqAqZglWyq227ksKw4p7O6YfyhdtvC58oJmivlLr6sFaTsER7mDcYce8sQpqm+XQ8IPbnOk0Z1l6g56euTwTnew49uy25M6U1xL0P8CgYEAxEXv2Kw+OVhHV5PX4BBHHj6we88FiDyMfwM8cvfOJ0datekf9X7ImZkmZEAVPJpWBMD+B0J0jzU2b4SLjfFVkzBHVOH2Ob0xCH2MWPAWtekin7OKizUlPbW5ZV8b0+Kq30DQ/4a7D3rEhK8UPqeuX1tHZox1MAqrgbq3zJj4yvcCgYEAktYPKPm4pYCdmgFrlZ+bA0iEPf7Wvbsd91F5BtHsOOM5PQQ7e0bnvWIaEXEad/2CG9lBHlBy2WVLjDEZthILpa/h6e11ao8KwNGY0iKBuebT17rxOVMqqTjPGt8CuD2994IcEgOPFTpkAdUmyvG4XlkxbB8F6St17NPUB5DGuhsCgYA//Lfytk0FflXEeRQ16LT1YXgV7pcR2jsha4+4O5pxSFw/kTsOfJaYHg8StmROoyFnyE3sg76dCgLn0LENRCe5BvDhJnp5bMpQldG3XwcAxH8FGFNY4LtV/2ZKnJhxcONkfmzQPOmTyedOzrKQ+bNURsqLukCypP7/by6afBY4dA=="}}`
```
aaronlehmann: Looks like we aren't using `timestampKeyJSON` anymore. I guess this is because the timestamp key is hardcoded to ECDSA right now. Could you please add a TODO comment saying to re-add test coverage with the RSA timestamp key once it's possible to use RSA for a timestamp key?

trustmanager/x509utils.go
```diff
@@ -125,8 +95,19 @@ func fingerprintCert(cert *x509.Certificate) CertID {
...
 	pemdata := pem.EncodeToMemory(&block)
 
+	keyType := ""
+	switch cert.PublicKeyAlgorithm {
+	case x509.RSA:
+		keyType = "RSA"
+	case x509.ECDSA:
+		keyType = "ECDSA"
+	default:
+		logrus.Debug("error while fingerprinting certificate. Got Unknown key type.")
```
aaronlehmann: Let's return an explicit error here.

client/cli_crypto_service.go
endophage: +1 for the string constants.

Having Cipher do the case conversion seems like it would be forcing a convention on everyone. Whatever somebody sets as the type when creating a key should be what they get back when calling Cipher.

trustmanager/x509utils.go
diogomonica: I had to rework a ton of things that expected this method to never error out, but done.

client/client_test.go
diogomonica: I tried. After 1 hour of debugging turns out that this is happening:

- We create two signers, one for the root key and one for other keys
- We indiscriminately call both signers with all keys and what works works (if signatures get appended great, if not, tough luck)
- We are able to handle the mode of Root RSA and targets/snapshots ECDSA, and root ECDSA and targets/snapshots ECDSA because they have different signers and are always aligned on the key type.
- Now the timestamps key comes in. Since the timestamp key is generated remotely, it has to be aligned with whatever the snapshot/targets are using. This means that if we have an RSA timestamp key, and generate ECDSA targets/snapshots, we will always fail because the cryptoservice that has access to that key is an ECDSA cryptoservice. The opposite is also true.

Currently in the code we're hardcoding ECDSA targets and snapshot keys.

client/cli_crypto_service.go
diogomonica: Agreed.

diogomonica:
@endophage @aaronlehmann addressed all the comments except refactoring the passphrase checking. We can do that tomorrow.

Also, need review of the remainer of https://github.com/endophage/gotuf/pull/9 to then be able to `godep save` on this branch and merge to master.

PR #46, opened 2015-07-13
Use uuid library from distribution

dmcgowan:
Distribution updated context library to use its own uuid library which does not panic on entropy exhaustion.
Updated to use latest context library from distribution.
Updated auth to match context and uuid library version.

Needed for https://github.com/docker/docker/pull/14546


endophage:
Ah thanks. I wrote this offline on the plane to Denver. Didn't know there was a uuid lib in distribution already.

LGTM.

PR #48, opened 2015-07-13
refactoring database schema out into its own file

endophage:
This makes it easy for Richard to use to bootstrap our staging and production databases (also for anyone else that wants to run a local db rather than a container)


NathanMcCauley:
LGTM

PR #51, opened 2015-07-13
Add ping endpoint

dmcgowan:
Ping endpoint added so clients can get the list of authentication challenges before making an initial request.

The clients could be coded without a ping endpoint but that requires closer integration of the token auth with the client which does not currently exist. This is the fastest approach to get authentication functional.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -76,6 +76,7 @@ func Run(ctx context.Context, addr, tlsCertFile, tlsKeyFile string, trust signed
 
 	r := mux.NewRouter()
 	// TODO (endophage): use correct regexes for image and tag names
+	r.Methods("GET").Path("/v2/").Handler(hand(handlers.MainHandler))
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -93,3 +95,14 @@ func Run(ctx context.Context, addr, tlsCertFile, tlsKeyFile string, trust signed
 
 	return err
 }
+
+func apiBase(ctx context.Context, w http.ResponseWriter, r *http.Request) *errors.HTTPError {
```
endophage: Can we put this in the handlers package somewhere for consistency.

diogomonica:
@endophage can you take a look again?

dmcgowan:
ping @endophage

endophage:
My error branch will change how the error is returned, but I'll deal with that when I rebase that branch so LGTM.

PR #53, opened 2015-07-14
Import notary signer, refactor notary-signer API again

diogomonica:
- Copied over rufus and renamed everything to notary-signer
- Refactored notary-signer API to only receive Algorithm on CreateKey.

@aaronlehmann: I refactored the notary-signer to remove all instances of Algorithm except on Create. For this to fully work we need persistence in notary-signer. In particular I called out the 6 places in the code where we need to add resolution from keyID to KeyInfo.

`// TODO(diogo): call resolve method from KeyID -> KeyInfo`

This resolution is hidden by notary-signer and allows people to configure different backends for different algorithm types on the backend, and users to not care about the signature type.

We also need to add viper configuration to notary-signer, like we did for notary-server and for notary-cli to receive configuration of what backend systems to use, where the TLS keys come in from etc.

Finally, it would be amazing to figure out if there is a better way than to have two SigningServices that are exactly the same but differ on types (on for PBs and the other one for TUF's keys).

Finally, note that all the tests pass. The reason why the build is red is due to lint failures on generated code.

@NathanMcCauley not sure what your magic was for rufus not to fail with lint on the protobufs generated code, but we need it again.


NathanMcCauley:
I excluded it from the golint with some nasty greps. I'll add it back here.

fixtures/rufus.key
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
aaronlehmann: Should these two keys be renamed so they don't refer to rufus?

signer/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+# Copy the local repo to the expected go path
+COPY . /go/src/github.com/docker/notary
+
+# Install notary-signer
+RUN go get github.com/docker/notary/cmd/notary-signer
+
+EXPOSE 4443
+
+#ENTRYPOINT rufus -cert /go/src/github.com/docker/notary/fixtures/rufus.pem -key /go/src/github.com/docker/notary/fixtures/rufus.key -debug -pkcs11 /usr/lib/x86_64-linux-gnu/opensc-pkcs11.so -pin 123456
+ENTRYPOINT rufus -cert /go/src/github.com/docker/notary/fixtures/rufus.pem -key /go/src/github.com/docker/notary/fixtures/rufus.key -debug -pkcs11 $LIBDIR/libsofthsm2.so -pin 1234
```
aaronlehmann: I think this needs to be updated to refer to notary-signer instead of rufus.

fixtures/rufus.key
diogomonica: Yes, I need to generate new certificates.

signer/Dockerfile
diogomonica: when I generate new certs I can fix this.

PR #54, opened 2015-07-14
Introduce a KeyStoreManager to abstract management of root and non-root key storage

aaronlehmann:
Do not merge, PR for review only.

This structure encapsulates what used to be "rootKeyStore" and
"privKeyStore". These are being moved out of NotaryRepository, so that
operations like listing keys, importing keys, and exporting keys aren't
tied to a NotaryRepository structure.

Initial draft for review. Will add import/export functionality next if
this looks good. Do not merge yet.

Signed-off-by: Aaron Lehmann <aaron.lehmann@docker.com>

PR #55, opened 2015-07-14
Introduce a KeyStoreManager to abstract management of root and non-root key storage

aaronlehmann:
Do not merge, PR for review only.

This structure encapsulates what used to be "rootKeyStore" and
"privKeyStore". These are being moved out of NotaryRepository, so that
operations like listing keys, importing keys, and exporting keys aren't
tied to a NotaryRepository structure.

Initial draft for review. Will add import/export functionality next if
this looks good. Do not merge yet.


aaronlehmann:
Supersedes https://github.com/docker/notary/pull/54, which was opened against the wrong branch

client/client.go
```diff
@@ -44,9 +45,7 @@ func (err *ErrRepoNotInitialized) Error() string {
 // Default paths should end with a '/' so directory creation works correctly
 const (
 	trustDir       string = "/trusted_certificates/"
-	privDir        string = "/private/"
 	tufDir         string = "/tuf/"
-	rootKeysDir    string = privDir + "/root_keys/"
 	rsaKeySize     int    = 2048 // Used for snapshots and targets keys
 	rsaRootKeySize int    = 4096 // Used for new root keys
```
diogomonica: should these constants be part of keystoremanager?

diogomonica:
@aaronlehmann good direction. Can we go one step further and have caStore and certificateStore also be stored by keystoremanager?

for example, `loadkeys()` should be called on keystoremanager creation. Keystoremanager should be able to validate an x509 certificate (we should refactor ValidateRoot to keystoremanager), etc.

aaronlehmann:
Sounds good. Working on moving caStore, certificateStore, and ValidateRoot into KeyStoreManager.

One issue I ran into is that ValidateRoot uses the GUN for its validation. I think you said earlier that the KeyStoreManager shouldn't need a GUN to be specified when it's created, because the GUN is a repository concept, not a key management concept. What do you think we should do about this?

aaronlehmann:
Maybe ValidateRoot should take another argument specifying the CommonName to validate, and when NotaryRepo uses it it will pass the GUN for this parameter?

aaronlehmann:
Pushed a commit to move caStore and certificateStore, and refactor ValidateRoot into KeyStoreManager. PTAL.

diogomonica:
This looks great.

aaronlehmann:
Closing this and reopening #54, since this is now ready to merge to master.

PR #56, opened 2015-07-14
Change fallback timestamp algorithtm to supported version

dmcgowan:
ECDSA timestamp keys currently not supported by crypto service, use support ED25519.


endophage:
LGTM (it's breaking in master right now)

PR #57, opened 2015-07-14
client side of multi TUF file atomic update

endophage:
Depends on https://github.com/endophage/gotuf/pull/5

Includes #56 


client/client.go
```diff
@@ -420,28 +421,18 @@ func (r *NotaryRepository) Publish(getPass passwordRetriever) error {
 	if err != nil {
 		return err
 	}
-
+	update := make(map[string]json.RawMessage)
```
diogomonica: Not sure if we should make more or less use of RawMessage, but in this specific case it is bleeding a bit into client.go that didn't really have explicit uses of it yet.

NathanMcCauley:
LGTM but the API makes me sad.

client/client.go
endophage: For now it's here. I intend to go through notary and gotuf and as much as possible change them out for `[]byte` but that's time consuming and of minimal impact. The `json` package is already imported.

endophage:
@NathanMcCauley what makes you sad?

endophage:
FYI, now that we have multi-update I'm planning to remove the single file update and make this the only way to update TUF files. Net change in number of HTTP endpoints is then 0. This is a more robust solution than the existing "UpdateHandler" as it requires more strictly structured data.

NathanMcCauley:
It makes me sad that we're passing in a map from string->raw json to gotuf in order to update the remote files.

I wish it took a list of TufFiles where TufFile was something closer to this object: https://github.com/endophage/gotuf/blob/master/data/roles.go#L54

Agreed that its a step forward as its more structure than we had before.

endophage:
What you're calling TufFile is currently represented as the partially parsed `Signed`, and the fully parsed objects `SignedRoot`, `SignedTargets`, `SignedSnapshot`, and `SignedTimestamp`.

It's interesting you reference Role because I'm quite unhappy with that. The code that handles them is ugly, especially root.json roles.

NathanMcCauley:
You're right, it should actually probably be something like this:

```
type TufRepo struct {
	Targets []Target
	Root     Root
	Timestamp      Timestamp
	Snapshot      Snapshot
}
```

Then, json encoding/serialzing would happen inside gotuf.

endophage:
I could simply add a `ToJSON` method on `Signed` if it's the marshalling in the notary client you want to remove. I have a similar `ToSigned` method on the `SignedRoot` etc... structs.

endophage:
I think the real problem is I need to bulk up the gotuf client. Currently we have a gotuf client that handles reading only really (a design I carried over from flynn, although the client has been fully rewritten), and a notary client that handles writing, and mostly just calls the gotuf client for reading.

NathanMcCauley:
Makes sense, I'm fine with spiffying this up later if you are.

endophage:
Yeah, I've been planning to put write functionality into the gotuf client for a while. It just hasn't been critical and at this point is going to be a cross-repo refactor, moving stuff out of notary's client and into gotuf's.

diogomonica:
@endophage any more functionality you want to add here?

endophage:
Just tests. Working on wrapping up the caching in gotuf right now. Then tests here.

PR #59, opened 2015-07-15
KeyStoreManager

aaronlehmann:
This structure encapsulates what used to be "rootKeyStore" "privKeyStore", and the certificate stores. These are being moved out of NotaryRepository, so that operations like listing keys, importing keys, and exporting keys aren't tied to a NotaryRepository structure.

This PR also moves the CryptoService and UnlockedCryptoService to a dedicated cryptoservice package.

Key import/export will be added in a subsequent PR.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/Sirupsen/logrus"
 	"github.com/docker/notary/client/changelist"
+	"github.com/docker/notary/keystoremanager"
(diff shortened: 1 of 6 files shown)


aaronlehmann:
Supersedes #54 and #55

diogomonica:
Congrats, you now have 3 PRs with the same content :v:

aaronlehmann:
GitHub wouldn't let me reopen #54 after the force-push :(

PR #62, opened 2015-07-15
Import/export

aaronlehmann:
Adds key import and export features

There are functions to export all keys (to a zip file), keys by GUN (to a zip file), and a specific root key (as PEM). There are corresponding functions to import from a zip or PEM file.

This PR also moves non-root keys to a `tuf_keys` subdirectory, so that the root and non-root keystores don't overlap on the filesystem.

### DIFF
--- a/keystoremanager/import_export.go
+++ b/keystoremanager/import_export.go
@@ -1,0 +1,147 @@
+package keystoremanager
+
+import (
+	"archive/zip"
+	"crypto/x509"
+	"encoding/pem"
+	"errors"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/docker/notary/trustmanager"
+)
+
+func moveKeysWithNewPassphrase(oldKeyStore, newKeyStore *trustmanager.KeyFileStore, outputPassphrase string) error {
+	// List all files but no symlinks
+	for _, f := range oldKeyStore.ListFiles(false) {
+		fullKeyPath := strings.TrimSpace(strings.TrimSuffix(f, filepath.Ext(f)))
+		relKeyPath := strings.TrimPrefix(fullKeyPath, oldKeyStore.BaseDir())
+		relKeyPath = strings.TrimPrefix(relKeyPath, string(filepath.Separator))
+
+		pemBytes, err := oldKeyStore.Get(relKeyPath)
+		if err != nil {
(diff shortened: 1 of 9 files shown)


keystoremanager/import_export.go
```diff
@@ -0,0 +1,351 @@
...
+	if err != nil {
+		return err
+	}
+
+	_, err = dest.Write(pemBytes)
+	return err
+}
+
+// checkRootKeyIsEncrypted makes sure the root key is encrypted. We have
+// internal assumptions that/ depend on this.
```
NathanMcCauley: type: that/

keystoremanager/import_export.go
```diff
@@ -0,0 +1,351 @@
...
+	pemBytes, err := ioutil.ReadAll(source)
+	if err != nil {
+		return err
+	}
+
+	if err = checkRootKeyIsEncrypted(pemBytes); err != nil {
+		return err
+	}
+
+	err = km.rootKeyStore.Add(keyID, pemBytes)
```
NathanMcCauley: maybe use the same error checking style within this method?

diogomonica:
LGTM

keystoremanager/import_export.go
```diff
@@ -0,0 +1,351 @@
...
+			return err
+		}
+
+		infoHeader, err := zip.FileInfoHeader(fi)
+		if err != nil {
+			return err
+		}
+
+		infoHeader.Name = relKeyPath
+		fileWriter, err := zipWriter.CreateHeader(infoHeader)
```
NathanMcCauley: nit: could you call this `zipFileEntryWriter` instead of `fileWriter`? It'd make it more clear what this writer is being used for.

aaronlehmann:
Addressed Nathan's comments (so far) and repushed.

NathanMcCauley:
If I'm reading this correctly, the password handling seems a little unsafe. If the user calls `import` with the password for the root key or the old (non-export) key passphrase for the targets, snapshots, or timestamps key this will leave those files unencrypted on the import machine.

aaronlehmann:
As we discussed outside this PR, this shouldn't be an issue. Currently, keys other than the root key can't be stored encrypted (except for import/export). If we ever fail to decrypt one of these keys while importing, the import process fails.

NathanMcCauley:
LGTM

PR #63, opened 2015-07-15
fixing database queries

endophage:
This is why I need a way to unit test against a real database. The queries were broken. Had to revert to an earlier form that I'm not particularly happy with but it'll work.


RichardScothern:
Tested on staging. LGTM.

PR #64, opened 2015-07-16
Change notary-signer to use gotuf's key types

aaronlehmann:
This causes notary-signer to use gotuf's Key interface instead of
defining its own redundant types.

We can go further with this in the future by removing the redundant
ED25519 implementation. This would be refactored into the cryptoservice
package, and notary-signer would be changed to use that package's
cryptoservice for key creation and signing operations.

### DIFF
--- a/signer/keys/keys.go
+++ b/signer/keys/keys.go
@@ -3,7 +3,7 @@ package keys
 import (
 	"errors"
 
-	"github.com/agl/ed25519"
+	"github.com/endophage/gotuf/data"
 	"github.com/miekg/pkcs11"
 )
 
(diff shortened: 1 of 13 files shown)


signer/keys/keys.go
```diff
@@ -16,18 +20,53 @@ var (
...
+}
+
+// Public implements a method of the data.Key interface
+func (rsa *HSMRSAKey) Public() []byte {
+	return rsa.public
+}
+
+// Private implements a method of the data.Key interface
+func (rsa *HSMRSAKey) Private() []byte {
+	// TODO(aaronl): get private key bytes from pkcs11?
```
diogomonica: This is not possible. We cant get private keys out of HSM's (that is kind of the point). You can remove the comment

signer/keys/keys.go
```diff
@@ -16,18 +20,53 @@ var (
...
+// Algorithm implements a method of the data.Key interface
+func (rsa *HSMRSAKey) Algorithm() data.KeyAlgorithm {
+	return data.RSAKey
+}
+
+// ID implements a method of the data.Key interface
+func (rsa *HSMRSAKey) ID() string {
+	if rsa.id == "" {
+		pubK := data.NewTUFKey(rsa.Algorithm(), rsa.Public(), nil)
+		data, err := cjson.Marshal(&pubK)
```
diogomonica: Why are we doing this? Why isn't this just `return pubK.ID()` at this point?

aaronlehmann: I copied TUFKey's ID method without thinking too hard about it. That is indeed a better idea.

aaronlehmann:
Re-pushed with comments addressed.

PR #65, opened 2015-07-16
Adding the not yummy kind of tofu

diogomonica:
This is ready to be merged.

The big change is a complete overhaul to ValidateRoot and adding TOFUS  (Trust On First Use over httpS) behavior.

All the other changes are related either because they make the interfaces better to work with, or because they needed better methods to operate on.

Also added tests for multiple different parts of the code base that I kept interacting with and weren't tested yet.


keystoremanager/keystoremanager.go
```diff
@@ -205,27 +209,26 @@ func (km *KeyStoreManager) ValidateRoot(root *data.Signed, dnsName string) error
...
+		// how the ID gets computed.
+		_, err = km.certificateStore.GetCertificateByCertID(leafID)
 		if err == nil && leafCert.Subject.CommonName == dnsName {
-			certs[keyID] = rootSigned.Keys[keyID]
+			validKeys[keyID] = rootSigned.Signed.Keys[keyID]
 		}
 
-		// Check to see if this leafCertificate has a chain to one of the Root CAs
-		// of our CA Store.
+		// Check to see if this leafCertificate has a chain to one of the Root // CAs of our CA Store.
```
NathanMcCauley: newline needed

trustmanager/x509filestore.go
```diff
@@ -213,10 +240,10 @@ func (s X509FileStore) GetVerifyOptions(dnsName string) (x509.VerifyOptions, err
 }
 
 func fileName(cert *x509.Certificate) (string, CertID, error) {
-	keyID, err := fingerprintCert(cert)
+	certID, err := fingerprintCert(cert)
```
NathanMcCauley: this function needs a new name

keystoremanager/keystoremanager.go
NathanMcCauley: Is this supposed to support building a chain that contains an intermediate we've never seen? Seems to require the intermediate to already be in the caStore?

keystoremanager/keystoremanager.go
```diff
@@ -155,79 +162,127 @@ attempts to validate the certificate by first checking for an exact match on
...
-	for _, keyID := range rootSigned.Roles["root"].KeyIDs {
-		// TODO(dlaw): currently assuming only one cert contained in
-		// public key entry. Need to fix when we want to pass in chains.
-		k, _ := pem.Decode([]byte(rootSigned.Keys[keyID].Public()))
-		decodedCerts, err := x509.ParseCertificates(k.Bytes)
+	// Iterate over every keyID for the root role inside of roots.json
+	validKeys := make(map[string]*data.PublicKey)
+	allCerts := make(map[string]*x509.Certificate)
+	logrus.Debugf("found the following root keys in roots.json: %v", rootSigned.Signed.Keys)
+	for _, keyID := range rootSigned.Signed.Roles["root"].KeyIDs {
```
aaronlehmann: Is the "root" entry guaranteed to exist at all times when this function is called? It might be a good idea to check anyway.

keystoremanager/keystoremanager.go
```diff
@@ -155,79 +162,127 @@ attempts to validate the certificate by first checking for an exact match on
...
-		// TODO(diogo): Assuming that first certificate is the leaf-cert. Need to
-		// iterate over all decodedCerts and find a non-CA one (should be the last).
-		leafCert := decodedCerts[0]
 
+		// Get all non-CA certificates in the decoded certificates
+		leafCerts := trustmanager.GetLeafCerts(decodedCerts)
+
+		// If we got no leaf certificates or we got more than one, fail
+		if len(leafCerts) != 1 {
+			logrus.Debugf("error while parsing root certificate with keyID: %s, %v", keyID, err)
```
aaronlehmann: Log message seems wrong? Also, should this be logged at a more visible level than debug?

keystoremanager/keystoremanager.go
```diff
@@ -155,79 +162,127 @@ attempts to validate the certificate by first checking for an exact match on
...
+			}
 		}
 
-		// Check to see if this leafCertificate has a chain to one of the Root CAs
-		// of our CA Store.
-		certList := []*x509.Certificate{leafCert}
-		err = trustmanager.Verify(km.caStore, dnsName, certList)
+		// Check to see if this leafCertificate has a chain to one of the Root
+		// CAs of our CA Store.
+		err = trustmanager.Verify(km.caStore, dnsName, decodedCerts)
```
aaronlehmann: Can we skip this if we did the TOFUS thing instead? Or maybe it's better to look for a chain to a root CA before TOFUS...

keystoremanager/keystoremanager.go
```diff
@@ -155,79 +162,127 @@ attempts to validate the certificate by first checking for an exact match on
...
+		return err
 	}
 
-	_, err = signed.VerifyRoot(root, 0, certs, 1)
+	// VerifyRoot returns a non-nil value if there is a root key rotation happening.
+	// If this happens, we should replace the old root of trust with the new one
+	if newRootKey != nil {
+		logrus.Debugf("got a new root key to rotate to: %s", newRootKey.ID())
+		// Retrieve the certificate associated with the new root key and trust it
+		newRootKeyCert := allCerts[newRootKey.ID()]
```
aaronlehmann: I think we should check to make sure this entry is present in the map. Something could have gone wrong during the loop above where we're supposed to add it to the map.

trustmanager/x509filestore_test.go
```diff
@@ -135,6 +135,52 @@ func TestRemoveCertX509FileStore(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("couldn't parse certificate: %v", err)
+		}
+		err = store.AddCert(cert)
+		if err != nil {
+			t.Fatalf("failed to load certificate: %v", err)
+		}
+	}
+
+	// Number of certificates should be 1 since we added the cert
```
aaronlehmann: Comment doesn't match code

trustmanager/x509filestore_test.go
```diff
@@ -135,6 +135,52 @@ func TestRemoveCertX509FileStore(t *testing.T) {
...
+	// Remove the cert from the store
+	err = store.RemoveAll()
+	if err != nil {
+		t.Fatalf("failed to remove all certificates: %v", err)
+	}
+	// Number of certificates should be 0 since we added and removed the cert
+	numCerts = len(store.GetCertificates())
+	if numCerts != 0 {
+		t.Fatalf("unexpected number of certificates in store: %d", numCerts)
+	}
```
aaronlehmann: If we use testify here, we could write this `assert.Len(t, numCerts, 0, "unexpected number of certificates in store")`. Not worth rewriting it, but I'm a fan of testify and I'm using it in new tests.

trustmanager/x509memstore.go
```diff
@@ -84,8 +84,23 @@ func (s X509MemStore) RemoveCert(cert *x509.Certificate) error {
...
+func (s *X509MemStore) RemoveAll() error {
+
+	for _, cert := range s.fingerprintMap {
+		if err := s.RemoveCert(cert); err != nil {
+			return err
+		}
+	}
+
+	s.fingerprintMap = make(map[CertID]*x509.Certificate)
+	s.nameMap = make(map[string][]CertID)
```
aaronlehmann: Seems like we either need these two lines or the loop above; is there some case where they're both needed?

trustmanager/x509memstore_test.go
```diff
@@ -106,14 +106,55 @@ func TestRemoveCert(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("couldn't parse certificate: %v", err)
+		}
+		err = store.AddCert(cert)
+		if err != nil {
+			t.Fatalf("failed to load certificate: %v", err)
+		}
+	}
+
+	// Number of certificates should be 1 since we added the cert
```
aaronlehmann: Comment doesn't match code

trustmanager/x509filestore.go
jfrazelle: you need a new name

diogomonica: agreed with Jess.

trustmanager/x509filestore_test.go
diogomonica: Let's port that later

diogomonica:
@aaronlehmann addressed your comments and re-based off of master.

client/client_root_validation_test.go
```diff
@@ -0,0 +1,269 @@
...
+	//
+	err = keyStoreManager.ValidateRoot(&testSignedRoot, "docker.com/notary")
+	assert.NoError(t, err)
+
+	//
+	// This call to ValidateRoot will fail since we are passing in a dnsName that
+	// doesn't match the CN of the certificate.
+	//
+	err = keyStoreManager.ValidateRoot(&testSignedRoot, "diogomonica.com/notary")
+	if assert.Error(t, err, "An error was expected") {
```
NathanMcCauley: why is the `if` conditional needed here?

keystoremanager/keystoremanager.go
```diff
@@ -166,79 +173,166 @@ attempts to validate the certificate by first checking for an exact match on
...
 
-		// Check to see if there is an exact match of this certificate.
-		// Checking the CommonName is not required since ID is calculated over
-		// Cert.Raw. It's included to prevent breaking logic with changes of how the
-		// ID gets computed.
-		_, err = km.trustedCertificateStore.GetCertificateByKeyID(leafID)
-		if err == nil && leafCert.Subject.CommonName == dnsName {
-			certs[keyID] = rootSigned.Keys[keyID]
+		// Validate that this leaf certificate has a CN that matches the exact gun
+		if leafCert.Subject.CommonName != dnsName {
```
NathanMcCauley: Why is this called `dnsName` here? Shouldn't it be `gun`?

trustmanager/keyfilestore.go
```diff
@@ -162,3 +162,8 @@ func listKeys(s LimitedFileStore) []string {
 	}
 	return keyIDList
 }
+
+// RemoveKey removes the key from the keyfilestore
+func (s *KeyFileStore) RemoveKey(name string) error {
```
NathanMcCauley: Should this function be added to the KeyStore interface?

trustmanager/keyfilestore_test.go
```diff
@@ -251,3 +251,53 @@ func TestGetDecryptedWithInvalidPassphrase(t *testing.T) {
...
 	}
 }
+
+func TestRemoveKey(t *testing.T) {
+	testName := "docker.com/notary/root"
+	testExt := "key"
+
+	// Temporary directory where test files will be created
+	tempBaseDir, err := ioutil.TempDir("", "notary-test-")
+	if err != nil {
```
NathanMcCauley: Could you update to use testify?

trustmanager/x509filestore.go
NathanMcCauley: method is called `fileName`.

It returns a path and a certificate ID.

Recommend `getRelPathAndCertIDFromx509` or some shiz

NathanMcCauley:
LGTM modulo nits

NathanMcCauley:
none of the nits are bad enough to block merge

trustmanager/x509filestore.go
jfrazelle: You know I was joking right

PR #68, opened 2015-07-16
updating errors to use distribution's errcode package

endophage:
@stevvooe 

## Important

In the interests of updating less code. This PR removes the UpdateHandler. This requires that the atomic update PR is merged before this one.


utils/http_test.go
```diff
@@ -1,30 +1,25 @@
...
 }
 
-func MockBetterErrorHandler(ctx context.Context, w http.ResponseWriter, r *http.Request) *errors.HTTPError {
-	return &errors.HTTPError{
-		HTTPStatus: http.StatusInternalServerError,
-		Code:       9999,
-		Err:        fmt.Errorf("TestError"),
-	}
+func MockBetterErrorHandler(ctx context.Context, w http.ResponseWriter, r *http.Request) errcode.Errors {
+	return errcode.Errors{errors.ErrUnknown.WithDetail("Test Error")}
```
stevvooe: @endophage Try this:

```
return errcode.Errors{
    errors.ErrUnknown.WithDetail("Test Error"),
}
```

@dsymonds Any ideas? This is a slice type that is failing a golint check for unkeyed fields on a struct literal:

```
utils/http_test.go:22: errcode.Errors composite literal uses unkeyed fields
```

endophage: We've fixed this in a different way, but for reference, there does appear to be an issue with `go vet` so I'm leaving this link to the failing CircleCI build https://circleci.com/gh/docker/notary/151

endophage:
@dmcgowan now that it's passing, take a look

errors/errors.go
```diff
@@ -1,22 +1,100 @@
...
+		Code:           1008,
+		Value:          "ERR_NO_CRYPTOSERVICE",
+		Message:        "The server does not have a signing service configured.",
+		Description:    "This error is returned if no signing service has been configured for the server and it has been asked to perform an operation that requires either signing, or key generation.",
+		HTTPStatusCode: http.StatusInternalServerError,
+	})
+	// ErrUnauthorized lint comment
+	ErrUnauthorized = errcode.Register(errGroup, errcode.ErrorDescriptor{
+		Code:           1009,
+		Value:          "ERR_UNAUTHORIZED",
```
stevvooe: No need to declare "ERR" in front of everything. That is redundant.

errors/errors.go
```diff
@@ -1,22 +1,100 @@
...
+	ErrGenericNotFound = errcode.Register(errGroup, errcode.ErrorDescriptor{
+		Code:           1007,
+		Value:          "ERR_GENERIC_NOT_FOUND",
+		Message:        "You have requested a resource that does not exist.",
+		Description:    "This error is returned when the user requests a non-specific resource that is not known to the server.",
+		HTTPStatusCode: http.StatusNotFound,
+	})
+	// ErrNoCryptoService lint comment
+	ErrNoCryptoService = errcode.Register(errGroup, errcode.ErrorDescriptor{
+		Code:           1008,
```
stevvooe: These are automatically assigned. If you define them here, you risk panic.

endophage: Oh, interesting, I didn't realize that.

stevvooe: @endophage Let me know if we need to update the godoc.

errors/errors.go
```diff
@@ -1,22 +1,89 @@
...
+	// ErrUnauthorized lint comment
+	ErrUnauthorized = errcode.Register(errGroup, errcode.ErrorDescriptor{
+		Value:          "UNAUTHORIZED",
+		Message:        "The server does not have a signing service configured.",
+		Description:    "This error is returned if no signing service has been configured for the server and it has been asked to perform an operation that requires either signing, or key generation.",
+		HTTPStatusCode: http.StatusUnauthorized,
+	})
+	// ErrUnknown lint comment
+	ErrUnknown = errcode.Register(errGroup, errcode.ErrorDescriptor{
+		Value:          "GENERAL_UNKNOWN",
```
stevvooe: Just use `errcode.ErrorCodeUnknown`. That is meant to be a common definition.

errors/errors.go
```diff
@@ -1,22 +1,89 @@
...
-	}
-	return fmt.Sprintf("%d: %s", he.Code, msg)
-}
+var (
+	// ErrNoStorage lint comment
+	ErrNoStorage = errcode.Register(errGroup, errcode.ErrorDescriptor{
+		Value:          "NO_STORAGE",
+		Message:        "The server is misconfigured and has no storage.",
+		Description:    "The server requires a storage backend to be configured to be capable of satisfying requests. This error is returned in the case that no such backend has been configured.",
+		HTTPStatusCode: http.StatusInternalServerError,
```
stevvooe: This is the default status. No need to define it.

errors/errors.go
endophage: Can you make it take detail? It would be useful to be able to output the unknown error as part of the message.

errors/errors.go
endophage: It's clearer to be explicit.

endophage:
@stevvooe updated to use ServeJSON

dmcgowan:
LGTM

I would like to nit though about what a variable comment should be and maybe define a better way to do description. "This error returns if" does not need to be in the description, likewise the comment should be more like "<ERROR NAME> returned if ..." and less like "<ERROR NAME> lint comment".

errors/errors.go
stevvooe: Does `errcode.ErrorCodeUnknown.WithDetail(x)` not work?

errors/errors.go
```diff
@@ -1,22 +1,89 @@
...
+	})
+	// ErrMalformedUpload lint comment
+	ErrMalformedUpload = errcode.Register(errGroup, errcode.ErrorDescriptor{
+		Value:          "MALFORMED_UPLOAD",
+		Message:        "The body of your request is malformed.",
+		Description:    "This error is raised when a user uploads new TUF data and we are unable to parse it as multipart/form-data.",
+		HTTPStatusCode: http.StatusBadRequest,
+	})
+	// ErrGenericNotFound lint comment
+	ErrGenericNotFound = errcode.Register(errGroup, errcode.ErrorDescriptor{
```
stevvooe: It looks like we should add notfound, unknown and unauthorized to the `errcode` package to avoid redefinition.

errors/errors.go
endophage: The detail doesn't get output into the message.

endophage: Actually, where does detail every get used? I guess what I wanted was WithArgs, but the current ErrorCodeUnknown still doesn't have a place to output extra content into the message.

errors/errors.go
endophage: That would probably be good. Re. my comments about the current Unknown, they should all take extra content to be output at the end of the message so we can specify what was not found, unknown, or unauthorized.

endophage:
@dmcgowan I didn't want to have to put the lint comments in at all... They should be identical to the Description entries and I didn't want to have to include all the text twice.

errors/errors.go
stevvooe: Detail gets added to the json error response. Typically, this can be some other structure, describing the failure. It isn't meant to augment the error message.

errors/errors.go
```diff
@@ -1,22 +1,89 @@
...
 
-// Error implements the error interface
-func (he *HTTPError) Error() string {
-	msg := ""
-	if he.Err != nil {
-		msg = he.Err.Error()
-	}
-	return fmt.Sprintf("%d: %s", he.Code, msg)
-}
+var (
```
stevvooe: You may be able to put a single comment here.

PR #69, opened 2015-07-16
Fix the staging deployment.

RichardScothern:
Add a service name and ephemeral ports to notary server.
Remove certs as TLS is terminated in the ELB.

Signed-off-by: Richard Scothern <richard.scothern@gmail.com>


endophage:
LGTM

PR #71, opened 2015-07-16
Unify cryptoservice

aaronlehmann:
Add ED25519 support to cryptoservice and x509utils. Add unit tests for cryptoservices that do sign and verify for all three supported algorithms.

Big refactor to make signer use cryptoservices:

- Add MemoryFileStore, a partial FileStore implementation that doesn't persist on disk.

- Create a KeyStore interface that allows pluggable key store types. Use this interface in the cryptoservice implementation.

- Add KeyMemoryStore, which uses MemoryFileStore to provide a KeyStore.

- Add GetKey and DeleteKey functions to cryptoservice.CryptoService.

- Refactor the hardware RSA signing service as a CryptoService.

- Replace custom ed25519 code with cryptoservice.CryptoService.


cryptoservice/crypto_service.go
```diff
@@ -24,16 +25,16 @@ const (
...
 
 // NewCryptoService returns an instance of CryptoService
-func NewCryptoService(gun string, keyStore *trustmanager.KeyFileStore, passphrase string) *CryptoService {
+func NewCryptoService(gun string, keyStore trustmanager.KeyStore, passphrase string) *CryptoService {
 	return &CryptoService{gun: gun, keyStore: keyStore, passphrase: passphrase}
 }
 
 // Create is used to generate keys for targets, snapshots and timestamps
-func (ccs *CryptoService) Create(role string, algorithm data.KeyAlgorithm) (*data.PublicKey, error) {
+func (ccs *CryptoService) Create(role string, algorithm data.KeyAlgorithm) (data.Key, error) {
```
diogomonica: Woudn't *data.PublicKey make this more obvious what the return is?

aaronlehmann: I changed the CryptoService interface so the return types are `data.Key` (an interface) rather than `*data.PublicKey` (pointer to a structure). This makes it possible to implement the hardware signing as a cryptoservice, because it doesn't use the PublicKey type (it needs a PKCS11 object handle, for example).

diogomonica: Got it. A small loss for unification. I still liked the PublicKey here better. It's easier to understand what is going on.

Now just looking at the method return args it isn't obvious to the caller what is expected to get out of this function. Do I get the private key or the full public key?!

cryptoservice/crypto_service.go
```diff
@@ -58,16 +64,29 @@ func (ccs *CryptoService) Create(role string, algorithm data.KeyAlgorithm) (*dat
 	if err != nil {
 		return nil, fmt.Errorf("failed to add key to filestore: %v", err)
 	}
-	return data.PublicKeyFromPrivate(*privKey), nil
+	return data.PublicKeyFromPrivate(privKey), nil
+}
+
+// GetKey returns a key by ID
```
diogomonica: should these operations be done on a keystore directly? Why on the crypto service?

diogomonica: I actually added key removal to keystore on my current branch.

cryptoservice/crypto_service.go
```diff
@@ -58,16 +64,29 @@ func (ccs *CryptoService) Create(role string, algorithm data.KeyAlgorithm) (*dat
...
 
 // Sign returns the signatures for the payload with a set of keyIDs. It ignores
 // errors to sign and expects the called to validate if the number of returned
 // signatures is adequate.
 func (ccs *CryptoService) Sign(keyIDs []string, payload []byte) ([]data.Signature, error) {
-	// Create hasher and hash data
-	hash := crypto.SHA256
-	hashed := sha256.Sum256(payload)
+	var sha256Sum [sha256.Size]byte
+	sha256Computed := false
```
diogomonica: This looks kind of ugly. Might as well do this individually for RSA and ECDSA:

```go
hash := crypto.SHA256
hashed := sha256.Sum256(payload)
```

cryptoservice/crypto_service_test.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+
+	"github.com/docker/notary/trustmanager"
+	"github.com/endophage/gotuf/data"
+	"github.com/endophage/gotuf/signed"
+)
+
+func TestED25519(t *testing.T) {
```
diogomonica: If you're testing create and sign you should also probably test Get and Delete? Also, we don't use the word "Delete" anywhere. I think we always use "Remove" instead.

cryptoservice/crypto_service.go
aaronlehmann: Well, my logic was that CryptoService is supposed to be a SigningService and KeyService:

```go
// CryptoService defines a unified Signing and Key Service as this
// will be most useful for most applications.
type CryptoService interface {
        SigningService
        KeyService
}
```

It seemed weird for a KeyService to only provide a Create method, especially since the description comment says "KeyService provides management of keys locally".

You're right that we could operate on the keystore directly. This would be a bit of a pain though, since notary-signer would have to keep track of CryptoService/KeyStore pairs instead of just keeping track of CryptoServices. I'd rather avoid the busy-work of redoing CryptoServiceIndex to keep track of both, but other than that I don't care much.

cryptoservice/crypto_service.go
aaronlehmann: I didn't want to repeat the hash for each key (for example, if we're signing with many RSA keys at once). But if you think it's cleaner to do it inside the loop, I'm ok with making the change.

cryptoservice/crypto_service_test.go
aaronlehmann: I'll change it to Remove and add more test coverage here.

trustmanager/keyfilestore.go
```diff
@@ -24,11 +40,73 @@ func NewKeyFileStore(baseDir string) (*KeyFileStore, error) {
...
+
+// ListKeys returns a list of unique PublicKeys present on the KeyFileStore.
+// There might be symlinks associating Certificate IDs to Public Keys, so this
+// method only returns the IDs that aren't symlinks
+func (s *KeyFileStore) ListKeys() []string {
+	return listKeys(s)
+}
+
+// NewKeyMemoryStore returns a new KeyMemoryStore which holds keys in memory
+func NewKeyMemoryStore() *KeyMemoryStore {
```
diogomonica: I don't see any tests for this new MemoryFileStore. Did I missed them somewhere?

trustmanager/x509utils.go
```diff
@@ -198,7 +199,26 @@ func ParsePEMPrivateKey(pemBytes []byte, passphrase string) (*data.PrivateKey, e
...
 		}
 
-		tufECDSAPrivateKey, err := ECDSAToPrivateKey(ecdsaPrivKey, data.ECDSAKey)
+		tufECDSAPrivateKey, err := ECDSAToPrivateKey(ecdsaPrivKey)
+		if err != nil {
+			return nil, fmt.Errorf("could not convert ecdsa.PrivateKey to data.PrivateKey: %v", err)
+		}
+
+		return tufECDSAPrivateKey, nil
+	case "ED25519 PRIVATE KEY":
```
diogomonica: Probably worth a comment stating how we're serializing ed25519 keys

trustmanager/keyfilestore.go
aaronlehmann: The notary-signer API tests use it quite heavily. But yes, a dedicated test suite would be good.

cryptoservice/crypto_service.go
diogomonica: Got it. Lets keep it this way for now. Maybe we'll see some other abstraction that makes sense in the future.

cryptoservice/crypto_service.go
diogomonica: That is a good point. Let's keep it this way.

cryptoservice/crypto_service_test.go
diogomonica: It's either now or next week when we try to bring code coverage to > 80% ;)

cryptoservice/crypto_service.go
aaronlehmann: I thought about that, but it would make ED25519 slower than it would be otherwise. If we expect people to use ED25519 in practice, I don't think we should compute a SHA256 sum and then throw it away. If we don't expect it to be used much in practice, that seems okay.

trustmanager/x509utils.go
aaronlehmann: Comment added

aaronlehmann:
I've made most of the suggested changes, except the new test coverage areas that were suggested. I'll work on adding these new tests next.

signer/api/api.go
```diff
@@ -63,34 +64,36 @@ func KeyInfo(sigServices signer.SigningServiceIndex) http.Handler {
 				return
 			}
 		}
+		key := &pb.PublicKey{KeyInfo: &pb.KeyInfo{KeyID: &pb.KeyID{ID: tufKey.ID()}, Algorithm: &pb.Algorithm{Algorithm: tufKey.Algorithm().String()}}, PublicKey: tufKey.Public()}
```
NathanMcCauley: This could be multiline to increase readability.

cryptoservice/crypto_service.go
NathanMcCauley: I'd just move the hash calculations in to the *sign methods.

signer/api/api.go
```diff
@@ -63,34 +64,36 @@ func KeyInfo(sigServices signer.SigningServiceIndex) http.Handler {
...
 		}
 
-		key, err := sigService.CreateKey()
+		tufKey, err := cryptoService.Create("", data.KeyAlgorithm(vars["Algorithm"]))
 		if err != nil {
 			w.WriteHeader(http.StatusInternalServerError)
 			w.Write([]byte(err.Error()))
 			return
 		}
+		key := &pb.PublicKey{KeyInfo: &pb.KeyInfo{KeyID: &pb.KeyID{ID: tufKey.ID()}, Algorithm: &pb.Algorithm{Algorithm: tufKey.Algorithm().String()}}, PublicKey: tufKey.Public()}
```
NathanMcCauley: multiline

signer/api/find_key.go
```diff
@@ -10,10 +12,10 @@ import (
 // FindKeyByID looks for the key with the given ID in each of the
 // signing services in sigServices. It returns the first matching key it finds,
 // or ErrInvalidKeyID if the key is not found in any of the signing services.
-func FindKeyByID(sigServices signer.SigningServiceIndex, keyID *pb.KeyID) (*pb.PublicKey, signer.SigningService, error) {
-	for _, service := range sigServices {
-		key, err := service.KeyInfo(keyID)
-		if err == nil {
+func FindKeyByID(cryptoServices signer.CryptoServiceIndex, keyID *pb.KeyID) (data.Key, signed.CryptoService, error) {
```
NathanMcCauley: maybe add comment explaining why a service is returned?

signer/api/api.go
```diff
@@ -166,24 +167,23 @@ func Sign(sigServices signer.SigningServiceIndex) http.Handler {
...
 		}
 
-		signer, err := sigService.Signer(sigRequest.KeyID)
-		if err == keys.ErrInvalidKeyID {
-			w.WriteHeader(http.StatusNotFound)
-			w.Write([]byte(err.Error()))
-			return
-		} else if err != nil {
-			// We got an unexpected error
+		tufKey := cryptoService.GetKey(sigRequest.KeyID.ID)
```
NathanMcCauley: Why do we have to call GetKey here -- didn't we have it on the return value from `FindKeyByID` above?

cryptoservice/crypto_service.go
aaronlehmann: I thought about this a bit, and I think we should change things so that we have separate data.PublicKey and data.PrivateKey interfaces. I'll work on this if I have time.

cryptoservice/crypto_service.go
aaronlehmann: Moved the hash calculations to *sign methods.

signer/api/api.go
aaronlehmann: Switched to a multiline format here and in other places with similar statements. Good suggestion.

signer/api/find_key.go
aaronlehmann: Comment added.

signer/api/api.go
aaronlehmann: Good point. This was the result of a poor translation of older code. I just fixed this both here and in rpc_api.go.

NathanMcCauley:
LGTM

PR #72, opened 2015-07-17
Update keyfilestore to use passwordRetriever

NathanMcCauley:
Will allow us to have a callback for password protection of all signing keys.

CC @diogomonica @aaronlehmann 


cmd/notary/tuf.go
```diff
@@ -269,56 +265,49 @@ func verify(cmd *cobra.Command, args []string) {
...
-		return nil, err
+		return "", false, err
 	}
 	term.DisableEcho(0, state)
 	defer term.RestoreTerminal(0, state)
 
 	stdin := bufio.NewReader(os.Stdin)
 
-	fmt.Printf("Enter root key passphrase: ")
+	fmt.Printf("Enter %s key passphrase: ", keyName)
```
aaronlehmann: Let's customize the prompt to make it clear whether we're setting a new password or entering a password to unlock an existing key.

cmd/notary/tuf.go
```diff
@@ -269,56 +265,49 @@ func verify(cmd *cobra.Command, args []string) {
...
-	return passphrase, nil
-}
+func passphraseRetriever(keyName string, createNew bool, numAttempts int) (string, bool, error) {
+	fmt.Printf("Retrieving passphrase for key %s: ", keyName)
 
-func getPassphrase(confirm bool) ([]byte, error) {
-	if pass := os.Getenv("NOTARY_ROOT_PASSPHRASE"); pass != "" {
-		return []byte(pass), nil
+	if (numAttempts > 3 && !createNew) {
+		return "", true, errors.New("Too many attempts")
```
aaronlehmann: I think the retry logic should be inside the password verifier, not passphraseRetriever. It would be easy for a call site that needs to retrieve a passphrase to forget to retry in a loop if the password is incorrect. It's also a little weird that the retry limit is enforced in passphraseRetriever when it's not the function actually doing the retries.

trustmanager/keyfilestore.go
```diff
@@ -11,20 +11,27 @@ const (
...
 // hold the keys.
-func NewKeyFileStore(baseDir string) (*KeyFileStore, error) {
+func NewKeyFileStore(baseDir string, passphraseRetriever PassphraseRetriever) (*KeyFileStore, error) {
 	fileStore, err := NewPrivateSimpleFileStore(baseDir, keyExtension)
 	if err != nil {
 		return nil, err
 	}
 
-	return &KeyFileStore{fileStore}, nil
+	return &KeyFileStore{fileStore, passphraseRetriever}, nil
```
aaronlehmann: Suggest named initializers:

```go
return &KeyFileStore{FileStore: fileStore, PassphraseRetriever: passphraseRetriever}, nil
```

trustmanager/keyfilestore.go
```diff
@@ -11,20 +11,27 @@ const (
...
+// for a given named key. If it should be treated as new passphrase (e.g. with
+// confirmation), createNew will be true. Attempts is passed in so that implementers
+// decide how many chances to give to a human, for example.
+type PassphraseRetriever func(keyName string, createNew bool, attempts int) (passphrase string, giveup bool, err error)
+
 // KeyFileStore persists and manages private keys on disk
 type KeyFileStore struct {
 	FileStore
+	PassphraseRetriever
 }
```
aaronlehmann: I didn't realize you could embed a func in a struct like this. It may be a little clearer to give it an explicit name.

trustmanager/keyfilestore.go
```diff
@@ -34,52 +41,65 @@ func (s *KeyFileStore) AddKey(name string, privKey *data.PrivateKey) error {
...
-
-// GetKey returns the PrivateKey given a KeyID
-func (s *KeyFileStore) GetKey(name string) (*data.PrivateKey, error) {
-	keyBytes, err := s.Get(name)
-	if err != nil {
-		return nil, err
+	attempts := 0
+	passphrase := ""
+	giveup := false
+	for (true) {
```
aaronlehmann: `for {`

cmd/notary/tuf.go
aaronlehmann: Er, obviously that's a stupid suggestion because passphraseRetriever doesn't know whether the password is correct or not. And I guess it's useful to have the passphraseRetriever decide how many times the retry should happen (for example, one that checks an environment variable shouldn't retry). So please ignore the comment above.

trustmanager/keyfilestore.go
```diff
@@ -106,54 +96,79 @@ func (s *KeyMemoryStore) ListKeys() []string {
...
 
-	// Gets an unencrypted PrivateKey.
-	privKey, err := ParsePEMPrivateKey(keyBytes, passphrase)
+	// See if the key is encrypted. If its encrypted we'll fail to parse the private key
+	privKey, err := ParsePEMPrivateKey(keyBytes, "")
 	if err != nil {
-		return nil, err
+		// We need to decrypt the key, lets get a passphrase
+		attempts := 0
+		for (true) {
```
aaronlehmann: `for {`

trustmanager/keyfilestore.go
```diff
@@ -106,54 +96,79 @@ func (s *KeyMemoryStore) ListKeys() []string {
...
-	return s.Add(name, pemPrivKey)
+// RemoveKey removes the key from the keystore
+func (s *KeyMemoryStore) RemoveKey(name string) error {
+	return remove(s, name)
 }
 
-func getKey(s LimitedFileStore, name string) (data.PrivateKey, error) {
-	keyBytes, err := s.Get(name)
+
+func addKey(s LimitedFileStore, passphraseRetriever PassphraseRetriever, name string, privKey data.PrivateKey) error {
```
aaronlehmann: We discussed the need to cache decrypted keys. That will address the current problem where we'd prompt for a password for each key.

As far as addKey, I feel like it's best to push the passphrase retrieval logic to the caller and pass in a password. That way, if a caller adds several keys, it can call the PassphraseRetriever and explicitly use that passphrase for each of the keys. The alternative of having this logic in the PassphraseRetriever callback means that the libnotary user needs to keep track of a lot of state to do their own password caching, and it would make it more complicated to use libnotary.

cmd/notary-signer/main.go
```diff
@@ -41,6 +41,13 @@ func init() {
 	flag.BoolVar(&debug, "debug", false, "show the version and exit")
 }
 
+func passphraseRetriever(keyName, alias string, createNew bool, attempts int) (passphrase string, giveup bool, err error) {
+
+	//TODO(mccauley) Read from config once we have locked keys in notary-signer
+	return "", false, nil;
```
aaronlehmann: Remove the `;`

cmd/notary/main.go
```diff
@@ -86,11 +86,13 @@ func init() {
 		fatalf("could not create Certificate X509FileStore: %v", err)
 	}
 
-	privKeyStore, err = trustmanager.NewKeyFileStore(finalPrivDir)
+
+	//TODO(mccauley): Appears unused? Remove it? Or is it here for early failure?
```
aaronlehmann: Weird. I think it should be removed.

cmd/notary/tuf.go
```diff
@@ -270,58 +272,82 @@ func verify(cmd *cobra.Command, args []string) {
...
-}
+func getNotaryPassphraseRetriever() (trustmanager.PassphraseRetriever) {
+	userEnteredTargetsSnapshotsPass := false
+	targetsSnapshotsPass := ""
 
-func getPassphrase(confirm bool) ([]byte, error) {
-	if pass := os.Getenv("NOTARY_ROOT_PASSPHRASE"); pass != "" {
-		return []byte(pass), nil
-	}
+	return func(keyID string, alias string, createNew bool, numAttempts int) (string, bool, error) {
```
aaronlehmann: Interesting way of using a closure to keep track of state. I don't actively dislike it, but since we use a global to store `retriever` anyway, I'm not sure why this is better than just having a global structure that stores the state and replacing `retriever` with a normal function.

Keep in mind that accesses to these state variables will need to be synchronized if it's possible for the retriever to be called by multiple goroutines. I don't think this is ever an issue in the notary client (but I could be wrong). It could become an issue if similar code is used in the Docker engine, for example.

trustmanager/keyfilestore.go
```diff
@@ -106,54 +110,97 @@ func (s *KeyMemoryStore) ListKeys() []string {
...
+	if passphrase != "" {
+		pemPrivKey, err = EncryptPrivateKey(privKey, passphrase)
+		if err != nil {
+			return err
+		}
 	}
 
-	// Convert PEM encoded bytes back to a PrivateKey
-	privKey, err := ParsePEMPrivateKey(keyBytes, "")
+	err = s.Add(name + "." + aliasExtension, []byte(alias))
```
aaronlehmann: I believe `s.Add` will add a .key extension, so the actual filename will end up being `<keyid>.alias.key`, which is misleading. We should probably change things so that the KeyStore adds the .key extension instead of the FileStore. A FileStore should just be about manipulating files, not deciding the filenames.

NathanMcCauley: Yeah, I think I'm going to instead embed the alias in the keyname itself. It'll make it easier to see what's going on when inspecting the directory.

trustmanager/keyfilestore.go
```diff
@@ -9,60 +9,64 @@ import (
...
 // KeyStore is a generic interface for private key storage
 type KeyStore interface {
 	LimitedFileStore
 
-	AddKey(name string, privKey data.PrivateKey) error
+	AddKey(name, alias string, privKey data.PrivateKey) error
 	GetKey(name string) (data.PrivateKey, error)
-	AddEncryptedKey(name string, privKey data.PrivateKey, passphrase string) error
-	GetDecryptedKey(name string, passphrase string) (data.PrivateKey, error)
+	GetKeyAlias(name string) (string, error)
```
aaronlehmann: Conceptually, do we really need to track the "alias" for each key? It seems like we're adding a decent amount of complexity - new files that need to be kept in sync, passed through with imports/exports, etc.

Stepping back, I think the only thing this is used for is to tell the passphraseRetriever what kind of key we're trying to decrypt, so it can check its cache and return the appropriate passphrase. But if we maintained the cache inside notary instead of pushing it out to whatever implements passphrase retrieval, it would be trivial to try all the passphrases we know about, and not have to care what kind of key we want to unlock.

What would you think about replacing the alias stuff with a cache inside the `KeyStore` that keeps track of every passphrase passphraseRetriever has returned and tries those passphrases when it needs to decrypt a key? If one of them works, great. If not, we call passphraseRetriever to prompt the user or check the environment variable. I suspect that would be a lot simpler than storing extra metadata.

trustmanager/keyfilestore.go
```diff
@@ -164,6 +211,6 @@ func listKeys(s LimitedFileStore) []string {
 }
 
 // RemoveKey removes the key from the keyfilestore
-func (s *KeyFileStore) RemoveKey(name string) error {
+func removeKey(s LimitedFileStore, name string) error {
 	return s.Remove(name)
```
aaronlehmann: We would also need to remove the alias file here, right? The way we need to think about stuff like this makes me feel even more strongly that we should avoid adding these extra files (see comment at the top of this file).

cmd/notary/tuf.go
NathanMcCauley: You caught me, I was being clever. There is a slight advantage to this model in that I can eventually use this method for the import/export flow where I create a closure for each of the keystores (one for import, one for export). 

I'm not sure how much the synchronization issues will affect us, as in the worst case the code will ask twice, but I don't imagine code with goroutines will be prompting the user multiple times. Even so, I think a synchronized version of passphraseRetriever could be implemented for that use case.

trustmanager/keyfilestore.go
```diff
@@ -106,54 +110,97 @@ func (s *KeyMemoryStore) ListKeys() []string {
...
+		for {
+			passphrase, giveup, err := passphraseRetriever(name, string(keyAlias), false, attempts)
+			// Check if the passphrase retriever got an error or if it is telling us to give up
+			if giveup || err != nil {
+				return nil, err
+			}
+
+			// Try to convert PEM encoded bytes back to a PrivateKey using the passphrase
+			privKey, err = ParsePEMPrivateKey(keyBytes, passphrase)
+			if err == nil {
```
aaronlehmann: Normally I would suggest checking what type of error is returned so we can distinguish decryption failure from other types of errors, but I've noticed that the PEM functions can return unexpected errors sometimes when you pass the wrong passphrase, so I think what this code is doing right now is actually best.

trustmanager/keyfilestore.go
aaronlehmann: I thought about suggesting that, but wouldn't that complicate retrieving a key? It would mean you can't just open a file with the key ID as the name; you'd have to list the whole directory and figure out which one matches the key ID.

NathanMcCauley: yes, it'll complicate things a bit, sadly

trustmanager/keyfilestore.go
NathanMcCauley: passphraseRetriever needs the context to know what password to ask the user to supply. I don't think we want to have the user to have to enter all of their password or randomly try passwords until one works...

client/client.go
```diff
@@ -300,7 +298,7 @@ func (r *NotaryRepository) GetTargetByName(name string) (*Target, error) {
 
 // Publish pushes the local changes in signed material to the remote notary-server
 // Conceptually it performs an operation similar to a `git rebase`
-func (r *NotaryRepository) Publish(getPass passwordRetriever) error {
+func (r *NotaryRepository) Publish(getPass trustmanager.PassphraseRetriever) error {
```
endophage: If the keyStoreManager already has the PassphraseRetriever, do we need to pass it in here?

client/client.go
```diff
@@ -356,12 +354,11 @@ func (r *NotaryRepository) Publish(getPass passwordRetriever) error {
 
 	// check if our root file is nearing expiry. Resign if it is.
 	if nearExpiry(r.tufRepo.Root) || r.tufRepo.Root.Dirty {
-		passphrase, err := getPass()
```
endophage: Looks like you've removed the one place we were using the PassphraseRetriever so unless I'm missing something, `Publish()` doesn't need to take it as an argument, which seems like it will make lots of people happy :-)

client/client_root_validation_test.go
```diff
@@ -22,6 +22,8 @@ type SignedRSARootTemplate struct {
 	RootPem string
 }
 
+var passphraseRetriever = func(string, string, bool, int) (string, bool, error) { return "passphrase", false, nil }
```
endophage: named arguments please. I have no idea what the `string, string, bool, int` represent. Would probably be good to name the return arguments as well.

endophage: Also, any reason it's a var, rather than `func passphraseRetriever(string, string ....`?

cmd/notary-server/config.json
```diff
@@ -3,10 +3,7 @@
 		"addr": ":4443"
 	},
 	"trust_service": {
-		"type": "remote",
```
endophage: We should leave config.json as is. I think Richard is using if for deployments. dev-config.json should represent an in memory setup for notary-server.

client/client_test.go
```diff
@@ -315,7 +316,7 @@ func testAddListTarget(t *testing.T, rootType data.KeyAlgorithm) {
 	var tempKey data.TUFKey
 	json.Unmarshal([]byte(timestampECDSAKeyJSON), &tempKey)
 
-	repo.KeyStoreManager.NonRootKeyStore().AddKey(filepath.Join(filepath.FromSlash(gun), tempKey.ID()), &tempKey)
+	repo.KeyStoreManager.NonRootKeyStore().AddKey(filepath.Join(filepath.FromSlash(gun), tempKey.ID()), "nonroot", &tempKey)
```
diogomonica: nonroot? Wouldnt this be timestamps?

cmd/notary-server/config.json
diogomonica: Did you commit this intentionally?

circle.yml
```diff
@@ -49,7 +49,7 @@ test:
         pwd: $BASE_STABLE
 
   # VET
-    - gvm use stable && go vet ./...:
+    - gvm use stable && test -z "$(go tool vet -printf=false . 2>&1 | grep -v Godeps/_workspace/src/ | tee /dev/stderr)":
```
diogomonica: why these changes to circle CI?

cmd/notary/tuf.go
```diff
@@ -14,12 +14,19 @@ import (
...
 	"github.com/docker/docker/pkg/term"
 	notaryclient "github.com/docker/notary/client"
+	"github.com/docker/notary/trustmanager"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
 
 // FIXME: This should not be hardcoded
-const hardcodedBaseURL = "https://notary-server:4443"
+const hardcodedBaseURL = "http://notary-server:4443"
```
diogomonica: not sure if I agree with the hardcodedBaseURL going to http

cmd/notary/tuf.go
```diff
@@ -82,7 +89,8 @@ func tufAdd(cmd *cobra.Command, args []string) {
 	targetName := args[1]
 	targetPath := args[2]
 
-	repo, err := notaryclient.NewNotaryRepository(viper.GetString("baseTrustDir"), gun, hardcodedBaseURL, getInsecureTransport())
+	repo, err := notaryclient.NewNotaryRepository(viper.GetString("baseTrustDir"), gun, hardcodedBaseURL,
```
diogomonica: do we want to merge getInsecureTransport?

cryptoservice/crypto_service_test.go
```diff
@@ -17,11 +17,13 @@ func TestCryptoService(t *testing.T) {
 	}
 }
 
+var passphraseretriever = func(string, string, bool, int) (string, bool, error) { return "", false, nil }
```
diogomonica: camel case

keystoremanager/import_export.go
```diff
@@ -75,43 +75,27 @@ func (km *KeyStoreManager) ImportRootKey(source io.Reader, keyID string) error {
...
-			if err != nil {
-				return err
-			}
-			err = newKeyStore.AddEncryptedKey(relKeyPath, privKey, outputPassphrase)
-		} else {
-			// Encrypted key - pass it through without
-			// decrypting
-			err = newKeyStore.Add(relKeyPath, pemBytes)
-		}
+		err = newKeyStore.AddKey(f, alias, pemBytes)
```
diogomonica: how do you know if the current key is encrypted or not, and if you have to encrypt it before storing?

keystoremanager/import_export_test.go
```diff
@@ -79,17 +82,21 @@ func TestImportExportZip(t *testing.T) {
 	passphraseByFile := make(map[string]string)
 
 	// Add non-root keys to the map. These should use the new passphrase
-	// because they were formerly unencrypted.
-	privKeyList := repo.KeyStoreManager.NonRootKeyStore().ListFiles(false)
+	// because the passwords were chosen by the newPassphraseRetriever.
```
diogomonica: again, if someone doesn't choose a passphrase for targets, we still want to encrypt it at rest on export.

circle.yml
NathanMcCauley: protobuf failures due to their errorf style. I was getting it on my branch, but somehow it isn't happening on master

keystoremanager/import_export_test.go
```diff
@@ -79,17 +82,21 @@ func TestImportExportZip(t *testing.T) {
...
+	privKeyList := repo.KeyStoreManager.NonRootKeyStore().ListKeys()
 	for _, privKeyName := range privKeyList {
-		relName := strings.TrimPrefix(privKeyName, tempBaseDir+string(filepath.Separator))
-		passphraseByFile[relName] = exportPassphrase
+		alias, err := repo.KeyStoreManager.NonRootKeyStore().GetKeyAlias(privKeyName)
+		if err != nil {
```
diogomonica: `assert.NoError` instead

client/client_test.go
NathanMcCauley: nonroot = targets, timestamps, and snapshots

cmd/notary-server/config.json
NathanMcCauley: i'll revert

keystoremanager/import_export_test.go
```diff
@@ -224,15 +222,21 @@ func TestImportExportGUN(t *testing.T) {
 
 	// Add keys non-root keys to the map. These should use the new passphrase
 	// because they were formerly unencrypted.
-	privKeyList := repo.KeyStoreManager.NonRootKeyStore().ListFiles(false)
+	privKeyList := repo.KeyStoreManager.NonRootKeyStore().ListKeys()
 	for _, privKeyName := range privKeyList {
-		relName := strings.TrimPrefix(privKeyName, tempBaseDir+string(filepath.Separator))
-		passphraseByFile[relName] = exportPassphrase
+		alias, err := repo.KeyStoreManager.NonRootKeyStore().GetKeyAlias(privKeyName)
+		if err != nil {
```
diogomonica: `assert.NoError`

keystoremanager/keystoremanager.go
```diff
@@ -150,19 +150,20 @@ func (km *KeyStoreManager) GenRootKey(algorithm, passphrase string) (string, err
...
+	km.rootKeyStore.AddKey(privKey.ID(), "root", privKey)
 
 	return privKey.ID(), nil
 }
 
-// GetRootCryptoService retreives a root key and a cryptoservice to use with it
-func (km *KeyStoreManager) GetRootCryptoService(rootKeyID, passphrase string) (*cryptoservice.UnlockedCryptoService, error) {
-	privKey, err := km.rootKeyStore.GetDecryptedKey(rootKeyID, passphrase)
+// GetRootCryptoService retrieves a root key and a cryptoservice to use with it
+// TODO(mccauley): remove this as its no longer needed once we have key caching in the keystores
```
diogomonica: We are not going to need cryptoservices?

client/client.go
NathanMcCauley: you're right, it can be removed

trustmanager/filestore.go
```diff
@@ -172,6 +172,10 @@ func (f *SimpleFileStore) list(path string, symlinks bool) []string {
 		matched, _ := filepath.Match("*"+f.fileExt, fi.Name())
 
 		if matched {
+			fp, err = filepath.Rel(path, fp)
```
diogomonica: comment?

client/client.go
NathanMcCauley: we're still going to need one passed in during NewNotaryRepository ;)

trustmanager/keyfilestore.go
```diff
@@ -4,64 +4,69 @@ import (
...
 
+	"errors"
+	"fmt"
 	"github.com/endophage/gotuf/data"
 )
 
 const (
-	keyExtension = "key"
+	keyExtension   = "key"
+	aliasExtension = "alias"
```
diogomonica: we we still have an alias extension?

trustmanager/keyfilestore.go
```diff
@@ -106,64 +110,122 @@ func (s *KeyMemoryStore) ListKeys() []string {
...
+		passphrase, giveup, err = passphraseRetriever(name, alias, true, attempts)
+		if err != nil {
+			attempts++
+			continue
+		}
+		if giveup {
+			return errors.New("obtaining passphrase failed")
+		}
+		if attempts > 10 {
+			return errors.New("sanity check on number of passphrase attempts exceeded")
```
diogomonica: remove "sanity check on"

cmd/notary/tuf.go
NathanMcCauley: i was thinking it'd best for the ci builds, but we can have that be its own config

cmd/notary/tuf.go
NathanMcCauley: its already on master, this is just a new line and param addition

trustmanager/keyfilestore.go
```diff
@@ -106,64 +110,122 @@ func (s *KeyMemoryStore) ListKeys() []string {
...
-}
-
-func getKey(s LimitedFileStore, name string) (data.PrivateKey, error) {
-	keyBytes, err := s.Get(name)
-	if err != nil {
-		return nil, err
+	attempts := 0
+	passphrase := ""
+	giveup := false
+	for {
```
endophage:
```go
for ; ; attempts++ {
```

cryptoservice/crypto_service_test.go
NathanMcCauley: fixed

trustmanager/keyfilestore.go
```diff
@@ -106,64 +110,122 @@ func (s *KeyMemoryStore) ListKeys() []string {
...
+		passphrase, giveup, err = passphraseRetriever(name, alias, true, attempts)
+		if err != nil {
+			attempts++
+			continue
+		}
+		if giveup {
+			return errors.New("obtaining passphrase failed")
+		}
+		if attempts > 10 {
```
endophage: If this has the potential to ask the user for the password every time passphraseRetriever is called, 10 seems like an abusive number of attempts.

keystoremanager/import_export.go
NathanMcCauley: Its now the user's decision on a case by case basis, as the callback will get called and they'll get to decide if it should be encrypted. We can write a callback method that guarantees encryption on export if we want.

keystoremanager/keystoremanager.go
NathanMcCauley: we're not going to need to the unlocked crypto service once i implement key caching in the keystores, b/c they'll just transparently be retrievable

trustmanager/keyfilestore.go
```diff
@@ -106,64 +110,122 @@ func (s *KeyMemoryStore) ListKeys() []string {
...
 	if err != nil {
 		return nil, err
 	}
 
+	// See if the key is encrypted. If its encrypted we'll fail to parse the private key
+	privKey, err := ParsePEMPrivateKey(keyBytes, "")
+	if err != nil {
+		// We need to decrypt the key, lets get a passphrase
+		attempts := 0
+		for {
```
endophage: `for attempts := 0; ; attempts++ {`

trustmanager/keyfilestore.go
```diff
@@ -106,64 +110,122 @@ func (s *KeyMemoryStore) ListKeys() []string {
...
-	if err != nil {
-		return nil, err
+	attempts := 0
+	passphrase := ""
+	giveup := false
+	for {
+		passphrase, giveup, err = passphraseRetriever(name, alias, true, attempts)
+		if err != nil {
+			attempts++
+			continue
```
endophage: This will go into an infinite loop at the moment if passphraseRetriever is consistently erroring because the `attempts > 10` check is never reached.

trustmanager/keyfilestore.go
NathanMcCauley: nope, vestigial

trustmanager/keyfilestore.go
```diff
@@ -106,64 +110,122 @@ func (s *KeyMemoryStore) ListKeys() []string {
...
+	if err != nil {
+		// We need to decrypt the key, lets get a passphrase
+		attempts := 0
+		for {
+			passphrase, giveup, err := passphraseRetriever(name, string(keyAlias), false, attempts)
+			// Check if the passphrase retriever got an error or if it is telling us to give up
+			if giveup || err != nil {
+				return nil, errors.New("obtaining passphrase failed")
+			}
+			if attempts > 10 {
```
endophage: Same comment as above, if user might get asked for password on every call, 10 attempts seems abusive.

trustmanager/keyfilestore_test.go
```diff
@@ -10,9 +11,18 @@ import (
 	"testing"
 )
 
+var passphraseRetriever = func(keyID string, alias string, createNew bool, numAttempts int) (string, bool, error) {
```
endophage: Need a test with a passphraseRetriever that only returns errors to test attempt cap.

trustmanager/keyfilestore.go
NathanMcCauley: 10 is just the sanity check where we'll give up unconditionally, see the definition for PassphraseRetriever -- the number of requests so far gets passed in and the callback gets to decide how many chances the user gets.

trustmanager/keyfilestore.go
NathanMcCauley: yup, that'll be better

trustmanager/keyfilestore_test.go
NathanMcCauley: will do

trustmanager/keyfilestore.go
endophage: With the change to the for statement above, this should probably be 
```go
if err == nil {
    break
} 
```

trustmanager/keyfilestore_test.go
NathanMcCauley: done, see TestGetDecryptedWithConsistentlyInvalidPassphrase

client/client.go
diogomonica: Yeah, but publish makes less sense. I think this is a good thing. Passwords are tied to NotaryRepositories.

cmd/notary-server/config.json
diogomonica: Actually I would like to discuss this. I think the master branch of notary should work with dev out of the box. You docker-compose up and run notary.

Production is a different beast, and we shouldn't be surprised if we have to change the configs for it to work there.

keystoremanager/import_export.go
diogomonica: @aaronlehmann

cmd/notary/main.go
```diff
@@ -86,11 +86,12 @@ func init() {
 		fatalf("could not create Certificate X509FileStore: %v", err)
 	}
 
-	privKeyStore, err = trustmanager.NewKeyFileStore(finalPrivDir)
+	//TODO(mccauley): Appears unused? Remove it? Or is it here for early failure?
+	privKeyStore, err = trustmanager.NewKeyFileStore(finalPrivDir,
+		func(string, string, bool, int) (string, bool, error) { return "", false, nil })
```
aaronlehmann: Let's remove this.

cmd/notary/tuf.go
```diff
@@ -270,58 +272,92 @@ func verify(cmd *cobra.Command, args []string) {
...
+			return "", false, errors.New("Passphrase too short")
+		}
+
+		fmt.Printf("Repeat passphrase for new %s key with id %s: ", alias, keyID)
+		confirmation, err := stdin.ReadBytes('\n')
+		fmt.Println()
+		if err != nil {
+			return "", false, err
+		}
+		confirmation = confirmation[0 : len(confirmation)-1]
```
aaronlehmann: Is len(confirmation) guaranteed to be positive here? I checked the docs for ReadBytes and I think the answer is yes, but I'd still err on the side of caution.

cmd/notary/tuf.go
```diff
@@ -270,58 +272,92 @@ func verify(cmd *cobra.Command, args []string) {
...
 
-	if !confirm {
-		return passphrase, nil
-	}
+		passphrase, err := stdin.ReadBytes('\n')
+		fmt.Println()
+		if err != nil {
+			return "", false, err
+		}
+		passphrase = passphrase[0 : len(passphrase)-1]
```
aaronlehmann: Is len(passphrase) guaranteed to be positive here? I checked the docs for ReadBytes and I think the answer is yes, but I'd still err on the side of caution.

cmd/notary/main.go
NathanMcCauley: tried removing and just found that its used by keys.go. I'll remove the comment.

trustmanager/keyfilestore.go
aaronlehmann: Rather than hardcoding 10 in multiple places, let's make this a constant at the top of the file.

trustmanager/keyfilestore.go
```diff
@@ -106,64 +109,120 @@ func (s *KeyMemoryStore) ListKeys() []string {
...
+	for _, file := range files {
+		lastPathSeparator := strings.LastIndexAny(file, "/\\")
+		filename := file[lastPathSeparator+1:]
+
+		if strings.HasPrefix(filename, name) {
+			aliasPlusDotKey := strings.TrimPrefix(filename, name+"_")
+			retVal := strings.TrimSuffix(aliasPlusDotKey, "."+keyExtension)
+			return retVal, nil
+		}
 	}
```
aaronlehmann: I'm not crazy about this way of doing things. It means some operations like removing every key will become O(n^2). It's probably not a practical concern since we shouldn't have many keys at a time.

IMHO it would be a little better to have a subdirectory for each key, so we could list just that subdirectory. But I don't think it's worth spending time changing it at this point.

trustmanager/keyfilestore.go
```diff
@@ -106,64 +109,120 @@ func (s *KeyMemoryStore) ListKeys() []string {
...
-func addEncryptedKey(s LimitedFileStore, name string, privKey data.PrivateKey, passphrase string) error {
-	encryptedPrivKey, err := EncryptPrivateKey(privKey, passphrase)
-	if err != nil {
-		return err
+func getKeyAlias(s LimitedFileStore, keyID string) (string, error) {
+	files := s.ListFiles(true)
+	name := strings.TrimSpace(strings.TrimSuffix(filepath.Base(keyID), filepath.Ext(keyID)))
+
+	for _, file := range files {
+		lastPathSeparator := strings.LastIndexAny(file, "/\\")
```
aaronlehmann: I suggest using using filepath.Separator at the second argument. That will avoid weird misbehavior if a filename somehow contains a '\' on UNUX.

keystoremanager/import_export.go
```diff
@@ -275,39 +253,26 @@ func (km *KeyStoreManager) ImportKeysZip(zipReader zip.Reader, passphrase string
...
 			return err
 		}
 
-		block, _ := pem.Decode(pemBytes)
-		if block == nil {
-			return ErrNoValidPrivateKey
-		}
-
-		if x509.IsEncryptedPEMBlock(block) {
-			return ErrNonRootKeyEncrypted
```
aaronlehmann: ErrNonRootKeyEncrypted is now unused and should be deleted

trustmanager/keyfilestore.go
aaronlehmann: I noticed that everywhere we call GetKeyAlias is right after a call to GetKey. It might simplify things to make GetKey return the alias as another return value (since it looks up the alias anyway). Then we wouldn't need GetKeyAlias.

aaronlehmann:
The import/export changes look correct to me, but I want to make sure everyone's onboard with the UX implications:

* When we export all keys, we'll prompt for new passwords and encrypt the exported keys with those passwords. When we import keys, we'll import them in the form they were exported, so they'll use the passphrases that were entered at export time. Does this makes sense? The asymmetry seems a little weird to me, but on the other hand, it would be awkward to change the passwords twice when exporting the keys from one repo and importing them into another repo.

* Unlike the bulk import/export functions to export/import a single root key don't reencrypt it. Its passphrase would be preserved through the export/import process. Do we want to change the root key functions so they behave the same was as the other functions?

cmd/notary-server/config.json
endophage: dev should still run the end to end services. There's not much difference functionally if the server is storing timestamp keys in memory, or if the signer is.

endophage:
@aaronlehmann in regards to your first bullet point, I think this is the right way to do it. It makes it much simpler to deploy keys to servers as they are now just files dropped in place, there is no command that needs to be run.

I think it would be good to enable root keys to be encrypted with a different password but I don't consider it critical. We can add an issue to do that and it can come as a feature at some later point.

NathanMcCauley:
@aaronlehmann we can land UX changes for import/export after freeze, we need this callback for docker integration

aaronlehmann:
I noticed that the cmd/notary-server/config.json change got merged along with this. Does that need to be reverted?

PR #80, opened 2015-07-17
Add a unit test for publish

aaronlehmann:
@diogomonica @NathanMcCauley 

This instantiates a temporary server, publishes some targets to it, and
makes sure we can pull back the correct targets from the server.

Also fixes a few problems with the client unit tests, error reporting in
the client, and logging in the server.


diogomonica:
LGTM

PR #83, opened 2015-07-18
Cooking the tofu to support rotations

diogomonica:
Changes the validate root logic to use TUF for rotation.

@NathanMcCauley

### DIFF
--- a/keystoremanager/keystoremanager.go
+++ b/keystoremanager/keystoremanager.go
@@ -34,14 +34,29 @@ const (
 	rsaRootKeySize    = 4096 // Used for new root keys
 )
 
-var (
-	// ErrValidationFail is returned when there is no trusted certificate in any of the
-	// root keys available in the roots.json
-	ErrValidationFail = errors.New("could not validate the path to a trusted root")
-	// ErrRootRotationFail is returned when we fail to do a full root key rotation
-	// by either failing to add the new root certificate, or delete the old ones
-	ErrRootRotationFail = errors.New("could not rotate trust to a new trusted root")
-)
+// ErrValidationFail is returned when there is no valid trusted certificates
+// being served inside of the roots.json
+type ErrValidationFail struct {
+	Reason string
+}
+
+// ErrValidationFail is returned when there is no valid trusted certificates
+// being served inside of the roots.json
+func (err ErrValidationFail) Error() string {
+	return fmt.Sprintf("could not validate the path to a trusted root: %s", err.Reason)
+}
+
+// ErrRootRotationFail is returned when we fail to do a full root key rotation
+// by either failing to add the new root certificate, or delete the old ones
+type ErrRootRotationFail struct {
+	Reason string
+}
+
+// ErrRootRotationFail is returned when we fail to do a full root key rotation
+// by either failing to add the new root certificate, or delete the old ones
+func (err ErrRootRotationFail) Error() string {
+	return fmt.Sprintf("could not rotate trust to a new trusted root: %s", err.Reason)
+}
 
 // NewKeyStoreManager returns an initialized KeyStoreManager, or an error
 // if it fails to create the KeyFileStores or load certificates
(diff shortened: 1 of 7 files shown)


keystoremanager/keystoremanager.go
```diff
@@ -34,14 +34,29 @@ const (
...
 )
 
-var (
-	// ErrValidationFail is returned when there is no trusted certificate in any of the
-	// root keys available in the roots.json
-	ErrValidationFail = errors.New("could not validate the path to a trusted root")
-	// ErrRootRotationFail is returned when we fail to do a full root key rotation
-	// by either failing to add the new root certificate, or delete the old ones
-	ErrRootRotationFail = errors.New("could not rotate trust to a new trusted root")
-)
```
aaronlehmann: I think having a custom error type here is good, but maybe keep predefined errors for the different reasons so a caller can distinguish between the reasons for the failure without resorting to string comparisons. This would also make the tests cleaner.

keystoremanager/keystoremanager.go
```diff
@@ -185,155 +200,169 @@ We shall call this: TOFUS.
...
+		// Convert all the valid certificates into public keys to do root validation
+		keysToCheck = trustmanager.CertsToKeys(allValidCerts)
+		logrus.Debugf("found no valid certificates for %s, entering TOFU mode", gun)
+	} else {
+		// Convert all the currently trusted certificates into public keys for root validation
+		keysToCheck = trustmanager.CertsToKeys(certsForCN)
+		logrus.Debugf("found valid certificates for %s", gun)
+	}
+
+	fmt.Printf("LEN: %d\n", len(certsForCN))
```
aaronlehmann: Needs to be removed before merge

endophage:
Does cert expiration ever get checked?

endophage:
The 2 ValidateRoot calls look correct to me. I'm happy that this is now honoring the TUF spec regarding accepting a root key rotation.

diogomonica:
@endophage X509FilteredStore doesn't load expired certificates.

endophage:
@diogomonica but what about the new certs that we parse from root.json?

diogomonica:
@endophage 

```go
// AddCert creates a filename for a given cert and adds a certificate with that name
func (s *X509FileStore) AddCert(cert *x509.Certificate) error {
	if cert == nil {
		return errors.New("adding nil Certificate to X509Store")
	}

	// Check if this certificate meets our validation criteria
	if !s.validate.Validate(cert) {
		return &ErrCertValidation{}
	}
	// Attempt to write the certificate to the file
	if err := s.addNamedCert(cert); err != nil {
		return err
	}

	return nil
}
```

endophage:
@diogomonica but don't we only add after we've already used the (possibly expired) cert to validate?

diogomonica:
@endophage good point. Just pushed a new commit with best effort rotation and expiration validation.

keystoremanager/keystoremanager.go
```diff
@@ -34,14 +34,29 @@ const (
...
+	return fmt.Sprintf("could not validate the path to a trusted root: %s", err.Reason)
+}
+
+// ErrRootRotationFail is returned when we fail to do a full root key rotation
+// by either failing to add the new root certificate, or delete the old ones
+type ErrRootRotationFail struct {
+	Reason string
+}
+
+// ErrRootRotationFail is returned when we fail to do a full root key rotation
```
NathanMcCauley: incorrect comment

keystoremanager/keystoremanager.go
```diff
@@ -181,159 +208,202 @@ Note that since we only allow trust data to be downloaded over an HTTPS channel
...
+		// Make sure the certificate is not expired
+		if time.Now().After(cert.NotAfter) {
+			logrus.Debugf("error leaf certificate is expired")
+			continue
+		}
+		validLeafCerts = append(validLeafCerts, cert)
+	}
+
+	if len(validLeafCerts) < 1 {
+		return validLeafCerts, errors.New("no valid leaf certificates found in any of the root keys")
```
NathanMcCauley: return nil instead of validLeafCerts

keystoremanager/keystoremanager.go
```diff
@@ -181,159 +208,202 @@ Note that since we only allow trust data to be downloaded over an HTTPS channel
...
 	}
 
-	logrus.Debugf("found the following root keys in roots.json: %v", rootRoles.KeyIDs)
+	logrus.Debugf("found the following root keys: %v", rootRoles.KeyIDs)
 	// Iterate over every keyID for the root role inside of roots.json
 	for _, keyID := range rootRoles.KeyIDs {
 		// Decode all the x509 certificates that were bundled with this
 		// Specific root key
-		decodedCerts, err := trustmanager.LoadCertBundleFromPEM([]byte(rootSigned.Signed.Keys[keyID].Public()))
+		decodedCerts, err := trustmanager.LoadCertBundleFromPEM([]byte(signedRoot.Signed.Keys[keyID].Public()))
```
NathanMcCauley: nil pointer checks on all dereferences

NathanMcCauley: @endophage should we check for this case on the server and fail requests if any of the cert bundles uploaded fail to parse?

keystoremanager/keystoremanager.go
```diff
@@ -181,159 +208,202 @@ Note that since we only allow trust data to be downloaded over an HTTPS channel
...
+
+func validRootLeafCerts(root *data.SignedRoot, gun string) ([]*x509.Certificate, error) {
+	// Get a list of all of the leaf certificates present in root
+	allLeafCerts, _ := parseAllCerts(root)
+	var validLeafCerts []*x509.Certificate
+
+	// Go through every leaf certificate and check that the CN matches the gun
+	for _, cert := range allLeafCerts {
+		// Validate that this leaf certificate has a CN that matches the exact gun
+		if cert.Subject.CommonName != gun {
```
NathanMcCauley: nil checks

keystoremanager/keystoremanager.go
```diff
@@ -181,159 +208,202 @@ Note that since we only allow trust data to be downloaded over an HTTPS channel
...
+	}
+
+	// Retrieve all the leaf certificates in root for which the CN matches the GUN
+	allValidCerts, err := validRootLeafCerts(signedRoot, gun)
+	if err != nil {
+		logrus.Debugf("error retrieving valid leaf certificates for: %s, %v", gun, err)
+		return &ErrValidationFail{Reason: "unable to retrieve valid leaf certificates"}
+	}
+
+	// If there are certificates for this specific GUN, we already have trusted
```
NathanMcCauley: switch comment for readability

trustmanager/x509utils.go
```diff
@@ -420,6 +430,32 @@ func EncryptPrivateKey(key data.PrivateKey, passphrase string) ([]byte, error) {
...
+		pemdata := pem.EncodeToMemory(&block)
+
+		var keyType data.KeyAlgorithm
+		switch cert.PublicKeyAlgorithm {
+		case x509.RSA:
+			keyType = data.RSAx509Key
+		case x509.ECDSA:
+			keyType = data.ECDSAx509Key
+		default:
+			logrus.Debugf("unknow certificate type found, ignoring")
```
NathanMcCauley: s/unknow/unknown/

keystoremanager/keystoremanager.go
```diff
@@ -181,159 +208,202 @@ Note that since we only allow trust data to be downloaded over an HTTPS channel
...
+		return &ErrValidationFail{Reason: "failed to validate integrity of roots"}
+	}
+
+	// Getting here means A) we had trusted certificates and both the
+	// old and new validated this root; or B) we had no trusted certificates but
+	// the new set of certificates has integrity (self-signed)
+	logrus.Debugf("entering root certificate rotation for: %s", gun)
+
+	var rotationFailure bool
+	// To support root certificate rotation, let trust only the certs present in root
```
NathanMcCauley: this means "in root variable, the tuf file passed in to this function"

keystoremanager/keystoremanager.go
```diff
@@ -181,159 +208,202 @@ Note that since we only allow trust data to be downloaded over an HTTPS channel
...
+	for certID, cert := range certsToRemove(certsForCN, allValidCerts) {
+		logrus.Debugf("removing certificate with certID: %s", certID)
+		err = km.trustedCertificateStore.RemoveCert(cert)
+		if err != nil {
+			logrus.Debugf("failed to remove trusted certificate with keyID: %s, %v", certID, err)
+			rotationFailure = true
+		}
+	}
+
+	// Now we add all the new certificates (even if they arleady already exist)
```
NathanMcCauley: arleady

keystoremanager/keystoremanager.go
```diff
@@ -181,159 +208,202 @@ Note that since we only allow trust data to be downloaded over an HTTPS channel
...
+			// If the error is already exists we don't fail the rotation
+			if _, ok := err.(*trustmanager.ErrCertExists); ok {
+				continue
+			}
+			logrus.Debugf("error adding new trusted certificate for: %s, %v", gun, err)
+			rotationFailure = true
+		}
+		// Getting the CertID for debug only, don't care about the error
+		certID, _ := trustmanager.FingerprintCert(cert)
+		logrus.Debugf("adding trust certificate for with certID %s for %s", certID, gun)
```
NathanMcCauley: s/for with/for/

keystoremanager/keystoremanager.go
endophage: Do you just mean validate that the certs parse?

keystoremanager/keystoremanager.go
```diff
@@ -181,159 +208,202 @@ Note that since we only allow trust data to be downloaded over an HTTPS channel
...
+		// we couldn't check if there are any certificates with this CN already
+		// trusted. Let's take the conservative approach and return a failed validation
+		if _, ok := err.(*trustmanager.ErrNoCertificatesFound); !ok {
+			logrus.Debugf("error retrieving trusted certificates for: %s, %v", gun, err)
+			return &ErrValidationFail{Reason: "unable to retrieve trusted certificates"}
+		}
+	}
+
+	// Retrieve all the leaf certificates in root for which the CN matches the GUN
+	allValidCerts, err := validRootLeafCerts(signedRoot, gun)
```
NathanMcCauley: run filteredFileStore semantics here

keystoremanager/keystoremanager.go
```diff
@@ -181,159 +208,202 @@ Note that since we only allow trust data to be downloaded over an HTTPS channel
...
+
+	// Getting here means A) we had trusted certificates and both the
+	// old and new validated this root; or B) we had no trusted certificates but
+	// the new set of certificates has integrity (self-signed)
+	logrus.Debugf("entering root certificate rotation for: %s", gun)
+
+	var rotationFailure bool
+	// To support root certificate rotation, let trust only the certs present in root
+	// First, we delete all the old certificates that aren't present in the root
+	for certID, cert := range certsToRemove(certsForCN, allValidCerts) {
```
NathanMcCauley: perform remove after addition

keystoremanager/keystoremanager.go
```diff
@@ -168,12 +183,24 @@ func (km *KeyStoreManager) GetRootCryptoService(rootKeyID, passphrase string) (*
...
+certificates for this particular GUN, we check the integrity of the root by
+making sure that it is validated by itself. This means that we will attempt to
+validate the root data with the certificates that are included in the root keys
+themselves.
 
-Currently this method operates on a Trust On First Use (TOFU) model: if we
+If this last steps succeeds, we attempt to do root rotation, by ensuring that
+we only trust the certificates that are present in the new root.
+
+This mechanism of operation is essentially Trust On First Use (TOFU): if we
```
NathanMcCauley: change the comment below to be 'public PKI' instead of 'web of trust'

NathanMcCauley:
LGTM after nits

keystoremanager/keystoremanager.go
diogomonica: Yeah, reject any data that doesn't have parsable certificates.

keystoremanager/keystoremanager.go
aaronlehmann: cert.Subject can't be nil here because it's not a pointer, and cert can't be nil here either.

trustmanager/x509utils.go
```diff
@@ -189,6 +187,18 @@ func GetLeafCerts(certs []*x509.Certificate) []*x509.Certificate {
...
 }
 
+// GetIntermediateCerts parses a list of x509 Certificates and returns all of the
+// ones marked as a CA, to be used as intermediates
+func GetIntermediateCerts(certs []*x509.Certificate) (intCerts []*x509.Certificate) {
+	for _, cert := range certs {
+		if !cert.IsCA {
+			continue
+		}
+		intCerts = append(intCerts, cert)
```
aaronlehmann: Cosmetic suggestion (feel free to ignore):

```go
if cert.IsCA {
    intCerts = append(intCerts, cert)
}
```

keystoremanager/keystoremanager.go
```diff
@@ -168,172 +183,239 @@ func (km *KeyStoreManager) GetRootCryptoService(rootKeyID, passphrase string) (*
...
+		return nil, errors.New("no valid leaf certificates found in any of the root keys")
+	}
+
+	return validLeafCerts, nil
+}
 
-	// allCerts will keep a list of all leafCerts that were found, and is used
-	// to aid on root certificate rotation
-	allCerts := make(map[string]*x509.Certificate)
+func parseAllCerts(signedRoot *data.SignedRoot) (map[string]*x509.Certificate, map[string][]*x509.Certificate) {
```
aaronlehmann: Please document what is being returned.

Is it intentional that the intCerts return value is unused (and therefore GetIntermediateCerts is effectively unused)? If so, maybe we should remove both, to simplify the code.

aaronlehmann:
I think the logic makes sense. I'm not an expert on the nuances of TUF key rotation, so @endophage should sign off on this as well. Noted a few nits in specific places.

trustmanager/x509utils.go
diogomonica: good catch

keystoremanager/keystoremanager.go
diogomonica: GetIntermediateCerts is used on line 380:

```go
intermediateCerts := trustmanager.GetIntermediateCerts(decodedCerts)
```

And it is correct that right now we don't use the second return argument of parsAllCerts, because we are not supporting CAs.

keystoremanager/keystoremanager.go
endophage: @aaronlehmann it can't be nil but can it be the empty object so CommonName would be `""`?

endophage:
The order of validation operation and rotation looks correct. I see the check has been added to ensure certificates appearing in the root.json aren't expired before using them to validate the self referential integrity of the root, so LGTM.

PR #87, opened 2015-07-19
removing single file append changelist

endophage:


diogomonica:
Removing code: Love it.

LGTM

PR #100, opened 2015-07-20
Key caching

aaronlehmann:
Adds a cache to the keystores. The purpose is to prevent the user from entering a passphrase multiple times if there are multiple target keys.

Adds a new test for the cache, and updates existing tests that are affected by the cache.

### DIFF
--- a/trustmanager/keyfilestore_test.go
+++ b/trustmanager/keyfilestore_test.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/docker/notary/pkg/passphrase"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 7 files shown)


trustmanager/keyfilestore_test.go
```diff
@@ -364,3 +365,77 @@ func TestRemoveKey(t *testing.T) {
...
+		return "password", false, nil
+	}
+
+	// Create our store
+	store, err := NewKeyFileStore(tempBaseDir, countingPassphraseRetriever)
+	if err != nil {
+		t.Fatalf("failed to create new key filestore: %v", err)
+	}
+
+	privKey, err := GenerateRSAKey(rand.Reader, 512)
```
diogomonica: Lets start using EC keys whenever possible.

NathanMcCauley:
LGTM

diogomonica:
LGTM except don't use RSA keys

aaronlehmann:
LGTM

PR #101, opened 2015-07-20
Move passphrase logic to its own package

dmcgowan:
The logic to retrieve passphrase is generic and may be used by directly by clients.


NathanMcCauley:
LGTM

PR #105, opened 2015-07-20
Use correct regular expression for repository names in HTTP handlers

### LINKED ISSUE #92: Use correct regexes for image and tag names
`server.go` currently does:

```go
	r.Methods("POST").Path("/v2/{imageName:.*}/_trust/tuf/").Handler(hand(handlers.AtomicUpdateHandler, "push", "pull"))
```

We need to change the regular expression to support only valid characters for tags and image names.

aaronlehmann:
Fixes #92


diogomonica:
@endophage PTAL

endophage:
Would be good to import the regex from distribution to ensure future consistency.

aaronlehmann:
I tried importing the distribution/registry/api/v2 package, but this causes a panic because of conflicting error definitions.

thaJeztah:
see https://github.com/docker/distribution/pull/690

aaronlehmann:
I tried the regexps from https://github.com/docker/distribution/pull/690, and they don't seem to work properly.

Here's what I tried:

```go
// RepositoryNameComponentRegexp restricts registry path component names to
// start with at least one letter or number, with following parts able to
// be separated by one period, dash or underscore.
var RepositoryNameComponentRegexp = `[a-zA-Z0-9]+(?:[._-][a-z0-9]+)*`

// RepositoryNameHostnameRegexp restricts the registry hostname component of a repository name to
// start with a component as defined by RepositoryNameComponentRegexp and followed by an optional port.
var RepositoryNameHostnameRegexp = RepositoryNameComponentRegexp + `(?::[0-9]+)?`

// RepositoryNameRegexp builds on RepositoryNameComponentRegexp to allow
// multiple path components, separated by a forward slash.
var RepositoryNameRegexp = `(?:` + RepositoryNameHostnameRegexp + `/` + RepositoryNameComponentRegexp + `/)*` + RepositoryNameComponentRegexp
```

This causes a test to fail with 404 errors. It's hard for me to follow exactly what the regexps are doing, but it looks like RepositoryNameRegexp is set up to allow multiple "<hostname>/<component>" matches instead of multiple components. When I tested the regexp with regexr.com, I found that:

* docker.com/notary does not match
* docker.com/notary/notary does match
* docker.com/notary/notary/notary does not match

The exact regexp I'm testing is `(?:[a-zA-Z0-9]+(?:[._-][a-z0-9]+)*(?::[0-9]+)?/[a-zA-Z0-9]+(?:[._-][a-z0-9]+)*/)*[a-zA-Z0-9]+(?:[._-][a-z0-9]+)*`

Unless it turns out that I'm doing something wrong here, it seems like the regexps in that PR are wrong, or are trying to do something different from what we're doing here.

thaJeztah:
Strange. Must admit I haven't looked into them thoroughly, but I knew that's the bit they were working on at the moment. Perhaps @stevvooe is able to have a quick look (or add a suggestion which regexp should be used)

Using a standard regexp would be great (we had some issues in the past where regexes in Docker and Distribution got out of sync )

stevvooe:
@aaronlehmann These have been in production since April. Modify [`routes_test.go`](https://github.com/docker/distribution/blob/master/registry/api/v2/routes_test.go#L31) to test these routes: they work fine.

Here is an [example](https://github.com/docker/distribution/blob/master/registry/api/v2/descriptors.go#L365) of defining the route.

aaronlehmann:
Pushed a new version that imports github.com/docker/distribution/registry/api/v2 to use its regexp. Thanks @stevvooe for explaining the import conflict and how to resolve it.

endophage:
Needs godep updates.

endophage:
It looks like the deps were updated with a `godep save ./...`, this kind of scares me because so many files get updated and there might be an errant piece of debugging or something else hiding in one of them. Can we try and limit the update to just the godeps that we require to be updated?

client/client_test.go
```diff
@@ -464,12 +465,12 @@ func testPublish(t *testing.T, rootType data.KeyAlgorithm) {
 		cryptoservice.NewCryptoService("", trustmanager.NewKeyMemoryStore(passphraseRetriever)))
 
 	r := mux.NewRouter()
-	r.Methods("POST").Path("/v2/{imageName:.*}/_trust/tuf/").Handler(hand(handlers.AtomicUpdateHandler, "push", "pull"))
-	r.Methods("GET").Path("/v2/{imageName:.*}/_trust/tuf/{tufRole:(root|targets|snapshot)}.json").Handler(hand(handlers.GetHandler, "pull"))
-	r.Methods("GET").Path("/v2/{imageName:.*}/_trust/tuf/timestamp.json").Handler(hand(handlers.GetTimestampHandler, "pull"))
-	r.Methods("GET").Path("/v2/{imageName:.*}/_trust/tuf/timestamp.key").Handler(hand(handlers.GetTimestampKeyHandler, "push", "pull"))
-	//r.Methods("POST").Path("/v2/{imageName:.*}/_trust/tuf/{tufRole:(root|targets|timestamp|snapshot)}.json").Handler(hand(handlers.UpdateHandler, "push", "pull"))
-	r.Methods("DELETE").Path("/v2/{imageName:.*}/_trust/tuf/").Handler(hand(handlers.DeleteHandler, "push", "pull"))
+	r.Methods("POST").Path("/v2/{imageName:" + v2.RepositoryNameRegexp.String() + "}/_trust/tuf/").Handler(hand(handlers.AtomicUpdateHandler, "push", "pull"))
```
stevvooe: Small nit: `imageName`. Are these always images?

stevvooe:
@endophage The Godeps.json file was in an odd state. We kind of "reset" things to get them working. The whole directory was blown out and recopied, so there should be no errant debugging statements.

client/client_test.go
endophage: for docker's use of notary, yes they are always image names, but for the general case of notary, they are GUNs, so we should probably update it.

endophage:
@aaronlehmann @stevvooe It still scares me but I took a detailed look through the dependency changes and can't see anything obviously dangerous.

LGTM

stevvooe:
LGTM

endophage:
Was just double checking this as Nathan and I updated the gotuf godep earlier. Noticed this file being added and it probably shouldn't be Godeps/_workspace/src/github.com/endophage/gotuf/client/client_test.go.remove

aaronlehmann:
Sigh, godep is really dumb to have imported that file, since it's not tracked by git. I'll submit a PR to remove it.

aaronlehmann:
Fixed in this PR. I forgot this PR hadn't been merged yet.

endophage:
Cool, LGTM.

Unless there is something super critical I'm missing, let's wait to merge until the Derek's docker/docker PR gets merged. Want to keep master as stable as possible until then.

aaronlehmann:
That is fine with me. I don't think there's any particular urgency here.

dmp42:
@aaronlehmann can you rebase?

aaronlehmann:
Rebased

thaJeztah:
Thanks @aaronlehmann !

PR #107, opened 2015-07-21
fix hash compare on verfy to be bytewise

NathanMcCauley:

### DIFF
--- a/cmd/notary/tuf.go
+++ b/cmd/notary/tuf.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 
+	"crypto/subtle"
 	"github.com/Sirupsen/logrus"
 	notaryclient "github.com/docker/notary/client"
 	"github.com/docker/notary/pkg/passphrase"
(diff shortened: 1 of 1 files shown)


endophage:
LGTM

cmd/notary/tuf.go
```diff
@@ -257,9 +258,10 @@ func verify(cmd *cobra.Command, args []string) {
 	}
 
 	// Create hasher and hash data
-	stdinHash := fmt.Sprintf("sha256:%x", sha256.Sum256(payload))
-	serverHash := fmt.Sprintf("sha256:%s", target.Hashes["sha256"])
-	if stdinHash != serverHash {
+	stdinHash := sha256.Sum256(payload)
+	serverHash := target.Hashes["sha256"]
```
aaronlehmann: Does this need a check to verify that there's a sha256 entry? The Sprintf would have been resilient to this case.

aaronlehmann:
LGTM, except please add a check that serverHash is not nil

aaronlehmann:
Actually subtle.ConstantTimeCompare won't crash if you pass in nil, so this is fine as is.

PR #108, opened 2015-07-21
Fixed but with loadCertsFromDir and added tests

diogomonica:
Fixed but with loadCertsFromDir, added tests and more debugging.

@endophage @NathanMcCauley 

### DIFF
--- a/trustmanager/x509utils.go
+++ b/trustmanager/x509utils.go
@@ -15,6 +15,7 @@ import (
 	"math/big"
 	"net/http"
 	"net/url"
+	"path/filepath"
 	"time"
 
 	"github.com/Sirupsen/logrus"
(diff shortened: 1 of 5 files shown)


trustmanager/x509utils.go
```diff
@@ -118,11 +119,26 @@ func fingerprintCert(cert *x509.Certificate) (CertID, error) {
...
 
 // loadCertsFromDir receives a store AddCertFromFile for each certificate found
-func loadCertsFromDir(s *X509FileStore) {
+func loadCertsFromDir(s *X509FileStore) error {
 	certFiles := s.fileStore.ListFiles(true)
-	for _, c := range certFiles {
-		s.AddCertFromFile(c)
+	for _, f := range certFiles {
+		// ListFiles returns relative paths
+		fullPath := filepath.Join(s.fileStore.BaseDir(), f)
```
endophage: should probably be `s.fileStore.GetPath(f)` given we have the function.

endophage:
LGTM.

Not too concerned either way with my one comment.

trustmanager/x509utils.go
diogomonica: Well, GetPath takes a gun so I would have to extract the extension.

So either I do `.Join` with `BaseDir` or I do `GetPath` with a `suffixTrim`.

NathanMcCauley:
LGTM

PR #112, opened 2015-07-22
Only skip TLS certificate verification if this is explicitly specified in the config

### LINKED ISSUE #111: Notary client doesn't verify TLS certificates
In `cmd/notary/tuf.go`, the notary client turns off TLS certificate verification by passing `NewNotaryRepository` a transport from `getInsecureTransport`.

Self-signed certificates should only be allowed if the user explicitly asks for this. We should not turn off verification unconditionally.

aaronlehmann:
@diogomonica @NathanMcCauley @endophage

Fixes #111


endophage:
Ah good change. Thank you. I put this in so I could run the client easily before the certs were fixed.

LGTM

diogomonica:
LGTM

PR #113, opened 2015-07-22
Return incorrect passphrase for signing failures

NathanMcCauley:
@endophage we need something like this, PTAL

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -47,6 +47,16 @@ func (err *ErrRepoNotInitialized) Error() string {
 	return "Repository has not been initialized"
 }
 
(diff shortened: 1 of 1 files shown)


endophage:
This looks good but also makes me realize something. We have to re-unify `saveMetadata` and `snapshot`. All the data needs to be signed and serialized before writing to disk otherwise we could end up only writing some of the data and being left in an inconsistent state.  As these functions are only used on `init` it's not a critical failure and relatively easy to recover from, but it would be difficult to explain to a user.

endophage:
Superseded by #114

PR #116, opened 2015-07-22
Return invalid password when cannot retrieve passphrase

dmcgowan:
Fixes mismatched expectation in docker client


diogomonica:
LGTM

PR #119, opened 2015-07-23
Abridge key ids on print, warn about key generation

NathanMcCauley:
@dmcgowan 


diogomonica:
LGTM.

PR #120, opened 2015-07-23
add ErrExpired to notary client to translate from gotuf ErrExpired

endophage:


diogomonica:
LGTM

PR #125, opened 2015-07-23
Implement an evil notary server 

NathanMcCauley:
Could be used to test client validations of TUF metadata.

DO NOT MERGE, its pretty dangerous. We should discuss how/if we'd ever want to merge this.


endophage:
Trying to think of a good way to safely include something like this and so far can't come up with anything...

diogomonica:
@NathanMcCauley what do you want to do this?

endophage:
I can't see this being merged any time soon. We'll keep the branch for reference but closing this PR.

PR #126, opened 2015-07-23
ErrSigVerifyFail isn't used, we should be retrying on ErrRoleThreshol…

endophage:
…d which means we didn't have enough keys to validate the signatures


dmcgowan:
LGTM

PR #129, opened 2015-07-24
Notary-cli cleanup and adding remove support - WIP

diogomonica:
WIP - DO NOT MERGE

@endophage PTAL . Remove seems to apply the delete changelists (they disappear locally), but no state appears to change on the server (when listing).


diogomonica:
Superceeded by #143

PR #130, opened 2015-07-25
Signer persistence

diogomonica:
Added encrypted key persistence to notary-signer using GORM.

Still have to workout how to do encryption key rotation.

@nathanmccauley


trustmanager/keydbstore.go
```diff
@@ -0,0 +1,153 @@
...
+	db         gorm.DB
+	passphrase string
+	encrypter  gojose.Encrypter
+	cachedKeys map[string]data.PrivateKey
+}
+
+// GormPrivateKey represents a PrivateKey in the database
+type GormPrivateKey struct {
+	gorm.Model
+	KeyID      string `sql:"not null;unique"`
```
NathanMcCauley: Should we have a primary key on this field?

trustmanager/keydbstore.go
```diff
@@ -0,0 +1,153 @@
...
+	Encryption string `sql:"not null"`
+	Algorithm  string `sql:"not null"`
+	Public     []byte `sql:"not null"`
```
NathanMcCauley: Why is `Public` a `[]byte` and `Private` a `string`?

trustmanager/keydbstore.go
```diff
@@ -0,0 +1,153 @@
...
+	cachedKeys := make(map[string]data.PrivateKey)
+
+	// Retreive the passphrase that will be used to encrypt the keys
+	passphrase, _, err := passphraseRetriever("", "", false, 0)
+	if err != nil {
+		return nil, err
+	}
+
+	// Setup our encrypted object
+	encrypter, err := gojose.NewEncrypter(gojose.A256GCMKW, gojose.A256GCM, []byte(passphrase))
```
NathanMcCauley: I'm not sure we should be using A128GCMKW here. This code will use the passphrase directly as the encryption key and do no key derivation. Hilariously (or maybe intentionally?), the random garbage you typed in the test is exactly 32 bytes which means it works as a 256 bit AES key.

If we're going to use passphrase for key encryption we need to be using something like PBES2-HS256+A128KW from https://tools.ietf.org/html/draft-miller-jose-jwe-protected-jwk-02#section-6.

trustmanager/keydbstore.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Encrypt the private key material
+	encryptedPrivKeyStr := encryptedKey.FullSerialize()
+
+	gormPrivKey := GormPrivateKey{
+		KeyID:      privKey.ID(),
+		Encryption: string(gojose.PBES2_HS512_A256KW),
```
NathanMcCauley: `PBES2_HS512_A256KW` isn't being used here. Can we instead grab the encryption algorithm from the result of the encrypter? It looks like its available on the encryptedKey object if you follow the structs down to the JSON web key object stored on it.

diogomonica: This was here until we realized that gojose didn't implement PBES. Forgot to change it.

Taking it from the encrypter sounds good.

trustmanager/keydbstore.go
diogomonica: We were going to use PBES, but I don't think in this particular case (rufus) it's worth it. We might as well just generate a random 256 bit key.

trustmanager/keydbstore.go
diogomonica: An artifact from gojose receiving a string for encryption and decryption (I think it should receive []byte).

The field that doesn't need to be encrypted/decrypted kept the right data-type :)

I'll change Private to byte[] and convert it before encryption/after decryption.

trustmanager/keystore.go
```diff
@@ -0,0 +1,58 @@
...
+
+type cachedKey struct {
+	alias string
+	key   data.PrivateKey
+}
+
+// PassphraseRetriever is a callback function that should retrieve a passphrase
+// for a given named key. If it should be treated as new passphrase (e.g. with
+// confirmation), createNew will be true. Attempts is passed in so that implementers
+// decide how many chances to give to a human, for example.
```
NathanMcCauley: Is this needed here. I think it is defined as `Retriever` in the `Passphrase` package.

trustmanager/keydbstore.go
NathanMcCauley: Got it. In that case, does it make sense to just ditch the passphrase retriever for this keystore?

If you really wanted to do PBES, we could use this lib instead of go-jose: https://github.com/dvsekhvalnov/jose2go

notarymysql/initial.sql
```diff
@@ -16,3 +16,19 @@ CREATE TABLE `timestamp_keys` (
...
+DROP TABLE IF EXISTS `private_keys`;
+CREATE TABLE `private_keys` (
+	`id` int(11) NOT NULL AUTO_INCREMENT,
+	`created_at` datetime NOT NULL,
+	`updated_at` datetime NOT NULL,
+	`deleted_at` datetime DEFAULT NULL,
+	`key_id`  varchar(255) NOT NULL,
+	`encryption`  varchar(255) NOT NULL,
+	`algorithm`  varchar(50) NOT NULL,
+	`public`  blob NOT NULL,
```
NathanMcCauley: why do `public` and `private` have different types?

trustmanager/keydbstore.go
diogomonica: Gorm is like rails and autoadds a primary key. I enforced uniqueness and added an index.

notarymysql/initial.sql
diogomonica: private is expected to be longer

diogomonica:
@NathanMcCauley addressed all comments. Also, added PBES2, support for multiple passphrases based on aliases and passphrase rotation.

trustmanager/keydbstore.go
```diff
@@ -0,0 +1,203 @@
...
+		return ErrKeyNotFound{}
+	}
+
+	// Get the current passphrase to use for this key
+	passphrase, _, err := s.retriever(dbPrivateKey.KeyID, dbPrivateKey.PassphraseAlias, false, 1)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("Got old passphrase: ", passphrase)
```
NathanMcCauley: remove

trustmanager/keydbstore.go
```diff
@@ -0,0 +1,203 @@
...
+		return err
+	}
+
+	// Get the new passphrase to use for this key
+	newPassphrase, _, err := s.retriever(dbPrivateKey.KeyID, newPassphraseAlias, false, 1)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("new passphrase: ", newPassphrase)
```
NathanMcCauley: remove

NathanMcCauley:
Looks like a change to godeps is needed for the new jose lib.

NathanMcCauley:
LGTM after build passes.

PR #131, opened 2015-07-27
Adding support for passphrases from env

diogomonica:
@endophage 

### DIFF
--- a/cmd/notary-signer/config.json
+++ b/cmd/notary-signer/config.json
@@ -1,0 +1,18 @@
+{
+	"server": {
+		"http_addr": ":4444",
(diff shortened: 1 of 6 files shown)


endophage:
LGTM.

Summary of implementation:

- NOTARY_SIGNER_DEFAULT_ALIAS represents the alias of the key currently being used for encryption. In the first release it will be "timestamp_1".
- "NOTARY_SIGNER_<alias>" is inspected to find the password for the alias. This password is used to generate the KEK for the JWEs.

NathanMcCauley:
LGTM

PR #142, opened 2015-07-28
Removed all local keystores, added configurable trust dir

diogomonica:
Started the work to support a better cli.
- Removed local truststores
- Added configurable base directory
- Added configurable remote server
- Fixed key listing/removal

### DIFF
--- a/cmd/notary/keys.go
+++ b/cmd/notary/keys.go
@@ -4,104 +4,63 @@ import (
 	"crypto/x509"
 	"fmt"
 	"math"
-	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
 	"time"
 
+	"github.com/docker/notary/keystoremanager"
 	"github.com/docker/notary/trustmanager"
 
 	"github.com/spf13/cobra"
-	"github.com/spf13/viper"
 )
 
-var subjectKeyID string
+func init() {
+	cmdKeys.AddCommand(cmdKeysRemoveRootKey)
+	cmdKeys.AddCommand(cmdKeysGenerateRootKey)
+}
 
 var cmdKeys = &cobra.Command{
 	Use:   "keys",
-	Short: "Operates on keys.",
-	Long:  "operations on signature keys and trusted certificate authorities.",
+	Short: "Operates on root keys.",
+	Long:  "operations on private root keys.",
 	Run:   keysList,
 }
 
-func init() {
-	cmdKeys.AddCommand(cmdKeysTrust)
-	cmdKeys.AddCommand(cmdKeysRemove)
-	cmdKeys.AddCommand(cmdKeysGenerate)
+var cmdKeysRemoveRootKey = &cobra.Command{
+	Use:   "remove [ keyID ]",
+	Short: "Removes the root key with the given keyID.",
+	Long:  "remove the root key with the given keyID from the local host.",
+	Run:   keysRemoveRootKey,
 }
 
-var cmdKeysRemove = &cobra.Command{
-	Use:   "remove [ Subject Key ID ]",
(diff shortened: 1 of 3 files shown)


cmd/notary/keys.go
```diff
@@ -4,104 +4,63 @@ import (
...
-		err = certificateStore.RemoveCert(cert)
-		if err != nil {
-			fatalf("failed to remove certificate from KeyStore")
-		}
-		return
+	keyID := args[0]
+	if len(keyID) != 64 {
+		fatalf("please enter a valid root key ID")
 	}
+	parseConfig()
```
aaronlehmann: Would it make sense to call parseConfig() in main so we don't have to call it in every command function?

cmd/notary/keys.go
```diff
@@ -4,104 +4,63 @@ import (
...
+	if err != nil {
+		fatalf("failed to create a new truststore manager with directory: %s", trustDir)
 	}
 
 	// List all the keys about to be removed
-	fmt.Println("Are you sure you want to remove the following keys? (yes/no)", gunOrID)
-	for _, k := range keyList {
-		printKey(k)
-	}
+	fmt.Printf("Are you sure you want to remove the following key? (yes/no)\n%s\n", keyID)
```
aaronlehmann: Move the (yes/no) to after the key ID?

cmd/notary/main.go
```diff
@@ -1,96 +1,71 @@
...
-	"github.com/docker/notary/trustmanager"
+	"github.com/docker/notary/pkg/passphrase"
 )
 
 const configFileName string = "config"
-const configPath string = ".docker/trust/"
-const trustDir string = "trusted_certificates/"
-const privDir string = "private/"
-const rootKeysDir string = "root_keys/"
+const defaultTrustDir string = ".notary/"
```
aaronlehmann: I realize that we don't want to tie this tool to docker, but this means that you need to override the default trust dir to use it with docker, right? Isn't that awkward UX?

cmd/notary/main.go
```diff
@@ -1,96 +1,71 @@
...
+const defaultServerURL = "https://notary-server:4443"
```
aaronlehmann: This is kind of weird too. notary-server won't exist unless someone adds it to a host file. Maybe it's better to have no default, and require the user to specify the server on the command line or in the config file. The error message would be more helpful than a message about not being able to resolve notary-server.

cmd/notary/main.go
diogomonica: Any proposals on what to do?

People could add a configuration file to config.json under ~./notary to always point it to ~/.docker/trust

cmd/notary/keys.go
diogomonica: parseConfig has to be run after `Command.Execute` from viper, or else the command line flags (verbose, basedir, etc) won't be available.

Not sure if there is an alternative to calling it within every command context.

cmd/notary/keys.go
diogomonica: done

cmd/notary/main.go
diogomonica: Not sure if having notary point at ~/.docker/trust by default is a good idea. People might screw things up by accident.

cmd/notary/main.go
diogomonica: We're thinking about adding a default server at thenotary.io that people can use as the default trust server. Unfortunately setting that up (and how we would enforce GUN ownership), is currently an unsolved issue.

aaronlehmann:
LGTM, we'll address the https://notary-server thing later.

PR #143, opened 2015-07-28
fixing remove

endophage:


NathanMcCauley:
needs tests

client/helpers.go
```diff
@@ -26,13 +27,16 @@ func getRemoteStore(baseURL, gun string, rt http.RoundTripper) (store.RemoteStor
...
-		if err != nil {
-			return err
+		switch c.Scope() {
+		case changelist.ScopeTargets:
+			err := applyTargetsChange(repo, c)
+			if err != nil {
+				return err
+			}
+		default:
+			logrus.Debug("scope not supported: ", c.Scope())
```
diogomonica: This is a change in behavior. Now we're not returning an error here.

cmd/notary/tuf.go
```diff
@@ -17,7 +17,7 @@ import (
 )
 
 // FIXME: This should not be hardcoded
-const hardcodedBaseURL = "https://notary-server:4443"
+const hardcodedBaseURL = "http://notary-server:4443"
```
diogomonica: remove!

client/helpers.go
endophage: We weren't before either. Unsupported changes should never be in the changelist so we skip them.

endophage:
@NathanMcCauley I would swear you have a bot that adds "needs tests" to every PR :-P

diogomonica:
@endophage aren't you Mr. 80%+ test coverage?

endophage:
@diogomonica  Yes, but "Demo Early"

PR #147, opened 2015-07-28
Import export cli

aaronlehmann:
Add key import and export commands

This adds four commands:

- notary keys export: export all keys, or keys for a particular GUN (with -g)
- notary keys export-root: export root key by ID
- notary keys import: import a zip file of keys
- notary keys import-root: import a single root key

Change "notary keys" command to "notary key"

This makes the full subcommand names smoother; for example "notary key generate" instead of "notary keys generate".

Add a "notary key list" subcommand to list keys, so "notary key" lists the possible subcommands, instead of needing to use "notary key -h".

@diogomonica

### DIFF
--- a/cmd/notary/keys.go
+++ b/cmd/notary/keys.go
@@ -22,9 +22,10 @@ func init() {
 	cmdKey.AddCommand(cmdKeyRemoveRootKey)
 	cmdKey.AddCommand(cmdKeyGenerateRootKey)
 
-	cmdKeyExport.Flags().StringVarP(&keysExportGUN, "gun", "g", "", "Globally unique name to export keys for. A new password will be set for all the keys. Output format is a zip archive.")
+	cmdKeyExport.Flags().StringVarP(&keysExportGUN, "gun", "g", "", "Globally unique name to export keys for.")
 	cmdKey.AddCommand(cmdKeyExport)
 	cmdKey.AddCommand(cmdKeyExportRoot)
+	cmdKeyExportRoot.Flags().BoolVarP(&keysExportRootChangePassphrase, "change-passphrase", "c", false, "set a new passphrase for the key being exported")
 	cmdKey.AddCommand(cmdKeyImport)
 	cmdKey.AddCommand(cmdKeyImportRoot)
 }
@@ -64,6 +65,8 @@ var cmdKeyExport = &cobra.Command{
 	Long:  "exports a collection of keys. The keys are reencrypted with a new passphrase. The output is a ZIP file.",
 	Run:   keysExport,
 }
+
+var keysExportRootChangePassphrase bool
 
 var cmdKeyExportRoot = &cobra.Command{
 	Use:   "export-root [ keyID ] [ filename ]",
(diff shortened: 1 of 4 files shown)


cmd/notary/keys.go
```diff
@@ -10,37 +11,81 @@ import (
...
 )
 
 func init() {
-	cmdKeys.AddCommand(cmdKeysRemoveRootKey)
-	cmdKeys.AddCommand(cmdKeysGenerateRootKey)
+	cmdKey.AddCommand(cmdKeyList)
+	cmdKey.AddCommand(cmdKeyRemoveRootKey)
+	cmdKey.AddCommand(cmdKeyGenerateRootKey)
+
+	cmdKeyExport.Flags().StringVarP(&keysExportGUN, "gun", "g", "", "Globally unique name to export keys for. A new password will be set for all the keys. Output format is a zip archive.")
```
endophage: For consistency with other commands it seems like this should be a positional argument and possibly export all has a flag `notary key export --all`

cmd/notary/keys.go
```diff
@@ -141,6 +186,140 @@ func keysGenerateRootKey(cmd *cobra.Command, args []string) {
...
+// keysExport exports a collection of keys to a ZIP file
+func keysExport(cmd *cobra.Command, args []string) {
+	if len(args) < 1 {
+		cmd.Usage()
+		fatalf("must specify output filename for export")
+	}
+
+	exportFilename := args[0]
+
+	parseConfig()
```
endophage: This looks like it could be called in main() rather than in every function.

diogomonica:
LGTM. Let's just add a flag to export root that allows to change the passphrase.

cmd/notary/keys.go
aaronlehmann: I made the same comment to Diogo in #142. He replied:

> parseConfig has to be run after Command.Execute from viper, or else the command line flags (verbose, basedir, etc) won't be available.

> Not sure if there is an alternative to calling it within every command context.

endophage: Ahhh, that makes sense.

cmd/notary/keys.go
aaronlehmann: Hmm, I'm not sure I follow. If we made those changes, would the supported forms be:

    notary key export docker.com/docker file.zip
    notary key export --all file.zip

I think that when we have positional arguments, we should make them consistent no matter what flags are specified. But maybe I'm misunderstanding your suggestion.

endophage: In line with other unix commands, possibly the output file should be specified with `-o`, with the output being written to stdout otherwise. Just seems odd that this is the only command that take the GUN as a positional argument.

aaronlehmann: So then would my examples above could look like:

    notary key export -o file.zip docker.com/docker   # export a specific GUN
    notary key export -o file.zip    # export all

I'm okay with this, but it seems a little strange for -o to be a positional argument when it's basically mandatory (if it's omitted, we'd either error out or spew binary garbage to the terminal).

How about making the GUN an optional positional argument that comes after the (mandatory) filename?

endophage: spewing binary garbage to the terminal makes it easy to `|` commands, if somebody has some desire to pipe it directly into an upload via curl or some other madness, so I think that behaviour is acceptable, possibly even desirable.

I feel a little icky about the absence of any specific instruction doing an export all, so I'd really prefer to see a `--all` flag.

aaronlehmann: I think it's better to specify - as the output file to make it go to a terminal. Defaulting to outputting binary stuff to stdout is pretty unfriendly. At the very least, we should check whether stdout is a terminal.

Is the following closer to what you're suggesting?

    notary key export -o file.zip docker.com/docker   # export a specific GUN
    notary key export -o file.zip --all   # export all
    notary key export -o file.zip    # return an error becuase you need either a GUN or --all
    notary key export -o file.zip --all docker.com/docker   # return an error because the GUN and --all are mutually exclusive

I'd still rather not have the GUN as a positional option if there are variants where it shouldn't be specified.

diogomonica: Agreed with aaron.

cmd/notary/keys.go
diogomonica: Sort of. I want a better option. Please find me one.

PR #149, opened 2015-07-28
Avoid printing "Passphrases do not match" when passphrase is too short

aaronlehmann:
Also, wrap the passphrase instructions paragraph at 80 columns, and
change the passphrase variable name in addKey to avoid a conflict with
the package name.


diogomonica:
LGTM

PR #152, opened 2015-07-29
Use the random reader passed as argument to GenerateED25519Key.

calavera:
Rather than the global source.

Signed-off-by: David Calavera <david.calavera@gmail.com>


tiborvass:
Makes sense, although this won't fix any entropy issues, just FYI :)

NathanMcCauley:
LGTM

PR #154, opened 2015-07-29
Added cli cert command, changed keylisting to be a map

diogomonica:
- Added a new cli cert command
- Changed the keylisting to be a map instead of list
- Fixed removal of normal keys, and force root keys deletion to need -r

### DIFF
--- a/cmd/notary/cert.go
+++ b/cmd/notary/cert.go
@@ -1,0 +1,138 @@
+package main
+
+import (
+	"crypto/x509"
+	"fmt"
+	"math"
+	"os"
+	"time"
+
+	"github.com/docker/notary/keystoremanager"
+	"github.com/docker/notary/trustmanager"
+
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	cmdCert.AddCommand(cmdCertList)
+
+	cmdCertRemove.Flags().StringVarP(&certRemoveGUN, "gun", "g", "", "Globally unique name to delete certificates for")
+	cmdCert.AddCommand(cmdCertRemove)
+}
+
+var cmdCert = &cobra.Command{
+	Use:   "cert",
+	Short: "Operates on certificates.",
+	Long:  `operations on certificates.`,
+}
+
+var cmdCertList = &cobra.Command{
+	Use:   "list",
+	Short: "Lists certificates.",
+	Long:  "lists root certificates known to notary.",
+	Run:   certList,
+}
+
+var certRemoveGUN string
+
+var cmdCertRemove = &cobra.Command{
+	Use:   "remove [ certID ]",
+	Short: "Removes the certificate with the given cert ID.",
+	Long:  "remove the certificate with the given cert ID from the local host.",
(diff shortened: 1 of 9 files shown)


cmd/notary/cert.go
```diff
@@ -0,0 +1,138 @@
...
+			fatalf("unable to retrieve certificate with cert ID: %s", certID)
+		}
+		certsToRemove = append(certsToRemove, cert)
+	} else {
+		// We got the -g flag, it's a GUN
+		certs, err := keyStoreManager.TrustedCertificateStore().GetCertificatesByCN(certRemoveGUN)
+		if err != nil {
+			fatalf("%v", err)
+		}
+		certsToRemove = append(certsToRemove, certs...)
```
aaronlehmann: Looks like we ignore `args[0]` in this case. We should check if both `-g` and the positional argument are specified, and report an error.

cmd/notary/cert.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// List all the keys about to be removed
+	fmt.Println("Are you sure you want to remove the following certificates?")
+	for _, cert := range certsToRemove {
+		// This error can't occur because we're getting certs off of an
+		// x509 store that indexes by ID.
+		certID, _ := trustmanager.FingerprintCert(cert)
+		fmt.Printf("%s - %s\n", cert.Subject.CommonName, certID)
+	}
+	fmt.Println("(yes/no)")
```
aaronlehmann: I'd change the first Println to say "The following certificates will be removed:" and the second one to say "Are you sure you want to remove these certificates? (yes/no)". But feel free to ignore... this is a matter of opinion.

cmd/notary/cert.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func printCert(cert *x509.Certificate) {
+	timeDifference := cert.NotAfter.Sub(time.Now())
+	certID, err := trustmanager.FingerprintCert(cert)
+	if err != nil {
+		fatalf("could not fingerprint certificate: %v", err)
+	}
+
+	fmt.Printf("%s %s (expires in: %v days)\n", cert.Subject.CommonName, certID, math.Floor(timeDifference.Hours()/24))
```
aaronlehmann: What if it's expired? Would we print a negative expiration time? Or is there something in the cert loading path that filters out expired certificates? Normally I'd worry about a race here, when we load a non-expired certificate and then it expires before we do this calculation, but we're using units of days here so I don't think we need to worry.

aaronlehmann:
Noted a few things, but looks good to me in general. I'd suggest checking for conflicting arguments in certRemove, but the other suggestions are optional.

cmd/notary/cert.go
diogomonica: We never have expired certs loaded into keystore. We always create FilteredX509FileStores.

cmd/notary/cert.go
diogomonica: Added

PR #156, opened 2015-07-29
Docs infrastructure

moxiegirl:
- Created an index.md file that serves to create a menu
- Created overview.md
- README for how to interact

Signed-off-by: Mary Anthony <mary@docker.com>


moxiegirl:
@diogomonica @NathanMcCauley gents seems to be some git magic on the repo where `git status` tells me I deleted VERSION then I do a `git checkout -- VERSION` and it tells me I deleted `version/version.go` --- in any case until this is resolve I couldn't rebase and squash. So, two commits.

PR #157, opened 2015-07-29
make key algorithm configurable for local development with in memory …

endophage:
…ed25519 crypto service

Also bugsnag 

@dmcgowan 


cmd/notary-server/dev-config.json
```diff
@@ -7,7 +7,8 @@
 	"trust_service": {
 		"type": "local",
 		"hostname": "",
-		"port": ""
+		"port": "",
+		"key_algorithm": "ed25519"
```
diogomonica: why ed?

dmcgowan:
So I take it now setting the algorithm to `"ed25519"` still works as expected since it is not attempting to hardcode to ecdsa?

LGTM

cmd/notary-server/dev-config.json
endophage: Only algo supported for local trust service

endophage: Note that's in dev-config.json which is set up to run a notary server locally, fully in memory, outside a container.

endophage:
@dmcgowan after this latest commit that should be the case.

endophage:
@NathanMcCauley @diogomonica @aaronlehmann can one of you LGTM or otherwise provide feedback on this so we can merge it.

errors/errors.go
```diff
@@ -10,70 +10,69 @@ import (
 // with the registry API
 const errGroup = "notary.api.v1"
 
+// These errors should be returned from contextHandlers only. They are
+// serialized and returned to a user as part of the generic error handling
+// done by the rootHandler
 var (
-	// ErrNoStorage lint comment
```
endophage: @dmcgowan worked out how to get rid of all these comments. Apparently lint accepts a single comment before the var/const block :-)

diogomonica:
LGTM

PR #160, opened 2015-07-30
Update expiration logic

dmcgowan:
Makes use of an update in gotuf to properly check expirations. Notary will now use the gotuf expiration checking function directly.


diogomonica:
LGTM

PR #162, opened 2015-07-30
Notary cli auth

dmcgowan:
Add implementation to do authentication in notary

This is somewhat incomplete since the following does not happen
- allow per endpoint username/password configuration
- obfuscation of password in configuration file
- properly handling failed username/password attempts (might need change in distribution client)

### DIFF
--- a/cmd/notary/tuf.go
+++ b/cmd/notary/tuf.go
@@ -5,12 +5,17 @@ import (
 	"crypto/tls"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
+	"net/url"
 	"os"
+	"time"
 
 	"crypto/subtle"
 
 	"github.com/Sirupsen/logrus"
+	"github.com/docker/distribution/registry/client/auth"
+	"github.com/docker/distribution/registry/client/transport"
 	notaryclient "github.com/docker/notary/client"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
@@ -77,7 +82,7 @@ func tufAdd(cmd *cobra.Command, args []string) {
 
 	parseConfig()
 
-	nRepo, err := notaryclient.NewNotaryRepository(trustDir, gun, remoteTrustServer, getTransport(), retriever)
+	nRepo, err := notaryclient.NewNotaryRepository(trustDir, gun, remoteTrustServer, getTransport(gun), retriever)
 	if err != nil {
 		fatalf(err.Error())
 	}
@@ -102,7 +107,7 @@ func tufInit(cmd *cobra.Command, args []string) {
 	gun := args[0]
 	parseConfig()
 
-	nRepo, err := notaryclient.NewNotaryRepository(trustDir, gun, remoteTrustServer, getTransport(), retriever)
+	nRepo, err := notaryclient.NewNotaryRepository(trustDir, gun, remoteTrustServer, getTransport(gun), retriever)
 	if err != nil {
 		fatalf(err.Error())
 	}
@@ -144,7 +149,7 @@ func tufList(cmd *cobra.Command, args []string) {
 	gun := args[0]
 	parseConfig()
 
-	nRepo, err := notaryclient.NewNotaryRepository(trustDir, gun, remoteTrustServer, getTransport(), retriever)
+	nRepo, err := notaryclient.NewNotaryRepository(trustDir, gun, remoteTrustServer, getTransport(gun), retriever)
 	if err != nil {
 		fatalf(err.Error())
 	}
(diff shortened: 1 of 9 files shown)


cmd/notary/tuf.go
```diff
@@ -266,13 +274,108 @@ func verify(cmd *cobra.Command, args []string) {
...
+}
+
+func (ps *passwordStore) Basic(u *url.URL) (string, string) {
+	var username, password string
+	stdin := bufio.NewReader(os.Stdin)
+	if ps.username != "" {
+		username = ps.username
+		password = ps.password
+	} else {
+		fmt.Fprintf(os.Stdout, "Enter username: ")
```
diogomonica: Maybe make it explicit that this is the docker hub username?

dmcgowan: This is not necessary the hub username, this could be basic auth requested directly by a notary server or basic auth requested by a token server. Either way these do not have to be the hub. We could possibly put the server address which the authentication challenge was received from.

cmd/notary/tuf.go
```diff
@@ -266,13 +274,89 @@ func verify(cmd *cobra.Command, args []string) {
...
+		logrus.Errorf("error processing password input: %s", err)
+		return "", ""
+	}
+	password := strings.TrimSpace(string(userIn))
+
+	return username, password
+}
+
+func getTransport(gun string, readOnly bool) http.RoundTripper {
+	tlsConfig := &tls.Config{
```
diogomonica: add comment that the default is false?

cmd/notary/tuf.go
```diff
@@ -266,13 +274,89 @@ func verify(cmd *cobra.Command, args []string) {
...
-		return &http.Transport{
-			TLSClientConfig: &tls.Config{
-				InsecureSkipVerify: true,
-			},
-		}
+type passwordStore struct {
+	anonymous bool
+}
+
+func (ps passwordStore) Basic(u *url.URL) (string, string) {
```
diogomonica: Why does this receive a `*url.URL`? Doesn't seem to get used anywhere.

cmd/notary/tuf.go
diogomonica: Shouldn't this TLS config enforce minimum of TLS1.0 and all the other good cipher stuff?

cmd/notary/tuf.go
diogomonica: This method doesn't seem to be used at all.

cmd/notary/tuf.go
```diff
@@ -266,13 +274,89 @@ func verify(cmd *cobra.Command, args []string) {
...
+		DisableKeepAlives:   true,
+	}
+
+	// TODO(dmcgowan): add notary specific headers
+	authTransport := transport.NewTransport(base)
+	pingClient := &http.Client{
+		Transport: authTransport,
+		Timeout:   5 * time.Second,
+	}
+	endpointStr := "https://notary.docker.io/v2/"
```
diogomonica: Will this being hardcoded require every notary action to request something to this URL? What if we're just using `notary list -s localhost:4443 diogo/repo`. Won't the way this is setup always try to get this Transport, create a request and ping the notary.docker.io server?

cmd/notary/tuf.go
dmcgowan: This is to implement the credential store interface, it will be called by the transport if/when basic auth is needed. The url can be used to select between different credentials, in this case I should actually display the hostname to the user so the user knows where their authentication credentials are being sent.

See
https://github.com/docker/distribution/blob/master/registry/client/auth/session.go#L198

cmd/notary/tuf.go
dmcgowan: Ahh yes, I need to update this to use that value.

cmd/notary/tuf.go
dmcgowan: Yes we should do that here, it was like that before but I should not be lazy here and set the proper configuration.

cmd/notary/tuf.go
diogomonica: But even if it uses that value, it seems like it will `fatalf` if it can't get a token from the remote auth server, and that will prevent someone from using their own notary-server without auth.

dmcgowan: This ping is only to get the `WWW-Authenticate` header, if auth is not supported, then the server will be treated as unauthenticated. These challenges will be used when attempting to make a request, meaning a token is only requested when the first request is made. If there is no authentication, then no token will ever be requested.

thaJeztah:
> obfuscation of password in configuration file

(Apologies for just dropping in here); I'd really be careful here, thinking of the bad publicity we got with the credentials stored in `.dockercfg`. I haven't read-up on the context here, just read that line  :smile:

endophage:
Superseded by https://github.com/docker/notary/pull/188

PR #163, opened 2015-07-30
Move key database to signer package

dmcgowan:
The key database is not generally used but only used by the signing service.
Move the implementation to the signer package to be imported by the signer.

ping @endophage


endophage:
LGTM

PR #164, opened 2015-07-30
don't want bugsnag config template in config.json

endophage:


aaronlehmann:
LGTM. Merging.

PR #166, opened 2015-07-31
compile the notary server binary statically 

jfrazelle:
compile the notary server binary statically in the notary server image so that it can then be copied from the container into a scratch docker image

ping @aaronlehmann then you can use `FROM scratch`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,7 +97,7 @@ binaries: ${PREFIX}/bin/notary-server ${PREFIX}/bin/notary ${PREFIX}/bin/notary-
 
 define template
 mkdir -p ${PREFIX}/cross/$(1)/$(2);
-GOOS=$(1) GOARCH=$(2) go build -o ${PREFIX}/cross/$(1)/$(2)/notary -a -tags "static_build netgo" -installsuffix netgo ${GO_LDFLAGS_STATIC} ./cmd/notary;
+GOOS=$(1) GOARCH=$(2) CGO_ENABLED=0 go build -o ${PREFIX}/cross/$(1)/$(2)/notary -a -tags "static_build netgo" -installsuffix netgo ${GO_LDFLAGS_STATIC} ./cmd/notary;
 endef
 
 cross:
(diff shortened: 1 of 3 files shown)


jfrazelle:
also added the ldflags to the one built in the signer dockerfile so that binary has them as well

jfrazelle:
didnt statically compile that one tho because its probably hopeless and will need `softhsm2` in the image anyways, can try tho..

aaronlehmann:
So that's what I tried to do with the registry at first, but it didn't work out because of the cgo use. Same issue with notary-signer. For notary-server I think it's theoretically possible, but I'm not sure it's worth making the notary-server container way different from notary-signer. Also, there may be weird issues that come up running in a "scratch" container, for example if something tries to shell out. And if we ever add a dependency that uses cgo, we'd have to redo the container.

What do you think?

Thanks for the other feedback as well. I'll act on it tomorrow.

jfrazelle:
what would you shell out to? but ya i see your point i will just add the ldflags

jfrazelle:
updated, im a big fan of static binaries tho bc i hate installing things haha

aaronlehmann:
I'll defer to others on this, but it seems a little weird to me to end up with notary-server being static and notary-signer being dynamic. I'd like them to be as consistent as possible. And I don't think there's much advantage to notary-server having a tiny container if it would otherwise use the same base layer as notary-signer.

jfrazelle:
i changed to both dynmic

jfrazelle:
just didnt change the PR title or branch name

aaronlehmann:
Oh, okay, I didn't actually look at the diff. Could you please squash the commits and change the commit message to reflect what's being done, and then I'll merge. Thanks!

jfrazelle:
will do!

jfrazelle:
done

jfrazelle:
Btw you already merged https://github.com/docker/notary/pull/168 and it doesn't work without this patch. Which I said....

jfrazelle:
MERGE MY PR
...

i'll wait...
waiting

aaronlehmann:
LGTM

NathanMcCauley:
LGTM

PR #167, opened 2015-07-31
Updates to Makefile

jfrazelle:
not everyone has godeps and all that crap on their host, namely me, i do not so
yeah.... this is better

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ GO_LDFLAGS=-ldflags "-w $(CTIMEVAR)"
 GO_LDFLAGS_STATIC=-ldflags "-w $(CTIMEVAR) -extldflags -static"
 GOOSES = darwin freebsd linux
 GOARCHS = amd64
+NOTARYDIR := /go/src/github.com/docker/notary
 
 # go cover test variables
 COVERDIR=.cover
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -15,6 +15,7 @@ GO_LDFLAGS=-ldflags "-w $(CTIMEVAR)"
 GO_LDFLAGS_STATIC=-ldflags "-w $(CTIMEVAR) -extldflags -static"
 GOOSES = darwin freebsd linux
 GOARCHS = amd64
+NOTARYDIR := /go/src/github.com/docker/notary
```
diogomonica: What if your go dir is not on /go ?

jfrazelle: its only used in the container

Makefile
```diff
@@ -97,12 +98,28 @@ binaries: ${PREFIX}/bin/notary-server ${PREFIX}/bin/notary ${PREFIX}/bin/notary-
...
+server-dockerfile:
+	@docker build --rm --force-rm -f notary-server-Dockerfile -t notary-server .
+
+signer-dockerfile:
+	@docker build --rm --force-rm -f notary-signer-Dockerfile -t notary-signer .
+
+docker-images: notary-dockerfile server-dockerfile signer-dockerfile
+
+shell: notary-dockerfile
+	docker run --rm -it -v $(CURDIR)/cross:$(NOTARYDIR)/cross -v $(CURDIR)/bin:$(NOTARYDIR)/bin notary bash
```
jfrazelle: @diogomonica here

diogomonica: got it

PR #169, opened 2015-07-31
hard fail if TLS is only partially configured

endophage:


endophage:
@aaronlehmann didn't want to import errors for that one thing. fmt was already imported.

PR #170, opened 2015-07-31
Support aliasMap for pretty printing on PhraseRetrieval. Better keyID…

NathanMcCauley:
… printing

@dmcgowan 

### DIFF
--- a/pkg/passphrase/passphrase.go
+++ b/pkg/passphrase/passphrase.go
@@ -22,7 +22,7 @@ import (
 type Retriever func(keyName, alias string, createNew bool, attempts int) (passphrase string, giveup bool, err error)
 
 const (
-	idBytesToDisplay            = 5
+	idBytesToDisplay            = 7
 	tufRootAlias                = "root"
 	tufTargetsAlias             = "targets"
 	tufSnapshotAlias            = "snapshot"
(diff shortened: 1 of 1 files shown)


pkg/passphrase/passphrase.go
```diff
@@ -46,23 +46,35 @@ var (
...
+	return PromptRetrieverWithInOut(os.Stdin, os.Stdout, nil)
 }
 
 // PromptRetrieverWithInOut returns a new Retriever which will provide a
 // prompt using the given in and out readers. The passphrase will be cached
 // such that subsequent prompts will produce the same passphrase.
-func PromptRetrieverWithInOut(in io.Reader, out io.Writer) Retriever {
+// aliasMap can be used to specify display names for TUF key aliases. If aliasMap
+// is nil, a sensible default will be used.
+func PromptRetrieverWithInOut(in io.Reader, out io.Writer, aliasMap map[string]string) Retriever {
```
aaronlehmann: We should really use functional args instead of having different functions for creating a PromptRetriever. See http://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

pkg/passphrase/passphrase.go
```diff
@@ -46,23 +46,35 @@ var (
 	// ErrTooManyAttempts is returned if the maximum number of passphrase
 	// entry attempts is reached.
 	ErrTooManyAttempts = errors.New("Too many attempts")
+
+	defaultAliasMap = map[string]string{
+		"root":       "root",
+		"targets":    "targets",
+		"snapshot":   "snapshot",
+		"timestamps": "timestamps",
+	}
```
aaronlehmann: This seems kind of silly. Why not just avoid the alias map lookup if aliasMap is nil?

PR #171, opened 2015-07-31
can't be so restrictive on notary's GUN matching in URLs

endophage:
The current image regex from distribution does not work for `hostname:port` or anything non-lowercase ascii alpha numeric. Rely on the authentication service to restrict users to only pushing valid names.

### DIFF
--- a/utils/http.go
+++ b/utils/http.go
@@ -7,6 +7,7 @@ import (
 	"github.com/docker/distribution/registry/api/errcode"
 	"github.com/docker/distribution/registry/api/v2"
 	"github.com/docker/distribution/registry/auth"
+	"github.com/docker/notary/errors"
 	"github.com/endophage/gotuf/signed"
 	"github.com/gorilla/mux"
 	"golang.org/x/net/context"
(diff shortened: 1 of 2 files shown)


utils/http.go
```diff
@@ -66,13 +67,28 @@ func (root *rootHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		if err, ok := err.(errcode.Error); ok {
+			logrus.Errorf(
+				"[Notary Server] %d %s %s",
+				err.Code.Descriptor().HTTPStatusCode,
+				r.Method,
+				r.URL.Path,
+			)
+		} else {
+			logrus.Errorf(
+				"[Notary Server] 5XX %s %s %s",
```
aaronlehmann: Will this always be a 5XX error? Can we log the actual error code?

endophage: In this instance, it will likely always be a 500, but ServeJSON actually makes that decision. I put 5XX because if we ever see this, it's something that needs to be fixed ASAP, as it's essentially an unexpected, unhandled error.

aaronlehmann:
I want to make some changes to the way logging works, but it would make sense to merge this first and open a new PR for the changes. So I'd like to merge this as is - any objections?

PR #173, opened 2015-07-31
Update notary godeps

dmcgowan:
Updated to master of gotuf
Fix awkward docker/docker dependency


Godeps/Godeps.json
```diff
@@ -77,18 +77,13 @@
 		},
 		{
 			"ImportPath": "github.com/docker/docker/pkg/tarsum",
-			"Comment": "v1.4.1-775-g70fbd45",
-			"Rev": "70fbd45a5c88f6f39a07b04f81a07721bf5f3eed"
+			"Comment": "v1.7.1",
+			"Rev": "786b29d4db80a6175e72b47a794ee044918ba734"
```
NathanMcCauley: Interesting, out of curiousity what's the dependency on tarsum?

dmcgowan: `github.com/docker/distribution/digest`

We can probably deprecate it in that package since we never shipped a distribution release with support for it, and older docker clients don't use that package. Would file an issue in distribution to change it.

diogomonica:
LGTM.

Godeps/Godeps.json
```diff
@@ -77,18 +77,13 @@
...
 		},
 		{
 			"ImportPath": "github.com/docker/docker/pkg/term",
-			"Comment": "v1.4.1-775-g70fbd45",
-			"Rev": "70fbd45a5c88f6f39a07b04f81a07721bf5f3eed"
-		},
-		{
-			"ImportPath": "github.com/docker/docker/vendor/src/code.google.com/p/go/src/pkg/archive/tar",
```
diogomonica: this was really ugly.

dmcgowan: yes it was, not sure how the old docker version which needed it got pulled in.

unless you are talking about the poor diff job which is also ugly :)

PR #175, opened 2015-07-31
check error in initializing remote store

endophage:


NathanMcCauley:
LGTM. Should this make it in to 1.8?

endophage:
Not critical for us. It only sanity checks the options and we know we're
passing correct params for the docker integration

PR #177, opened 2015-08-03
Import and export symlinks in keystore

aaronlehmann:
- Export symlinks by encoding them in the zip file.

- Detect symlinks in a zip file on import and create them on the local
  filesystem.

- Add test coverage.

### DIFF
--- a/cmd/notary/keys.go
+++ b/cmd/notary/keys.go
@@ -258,8 +258,8 @@ func keysExport(cmd *cobra.Command, args []string) {
 	exportFile.Close()
 
 	if err != nil {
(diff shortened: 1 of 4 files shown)


diogomonica:
Seems like a decent intermediate compromise while we don't build a better solution into notary itself.

@endophage, just so we have it here:

- Change cryptosigners to receive PublicKeys instead of keyID strings, decode if the publickey is a valid X509 Certificate, and if so, use the public key material from the x509 certificate to calculate a new ID that we actually pass to `.GetKey()`.

endophage:
@diogomonica Sounds good!

PR #174, opened 2015-07-31
Change logging to use contexts effectively

aaronlehmann:
Use the github.com/docker/distribution/context package to get a logger
that adds a unique ID and useful information about each HTTP request.

Use this logger in HTTP handlers instead of using logrus or the log
package directly.

Remove [Notary Server] and [Notary Signer] prefixes from log messages.

The distribution/context package was already vendored, so there are no
Godeps changes necessary.

Sample output:

    notaryserver_1 | time="2015-07-31T23:02:01Z" level=debug msg="retrieving timestamp key for docker.com/docker"
    notarysigner_1 | time="2015-07-31T23:02:01Z" level=debug msg="generated ECDSA key with keyID: ea89e7dc49a13feab1e5ed349760b148c3c6ebd86968b2bc6cb0d003a8b79f78"
    notarysigner_1 | time="2015-07-31T23:02:01Z" level=debug msg="generated new ecdsa key for role:  and keyID: ea89e7dc49a13feab1e5ed349760b148c3c6ebd86968b2bc6cb0d003a8b79f78"
    notarysigner_1 | time="2015-07-31T23:02:01Z" level=info msg="CreateKey: Created KeyID ea89e7dc49a13feab1e5ed349760b148c3c6ebd86968b2bc6cb0d003a8b79f78"
    notaryserver_1 | time="2015-07-31T23:02:01Z" level=debug msg="Creating new timestamp key for docker.com/docker. With algo: ecdsa"
    notaryserver_1 | time="2015-07-31T23:02:01Z" level=debug msg="Inserting timestamp key for docker.com/docker"
    notaryserver_1 | time="2015-07-31T23:02:01Z" level=debug msg="200 GET timestamp key" docker.com/docker=gun http.request.host="192.168.99.100:4443" http.request.id=a720da02-4312-48ae-b122-6d4bce9d3b20 http.request.method=GET http.request.remoteaddr="192.168.99.1:58178" http.request.uri="/v2/docker.com/docker/_trust/tuf/timestamp.key" http.request.useragent="Go 1.1 package http"
[...]

PR #178, opened 2015-08-05
server side validation during updates

endophage:
This adds server side validation of root, top level targets, and snapshots during an update (timestamps are not validated because the server will generate them).

It validates:
- that a root rotation is correct if detected.
- that the signatures on all items being updated are correct.
- that the hashes in snapshot are correct for the things being updated and unchanged for the things that aren't being updated.

This also vendors an updated gotuf with some minor improvements: https://github.com/endophage/gotuf/pull/40

In before @NathanMcCauley: working on tests 

### DIFF
--- a/Godeps/_workspace/src/github.com/endophage/gotuf/store/httpstore.go
+++ b/Godeps/_workspace/src/github.com/endophage/gotuf/store/httpstore.go
@@ -14,10 +14,12 @@ import (
 	"github.com/Sirupsen/logrus"
 )
 
-type ErrServerUnavailable struct{}
+type ErrServerUnavailable struct {
+	code int
+}
 
 func (err ErrServerUnavailable) Error() string {
-	return "Unable to reach trust server at this time."
+	return fmt.Sprintf("Unable to reach trust server at this time: %d.", err.code)
 }
 
 type ErrShortRead struct{}
@@ -85,13 +87,15 @@ func (s HTTPStore) GetMeta(name string, size int64) ([]byte, error) {
 		return nil, err
 	}
 	defer resp.Body.Close()
+	if resp.StatusCode == http.StatusNotFound {
+		return nil, ErrMetaNotFound{}
+	} else if resp.StatusCode != http.StatusOK {
+		return nil, ErrServerUnavailable{code: resp.StatusCode}
+	}
 	if resp.ContentLength > size {
 		return nil, ErrMaliciousServer{}
 	}
 	logrus.Debugf("%d when retrieving metadata for %s", resp.StatusCode, name)
-	if resp.StatusCode == http.StatusNotFound {
-		return nil, ErrMetaNotFound{}
-	}
 	b := io.LimitReader(resp.Body, size)
 	body, err := ioutil.ReadAll(b)
 	if resp.ContentLength > 0 && int64(len(body)) < resp.ContentLength {
@@ -113,8 +117,17 @@ func (s HTTPStore) SetMeta(name string, blob []byte) error {
 	if err != nil {
 		return err
 	}
-	_, err = s.roundTrip.RoundTrip(req)
-	return err
+	resp, err := s.roundTrip.RoundTrip(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode == http.StatusNotFound {
+		return ErrMetaNotFound{}
+	} else if resp.StatusCode != http.StatusOK {
+		return ErrServerUnavailable{code: resp.StatusCode}
+	}
+	return nil
 }
 
 func (s HTTPStore) SetMultiMeta(metas map[string][]byte) error {
@@ -140,8 +153,17 @@ func (s HTTPStore) SetMultiMeta(metas map[string][]byte) error {
 	if err != nil {
 		return err
 	}
-	_, err = s.roundTrip.RoundTrip(req)
-	return err
+	resp, err := s.roundTrip.RoundTrip(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode == http.StatusNotFound {
+		return ErrMetaNotFound{}
+	} else if resp.StatusCode != http.StatusOK {
+		return ErrServerUnavailable{code: resp.StatusCode}
+	}
+	return nil
 }
 
 func (s HTTPStore) buildMetaURL(name string) (*url.URL, error) {
@@ -188,6 +210,12 @@ func (s HTTPStore) GetTarget(path string) (io.ReadCloser, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer resp.Body.Close()
+	if resp.StatusCode == http.StatusNotFound {
+		return nil, ErrMetaNotFound{}
+	} else if resp.StatusCode != http.StatusOK {
+		return nil, ErrServerUnavailable{code: resp.StatusCode}
+	}
 	return resp.Body, nil
 }
 
@@ -205,6 +233,11 @@ func (s HTTPStore) GetKey(role string) ([]byte, error) {
 		return nil, err
 	}
 	defer resp.Body.Close()
+	if resp.StatusCode == http.StatusNotFound {
+		return nil, ErrMetaNotFound{}
+	} else if resp.StatusCode != http.StatusOK {
+		return nil, ErrServerUnavailable{code: resp.StatusCode}
+	}
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
 		return nil, err
--- a/client/client.go
+++ b/client/client.go
@@ -456,17 +456,20 @@ func (r *NotaryRepository) bootstrapRepo() error {
 	if err != nil {
 		return err
 	}
-	root := &data.Signed{}
+	root := &data.SignedRoot{}
 	err = json.Unmarshal(rootJSON, root)
 	if err != nil {
 		return err
 	}
-	tufRepo.SetRoot(root)
+	err = tufRepo.SetRoot(root)
+	if err != nil {
+		return err
+	}
 	targetsJSON, err := r.fileStore.GetMeta("targets", 0)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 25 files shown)


endophage:
83.6% test coverage on validation.go :-)

endophage:
Realised I hadn't tested for what will be the most common case: only pushing targets and snapshots. Wrote test, test failed, fixed code, better now.

Godeps/_workspace/src/github.com/endophage/gotuf/store/httpstore.go
```diff
@@ -14,10 +14,12 @@ import (
...
 )
 
-type ErrServerUnavailable struct{}
+type ErrServerUnavailable struct {
+	code int
+}
 
 func (err ErrServerUnavailable) Error() string {
-	return "Unable to reach trust server at this time."
+	return fmt.Sprintf("Unable to reach trust server at this time: %d.", err.code)
```
aaronlehmann: Let's make clear that the integer is an error code and not a timestamp.

Godeps/_workspace/src/github.com/endophage/gotuf/store/httpstore.go
```diff
@@ -113,8 +117,17 @@ func (s HTTPStore) SetMeta(name string, blob []byte) error {
...
-	_, err = s.roundTrip.RoundTrip(req)
-	return err
+	resp, err := s.roundTrip.RoundTrip(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode == http.StatusNotFound {
+		return ErrMetaNotFound{}
+	} else if resp.StatusCode != http.StatusOK {
```
aaronlehmann: Is the spec explicit about 200 being the only acceptable success code for these POST operations?

client/client.go
```diff
@@ -557,6 +560,9 @@ func (r *NotaryRepository) bootstrapClient() (*tufclient.Client, error) {
 			return nil, store.ErrMetaNotFound{}
 		}
 	}
+	// can't just unmarshal into SignedRoot because validate root
+	// needs the root.Signed field to still be []byte for signature
+	// validation
```
aaronlehmann: Makes sense; thanks for adding this comment.

server/handlers/validation.go
```diff
@@ -0,0 +1,377 @@
...
+
+// ErrBadHeirarchy represents a missing snapshot at this current time.
+// When delegations are implemented it will also represent a missing
+// delegation parent
+type ErrBadHeirarchy struct {
+	msg string
+}
+
+func (err ErrBadHeirarchy) Error() string {
+	return fmt.Sprintf("Heirarchy of updates in incorrect: %s", err.msg)
```
aaronlehmann: hierarchy

server/handlers/validation.go
```diff
@@ -0,0 +1,377 @@
...
+}
+
+func (err ErrValidation) Error() string {
+	return fmt.Sprintf("An error occurred during validation: %s", err.msg)
+}
+
+// ErrBadHeirarchy represents a missing snapshot at this current time.
+// When delegations are implemented it will also represent a missing
+// delegation parent
+type ErrBadHeirarchy struct {
```
aaronlehmann: s/Heirarchy/Hierarchy/ on this type and the docstring

server/handlers/validation.go
```diff
@@ -0,0 +1,377 @@
...
+	return fmt.Sprintf("Heirarchy of updates in incorrect: %s", err.msg)
+}
+
+// ErrBadRoot represents a failure validating the root
+type ErrBadRoot struct {
+	msg string
+}
+
+func (err ErrBadRoot) Error() string {
+	return fmt.Sprintf("The root being updated is invalid: %s", err.msg)
```
aaronlehmann: Might be more user-friendly to say "The root metadata file" or something like that.

server/handlers/validation.go
```diff
@@ -0,0 +1,377 @@
...
+	targetsRole := data.RoleName(data.CanonicalTargetsRole)
+	snapshotRole := data.RoleName(data.CanonicalSnapshotRole)
+
+	// check that the necessary roles are present:
+	roles := make(map[string]storage.MetaUpdate)
+	for _, v := range updates {
+		roles[v.Role] = v
+	}
+	if err := heirarchyOK(roles); err != nil {
+		logrus.Error("ErrBadHeirarchy: ", err.Error())
```
aaronlehmann: s/Heirarchy/Hierarchy/

server/handlers/validation.go
```diff
@@ -0,0 +1,377 @@
...
+// 1. the user is updating only the snapshot.
+// 2. the user is updating a targets (incl. delegations) or
+//    root metadata. This requires they also provide a new
+//    snapshot.
+// N.B. users should never be updating timestamps. The server
+//      always handles timestamping. If the user does send a
+//      timestamp, the server will replace it on next
+//      GET timestamp.jsonshould it detect the current
+//      snapshot has a different hash to the one in the timestamp.
+func heirarchyOK(roles map[string]storage.MetaUpdate) error {
```
aaronlehmann: s/heirarchy/hierarchy/

server/handlers/validation.go
aaronlehmann: I suggest passing in a context to this function and using ctxu.GetLogger().Error instead of logrus.Error. Then this log line will indicate which request experienced the error. See https://github.com/docker/notary/pull/174

server/handlers/validation.go
```diff
@@ -0,0 +1,377 @@
...
+	parsedNewRoot := &data.SignedRoot{}
+
+	if oldRoot != nil {
+		parsedOldRoot = &data.SignedRoot{}
+		err := json.Unmarshal(oldRoot, parsedOldRoot)
+		if err != nil {
+			// TODO(david): if we can't read the old root should we continue
+			//             here to check new root self referential integrity?
+			//             This would permit recovery of a repo with a corrupted
+			//             root.
```
aaronlehmann: Hmm. My instinct is that we should error out if the old root is corrupted, but I can't think of any obvious attacks that continuing here would open up.

server/handlers/validation_test.go
```diff
@@ -0,0 +1,893 @@
...
+			Role:    "snapshot",
+			Version: 1,
+			Data:    snapshot,
+		},
+		{
+			Role:    "timestamp",
+			Version: 1,
+			Data:    timestamp,
+		},
+	}
```
aaronlehmann: Can we break this out of the test cases? This exact variant is used in many functions.

server/handlers/validation_test.go
```diff
@@ -0,0 +1,893 @@
...
+
+	err = validateUpdate("testGUN", updates, store)
+	assert.Error(t, err)
+	assert.IsType(t, ErrBadSnapshot{}, err)
+}
+
+// ### End snapshot hash mismatch negative tests ###
```
aaronlehmann: The test cases look really extensive. Out of curiousity, what level of test coverage are you seeing?

aaronlehmann: Never mind - saw your comment above.

aaronlehmann:
The code looks correct to me, but I think I'm missing the "why?". I'd like to better understand what benefits server-side validation brings and what kind of situations it's meant to address.

My other question is why the server-side validation code is distinct from the client-side validation code. I would think that both the client and server sides should do essentially the same thing to validate a set of documents. I'm a little concerned about having two separate sets of checks, in case they handle edge cases differently.

Godeps/_workspace/src/github.com/endophage/gotuf/store/httpstore.go
endophage: It is the only acceptable status we currently return.

endophage:
@aaronlehmann it's really just a case of "sanitize your inputs" and in this case, may reduce the on call events because it will be that much harder for a user to persist bad data. If the client is having problems, nobody is getting paged, but if somebody pushes data that causes errors on the server, somebody (probably myself or @diogomonica) will be getting paged.

aaronlehmann:
That makes sense, but can you comment on why this is a separate set of checks from the ones the client performs? I'm sure there's a good reason; I'm just curious to understand this better.

endophage:
As far as why server and client are different, they validate different things in different ways (there is some overlap).

This validation for example ensures that root.json actually contains the 4 base roles. A client doesn't care about that, a client only cares that the keys are available when it needs to check a signature.

It will eventually (when we add delegation support) compare the previous snapshot against the new one and ensure only the updated roles have changed and all others have stayed the same (or possibly been removed, lots of complexity here).

We may be able to merge some of it, but generally the clients don't actually do any validation, they just check signatures, and this is sharing the same cryptographic verification code.

server/handlers/validation_test.go
endophage: I thought about doing that last night and there isn't that much improvement to be had. I'd already written a lot of the cases at that point so I just didn't feel it was worth the effort.

aaronlehmann: I just feel that it's easier to read the tests if this gets split out; many of them would be a lot more succinct. But I don't feel especially strongly.

server/handlers/validation.go
NathanMcCauley: If we can't parse the root on the server what do we expect the client to do?

server/handlers/validation.go
```diff
@@ -0,0 +1,377 @@
...
+	err = signed.VerifyRoot(newSigned, newRootRole.Threshold, newKeys)
+	if err != nil {
+		return err
+	}
+	root, err := data.RootFromSigned(newSigned)
+	if err != nil {
+		return err
+	}
+	// at a minimum, check the 4 required roles are present
+	for _, r := range []string{rootRole, targetsRole, snapshotRole, timestampRole} {
```
NathanMcCauley: This seems like it'd be a good check for the client to do.

NathanMcCauley:
I add a new issue to move some of these validations to client side. We can look at that as a followup.

PR #180, opened 2015-08-06
some cleanup and default server should be staging

jfrazelle:
...or is this prod, idk idk ;)

### DIFF
--- a/cmd/notary/main.go
+++ b/cmd/notary/main.go
@@ -15,16 +15,20 @@ import (
 	"github.com/docker/notary/version"
 )
 
-const configFileName string = "config"
-const defaultTrustDir string = ".notary/"
-const defaultServerURL = "https://notary-server:4443"
-const idSize = 64
+const (
+	configFileName   string = "config"
+	defaultTrustDir  string = ".notary/"
+	defaultServerURL        = "https://notary.docker.io"
+	idSize                  = 64
+)
 
-var rawOutput bool
-var trustDir string
-var remoteTrustServer string
-var verbose bool
-var retriever passphrase.Retriever
+var (
+	rawOutput         bool
+	trustDir          string
+	remoteTrustServer string
+	verbose           bool
+	retriever         passphrase.Retriever
+)
 
 func init() {
 	retriever = passphrase.PromptRetriever()
--- a/cmd/notary/main.go
+++ b/cmd/notary/main.go
@@ -3,7 +3,6 @@ package main
 import (
 	"fmt"
 	"os"
-	"os/user"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


jfrazelle:
wtf is a GUN

jfrazelle:
`* fatal: must specify a GUN`

jfrazelle:
i dunno how to use godeps so idk how to add that..

cmd/notary/main.go
```diff
@@ -11,20 +10,25 @@ import (
...
 )
 
-const configFileName string = "config"
-const defaultTrustDir string = ".notary/"
-const defaultServerURL = "https://notary-server:4443"
-const idSize = 64
+const (
+	configFileName   string = "config"
+	defaultTrustDir  string = ".notary/"
+	defaultServerURL        = "https://notary.docker.io"
```
NathanMcCauley: This is the default notary server in Docker, but for dev I think we just want to be able to point to our local instances so this should be `https://notary-server:4443`

jfrazelle: Maybe it should be a compile time variable :/

jfrazelle: We do this a lot in the docker release scripts but something like 

if [ "$(cat NOTARY_VERSION)" = *-dev]; then
     # add -X notaryservervar is dev server
else
    # make it prod
     # for a real release version
fi

something like that

Btw this is from my phone so that probably looks like poo

NathanMcCauley: i like this plan

diogomonica: Note that there is no current public notary server. We would have to deal with GUN ownership for that to be a reality.

jfrazelle: if i as a docker user want to use `notary` the cmd line tool to manage my
keys what server do i point it at, because I think we can all agree (at
least for the time being, the most popular use case for the notary cmd line
tool will be with docker.

diogomonica: You would point it at `notary.docker.io`, probably by providing the `-s` flag. I don't think we should point it at it by default though, because it would be really easy for someone to screw up their repository on accident.

jfrazelle: ok ill close then but i dont really agree

jfrazelle: maybe you should not have a default when you ship because otherwise you are going to be getting a whole lot of tickets with the title 
"cannot connect to http://notary-server:4443"

diogomonica: I'm ok with removing the default. It's just been useful for testing

jfrazelle: well u can make it a compile time var

PR #185, opened 2015-08-13
WIP: Update dockerfiles so we can deploy notary services without ops involvement.

endophage:
Still need to update notary-signer. Want to open the PR so I can get Aaron to take a look and make sure I'm doing it right before I continue.


endophage:
This was useful to understand how to get to self sufficient deployments, but ultimately it's unnecessary to have these changes in this repo. Closing.

PR #188, opened 2015-08-13
Notary cli auth

endophage:
Supersedes https://github.com/docker/notary/pull/162

Adds token auth capabilities to notary cli. For any interactions that require a user to be authenticated (private repos, pushes, etc...) the user will always be asked for username and password (@thaJeztah, no caching of credentials :-).

~~Authentication is enabled with the `--auth` or short form `-a` flags. Valid values are "token" to enable docker style token authentication, or "" to disable authentication. We plan to support HTTP basic auth at some point in the future but it is not an immediate requirement.~~ As @dmcgowan points out below, auth is required or not by the server and there's not point in having a CLI flag, it just confuses things. It has been removed.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -76,6 +76,16 @@
 			"Rev": "fed58bd2d3c096055c0e69c2fb86c9a4965d1b8b"
 		},
 		{
+			"ImportPath": "github.com/docker/distribution/registry/client/auth",
+			"Comment": "v2.0.0-353-gfed58bd",
+			"Rev": "fed58bd2d3c096055c0e69c2fb86c9a4965d1b8b"
+		},
+		{
+			"ImportPath": "github.com/docker/distribution/registry/client/transport",
+			"Comment": "v2.0.0-353-gfed58bd",
+			"Rev": "fed58bd2d3c096055c0e69c2fb86c9a4965d1b8b"
+		},
+		{
 			"ImportPath": "github.com/docker/distribution/uuid",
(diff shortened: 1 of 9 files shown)


thaJeztah:
Awesome! thanks @endophage !

endophage:
@diogomonica @NathanMcCauley I mentioned to one/both of you I wasn't sure where the URL for authentication came from and thought it might be hard coded into the token library. Good news! I've just been looking around and realized it's provided by the server as a configurable part of the auth challenge response.

We're good to merge this once it has been reviewed.

@dmcgowan @jlhawn as the most knowledgeable people on the token auth can you take a look? Derek: this has largely remained unchanged from your PR, just made the auth URL a CLI flag, and made it possible to disable/enable auth (and choose a type of auth) with another flag.

ALL: please vet the choice of CLI flag names. I'm not entirely happy with the ones I chose but I didn't want to get hung up on them while I was testing functionality.

endophage:
@dmcgowan ah, very good point. I'll take out the flag and conditional.

dmcgowan:
LGTM

As a follow up we should create a tracking issue to remove the pinging.

endophage:
Truly confirmed I was testing against staging this time. Everything looks to be working well. No deployments in the evening so I'll do a second pass in the morning and if everything is still OK, merge and deploy production before lunchtime.

N.B. there are no server changes in this PR, but I was making sure it works correctly against the updated master with server validation and improved logging. Both of which need to get out to prod.

PR #189, opened 2015-08-17
Don't blindly overwrite ctx with result of Authorized

aaronlehmann:
If Authorized returns an error, it returns nil for ctx. This means the
deferred function will dereference a nil pointer.


endophage:
LGTM

dmp42:
LGTM

Issue #144, opened 2015-07-28
File Changelist Iterator


endophage:
Create an iterator for the file changelist (also add appropriate iterator functions to the Changelist interface). It's inefficient to read all the changes into memory if there are a lot.

endophage:
Fixed in #191

PR #191, opened 2015-08-24
Add ChangeIterator interface for Changelists

davidwilliamson:
+ Ref https://github.com/docker/notary/issues/144
    + Create ChangeIterator interface
    + Implement ChangeIterator interface for memChangeList
    + Implement ChangeIterator interface for fileChangeList
    + Add iterator test case to changelist_test
    + Add iterator test case to file_changelist_test
    + Change func applyChangelist to use iterator per PR comment
    + Remove redundant defer statement in file_changelist_test.go (PR comment)
    + Change Next error check to simple array bounds check (PR comment)
    + Add negative unit test cases to increase code coverage

### DIFF
--- a/client/changelist/changelist.go
+++ b/client/changelist/changelist.go
@@ -32,3 +32,28 @@ func (cl *memChangelist) Clear(archive string) error {
 func (cl *memChangelist) Close() error {
 	return nil
 }
+
+func (cl *memChangelist) NewIterator() (ChangeIterator, error) {
+	return &MemChangeListIterator{index: 0, collection: cl.changes}, nil
+}
+
+// MemChangeListIterator is a concrete instance of ChangeIterator
+type MemChangeListIterator struct {
+	index      int
+	collection []Change // Same type as memChangeList.changes
+}
+
+// Next returns the next Change
+func (m *MemChangeListIterator) Next() (item Change, err error) {
+	if m.index >= len(m.collection) {
+		return nil, IteratorBoundsError(m.index)
+	}
+	item = m.collection[m.index]
+	m.index++
+	return item, err
+}
+
+// HasNext indicates whether the iterator is exhausted
+func (m *MemChangeListIterator) HasNext() bool {
+	return m.index < len(m.collection)
+}
--- a/client/changelist/changelist_test.go
+++ b/client/changelist/changelist_test.go
@@ -29,3 +29,38 @@ func TestMemChangelist(t *testing.T) {
 	cs = cl.List()
 	assert.Equal(t, 0, len(cs), "List should be empty")
 }
+
+func TestMemChangeIterator(t *testing.T) {
+	cl := memChangelist{}
+	it, err := cl.NewIterator()
+	assert.Nil(t, err, "Non-nil error from NewIterator")
+	assert.False(t, it.HasNext(), "HasNext returns false for empty ChangeList")
+
+	c1 := NewTufChange(ActionCreate, "t1", "target1", "test/targ1", []byte{1})
+	cl.Add(c1)
+
+	c2 := NewTufChange(ActionUpdate, "t2", "target2", "test/targ2", []byte{2})
+	cl.Add(c2)
+
+	c3 := NewTufChange(ActionUpdate, "t3", "target3", "test/targ3", []byte{3})
+	cl.Add(c3)
+
+	cs := cl.List()
+	index := 0
+	it, _ = cl.NewIterator()
+	for it.HasNext() {
+		c, err := it.Next()
+		assert.Nil(t, err, "Next err should be false")
+		assert.Equal(t, c.Action(), cs[index].Action(), "Action mismatch")
+		assert.Equal(t, c.Scope(), cs[index].Scope(), "Scope mismatch")
+		assert.Equal(t, c.Type(), cs[index].Type(), "Type mismatch")
+		assert.Equal(t, c.Path(), cs[index].Path(), "Path mismatch")
+		assert.Equal(t, c.Content(), cs[index].Content(), "Content mismatch")
+		index++
+	}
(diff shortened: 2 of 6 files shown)


endophage:
Can you add the iterator to the [memchangelist](https://github.com/docker/notary/blob/master/client/changelist/changelist.go) too and then replace the iteration [here](https://github.com/docker/notary/blob/master/client/helpers.go#L31) to use the iterator?

davidwilliamson:
@endophage updated per your comment.

client/changelist/file_changelist_test.go
```diff
@@ -78,3 +78,47 @@ func TestListOrder(t *testing.T) {
 	assert.Equal(t, c2.Path(), cs[1].Path(), "Path 2 mismatch")
 	assert.Equal(t, c2.Content(), cs[1].Content(), "Content 2 mismatch")
 }
+
+func TestFileChangeIterator(t *testing.T) {
+	tmpDir, err := ioutil.TempDir("/tmp", "test")
+	defer os.RemoveAll(tmpDir)
```
endophage: Looks like we have this same defer at line 89. One of them should be removed. Generally I put the defer after the error because the cleanup shouldn't be required if the command didn't complete successfully.

davidwilliamson:
@endophage Removed redundant defer statements in file_changelist_test.go per your PR comment

endophage:
LGTM. Test coverage looks great. Builds and runs correctly when publishing multiple changes.

@dmcgowan @aaronlehmann can one of you take a quick look and LGTM or comment.

client/changelist/changelist.go
```diff
@@ -32,3 +36,30 @@ func (cl *memChangelist) Clear(archive string) error {
 func (cl *memChangelist) Close() error {
 	return nil
 }
+
+func (cl *memChangelist) NewIterator() (ChangeIterator, error) {
+	return &MemChangeListIterator{index: 0, collection: cl.changes}, nil
```
aaronlehmann: The `index: 0` can be omited because fields default to the 0 value, but it doesn't hurt to have it here.

client/changelist/changelist.go
```diff
@@ -32,3 +36,30 @@ func (cl *memChangelist) Clear(archive string) error {
...
+	collection []Change // Same type as memChangeList.changes
+}
+
+// Next returns the next Change
+func (m *MemChangeListIterator) Next() (item Change, err error) {
+	defer func() {
+		if e := recover(); e != nil {
+			err = fmt.Errorf("Error in Next(): %v ", e)
+		}
+	}()
```
aaronlehmann: This seems like a convoluted way of doing a bounds check. I think it would be much better to explicitly check the slice bounds with `len`.

client/changelist/changelist_test.go
```diff
@@ -29,3 +29,36 @@ func TestMemChangelist(t *testing.T) {
...
+	cl.Add(c2)
+
+	c3 := NewTufChange(ActionUpdate, "t3", "target3", "test/targ3", []byte{3})
+	cl.Add(c3)
+
+	cs := cl.List()
+	index := 0
+	it, _ = cl.NewIterator()
+	for it.HasNext() {
+		c, err := it.Next()
```
aaronlehmann: Instead of iterating this way, you could do something like:

    for c, err := it.Next(); err != nil {

Then `HasNext` might be unnecessary.

client/changelist/file_changelist.go
```diff
@@ -94,6 +115,41 @@ func (cl FileChangelist) Close() error {
...
+	f := m.collection[m.index]
+	m.index++
+	item, err = unmarshalFile(m.dirname, f)
+	return
+}
+
+// HasNext indicates whether iterator is exhausted
+func (m *FileChangeListIterator) HasNext() bool {
+	return m.index < len(m.collection)
+}
```
aaronlehmann: See comments above about bounds checking and `HasNext`.

client/changelist/changelist_test.go
endophage: From the relatively small number of custom iterator examples I've found, having a check for the next element seems to be more popular than not. It makes it clear whether the loop is ending because of an error state, or because all iterations were completed.

client/changelist/changelist.go
endophage: Ah, good catch. We should define an error indicating Next was called after the iterator has already returned all items in the collection.

client/changelist/file_changelist.go
endophage: For the FileChangeListIterator, @davidwilliamson and I chatted this morning and we're going to add a bounds check to return a specific error in the instance that `Next` is called too many times, but also keep the defer to capture any filesystem errors, as unlikely as they may be.

aaronlehmann: Won't filesystem errors be returned in the "err" return value? If there's a panic somewhere, that seems like a bug that we should fix instead of arbitrarily adding a recover here.

This code would make sense in a language that uses exceptions for error handling, but this isn't how errors are handled in Go.

endophage: Ah you're right. I should just defer to you :-P

@davidwilliamson at the point where Go actually reads the file, the Go devs already put in this defer...recover step: http://golang.org/src/io/ioutil/ioutil.go#L23

davidwilliamson: Agreed. The Next() code needs to only do a simple array bounds check, as @aaronlehmann said.

davidwilliamson:
@endophage @aaronlehmann Updated based on PR comments. Also beefed up unit tests, particularly negative test cases for file_changelist.go (now 87% coverage)

Please take a look when you get a chance.

aaronlehmann:
LGTM

endophage:
LGTM

PR #192, opened 2015-09-02
replace godeps w bash for better control

jfrazelle:
you are welcome

```
$ git diff --shortstat  HEAD~1
 895 files changed, 325 insertions(+), 76322 deletions(-)
```


jfrazelle:
how do i change gopath on circle

jfrazelle:
eff circle, it passes locally, i even changed the config

jfrazelle:
flip

take it or leave it im over it

endophage:
@NathanMcCauley @diogomonica @dmcgowan Thoughts? My opinion is godeps has largely been working well for us.

jfrazelle:
but look at the minussss

jfrazelle:
i dont care either way just kinda gross having all the docker src in your
git branch too

endophage:
Mentioned it to @jfrazelle, thought about this a lot last night and the only thing that I really need/want is an equivalent of `godep update github.com/...` otherwise when I'm working on gotuf it's a pain to keep updating the dependencies so I can `docker-compose build`.

I think she's having a think about how to add that functionality.

jfrazelle:
so if gotuf switched to github.com/mattn/go-sqlite3  instead of  code.google.com/p/gosqlite we could get rid of another dep

diogomonica:
I don't usually like witch hunts, but this one is pretty great.

endophage:
@jfrazelle if the mattn one is the same one docker/docker uses I know there was some kind of issue with it. Possibly docker is using an older version that wasn't database/sql compatible?

jfrazelle:
yes it is

jfrazelle:
nah see https://github.com/docker/docker/pull/16054

endophage:
Cool, I'm happy to use that one instead. <3 database/sql interface

jfrazelle:
omg i won

jfrazelle:
the circle ci war

jfrazelle:
JESS: 1, CIRCLECI: 0

jfrazelle:
now that i finally got this to work, i think we should maybe try this 0:) https://github.com/FiloSottile/gvt

jfrazelle:
like a combo so there is still a lot of minus, but it compiles with the vendoring experiment as well

PR #195, opened 2015-09-10
Make static cross darwin builds work

### LINKED ISSUE #194: Cross binaries don't work on OS X
Do this:

```
osx$ make shell
shell-container$ make cross
shell-container$ exit
osx$ cross/darwin/amd64/notary init example.com/scripts
* fatal: cannot get current user: user: Current not implemented on darwin/amd64
```

Looks like there is no darwin support for finding the home dir for static builds.

This library looks like a good solution: https://github.com/mitchellh/go-homedir.

jbeda:
Fixes #194.  Static builds couldn't find homedir on darwin.

### DIFF
--- a/cmd/notary/main.go
+++ b/cmd/notary/main.go
@@ -3,11 +3,11 @@ package main
 import (
 	"fmt"
 	"os"
-	"os/user"
 	"path/filepath"
 	"strings"
 
 	"github.com/Sirupsen/logrus"
+	"github.com/mitchellh/go-homedir"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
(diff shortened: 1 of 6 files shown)


cmd/notary/main.go
```diff
@@ -37,14 +37,11 @@ func parseConfig() {
...
+		// Get home directory for current user
+		homeDir, err := homedir.Dir()
 		if err != nil {
-			fatalf("cannot get current user: %v", err)
+			fatalf("cannot get current user home directory: %v", err)
 		}
-
-		// Get home directory for current user
-		homeDir := usr.HomeDir
 		if homeDir == "" {
```
NathanMcCauley: `go-homedir` returns an `err` if it can't find a homedir. This check can safely be removed.

NathanMcCauley:
LGTM after duplicate error check is removed.

NathanMcCauley:
Thanks @jbeda

cmd/notary/main.go
jbeda: I left this in deliberately.  It is possible that go-homedir might (at some point in the future if not now) return no error and a blank string.  That could lead to the path being based, silently, on the current dir instead of the home dir.

I'm happy to remove it but it felt more defensive this way.  Let me know.

NathanMcCauley: Yeah, the defensiveness makes sense.

NathanMcCauley:
cc @diogomonica @endophage

PR #196, opened 2015-09-19
Minor improvements on server url

ggtools:
- Add the `--server` parameter for `verify``
- Add and environment variable to specify the default server URL

### DIFF
--- a/cmd/notary/main.go
+++ b/cmd/notary/main.go
@@ -72,6 +72,11 @@ func parseConfig() {
 }
 
 func main() {
+	serverURL := os.Getenv("NOTARY_SERVER_URL")
+	if serverURL == "" {
+		serverURL = defaultServerURL
+	}
+
 	var notaryCmd = &cobra.Command{
 		Use:   "notary",
 		Short: "notary allows the creation of trusted collections.",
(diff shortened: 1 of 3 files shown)


NathanMcCauley:
LGTM

thaJeztah:
should the environment variable be documented somewhere?

diogomonica:
LGTM

@thaJeztah can you please confirm the right location for these docs? The more I look at this, the less I want the same structure as docker/docker, where I can't find anything in :(

thaJeztah:
Well, I think it needs to be documented *somewhere*, but I'm not sure the current amount of docs here needs a whole page.

ping @moxiegirl any thoughts here? You've been doing a lot of research into the docs/navigation structure.

diogomonica:
@moxiegirl in particular, any chance we can do a simpler version of the docs for notary, instead of the docker style?

Maybe something like this? http://lyndaoleary.github.io/docker-api-docs/#list-containers

moxiegirl:
@diogomonica By "docker look" You mean the theme? Yes, I'm working on that right now. I have a redesign from Josephine on the UX team. It is similar to the Readme.io look that seems to be rising in fashion.  I'm about to close on a contract with a scripting team to implement this shortly.

@diogomonica I'd prefer we talk in person about your concern's with the docs structure.  @thaJeztah isn't really responsible for that..

ggtools:
@thaJeztah I agree that, at the moment, having of whole page for an environment variable is way overkill. However in the current documentation state I didn't find a *natural* place to put it. Beside, it does not seem totally awkward to have, in the future, a page dedicated to the cli usage.

moxiegirl:
LGTM on current format. If we need to move bits around we can do that before the release.

PR #202, opened 2015-10-05
Mention where the notary binary is

vikstrous:
Normally go users expect binaries to be put in $GOPATH/bin, so it's worth mentioning where the binaries are output to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,8 @@ docker-compose build
 docker-compose up -d
 ```
 
+The `notary` binary can be found in `./bin/notary`.
+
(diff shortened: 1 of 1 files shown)


NathanMcCauley:
@vikstrous Maybe we should be putting things into $GOPATH/bin then?

vikstrous:
I assumed that there's a reason they are put in ./bin. You can use `go install`. It might make sense to just add another make target called `install` and use that in the readme.

endophage:
@vikstrous is right, the best solution would be to add `make install` that puts the binaries in `$GOPATH/bin`.

Issue #190, opened 2015-08-18
Add health check to notary server


endophage:
Use the distribution health check package to add health checks to notary server

avinson:
when implementing this, we also need a way to do health checks against the main port and not the debug port. there was recently a distribution PR that got merged for this: https://github.com/docker/distribution/pull/874

endophage:
Not the best place for this but a few things I'd love to see changed in the distribution health check package:

1. Always return all the service names with "OK" if there was no error. It's currently annoying to tell if all the service health checks have been registered correctly.
2. Periodic checks do not do an immediate check. You have to wait until the end of the first period until a check is done for the first time. With a periodic check it should immediately do the first check on registration.

PR #205, opened 2015-10-08
Adding a DB check to the signer health check

cyli:
Part of addressing #190, since the health check for notary server needs to make some sort of RPC call to the signer (to see if it is reachable and pokeable).

This also involved a large change to the dependencies after regenerating the protocol code.   The PR without the dependency changes can be viewed here:  https://github.com/docker/notary/compare/master...cyli:2740fd0

I plan on adding the health checks to notary-server in a separate PR.


diogomonica:
LGTM

PR #209, opened 2015-10-10
Strawman for targets and snapshot key rotation

endophage:
In b4 @NathanMcCauley: tests?

Will add tests, looking for feedback. The interface is super simple, `notary key rotate [GUN]`, but seems to make sense based on the current method of operation (as far as how targets and snapshot keys work).

Is the `key` subcommand the right place to put this? Not sure, feel a little out of place contextually (`key rotate` is the only `key` sub-subcommand that modifies TUF data), but intuitively the CLI command is better than anything else I've come up with.

p.s. the server validation is awesome, it helped me catch some errors when I was debugging the rotation.


endophage:
Tests added

diogomonica:
LGTM

PR #210, opened 2015-10-10
Adding client-side root-ca server and server config

diogomonica:
- Allows loading a config from the command-line using `-configFile` or `-c`.
- Config now checks for remote server's address under `remote_server.addr`, instead of ENV variable
- Config now checks for root CA and adds it to our transport, instead of needing to disable TLS verification

@cyli @endophage 


cmd/notary/main.go
```diff
@@ -45,17 +51,27 @@ func parseConfig() {
...
 	} else {
 		logrus.Debugf("trust directory provided: %s", trustDir)
 	}
 
-	// Setup the configuration details
+	// If there was a commandline configFile set, we parse that.
+	// If there wasn't we attempt to find it on the default location ~/.notary/config
+	if configFile != "" {
+		configFileExt = strings.TrimPrefix(filepath.Ext(configFile), ".")
+		configFileName = strings.TrimSuffix(filepath.Base(configFile), filepath.Ext(configFile))
```
cyli: Would this mean the config file name would end with a `.`?  Since the `.` is trimmed off the `configFileExt`?

cyli:
Tested these options, and LGTM.  :)  Do we want to update the README too with the new options?

diogomonica:
I'll do that. @cyli should I update the examples to all use `-c cmd/notary/config.json` so it works out of the box?

cyli:
@diogomonica: That sounds great, thanks!  :D

vikstrous:
What is the path relative to? Wouldn't it be better to have it relative to the location of the config file instead of PWD?

cmd/notary/tuf.go
```diff
@@ -357,11 +359,23 @@ func (ps passwordStore) Basic(u *url.URL) (string, string) {
 }
 
 func getTransport(gun string, readOnly bool) http.RoundTripper {
+	// Attempt to get a root CA from the config file. Nil is the host defaults.
+	rootPool := x509.NewCertPool()
+	rootCAFile := viper.GetString("remote_server.root_ca")
+	if rootCAFile != "" {
+		rootCert, err := trustmanager.LoadCertFromFile(viper.GetString("remote_server.root_ca"))
```
vikstrous: You can use `rootCAFile` here. Also, https://github.com/docker/notary/pull/210#issuecomment-147118693

diogomonica: You mean instead of `rootCert`? They are different types. `LoadCertFromFile` returns `*x509.Certificate` and `rootCaFile` is a string.

vikstrous: No, I mean you are calling `viper.GetString("remote_server.root_ca")` twice.

diogomonica: Ah, my bad.

diogomonica:
@vikstrous it's like any other utility, it receives a config file (either full path or relative to ".").

vikstrous:
What I'm wondering is if it matters where you call `notary` from. Is the path to the CA relative the PWD at the time you call notary? If that's the case, `notary` can't be called from other directories.

cmd/notary/main.go
```diff
@@ -7,24 +7,30 @@ import (
 	"strings"
 
 	"github.com/Sirupsen/logrus"
-	"github.com/mitchellh/go-homedir"
-	"github.com/spf13/cobra"
-	"github.com/spf13/viper"
-
 	"github.com/docker/notary/pkg/passphrase"
 	"github.com/docker/notary/version"
+	homedir "github.com/mitchellh/go-homedir"
```
endophage: You're breaking the nice layout we had of package groups :-P

```
import (
    // core packages

    // 3rd party packages

    // same repo packages
)

Comment (diogomonica) [inline cmd/notary/main.go]: Talk to GoImport.

File: cmd/notary/config.json [hunk 4]
```diff
@@ -0,0 +1,7 @@
+{
+	"remote_server": {
+		"url": "https://localhost:4443",
```
cyli: Setting this to "https://notary-server:4443" might make this more friendly for those whose docker hosts are not on localhost (since the README already suggests adding a `notary-server` entry to /etc/hosts).

cmd/notary-server/dev-config.json
```diff
@@ -1,20 +0,0 @@
-{
```
endophage: Why is this being deleted? Accident?

cmd/notary/config.json
endophage: The hardcoded default is `https://notary-server:4443` so it would make sense to use that here too.

cmd/notary/tuf.go
```diff
@@ -357,11 +360,23 @@ func (ps passwordStore) Basic(u *url.URL) (string, string) {
 }
 
 func getTransport(gun string, readOnly bool) http.RoundTripper {
+	// Attempt to get a root CA from the config file. Nil is the host defaults.
+	rootPool := x509.NewCertPool()
+	rootCAFile := viper.GetString("remote_server.root_ca")
+	if rootCAFile != "" {
+		rootCert, err := trustmanager.LoadCertFromFile(filepath.Join(configPath, rootCAFile))
```
cyli: Non-blocking:  This does not work if an absolute path is passed - it will just join the configPath with the absolute path, which doesn't exist.  Probably if we wanted to support absolute paths as well as relative paths, there could be a `IsAbs` check first?

Otherwise, if we didn't want to support absolute paths, we should probably that the config file only supports paths relative to the config path.

notary-signer-Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM diogomonica/golang-softhsm2
+FROM dockersecurity/golang-softhsm2
```
cyli: Since `docker.io/dockersecurity/golang-softhsm2` doesn't exist, `docker-compose up` will fail.  Can we add another instruction to the README to `cd signer; docker build -t dockersecurity/golang-softhsm2 .`?

cmd/notary/config.json
diogomonica: the problem is that notary-server isn't set in /etc/hosts by default, and I wanted something that works out of the box.

cmd/notary-server/dev-config.json
diogomonica: We no longer need it. The only difference between configs was the TLS config. Now we have a way for everything to work out of the box for mac and linux, so we no longer need dev.

endophage: Ah, awesome!

cmd/notary/config.json
cyli: That's true - it doesn't for users of `docker-machine`, though, unless they SSH forward or remap localhost to the VM.  I'm fine either way, so long as we document a single way to get things to work on both docker-on-localhost and docker-machine.

If we want to go with localhost, then I'd suggest SSH-forwarding for those who use docker-machine (since remapping localhost seems kind of ugly):  `ssh docker@$(docker-machine ip default) -L 4443:localhost:4443` ?

notary-signer-Dockerfile
cyli: Oops, never mind, thanks for public-izing that repo, @diogomonica!

diogomonica: Fixed. The repo was private.

cmd/notary/tuf.go
diogomonica: Fixed

cmd/notary/config.json
diogomonica: You're right. Forgot about macs. Everything is notary-server now.

cmd/notary-server/dev-config.json
cyli: I have no opinions on whether we should also keep a `dev-config.json`, but I think the one being deleted uses a local trust, as opposed to a remote notary-signer service in the regular `config.json`.

diogomonica: Indifferent if you're using `docker-compose up`

README.md
```diff
@@ -42,7 +42,8 @@ Prerequisites:
 - [docker and docker-compose](http://docs.docker.com/compose/install/)
 - [Notary server configuration](#configure-notary-server)
 
-As setup, lets build notary and then start up a local notary-server.
+As setup, lets build notary and then start up a local notary-server (don't forget
+to add `127.0.0.1 notary-server` to your `/etc/hosts`).
```
cyli: Non-blocking: "or `<docker-machine IP> notary-server` if you are using docker-machine" maybe?  Probably overly pedantic.

cyli:
LGTM - everything seems to work as it should.  I tested:

- Using the config:
    - `bin/notary -c ./cmd/notary/config.json ...` succeeds
    - `bin/notary -c cmd/notary/config.json ...` succeeds
    - `bin/notary -c ~/Go/src/.../cmd/notary/config.json ...` succeeds
    - `bin/notary -c $(pwd)/cmd/notary/config.json ...` succeeds
- HTTP still works (`bin/notary -s http://notary-server:4443 ...` suceeds)
- Deleting the root-ca from `cmd/notary` results in a cert error
- moving root-a and config to a `/tmp/tempnotary` directory and setting that as the config path (`bin/notary -d /tmp/tempnotary ...` succeeds)
- moving the root-ca to my home directory and changing `config.json` to point to:
    - `../root-ca.crt` succeeds
    - `/<abspath>/root-ca.crt` succeeds
    - `~/Go/src/.../root-ca.crt` fails, but it looks like [Go `filepath` does not support expanding tilde out of the box](https://stackoverflow.com/questions/17609732/expand-tilde-to-home-directory), so I think this may be just be fine?

PR #211, opened 2015-10-10
Convert the server DB tests to use GORM with SQLite3 instead of sqlmock

cyli:
In preparation for gorm-izing the DB itself.

This way we have passing tests we know work against the previous non-gorm storage package before gorm-izing the storage package.

### DIFF
--- a/notarymysql/migrate.sql
+++ b/notarymysql/migrate.sql
@@ -1,18 +1,16 @@
 -- This migrates initial.sql to tables that are needed for GORM
 
 ALTER TABLE `tuf_files`
-ADD COLUMN `created_at` timestamp NULL AFTER `id`,
-ADD COLUMN `updated_at` timestamp NULL AFTER `created_at`,
-ADD COLUMN `deleted_at` timestamp NULL AFTER `updated_at`,
+ADD COLUMN `created_at` timestamp NULL DEFAULT NULL AFTER `id`,
+ADD COLUMN `updated_at` timestamp NULL DEFAULT NULL AFTER `created_at`,
+ADD COLUMN `deleted_at` timestamp NULL DEFAULT NULL AFTER `updated_at`,
 MODIFY `id` int(10) unsigned AUTO_INCREMENT;
 
 ALTER TABLE `timestamp_keys`
 ADD COLUMN `id` int(10) unsigned AUTO_INCREMENT FIRST,
-ADD COLUMN `created_at` timestamp NULL AFTER `id`,
-ADD COLUMN `updated_at` timestamp NULL AFTER `created_at`,
-ADD COLUMN `deleted_at` timestamp NULL AFTER `updated_at`,
+ADD COLUMN `created_at` timestamp NULL DEFAULT NULL AFTER `id`,
+ADD COLUMN `updated_at` timestamp NULL DEFAULT NULL AFTER `created_at`,
+ADD COLUMN `deleted_at` timestamp NULL DEFAULT NULL AFTER `updated_at`,
 DROP PRIMARY KEY,
 ADD PRIMARY KEY (`id`),
 ADD UNIQUE (`gun`);
-
-
--- a/notarymysql/migrate.sql
+++ b/notarymysql/migrate.sql
@@ -1,0 +1,18 @@
+-- This migrates initial.sql to tables that are needed for GORM
+
+ALTER TABLE `tuf_files`
+ADD COLUMN `created_at` timestamp NULL AFTER `id`,
+ADD COLUMN `updated_at` timestamp NULL AFTER `created_at`,
+ADD COLUMN `deleted_at` timestamp NULL AFTER `updated_at`,
+MODIFY `id` int(10) unsigned AUTO_INCREMENT;
+
+ALTER TABLE `timestamp_keys`
+ADD COLUMN `id` int(10) unsigned AUTO_INCREMENT FIRST,
+ADD COLUMN `created_at` timestamp NULL AFTER `id`,
+ADD COLUMN `updated_at` timestamp NULL AFTER `created_at`,
+ADD COLUMN `deleted_at` timestamp NULL AFTER `updated_at`,
+DROP PRIMARY KEY,
+ADD PRIMARY KEY (`id`),
+ADD UNIQUE (`gun`);
+
+
--- a/server/storage/database_test.go
+++ b/server/storage/database_test.go
@@ -260,10 +260,6 @@ func TestMySQLGetCurrent(t *testing.T) {
 func TestMySQLDelete(t *testing.T) {
 	gormDB, dbStore := SetUpSQLite(t)
 
-	// Not testing deleting from an empty table, because that's not an error
-	// in SQLite3
-
-	// use UpdateCurrent to create one and test GetCurrent
 	tuf := SampleTUF(0)
 	query := gormDB.Create(&tuf)
 	assert.NoError(t, query.Error, "Creating a row in an empty DB failed.")
--- a/server/storage/database_test.go
+++ b/server/storage/database_test.go
@@ -38,6 +38,25 @@ func (g GormTimestampKey) TableName() string {
 	return "timestamp_keys"
 }
 
+// SampleTUF returns a sample GormTUFFile with the given Version (ID will have
+// to be set independently)
+func SampleTUF(version int) GormTUFFile {
+	return GormTUFFile{
+		Gun:     "testGUN",
+		Role:    "root",
+		Version: version,
+		Data:    []byte("1"),
+	}
+}
+
+func SampleUpdate(version int) MetaUpdate {
+	return MetaUpdate{
+		Role:    "root",
+		Version: version,
+		Data:    []byte("1"),
+	}
+}
+
 // SetUpSQLite creates a sqlite database for testing
 func SetUpSQLite(t *testing.T) (*gorm.DB, *MySQLStorage) {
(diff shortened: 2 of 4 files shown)


endophage:
Love the test changes coming first, looks good at a quick glance! Will look closer over the weekend.

server/storage/database_test.go
```diff
@@ -2,290 +2,330 @@ package storage
...
 
-	//err = db.Close()
-	//assert.Nil(t, err, "Expectation not met: %v", err)
+	dbStore.DB.Close()
+}
+
+func TestMySQLDelete(t *testing.T) {
+	gormDB, dbStore := SetUpSQLite(t)
+
+	// Not testing deleting from an empty table, because that's not an error
```
cyli: I plan on adding this test back in when the `database.go` implementation has been converted, because we can check the number of rows affected to see if the delete actually happened.

cyli:
Actually I'm wondering if I approached this wrong - should I have written these tests only in terms of the `MetaStore` api (like the memory storage), so that the tests can be shared between any storage implementation (such as both sql and memory)?

endophage:
@cyli that is the optimal way to do it but you sometimes end up with split code as the setup can call the interface functions and be shared, but sometimes you want the validation to dig into the internals of an implementation.

Go generally favours a bit of repetition here and there (one of the necessities of not having generics, and interfaces don't fit everything) so this is fine.

server/storage/database_test.go
```diff
@@ -2,290 +2,326 @@ package storage
...
-		"testGUN",
-		update,
-	)
-	assert.Nil(t, err, "UpdateCurrent errored unexpectedly: %v", err)
-
-	// There's a bug in the mock lib
-	//err = db.Close()
-	//assert.Nil(t, err, "Expectation not met: %v", err)
+// GormTUFFile represents a TUF file in the database
+type GormTUFFile struct {
```
endophage: Given we'll want to use the models when we do the full GORM integration, it would be good to put them in their own non-test file within the storage package. If it doesn't conflict with any else in the storage package, it would also be good to drop the `Gorm` prefix.

Rather than putting an `ID` field in, let's use the [`gorm.Model`](https://github.com/jinzhu/gorm#conventions) embed and fix the databases in stage/prod to have the `created_at`, `updated_at`, and `deleted_at` columns.

server/storage/database_test.go
```diff
@@ -2,290 +2,326 @@ package storage
...
+	return "tuf_files"
 }
 
-func TestMySQLUpdateMany(t *testing.T) {
-	db, err := sqlmock.New()
-	assert.Nil(t, err, "Could not initialize mock DB")
-	s := NewMySQLStorage(db)
-	update1 := MetaUpdate{
+// GormTimestampKey represents a single timestamp key in the database
+type GormTimestampKey struct {
```
endophage: Let's add `gorm.Model` embeds here too and drop the Gun as the primary key. The unique constraint is sufficient. Soft deletes would be really useful to have on timestamp keys so we can inspect how frequently people are doing TS key rotations (once we add those).

server/storage/database_test.go
```diff
@@ -2,290 +2,326 @@ package storage
...
-		Role:    "root",
-		Version: 0,
-		Data:    []byte("1"),
+// SetUpSQLite creates a sqlite database for testing
+func SetUpSQLite(t *testing.T) (*gorm.DB, *MySQLStorage) {
+	tempBaseDir, err := ioutil.TempDir("", "notary-test-")
+	defer os.RemoveAll(tempBaseDir)
+
+	// We are using SQLite for the tests
+	db, err := sql.Open("sqlite3", tempBaseDir+"test_db")
```
endophage: Gorm's open function is surprisingly versatile, there's no need to do a `sql.Open`, you can just pass these same arguments to a `gorm.Open`

server/storage/database_test.go
```diff
@@ -2,290 +2,326 @@ package storage
...
-	assert.Nil(t, err, "Could not initialize mock DB")
-	s := NewMySQLStorage(db)
-	update := MetaUpdate{
-		Role:    "root",
-		Version: 0,
-		Data:    []byte("1"),
+// SetUpSQLite creates a sqlite database for testing
+func SetUpSQLite(t *testing.T) (*gorm.DB, *MySQLStorage) {
+	tempBaseDir, err := ioutil.TempDir("", "notary-test-")
+	defer os.RemoveAll(tempBaseDir)
```
endophage: I don't think this is going to do what you want. This will delete the directory when `SetUpSQLite` exits, I think you want to delete it when the test that called this function exits. Possibly move this line and the line above to the individual test functions and pass in the `tempBaseDir` to this function.

server/storage/database_test.go
```diff
@@ -2,290 +2,326 @@ package storage
...
+
+	// We are using SQLite for the tests
+	db, err := sql.Open("sqlite3", tempBaseDir+"test_db")
+	assert.NoError(t, err)
+
+	// Create the DB tables
+	gormDB, _ := gorm.Open("sqlite3", db)
+	query := gormDB.CreateTable(&GormTUFFile{})
```
endophage: Move this line and everything below into its own function as its DB independent. Would also be good to move it outside a test file so we can reuse it in a cmd to initialize the DB.

server/storage/database_test.go
```diff
@@ -2,290 +2,326 @@ package storage
...
+	query := gormDB.Find(&rows)
+	assert.NoError(t, query.Error)
+	assert.Equal(t, []GormTUFFile{gorm1, gorm2}, rows)
+
+	dbStore.DB.Close()
+}
+
+// TestMySQLUpdateManyDuplicateRollback asserts that inserting duplicate
+// updates fails.
+func TestMySQLUpdateManyDuplicateRollback(t *testing.T) {
```
endophage: I love this test!

server/storage/database_test.go
cyli: I originally thought so too, but since the DB seemed fine during the tests, I got confused and thought maybe I misunderstood how this worked.  Thanks for noticing!  :)

endophage: Yeah, I'm equally confused why we didn't see problems. Unfortunately Gorm (and other DB drivers I've used), appear to fail silently in some edge cases. There's probably something around once it has opened the file handle, sqlite won't fail if it suddenly can't write. Might explain why we were seeing ID=0 being returned on querying the DB... but doesn't explain why the other fields were still populated.

server/storage/database_test.go
cyli: There is currently no unique constraint on the `timestamp_keys` structure - I think uniqueness was enforced by the GUN being the primary key.   To clarify, do you mean add a unique constraint to the Gun column?

Also, would this mean that this is incompatible with any existing tables?  (can you migrate from one primary key to another, and if so, should we be writing those migrations?)

endophage: You're absolutely correct, uniqueness was being enforced by the fact `Gun` was the primary key. We will need to add a separate uniqueness constraint.

Because there are no foreign key relationships, we can do this migration of primary key. Sadly we won't be able to do it in Gorm, it will have to be a hand crafted sql file as it's a little fiddly. You can take a crack at it and add it to this PR or I can send you a PR against this branch on your fork. Let me know.

Once we merge this PR, I'll look at updating the DB. I believe the changes proposed will be compatible with the existing production code but I'll take another look when the time comes.

cyli:
@endophage I think all the comments are addressed (if I understood them correctly, anyway!  :)), and I made a stab at adding the migration SQL.

I tried running the `initial.sql` + `migrate.sql`, vs just using the `gorm` create table functions, and there seems to be no difference in schema (the gorm ID seems to be `int(10)` and `initial.sql` creates IDs with `int(11)`, so I just altered them to be `int(10)`).

However, the charset seems to be different - when created with `initial.sql`, the charset as displayed by `show status table where name = '<tablename>'` of the tables is `utf8_general_ci`.  The charset produced using the gorm is `utf8_unicode_ci`.  

Any suggestions on fixing this?  I'm passing the default charset to the table options when creating the GORM table.

notarymysql/migrate.sql
```diff
@@ -0,0 +1,18 @@
+-- This migrates initial.sql to tables that are needed for GORM
+
+ALTER TABLE `tuf_files`
+ADD COLUMN `created_at` timestamp NULL AFTER `id`,
```
endophage: We had some bugs with just setting `NULL`, it turned out, as redundant as it seems, these need to be `NULL DEFAULT NULL` for MySQL to behave as expected. It's something special to do with the timestamp type, can't remember exactly what is was now.

endophage:
The int(10) vs int(11) stuff looks fine. Good to just make it uniform with Gorm so nobody looks back and wonders if there was some special reason for it to differ.

Not sure why the difference in charset... A `SHOW CREATE TABLE <tablename>` on Gorm vs manually created tables just shows "charset=utf8" for both... This [SO post](http://stackoverflow.com/questions/766809/whats-the-difference-between-utf8-general-ci-and-utf8-unicode-ci) has a good first answer, just sorting/comparison differences between the 2 so I'm not particularly concerned.

endophage:
LGTM!

cyli:
@endophage Oops, sorry!  It collate, not charset.  Should I add that to the gorm table options, or do we not care?

endophage:
@cyli yeah, I had a feeling it might have been collation. Don't really care in this instance. The only thing we might sort on that it would affect is the `gun`, and I don't see us every actually sorting on that for anything but internal analytics, at which point we can deal with sorting possibly being a little out of order.

PR #212, opened 2015-10-11
add circle ci badge

vikstrous:
This makes it easier for new people to see that we are using circle ci and it shows that our tests pass on master.


cyli:
Yay badge!  :D  Great idea!

PR #215, opened 2015-10-12
Use gorm for storing notary-server information

cyli:
This PR:

1.  Uses a GORM implementation for the notary-server store
1.  Changes the default mysql container to run both the initial and migrate sql
1.  Hopefully enables notary-server to be SQL-dialect agnostic.  Currently still errors if anything other than mysql is returned.

Caveat:  This does slightly less error checking with existing entries in the DB - previously, the code checked for a MySQL error code 1022 even after first doing a read to make sure the entry isn't there, but if we're SQL-DB agnostic, we're probably going to have to support all the different kinds of SQL error codes ourselves if we wanted to do the same check.  Should I add this back in?  I also do not know how to trigger this race condition for testing, to make sure we're covering it.


cyli:
Oops, I should add a health check.

server/storage/database.go
```diff
@@ -1,177 +1,165 @@
...
-			if err.Number == 1022 { // duplicate key error
-				return &ErrOldVersion{}
-			}
-		}
-		// need to check error type for duplicate key exception
-		// and return ErrOldVersion if duplicate
-		return err
-	}
-	return nil
+	return db.Create(&TUFFile{
```
endophage: Might be good to capture the error and check a key error in case a race condition has caused this to be written by another process. Will allow us to raise the ErrOldVersion and handle it more consistently.

cyli: This was the caveat in the PR description - should we be capturing error codes for every DB we support?  Gorm doesn't extract duplicate entry at all.

endophage: Sorry, didn't make the mental connection. That makes sense. Ignore my initial comment :-P

server/storage/database.go
```diff
@@ -1,177 +1,165 @@
...
+	if tx.Error != nil {
+		return tx.Error
+	}
 
-	tx, err := db.Begin()
-	if err != nil {
+	rollback := func(err error) error {
+		query := tx.Rollback()
+		if query.Error != nil {
+			logrus.Panic("Failed on Tx rollback with error: ", err.Error())
```
endophage: I think you meant to log `query.Error` here?

endophage: Also shouldn't panic once we're into real runtime code (as opposed to initialization code). It'll take down the server.

server/storage/database.go
```diff
@@ -1,177 +1,165 @@
...
+	tx := db.Begin()
+	if tx.Error != nil {
+		return tx.Error
+	}
 
-	tx, err := db.Begin()
-	if err != nil {
+	rollback := func(err error) error {
+		query := tx.Rollback()
+		if query.Error != nil {
```
endophage: Can simplify a little with an idiomatic Go pattern:

```
if err := tx.Rollback().Error; err != nil {
    logrus.Panic(...)
}
```

server/storage/database.go
cyli: We can make a best effort translation function.  Maybe something like:

```
// translateOldVersionError captures DB errors, and attempts to translate
// duplicate entry - currently only supports MySQL and Sqlite3
func translateOldVersionError(err error) error {
    if err, ok := err.(*mysql.MySQLError); ok {
        // https://dev.mysql.com/doc/refman/5.5/en/error-messages-server.html
        // 1022 = Can't write; duplicate key in table '%s'
        // 1062 = Duplicate entry '%s' for key %d
        if err.Number == 1022 || err.Number 1062 == {
            return &ErrOldVersion{}
        }
    } else if err, ok := err.(*sqlite3.Error); ok {
        // https://godoc.org/github.com/mattn/go-sqlite3#pkg-variables
        if (err.Code == sqlite3.ErrConstraint &&
            err.ExtendedCode == sqlite3.ErrConstraintUnique) {
            return &ErrOldVersion{}
        }
    }
    return err
}
```
?

server/storage/database.go
```diff
@@ -1,177 +1,165 @@
...
-			if rbErr != nil {
-				logrus.Panic("Failed on Tx rollback with error: ", err.Error())
-			}
-			if err, ok := err.(*mysql.MySQLError); ok && err.Number == 1022 { // duplicate key error
-				return &ErrOldVersion{}
-			}
-			return err
+		// it's previously been added, which means it's a duplicate entry
+		// in the same transaction
+		if added[row.ID] {
```
endophage: Just to avoid any weirdness (though this should be fine): 

```
if _, ok := added[row.ID]; ok {
    return rollback(&ErrOldVersion{})
}
```

server/storage/database_test.go
```diff
@@ -31,33 +31,35 @@ func SampleUpdate(version int) MetaUpdate {
 }
 
 // SetUpSQLite creates a sqlite database for testing
-func SetUpSQLite(dbDir string, t *testing.T) (*gorm.DB, *MySQLStorage) {
-	db, err := gorm.Open("sqlite3", dbDir+"test_db")
+func SetUpSQLite(dbDir string, t *testing.T, createTable bool) (*gorm.DB, *SQLStorage) {
```
endophage: Just for consistency, it's preferable to pass the testing argument first.

endophage:
Only thing I'm concerned about is the Panic. Everything else is stylisitc and would be fine as is.

server/storage/database.go
endophage: Hmmmm, yeah, that might be good. ErrOldVersion helps provide more meaningful feedback to users. Can do both: best effort translation using type switch :-)

server/storage/database.go
cyli: Ah ok, I was just directly translating from https://github.com/docker/notary/blob/master/server/storage/database.go#L90.  Should the rollback error be logged instead?

endophage: Ah, yeah, I must have copy pasted and not fixed it. Should definitely log the rollback error here.

cyli:
This seems to be failing because 2 types of sqlite3 get imported in handler validation checking (gotuf imports a different one).

Waiting on https://github.com/endophage/gotuf/pull/46.

cyli:
Apologies for the large change due to Gotuf update.  https://github.com/docker/notary/compare/master...cyli:e786d62  is the non-Godeps changes.

diogomonica:
LGTM.

Issue #213, opened 2015-10-12
make fails with “Makefile:52: *** target pattern contains no '%'.  Stop.”


mtrmac:
PR #207 adds rules like
```${GOPATH}/bin/notary-server: NOTARY_VERSION $(shell find . -type f -name '*.go')```
which break when GOPATH contains multiple :-separated entries with
```Makefile:52: *** target pattern contains no '%'.  Stop.```
(regardless of which target is used.)

I’m not entirely sure what a proper fix is at the moment, this is a tangle of several issues.

* https://golang.org/pkg/go/build/ suggests that $gopath_entry/bin should be only used for commands built from the corresponding $gopath_entry/src, which is not generally the case for a checkout of notary.
* This can be worked around by something like ```(unset GOPATH; make binaries)```, only using the Godeps-managed dependencies
* OTOH ```(make build)``` is still available, requires running setting GOPATH manually to point at Godeps, and the standard way to do that, ```GOPATH=`godep path`:$GOPATH```, causes a multi-entry GOPATH to be set.
* (make build) could arguably be dropped or reworked as (make binaries) has been, but this also breaks (go test), for the same reason.

AFAICT either all or none of the make rules should call ```godep```; and using ```${GOPATH}``` is wrong in any case.

(Filing at least a bug without a PR for now to hopefully save others the time figuring out the cause of the breakage.)

vikstrous:
Sorry about that. I'll revert it for now until we have a better solution https://github.com/docker/notary/pull/214

cyli:
Looks like this issue was resolved; closing.  Should we re-open #202? (putting the compiled binaries into `$GOPATH/bin`)

PR #216, opened 2015-10-13
Use canonical import paths instead of forked repo name

mtrmac:
The code referes to canonical import paths, so the code needs to be
available there even if CircleCI is building from a forked repository.

Otherwise the tests fail, e.g. https://circleci.com/gh/mtrmac/notary/3 vs. https://circleci.com/gh/mtrmac/notary/4 .

PR #217, opened 2015-10-13
Use Makefile targets for test/pre to avoid duplication

mtrmac:
Thus, we will actually test that the Makefile works.

Doesn’t yet convert test/override because the two already differ and
it’s not immediately obvious to me which one to use.

This should help to avoid future regressions like #213 , (well, apart from not converting the actual test execution ☺).

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -45,15 +45,15 @@ test:
     - gvm use stable && go version
 
   # FMT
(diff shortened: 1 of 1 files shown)


mtrmac:
(The test failure above will be resolved by merging #216.)

endophage:
#216 merged, please rebase

diogomonica:
LGTM. Thanks @mtrmac

endophage:
LGTM pending circleci passing

mtrmac:
Rebased, now passes.

PR #218, opened 2015-10-14
updating gotuf in godeps

endophage:
That is all

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -96,7 +96,7 @@
 		},
 		{
 			"ImportPath": "github.com/endophage/gotuf",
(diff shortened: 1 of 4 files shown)


diogomonica:
Tests failing. Seem to be a real failure.

endophage:
Failure was because downloadSigned obscured the raw data from the callers, which they required to populate cache. This mean the root never got cached (and it's the only thing that needs to be), so update was failing on the check of the snapshot hash for root vs the cache (the first time through, it fails and downloads the roots again, but when it fails on the update retry, it errors out of the update).

diogomonica:
LGTM, but I've been wrong in the past.

PR #219, opened 2015-10-15
Healthchecks for notary-server

### LINKED ISSUE #190: Add health check to notary server

cyli:
Fixes #190.

This PR makes notary-server unhealthy if:
~~- the connection to the trust service is dead~~
~~- it takes too long to answer an RPC call~~
~~- the trust signer service is unhealthy~~
~~- the trust key manager service is unhealthy~~
- the DB does not have the right tables

Update: It logs an error if the following are true:
- the connection to the trust service is dead
- it takes too long to answer an RPC call
- the trust signer service is unhealthy
- the trust key manager service is unhealthy

Since the server can operate with the signer down, but someone should get paged.

Note that the DB table query times out after a little while, the GRPC one seemed to hang forever (by which I mean about 10 minutes by wall clock).  It keeps trying to reconnect, but meanwhile the outstanding RPC call blocks.

The timeout should probably be in the distribution/health package, but since passing timeout arguments to all the functions breaks the existing API, maybe the health package should be split out and then projects can pin to a particular version or not?

### DIFF
--- a/cmd/notary-server/main.go
+++ b/cmd/notary-server/main.go
@@ -113,21 +113,20 @@ func main() {
 			viper.GetString("trust_service.port"),
 			viper.GetString("trust_service.tls_ca_file"),
 		)
+		minute := 1 * time.Minute
 		health.RegisterPeriodicFunc(
 			"Trust operational",
 			// If the trust service fails, the server is degraded but not
 			// exactly unheatlthy, so always return healthy and just log an
 			// error.
 			func() error {
-				logrus.Info("Getting health")
-				err := trust.(*signer.NotarySigner).CheckHealth(5)
-				logrus.Info("Got health")
+				err := trust.(*signer.NotarySigner).CheckHealth(minute)
 				if err != nil {
 					logrus.Error("Trust not fully operational: ", err.Error())
 				}
 				return nil
 			},
-			time.Second*5)
+			minute)
 	} else {
 		logrus.Info("Using local signing service")
 		trust = signed.NewEd25519()
--- a/cmd/notary-server/main.go
+++ b/cmd/notary-server/main.go
@@ -108,11 +108,12 @@ func main() {
 	var trust signed.CryptoService
 	if viper.GetString("trust_service.type") == "remote" {
 		logrus.Info("Using remote signing service")
-		trust = signer.NewNotarySigner(
+		notarySigner := signer.NewNotarySigner(
 			viper.GetString("trust_service.hostname"),
 			viper.GetString("trust_service.port"),
 			viper.GetString("trust_service.tls_ca_file"),
 		)
+		trust = notarySigner
 		minute := 1 * time.Minute
 		health.RegisterPeriodicFunc(
 			"Trust operational",
@@ -120,7 +121,7 @@ func main() {
 			// exactly unheatlthy, so always return healthy and just log an
 			// error.
 			func() error {
-				err := trust.(*signer.NotarySigner).CheckHealth(minute)
+				err := notarySigner.CheckHealth(minute)
 				if err != nil {
 					logrus.Error("Trust not fully operational: ", err.Error())
 				}
--- a/cmd/notary-server/main.go
+++ b/cmd/notary-server/main.go
@@ -114,10 +114,10 @@ func main() {
 			viper.GetString("trust_service.tls_ca_file"),
 		)
 		health.RegisterPeriodicFunc(
-			"Key Manager Healthy", trust.(*signer.NotarySigner).KMHealth,
+			"Key manager operational", trust.(*signer.NotarySigner).KMHealth,
 			time.Second*60)
 		health.RegisterPeriodicFunc(
-			"Signer Healthy", trust.(*signer.NotarySigner).SHealth,
+			"Signer operational", trust.(*signer.NotarySigner).SHealth,
 			time.Second*60)
 	} else {
 		logrus.Info("Using local signing service")
@@ -133,7 +133,7 @@ func main() {
 			return // not strictly needed but let's be explicit
 		}
 		health.RegisterPeriodicFunc(
-			"DB Operation", store.CheckHealth, time.Second*60)
+			"DB operational", store.CheckHealth, time.Second*60)
 		ctx = context.WithValue(ctx, "metaStore", store)
 	} else {
 		logrus.Debug("Using memory backend")
--- a/cmd/notary-server/main.go
+++ b/cmd/notary-server/main.go
@@ -9,9 +9,11 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/bugsnag/bugsnag-go"
+	"github.com/docker/distribution/health"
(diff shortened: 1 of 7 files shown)


signer/signer_trust.go
```diff
@@ -85,3 +95,56 @@ func (trust *NotarySigner) GetKey(keyid string) data.PublicKey {
...
+	done := make(chan struct{})
+	go func() {
+		status, err := check(context.Background(), &pb.Void{})
+		if len(status.Status) > 0 {
+			err = fmt.Errorf("%s not healthy", serviceName)
+		}
+		select {
+		case <-done:
+		case channel <- err:
+		}
```
aaronlehmann: Another way to do this would be to make `channel` a buffered channel (`make(chan error, 1)`). Then instead of this select statement, you could just do `channel <- err`. If the timeout has already elapsed and the outer function has returned, it's fine, because the channel is buffered so the write to it won't block. This would make the `done` channel unnecessary.

signer/signer_trust.go
```diff
@@ -85,3 +95,56 @@ func (trust *NotarySigner) GetKey(keyid string) data.PublicKey {
...
+	}()
+	var returnErr error
+	select {
+	case err := <-channel:
+		returnErr = err
+	case <-time.After(time.Second * time.Duration(timeout)):
+		returnErr = fmt.Errorf("Timed out connecting to %s after %s", serviceName, timeout)
+	}
+	close(done)
+	close(channel)
```
aaronlehmann: I don't think it's correct to close `channel` here. A write to a closed channel panics, and this could try to close `channel` at the same time that the select statement above tries to write to it.

signer/signer_trust.go
cyli: Thank you!  :D  I'm very new to Go, have absolutely never used channels before, so really appreciate someone with more knowledge looking over it!  The channel still be GCed even though there is buffered data in it, right?

signer/signer_trust.go
cyli: Good point, thanks!

signer/signer_trust.go
aaronlehmann: Correct :+1:

notary-server-Dockerfile
```diff
@@ -19,4 +19,4 @@ RUN go install \
     ${NOTARYPKG}/cmd/notary-server
 
 ENTRYPOINT [ "notary-server" ]
-CMD [ "-config", "cmd/notary-server/config.json" ]
+CMD [ "-config", "cmd/notary-server/config.json", "-debug"]
```
endophage: Shouldn't have `-debug` set in the Dockerfile (incase somebody deploys it to production unintentionally)

signer/signer_trust.go
```diff
@@ -85,3 +96,41 @@ func (trust *NotarySigner) GetKey(keyid string) data.PublicKey {
...
+
+// Generalized function that can check the health of both the signer client
+// and the key management client.
+func (trust *NotarySigner) checkServiceHealth(
+	serviceName string, check rpcHealthCheck, timeout time.Duration) error {
+
+	// Do not bother starting checking at all if the connection is broken.
+	if trust.clientConn.State() != grpc.Idle &&
+		trust.clientConn.State() != grpc.Ready {
+		return errors.New("Not currently connected to trust server.")
```
endophage: Not important: if `fmt` is already imported, it's always an option to do a `fmt.Errorf` with no variable substitution, just saves importing `errors` for one-offs like this..

notary-server-Dockerfile
cyli: Should I remove `-debug` from the signer Dockerfile then?

signer/signer_trust.go
```diff
@@ -85,3 +96,41 @@ func (trust *NotarySigner) GetKey(keyid string) data.PublicKey {
 	}
 	return data.NewPublicKey(data.KeyAlgorithm(publicKey.KeyInfo.Algorithm.Algorithm), publicKey.PublicKey)
 }
+
+// CheckHealth returns true if trust is healthy - that is if it can connect
+// to the trust server, and both the key management and signer services are
+// healthy
+func (trust *NotarySigner) CheckHealth(timeout time.Duration) error {
```
endophage: I think in this particular concrete implementation, as we know that the Key Manager and Signer are the same gRPC service, we probably don't need to check both, a check on one implies the other is up and available. Let me know if I'm missing some benefit of checking them separately.

signer/signer_trust.go
```diff
@@ -1,20 +1,30 @@
...
 	"google.golang.org/grpc/credentials"
 )
 
+// The only thing needed from grpc.ClientConn is it's state.
+type checkableConnectionState interface {
+	State() grpc.ConnectivityState
+}
+
 // NotarySigner implements a RPC based Trust service that calls the Notary-signer Service
 type NotarySigner struct {
```
endophage: Note for future: we should look at making this unexported. Only the NewNotarySigner constructor should be exposed and every treats this as a CryptoService.

cmd/notary-server/main.go
```diff
@@ -111,6 +113,20 @@ func main() {
...
 			viper.GetString("trust_service.tls_ca_file"),
 		)
+		minute := 1 * time.Minute
+		health.RegisterPeriodicFunc(
+			"Trust operational",
+			// If the trust service fails, the server is degraded but not
+			// exactly unheatlthy, so always return healthy and just log an
+			// error.
+			func() error {
+				err := trust.(*signer.NotarySigner).CheckHealth(minute)
```
endophage: Sorry, going outside the scope of your commits!

If above, rather than having `trust = signer.NewNotarySigner...` we do:
``` 
notarySigner := signer.NewNotarySigner(...)
trust = notarySigner
```
we'll eventually be able to un-export NotarySigner and won't need to do the cast here.

endophage:
Other than the debug flag, LGTM

Everything else is just notes, mostly about things I'd like to tighten up so we don't accidentally abuse structs we've exposed in the future.

signer/signer_trust.go
cyli: Ah, ok, I originally thought that maybe if we provide a new constructor (since the struct itself doesn't say whether they are the same server) that had separate endpoints, checking both would make sense, but you're right, we can check both services later if we add that.

notary-server-Dockerfile
endophage: Yes, I guess I've never taken a look inside these Dockerfiles, Diogo added them a while back. It's actually not critical for the moment as these values get overwritten when Ops wrap their own Dockerfile around this, but I've been trying to work out how we can become more self sufficient with deploys which will involve unwrapping the Ops stuff and using the images built by these Dockerfiles directly.

signer/signer_trust.go
cyli: As per discussion with @endophage, unexporting this does not break any tests, but it does break lint because `NewNotarySigner` returns an object that is unexported, so we are leaving this as is for now.

cyli:
@endophage Have addressed the review comments other than unexporting NotarySigner, if you wanted to take another look.

endophage:
LGTM. 

Love the tests. Big fan of lots of short tests.

PR #228, opened 2015-10-19
Do not use the viper singleton instance everywhere

cyli:
Well, this still has a singleton instance.  But not the `viper` package's private singleton instance.This way we can start injecting different configurations into helper functions in tests.


NathanMcCauley:
LGTM

PR #229, opened 2015-10-19
Factor out TLS configuration code for server and TLS

cyli:
So that mutual authentication can be supported between notary-server and notary-signer as well as notary and notary-server.

### DIFF
--- a/cmd/notary-signer/main.go
+++ b/cmd/notary-signer/main.go
@@ -1,8 +1,6 @@
 package main
 
 import (
-	"crypto/rand"
-	"crypto/tls"
 	"database/sql"
 	"errors"
 	_ "expvar"
@@ -22,6 +20,7 @@ import (
 	"github.com/docker/notary/cryptoservice"
 	"github.com/docker/notary/signer"
 	"github.com/docker/notary/signer/api"
+	"github.com/docker/notary/utils"
 	"github.com/docker/notary/version"
 	"github.com/endophage/gotuf/data"
 	_ "github.com/go-sql-driver/mysql"
@@ -103,20 +102,10 @@ func main() {
 		log.Fatalf("Certificate and key are mandatory")
 	}
 
-	tlsConfig := &tls.Config{
-		MinVersion:               tls.VersionTLS12,
-		PreferServerCipherSuites: true,
-		CipherSuites: []uint16{
-			tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
-			tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
-			tls.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,
-			tls.TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA,
-			tls.TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA,
-			tls.TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA,
-			tls.TLS_RSA_WITH_AES_128_CBC_SHA,
-			tls.TLS_RSA_WITH_AES_256_CBC_SHA},
+	tlsConfig, err := utils.ConfigureServerTLS(certFile, keyFile, false, "")
+	if err != nil {
+		logrus.Fatalf("Unable to set up TLS: %s", err.Error())
 	}
-	tlsConfig.Rand = rand.Reader
 
 	cryptoServices := make(signer.CryptoServiceIndex)
 
--- a/cmd/notary-signer/main.go
+++ b/cmd/notary-signer/main.go
@@ -102,7 +102,10 @@ func main() {
 		log.Fatalf("Certificate and key are mandatory")
 	}
 
-	tlsConfig, err := utils.ConfigureServerTLS(certFile, keyFile, false, "")
+	tlsConfig, err := utils.ConfigureServerTLS(&utils.ServerTLSOpts{
+		ServerCertFile: certFile,
+		ServerKeyFile:  keyFile,
+	})
 	if err != nil {
 		logrus.Fatalf("Unable to set up TLS: %s", err.Error())
 	}
--- a/cmd/notary/tuf.go
+++ b/cmd/notary/tuf.go
@@ -371,8 +371,10 @@ func getTransport(gun string, readOnly bool) http.RoundTripper {
 	if mainViper.IsSet("remote_server.skipTLSVerify") {
 		insecureSkipVerify = mainViper.GetBool("remote_server.skipTLSVerify")
 	}
-	tlsConfig, err := utils.ConfigureClientTLS(
-		rootCAFile, "", insecureSkipVerify, "", "")
+	tlsConfig, err := utils.ConfigureClientTLS(&utils.ClientTLSOpts{
+		RootCAFile:         rootCAFile,
+		InsecureSkipVerify: insecureSkipVerify,
+	})
 	if err != nil {
 		logrus.Fatal("Unable to configure TLS: ", err.Error())
 	}
--- a/cmd/notary/tuf.go
+++ b/cmd/notary/tuf.go
@@ -367,9 +367,12 @@ func getTransport(gun string, readOnly bool) http.RoundTripper {
 		}
 	}
 
+	insecureSkipVerify := false
+	if mainViper.IsSet("remote_server.skipTLSVerify") {
+		insecureSkipVerify = mainViper.GetBool("remote_server.skipTLSVerify")
+	}
 	tlsConfig, err := utils.ConfigureClientTLS(
-		rootCAFile, "", mainViper.GetBool("remote_server.skipTLSVerify"),
-		"", "")
+		rootCAFile, "", insecureSkipVerify, "", "")
 	if err != nil {
 		logrus.Fatal("Unable to configure TLS: ", err.Error())
 	}
--- a/cmd/notary/tuf.go
+++ b/cmd/notary/tuf.go
@@ -3,8 +3,6 @@ package main
 import (
 	"bufio"
 	"crypto/sha256"
-	"crypto/tls"
-	"crypto/x509"
 	"fmt"
 	"io/ioutil"
 	"net"
@@ -22,7 +20,7 @@ import (
 	"github.com/docker/distribution/registry/client/transport"
 	"github.com/docker/docker/pkg/term"
 	notaryclient "github.com/docker/notary/client"
-	"github.com/docker/notary/trustmanager"
+	"github.com/docker/notary/utils"
 	"github.com/spf13/cobra"
 )
 
@@ -360,7 +358,6 @@ func (ps passwordStore) Basic(u *url.URL) (string, string) {
 
 func getTransport(gun string, readOnly bool) http.RoundTripper {
 	// Attempt to get a root CA from the config file. Nil is the host defaults.
-	rootPool := x509.NewCertPool()
 	rootCAFile := mainViper.GetString("remote_server.root_ca")
 	if rootCAFile != "" {
 		// If we haven't been given an Absolute path, we assume it's relative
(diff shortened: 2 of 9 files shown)


utils/tls_config.go
```diff
@@ -0,0 +1,96 @@
...
+
+	return tlsConfig, nil
+}
+
+// ConfigureClientTLS generates a tls configuration for clients using the
+// provided.
+func ConfigureClientTLS(rootCA, serverName string, skipVerify bool, clientCert, clientKey string) (*tls.Config, error) {
+	tlsConfig := &tls.Config{
+		InsecureSkipVerify: skipVerify,
+		MinVersion:         tls.VersionTLS12,
```
cyli: To reviewer:  Is this ok?  I noticed the notary client seems to support `tls.VersionTLS10`.  Should we also set cipher suites for the client?

utils/tls_config.go
```diff
@@ -0,0 +1,96 @@
...
+			return nil, fmt.Errorf("No certificates in %s", caCertDir)
+		}
+		tlsConfig.ClientCAs = certStore.GetCertificatePool()
+	}
+
+	return tlsConfig, nil
+}
+
+// ConfigureClientTLS generates a tls configuration for clients using the
+// provided.
```
NathanMcCauley: provided params?

utils/tls_config.go
NathanMcCauley: Yes, would be nice to set certs for the clients. Generally, we should do everything that's present [here](https://github.com/docker/docker/blob/master/pkg/tlsconfig/config.go) in Docker's TLS setup.

utils/tls_config.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+		tlsConfig.ClientCAs = certStore.GetCertificatePool()
+	}
+
+	return tlsConfig, nil
+}
+
+// ConfigureClientTLS generates a tls configuration for clients using the
+// provided parameters.
+func ConfigureClientTLS(rootCA, serverName string, skipVerify bool, clientCert, clientKey string) (*tls.Config, error) {
```
NathanMcCauley: nit: I prefer to use golang's param name of InsecureSkipVerify since its more obviously the wrong thing to do.

utils/tls_config.go
```diff
@@ -0,0 +1,104 @@
...
+		if err != nil {
+			return nil, fmt.Errorf(
+				"Could not load root ca file. %s", err.Error())
+		}
+		rootPool := x509.NewCertPool()
+		rootPool.AddCert(rootCert)
+		tlsConfig.RootCAs = rootPool
+	}
+
+	if clientCert != "" && clientKey != "" {
```
NathanMcCauley: any usefulness in erroring if one is set and the other isn't?

cmd/notary/tuf.go
```diff
@@ -368,19 +365,13 @@ func getTransport(gun string, readOnly bool) http.RoundTripper {
...
 	}
 
-	// skipTLSVerify is false by default so verification will
-	// be performed.
-	tlsConfig := &tls.Config{
-		InsecureSkipVerify: mainViper.GetBool("remote_server.skipTLSVerify"),
-		MinVersion:         tls.VersionTLS10,
-		RootCAs:            rootPool,
+	tlsConfig, err := utils.ConfigureClientTLS(
+		rootCAFile, "", mainViper.GetBool("remote_server.skipTLSVerify"),
```
NathanMcCauley: What causes `remote_server.skipTLSVerify` to default to `false`? I think I'd rather see this value set to false here and then only set if the viper parser passes us something. Seems better than relying on viper to default values to false. (I don't see a default explicitly being set for this, sorry if I missed it)

utils/tls_config.go
cyli: Possibly - I was planning on doing the check in the caller, similarly to `ConfigureServerTLS`'s not checking the server cert/key (`notary-signer` [only allows full TLS configuration](https://github.com/docker/notary/blob/master/cmd/notary-signer/main.go#L103) while `notary-server` [allows no TLS configuration](https://github.com/docker/notary/blob/master/server/server.go#L69)).  I have no strong opinions though, if you feel erroring here would be better.

cmd/notary/tuf.go
cyli: You're right, I was just relying on the viper documentation saying that `Get*` defaults to the zero value.  I'll do an explicit check, thanks!

utils/tls_config.go
NathanMcCauley: sounds good, fine to do it earlier.

trustmanager/x509filestore_test.go
```diff
@@ -82,6 +124,21 @@ func TestAddCertFromFileX509FileStore(t *testing.T) {
...
+	tempDir, err := ioutil.TempDir("", "cert-test")
+	assert.NoError(t, err)
+	defer os.RemoveAll(tempDir)
+
+	store, err := NewX509FileStore(tempDir)
+	assert.NoError(t, err, "failed to create a new X509FileStore: %v", store)
+	assert.True(t, store.Empty())
+
+	err = store.AddCertFromFile("../fixtures/root-ca.crt")
+	assert.NoError(t, err, "failed to add certificate from file")
```
diogomonica: Why are we passing a string to `assert.NoError`. This shouldn't return an error at all?

trustmanager/x509filestore_test.go
```diff
@@ -82,6 +124,21 @@ func TestAddCertFromFileX509FileStore(t *testing.T) {
...
 }
 
+// TestNewX509FileStoreEmpty verifies the behavior of the Empty function
+func TestNewX509FileStoreEmpty(t *testing.T) {
+	tempDir, err := ioutil.TempDir("", "cert-test")
+	assert.NoError(t, err)
+	defer os.RemoveAll(tempDir)
+
+	store, err := NewX509FileStore(tempDir)
+	assert.NoError(t, err, "failed to create a new X509FileStore: %v", store)
```
diogomonica: Same thing `assert.NoError` not sure why we're passing in the string.

utils/tls_config.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+
+		certStore, err := trustmanager.NewX509FileStore(caCertDir)
+		if err != nil {
+			return nil, err
+		}
+		if certStore.Empty() {
+			return nil, fmt.Errorf("No certificates in %s", caCertDir)
+		}
+		tlsConfig.ClientCAs = certStore.GetCertificatePool()
```
diogomonica: Awesome seeing this being useful. I always knew we were going to use it for something else one day ;)

utils/tls_config.go
```diff
@@ -0,0 +1,104 @@
...
+func ConfigureClientTLS(rootCA, serverName string, insecureSkipVerify bool, clientCert, clientKey string) (*tls.Config, error) {
+	tlsConfig := &tls.Config{
+		InsecureSkipVerify: insecureSkipVerify,
+		MinVersion:         tls.VersionTLS12,
+		CipherSuites:       clientCipherSuites,
+		ServerName:         serverName,
+	}
+
+	if rootCA != "" {
+		rootCert, err := trustmanager.LoadCertFromFile(rootCA)
```
diogomonica: What about using an X509FileStore here too? You could get a new truststore, call `AddCertFromFile` (or something like that), and then call `.GetCertPool`?

utils/tls_config.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+		tlsConfig.ClientCAs = certStore.GetCertificatePool()
+	}
+
+	return tlsConfig, nil
+}
+
+// ConfigureClientTLS generates a tls configuration for clients using the
+// provided parameters.
+func ConfigureClientTLS(rootCA, serverName string, insecureSkipVerify bool, clientCert, clientKey string) (*tls.Config, error) {
```
diogomonica: Totally up to you, but I find `tlsConfig, err := utils.ConfigureClientTLS(RootCAs: rootPool, rootCAFile, "", insecureSkipVerify, "", "")` incredibly hard to read.

That invocation has 3 "", where I have no idea what the params are supposed to be.

Maybe we can create our own struct to pass in these params, and just have `ConfigureClientTLS` receive that struct, and we can create the config `tlsConfig` style before calling the method. That would avoid method invocations with "" in them.

utils/tls_config.go
diogomonica: Agreed. This should be a `||` so it fails earlier.

diogomonica:
LGTM. 

Added a few comments that we should probably address, but nothing that blocks the merge.

Good testing coverage! We should add a test where we add a keyfile and no cert to the config, and a test where we add a cert and no keyfile (and see that both should fail).

trustmanager/x509filestore_test.go
cyli: I just noticed some of the other tests in notary were were passing strings to assert functions, to state why things were failing.  I wasn't sure whether the standard practice was to include strings about what was happening when the error occurred, or no.  I'm happy to take them out.

utils/tls_config.go
cyli: If we create an `X509FileStore` and add a cert, it will attempt to persist the cert to a file in the directory the `X509FileStore` was created with.

cmd/notary-signer/main.go
```diff
@@ -103,20 +102,13 @@ func main() {
...
-			tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
-			tls.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,
-			tls.TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA,
-			tls.TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA,
-			tls.TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA,
-			tls.TLS_RSA_WITH_AES_128_CBC_SHA,
-			tls.TLS_RSA_WITH_AES_256_CBC_SHA},
+	tlsConfig, err := utils.ConfigureServerTLS(&utils.ServerTLSOpts{
+		ServerCertFile: certFile,
+		ServerKeyFile:  keyFile,
```
mtrmac: AFAICS these loaded files are not actually used; tlsconfig→server→server.ListenAndServeTLS overrides them.

ListenAndServeTLS should be called with ("", "") to keep the configuration from this call.

cyli: Yep, I do that in my next PR :)  I think I just call `ListenAndServe` actually

utils/tls_config.go
```diff
@@ -0,0 +1,125 @@
...
+	tls.TLS_RSA_WITH_AES_128_CBC_SHA,
+}...)
+
+// ServerTLSOpts generates a tls configuration for servers using the
+// provided parameters.
+type ServerTLSOpts struct {
+	ServerCertFile    string
+	ServerKeyFile     string
+	RequireClientAuth bool
+	ClientCADirectory string
```
mtrmac: Is there any actual (or planned) need for having a directory here? The only user so far are tests.

This makes using a single file awkward (as the tests already manifest), and the asymmetry between ServerTLSOpts and ClientTLSOpts is awkward.

grpc/credentials.NewClientTLSFromFile is using x509.CertPool.AppendCertsFromPEM, loading multiple CAs from a single file; this would support the same functionality and be easier to use in the simple case.

signer/signer_trust.go
```diff
@@ -30,10 +31,14 @@ type NotarySigner struct {
 func NewNotarySigner(hostname string, port string, tlscafile string) *NotarySigner {
 	var opts []grpc.DialOption
 	netAddr := net.JoinHostPort(hostname, port)
-	creds, err := credentials.NewClientTLSFromFile(tlscafile, hostname)
+	tlsConfig, err := utils.ConfigureClientTLS(&utils.ClientTLSOpts{
+		RootCAFile: tlscafile,
```
mtrmac: credentials.NewClientTLSFromFile supports multiple CAs in a single file; this (through trustmanager.LoadCertFromFile explicitly (and silently!) throws away any certificates after the first one.

Does this difference matter?

utils/tls_config.go
cyli: I think multiple certs from a single cert does make more sense, thank you for suggestion!

signer/signer_trust.go
cyli: Even if it didn't, it would probably make sense to be able to load multiple CAs from a single file for the root CA pool, as you suggested for the client CA pool.  Thanks for pointing out this backwards-incompatible change!

cmd/notary-signer/main.go
cyli: @mtrmac Would you be ok with my addressing this in the next PR?

signer/signer_trust.go
cyli: I have updated `ConfigureClientTLS` to use `CertPool.AppendCertsFromPEM` in 09dc607, so hopefully that addresses this issue.  :)

cmd/notary-signer/main.go
mtrmac: I’d slightly prefer if each commit made the code better in every possible aspect, but that’s just my pathological pedantry :) As long as it does not break the build and an improvement is coming, works for me.

cyli:
@mtrmac Yes of course!  :D  Thank you for all your suggestions!  Merging this and starting with the next PR that addresses the `ListenAndServeTLS issue`.

PR #233, opened 2015-10-22
Use make targets in circleci, more complete coverage

### LINKED ISSUE #222: CircleCI file should make use of the Makefile
So we can be sure that what's running locally matches what circleci runs (for instance, that all the same parameters are passed - currently circleci runs with `-race` and the Makefile does not)

cyli:
As per #213, #217, and #222, attempting to make the CircleCI config use make targets.  To that end I factored the `cover` a bit so that the testing/coverage part is the same across `make cover` and `make ci`.

We were also running short tests before on CI, which doesn't hit all code paths (for instance in the client.go tests we don't test with RSA keys)

cc @mtrmac (although apologies, in #213 you suggested using godep in all targets or none - I haven't done that here since not using godep lets us test various versions for development, but not using godep seems to fail in circleci - maybe we should just do a godep restore in circleci?)

This seems to work (see https://codecov.io/github/cyli/notary?branch=codecov), but the code coverage numbers as displayed in codecov.io are different than from `go tool cover --html`, which I'm still trying to work out.

Fixes #222

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,9 +87,9 @@ gen-cover:
 	@rm -rf "$(COVERDIR)"
 	@mkdir -p "$(COVERDIR)"
 	$(foreach PKG,$(PKGS),$(call gocover,$(PKG)))
-	@echo "mode: $(COVERMODE)" > "$(COVERPROFILE)"
 
 cover: gen-cover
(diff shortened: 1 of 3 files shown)


cyli:
Thanks to @vikstrous for suggesting `-coverpkg` to make sure we can account for cross-package test coverage as well.

PR #234, opened 2015-10-24
Remove symlinking root keys

cyli:
This may break backwards compatibility with import/export, but as per discussion with @diogomonica, no one is really using import/export yet, so that is probably ok.

Update: I have tested this against an existing `.notary` directory with a bunch of symlinked keys (not importing and exporting though), and regular init/add/publish/list operations work.  Switching back to the older version of notary client continued to function, since the symlinks weren't removed.


mtrmac:
l like the general direction of this: the symlinks and the two different kinds of key IDs occupying the same namespace are confusing, and confusing is risky.

Shouldn’t this end up being a much more thorough purge, though? I.e. actually drop the export/import code which expects and handles symlinks, drop `FileStore.Link` entirely, probably drop the `symlinks` parameter to `FileStore.ListFiles`, and perhaps even more. Leaving the code around, ostensibly unused but still reachable, still risks surprises.

BTW the import backwards compatibility should be easy to enough to handle, just ignore symlinks within imported archives. Hopefully we won’t need to support exporting from a newer version to be imported to an older one. (But should we start thinking about adding version indication to export archives?)

cyli:
@mtrmac Oh oops, sorry, I forgot about my stashed changes, thanks!

diogomonica:
LGTM. @cyli feel free to merge.

@mtrmac agreed on versioning, but I think what we're going to do is stick to a `tar cvf` or `cp -r` are the best ways for you to backup keys (even if notary provides you the convenience of `notary export` and `notary import`). 

We should never had symlinks in the first place.

PR #237, opened 2015-10-26
Support mutual authentication between the server and signer

cyli:
This is the follow-up PR to #229

### DIFF
--- a/cmd/notary-signer/config.json
+++ b/cmd/notary-signer/config.json
@@ -3,7 +3,8 @@
 		"http_addr": ":4444",
 		"grpc_addr": ":7899",
 		"cert_file": "./fixtures/notary-signer.crt",
-		"key_file": "./fixtures/notary-signer.key"
+		"key_file": "./fixtures/notary-signer.key",
+        "client_ca_file": "./fixtures/notary-server.crt"
 	},
 	"crypto": {
 		"pkcslib": "/usr/local/lib/softhsm/libsofthsm2.so"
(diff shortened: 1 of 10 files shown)


cmd/notary-signer/main.go
```diff
@@ -191,7 +204,7 @@ func main() {
 		log.Println("HTTP server listening on", httpAddr)
 	}
 
-	err = server.ListenAndServeTLS(certFile, keyFile)
+	err = server.ListenAndServe()
```
cyli: Addresses @mtrmac's [comment in #229](https://github.com/docker/notary/pull/229#discussion_r42788141)

mtrmac: I’m afraid this doesn’t work: with this change the signer is serving over unencrypted HTTP instead of HTTPS.

(ListenAndServe has no equivalent of the tls.NewListener call in ListenAndServeTLS)

cyli: Ah ok, thanks!  I should have just read the golang function for both of these.  Is it preferable to call `ListenAndServeTLS` with the two blank lines, or just to set up the listener and call `Serve`, in this case?

diogomonica:
Rebase needed

cmd/notary-signer/config.json
```diff
@@ -3,7 +3,8 @@
 		"http_addr": ":4444",
 		"grpc_addr": ":7899",
 		"cert_file": "./fixtures/notary-signer.crt",
-		"key_file": "./fixtures/notary-signer.key"
+		"key_file": "./fixtures/notary-signer.key",
+        "client_ca_file": "./fixtures/notary-server.crt"
```
diogomonica: not aligned?

diogomonica:
LGTM

PR #240, opened 2015-10-26
Add the health handler to the main server

### LINKED ISSUE #230: Expose health checks via the HTTP API at /notary-server/status

cyli:
Fixes #230.  Some healthcheck package refactoring may be necessary to avoid the wrapper function around health.StatusHandler.

@vikstrous Would this be ok as the endpoint to check if this is the notary server?  It either 200's or 503's.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 
 	"github.com/Sirupsen/logrus"
+	"github.com/docker/distribution/health"
 	"github.com/docker/distribution/registry/auth"
 	"github.com/endophage/gotuf/data"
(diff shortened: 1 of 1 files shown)


vikstrous:
I don't think you need the underscore, but it's up to you. It's fine either way.

cyli:
@vikstrous I don't particularly have a strong opinion - I was just copying the [elasticsearch health api](https://www.elastic.co/guide/en/elasticsearch/reference/1.4/cluster-health.html) where cluster health/info endpoints are at `_cluster/` or `_nodes/`.

vikstrous:
okay LGTM

PR #242, opened 2015-10-27
Unify root nonroot keystore

cyli:
This leaves handling the `root_keys` and `tuf_keys` subdirectories to the `KeyFileStore` rather than having `KeyStoreManager` (which is going to be removed in the future) have a non-root `KeyFileStore` and a root `KeyFileStore`.

We cannot do away with the `root_keys` and `tuf_keys` subdirectory completely because we still have to be able to read repos in that format.  And if we see a `tuf_keys` subdirectory, we can't be sure if it's the old-style file structure, or if someone added a GUN that started with `tuf_keys`.

This also makes the following changes:
1. `KeyStore` now has `ImportKey` and `ExportKey` functions, so each `KeyStore` can decide for itself whether and how to import/export keys.
2. After discussion with @endophage, importing a root key now requires you to decrypt the key, to make sure that it's a valid key.  This makes the `ImportKey` function in `KeyStore` less ugly, because you don't have to provide the key ID.

Not sure how importing/exporting will work in the future if there are multiple `KeyStore`'s.

### DIFF
--- a/cmd/notary/tuf.go
+++ b/cmd/notary/tuf.go
@@ -123,7 +123,12 @@ func tufInit(cmd *cobra.Command, args []string) {
 		fatalf(err.Error())
 	}
 
-	keysMap := nRepo.KeyStoreManager.RootKeyStore().ListKeys()
+	var keysMap map[string]string
+	for k, role := range nRepo.KeyStoreManager.KeyStore.ListKeys() {
+		if role == "root" {
+			keysMap[k] = role
+		}
+	}
 
 	var rootKeyID string
 	if len(keysMap) < 1 {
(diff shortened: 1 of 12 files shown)


cmd/notary/tuf.go
```diff
@@ -123,7 +123,12 @@ func tufInit(cmd *cobra.Command, args []string) {
 		fatalf(err.Error())
 	}
 
-	keysMap := nRepo.KeyStoreManager.RootKeyStore().ListKeys()
+	var keysMap map[string]string
```
endophage: Initializing like this raises a panic when you try and assign into it because memory hasn't been allocated. At this point keysMap is a nil pointer (map and slice variables are implicitly pointers to the underlying data, which is why they can take the `nil` value). Needs to be initialized as `keysMap := make(map[string]string)`

endophage:
This feels like a good step in the right direction. I'll take another look in the morning when my eyes are fresher.

endophage:
LGTM

PR #243, opened 2015-10-28
creating concrete types for the various key ciphers

endophage:
In the interests of providing strong type safety and moving towards a model that mimics Go's key objects (private keys should implement `crypto.Signer`) this refactors our key related code to use concrete typed keys specific to each possible supported algorithm. 


keystoremanager/keystoremanager_test.go
```diff
@@ -220,7 +220,7 @@ func TestValidateSuccessfulRootRotation(t *testing.T) {
 	}
 }
 
-func testValidateSuccessfulRootRotation(t *testing.T, keyAlg data.KeyAlgorithm, rootKeyType data.KeyAlgorithm) {
+func testValidateSuccessfulRootRotation(t *testing.T, keyAlg string, rootKeyType string) {
```
diogomonica: remove one of the "string"

cryptoservice/crypto_service.go
```diff
@@ -33,7 +33,7 @@ func NewCryptoService(gun string, keyStore trustmanager.KeyStore) *CryptoService
 }
 
 // Create is used to generate keys for targets, snapshots and timestamps
-func (ccs *CryptoService) Create(role string, algorithm data.KeyAlgorithm) (data.PublicKey, error) {
+func (ccs *CryptoService) Create(role string, algorithm string) (data.PublicKey, error) {
```
diogomonica: two "string" in a row

keystoremanager/keystoremanager_test.go
```diff
@@ -306,7 +306,7 @@ func TestValidateRootRotationMissingOrigSig(t *testing.T) {
 	}
 }
 
-func testValidateRootRotationMissingOrigSig(t *testing.T, keyAlg data.KeyAlgorithm, rootKeyType data.KeyAlgorithm) {
+func testValidateRootRotationMissingOrigSig(t *testing.T, keyAlg string, rootKeyType string) {
```
diogomonica: Two "string" in a row

keystoremanager/keystoremanager_test.go
```diff
@@ -389,7 +389,7 @@ func TestValidateRootRotationMissingNewSig(t *testing.T) {
 	}
 }
 
-func testValidateRootRotationMissingNewSig(t *testing.T, keyAlg data.KeyAlgorithm, rootKeyType data.KeyAlgorithm) {
+func testValidateRootRotationMissingNewSig(t *testing.T, keyAlg string, rootKeyType string) {
```
diogomonica: and again

signer/api/rsa_hardware_crypto_service.go
```diff
@@ -23,7 +23,7 @@ type RSAHardwareCryptoService struct {
 }
 
 // Create creates a key and returns its public components
-func (s *RSAHardwareCryptoService) Create(role string, algo data.KeyAlgorithm) (data.PublicKey, error) {
+func (s *RSAHardwareCryptoService) Create(role string, algo string) (data.PublicKey, error) {
```
diogomonica: string string

signer/signer.go
```diff
@@ -16,7 +15,7 @@ type SigningService interface {
 
 // CryptoServiceIndex represents a mapping between a service algorithm string
 // and a CryptoService
-type CryptoServiceIndex map[data.KeyAlgorithm]signed.CryptoService
+type CryptoServiceIndex map[string]signed.CryptoService
```
diogomonica: Everything looks cleaner until you get to these things.

`map[data.KeyAlgorithm]` is super easy to understand. `map[string]` you have to wonder what `string` is.

signer/signer_trust.go
```diff
@@ -77,12 +77,12 @@ func (trust *NotarySigner) Sign(keyIDs []string, toSign []byte) ([]data.Signatur
 }
 
 // Create creates a remote key and returns the PublicKey associated with the remote private key
-func (trust *NotarySigner) Create(role string, algorithm data.KeyAlgorithm) (data.PublicKey, error) {
-	publicKey, err := trust.kmClient.CreateKey(context.Background(), &pb.Algorithm{Algorithm: algorithm.String()})
+func (trust *NotarySigner) Create(role string, algorithm string) (data.PublicKey, error) {
```
diogomonica: string string

trustmanager/x509utils.go
```diff
@@ -507,18 +505,18 @@ func X509PublicKeyID(certPubKey data.PublicKey) (string, error) {
...
+	pubKeyBytes, err := x509.MarshalPKIXPublicKey(cert.PublicKey)
+	if err != nil {
+		return "", err
+	}
 
+	var key data.PublicKey
 	switch certPubKey.Algorithm() {
 	case data.ECDSAx509Key:
-		finalAlgorithm = data.ECDSAKey
+		key = data.NewECDSAPublicKey(pubKeyBytes)
```
diogomonica: Not related to your changes, but it feels like that we blindly create a new key with `pubKeyBytes`, independently of what `certPubKey.Algorithm()` was.

What I mean is: it might be possible to pass an RSA key, with a the Algorithm of `data.ECDSAx509Key`, and this method would still succeed in returning an ID, correct?

tuf/data/keys.go
```diff
@@ -34,38 +28,177 @@ type KeyPair struct {
...
+	parsed := make(map[string]tufKey)
+	err := json.Unmarshal(data, &parsed)
+	if err != nil {
+		return err
+	}
+	final := make(map[string]PublicKey)
+	for k, tk := range parsed {
+		final[k] = typedPublicKey(tk)
+	}
+	*ks = final
```
diogomonica: Interesting.

tuf/data/keys.go
```diff
@@ -34,38 +28,177 @@ type KeyPair struct {
...
+	return nil
+}
+
+// KeyList represents a list of keys
+type KeyList []PublicKey
```
diogomonica: Why would we ever use a `KeyList` over `Keys`?

tuf/signed/ed25519.go
```diff
@@ -50,7 +50,7 @@ func (e *Ed25519) Sign(keyIDs []string, toSign []byte) ([]data.Signature, error)
 }
 
 // Create generates a new key and returns the public part
-func (e *Ed25519) Create(role string, algorithm data.KeyAlgorithm) (data.PublicKey, error) {
+func (e *Ed25519) Create(role string, algorithm string) (data.PublicKey, error) {
```
diogomonica: string string

tuf/signed/interface.go
```diff
@@ -20,7 +20,7 @@ type KeyService interface {
 	// the private key into the appropriate signing service.
 	// The role isn't currently used for anything, but it's here to support
 	// future features
-	Create(role string, algorithm data.KeyAlgorithm) (data.PublicKey, error)
+	Create(role string, algorithm string) (data.PublicKey, error)
```
diogomonica: string string

tuf/signed/sign_test.go
```diff
@@ -29,7 +29,7 @@ func (mts *FailingCryptoService) Sign(keyIDs []string, _ []byte) ([]data.Signatu
 	return sigs, nil
 }
 
-func (mts *FailingCryptoService) Create(_ string, _ data.KeyAlgorithm) (data.PublicKey, error) {
+func (mts *FailingCryptoService) Create(_ string, _ string) (data.PublicKey, error) {
```
diogomonica: string string

tuf/signed/sign_test.go
```diff
@@ -56,7 +56,7 @@ func (mts *MockCryptoService) Sign(keyIDs []string, _ []byte) ([]data.Signature,
 	return sigs, nil
 }
 
-func (mts *MockCryptoService) Create(_ string, _ data.KeyAlgorithm) (data.PublicKey, error) {
+func (mts *MockCryptoService) Create(_ string, _ string) (data.PublicKey, error) {
```
diogomonica: string string

diogomonica:
What happens with notary-signer's `keys.go`? There is an `HSMKey` that should probably not exist.

Are these new key types HW friendly?

tuf/data/keys.go
```diff
@@ -34,38 +28,177 @@ type KeyPair struct {
...
+	case ECDSAx509Key:
+		return &ECDSAx509PublicKey{tufKey: tk}
+	case RSAKey:
+		return &RSAPublicKey{tufKey: tk}
+	case RSAx509Key:
+		return &RSAx509PublicKey{tufKey: tk}
+	case ED25519Key:
+		return &ED25519PublicKey{tufKey: tk}
+	}
+	return &UnknownPublicKey{tufKey: tk}
```
NathanMcCauley: Why not error? Seems unsafe to have this catch-all.

signer/signer.go
endophage: @cyli's work on keystores and the follow up work I'm doing to make keys into crypto.Signer implementers will make this type disappear. A single cryptoservice will be able to handle many keystores with many different types of key.

trustmanager/x509utils.go
endophage: Yes, and this was true before. I don't think it causes any meaningful vulnerabilities but we should sit down and talk through the possibilities.

I think the only place it would be a problem would be if somebody was running a modified version of the code that set the wrong key type. As long as the initial keys are good (which they will be as long as the user isn't running a modified binary, in which case all safety has gone out the window anyway), I think key continuity should ensure an attacker doesn't have an angle related to setting the wrong key type (particularly as we closed any holes similar to that JWT problem).

tuf/data/keys.go
endophage: This is the idiomatic way to implement `json.Unmarshaller`. It's expected that you don't unmarshal multiple times into the same instance.

tuf/data/keys.go
endophage: This unfortunately was the result of me not thinking about uniformity enough (and not having enough of a concrete idea of this refactor) when I was implementing the root key rotations. The changelist objects for root key rotations use lists of keys rather than a map... I'm kicking myself now...

tuf/data/keys.go
endophage: In the context of TUF, it's incorrect to error on a key or signature type you don't recognize. You simply won't be able to use it for signing/validating signatures, but it doesn't mean the data is invalid. We already handled this case in the same manner, it was just less explicit and hidden under the fact everything was a generic `TUFKey` or `Signature` object.

endophage:
@diogomonica these keys are specifically for use with in memory private keys. Each HSM would be fronted by a concrete KeyStore implementation that would implement its own concrete PrivateKey. In the follow up work that will make PrivateKey implement crypto.Signer, this will enable the HSM code to easily control all aspects of signing.

trustmanager/x509utils.go
cyli: Likewise the public key doesn't have to match the private key when constructing a private key, but I wasn't sure if this could cause problems.

endophage: Yeah, it's another existing problem. If a keystore gets corrupted, it would be possible for us to identify a private key with the wrong ID given that the IDs are based on the public component only. If we wanted to be really careful, we could validate everything we sign with the associated public key, but if we're doing things right, that level of paranoia shouldn't be necessary.

diogomonica:
LGTM

cyli:
LGTM since the other branch seems to do some additional testing (marshalling private bytes into their respective crypto private keys, etc.)  :)

PR #246, opened 2015-10-29
Private keys implement crypto.Signer

endophage:
This builds on top of #243 to make all private key implementers also implement crypto.Signer. This moves towards making additional signing options much more modular as current the cryptoservice has to have at least some knowledge of the various signing mechanisms. This implementation allows us to push specifics of signing implementations into small cohesive types.

TODO:

* ~~New___PrivateKey functions need to return an error in case parsing of the private key fails.~~
* ~~#247 should be merged before this then we should update cryptoservice.Sign to call the private key signing functions.~~

### DIFF
--- a/cryptoservice/certificate.go
+++ b/cryptoservice/certificate.go
@@ -1,10 +1,7 @@
 package cryptoservice
 
 import (
-	"crypto"
-	"crypto/ecdsa"
 	"crypto/rand"
-	"crypto/rsa"
 	"crypto/x509"
 	"fmt"
 
@@ -15,28 +12,11 @@ import (
 // GenerateCertificate generates an X509 Certificate from a template, given a GUN
 func GenerateCertificate(rootKey data.PrivateKey, gun string) (*x509.Certificate, error) {
 
-	algorithm := rootKey.Algorithm()
-	var (
-		publicKey  crypto.PublicKey
-		privateKey crypto.PrivateKey
-		err        error
-	)
-	switch algorithm {
-	case data.RSAKey:
-		var rsaPrivateKey *rsa.PrivateKey
-		rsaPrivateKey, err = x509.ParsePKCS1PrivateKey(rootKey.Private())
-		privateKey = rsaPrivateKey
-		publicKey = rsaPrivateKey.Public()
-	case data.ECDSAKey:
-		var ecdsaPrivateKey *ecdsa.PrivateKey
-		ecdsaPrivateKey, err = x509.ParseECPrivateKey(rootKey.Private())
-		privateKey = ecdsaPrivateKey
-		publicKey = ecdsaPrivateKey.Public()
+	switch rootKey.(type) {
+	case *data.RSAPrivateKey, *data.ECDSAPrivateKey:
+		// go doesn't fall through
 	default:
-		return nil, fmt.Errorf("only RSA or ECDSA keys are currently supported. Found: %s", algorithm)
-	}
-	if err != nil {
-		return nil, fmt.Errorf("failed to parse root key: %s (%v)", gun, err)
+		return nil, fmt.Errorf("only bare RSA or ECDSA keys (not x509 variants) are currently supported. Found: %s", rootKey.Algorithm())
 	}
 
 	template, err := trustmanager.NewCertificate(gun)
@@ -44,7 +24,7 @@ func GenerateCertificate(rootKey data.PrivateKey, gun string) (*x509.Certificate
 		return nil, fmt.Errorf("failed to create the certificate template for: %s (%v)", gun, err)
 	}
 
-	derBytes, err := x509.CreateCertificate(rand.Reader, template, template, publicKey, privateKey)
+	derBytes, err := x509.CreateCertificate(rand.Reader, template, template, rootKey.CryptoSigner().Public(), rootKey.CryptoSigner())
 	if err != nil {
 		return nil, fmt.Errorf("failed to create the certificate for: %s (%v)", gun, err)
 	}
--- a/cryptoservice/crypto_service.go
+++ b/cryptoservice/crypto_service.go
@@ -1,17 +1,11 @@
 package cryptoservice
 
 import (
-	"crypto"
-	"crypto/ecdsa"
 	"crypto/rand"
-	"crypto/rsa"
-	"crypto/sha256"
-	"crypto/x509"
 	"fmt"
 	"path/filepath"
 
 	"github.com/Sirupsen/logrus"
-	"github.com/agl/ed25519"
 	"github.com/docker/notary/trustmanager"
 	"github.com/docker/notary/tuf/data"
 )
@@ -91,18 +85,14 @@ func (ccs *CryptoService) RemoveKey(keyID string) error {
 func (ccs *CryptoService) Sign(keyIDs []string, payload []byte) ([]data.Signature, error) {
 	signatures := make([]data.Signature, 0, len(keyIDs))
 	for _, keyid := range keyIDs {
-		var (
-			privKey data.PrivateKey
-			err     error
-			keyName = keyid
-		)
+		keyName := keyid
 
 		// Try to get the key first without a GUN (in which case it's a root
 		// key).  If that fails, try to get the key with the GUN (non-root
 		// key).  If that fails, then we don't have the key.
-		privKey, _, err = ccs.keyStore.GetKey(keyName)
+		privKey, _, err := ccs.keyStore.GetKey(keyName)
 		if err != nil {
-			keyName := filepath.Join(ccs.gun, keyid)
+			keyName = filepath.Join(ccs.gun, keyid)
 			privKey, _, err = ccs.keyStore.GetKey(keyName)
 			if err != nil {
 				logrus.Debugf("error attempting to retrieve key ID: %s, %v", keyid, err)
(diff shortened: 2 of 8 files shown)


tuf/data/keys.go
```diff
@@ -317,113 +327,226 @@ func NewED25519PublicKey(public []byte) *ED25519PublicKey {
 }
 
 // Private key types
+type privateKey struct {
+	private []byte
+}
+
+type signer struct {
+	Signer crypto.Signer
```
cyli: Non-blocking - not sure if it might be kind of confusing to have multiple ways of getting this - since it's already exposed on every PrivateKey with CryptoSigner(), maybe it'd be simpler to not expose this?

endophage: With the `signer` type being unexported, it already makes the `crypto.Signer` effectively unexported on the concrete private key types, but for clarity your suggestion makes a lot of sense. Update incoming.

tuf/data/keys.go
```diff
@@ -317,113 +327,226 @@ func NewED25519PublicKey(public []byte) *ED25519PublicKey {
...
+	hashed := sha256.Sum256(msg)
+	sigASN1, err := k.Signer.Sign(rand, hashed[:], opts)
+	if err != nil {
+		return nil, err
+	}
+
+	return convertECDSASig(sigASN1, ecdsaPrivKey.Params().BitSize)
+}
+
+func convertECDSASig(sigASN1 []byte, bitSize int) ([]byte, error) {
```
cyli: Sorry, this is just me being a super newb.  :)  Is this to make it conform to https://tools.ietf.org/html/rfc3278#section-8.2?  If so, would make sense to add that as a comment?

cryptoservice/certificate.go
```diff
@@ -15,28 +12,11 @@ import (
...
-		publicKey = ecdsaPrivateKey.Public()
+	switch rootKey.(type) {
+	case *data.RSAPrivateKey, *data.ECDSAPrivateKey:
+		// go doesn't fall through
 	default:
-		return nil, fmt.Errorf("only RSA or ECDSA keys are currently supported. Found: %s", algorithm)
-	}
-	if err != nil {
-		return nil, fmt.Errorf("failed to parse root key: %s (%v)", gun, err)
+		return nil, fmt.Errorf("only RSA or ECDSA keys are currently supported. Found: %s", rootKey.Algorithm())
```
cyli: Non-blocking:  Maybe specify that they have to be bare keys, not x509 keys, since we're specifically excluding the x509 ones?

tuf/data/keys.go
endophage: It's possible that this format conforms to that spec, I'm actually not sure. This is the way Diogo/Aaron implemented serialization of ECDSA signatures, which I believe was copied from docker/libtrust.

cryptoservice/certificate.go
endophage: Wanted to run that by Diogo. There's no reason to exclude them from a technical perspective.

cyli:
Other than the doc nitpicks this looks great to me!  :D

cryptoservice/certificate.go
diogomonica: I think we should keep this clean: bare keys only.

cryptoservice/certificate.go
```diff
@@ -44,7 +24,7 @@ func GenerateCertificate(rootKey data.PrivateKey, gun string) (*x509.Certificate
 		return nil, fmt.Errorf("failed to create the certificate template for: %s (%v)", gun, err)
 	}
 
-	derBytes, err := x509.CreateCertificate(rand.Reader, template, template, publicKey, privateKey)
+	derBytes, err := x509.CreateCertificate(rand.Reader, template, template, rootKey.CryptoSigner().Public(), rootKey.CryptoSigner())
```
diogomonica: shouldn't a `rootKey` have a `.Public()`? Why do we have to go trough a cryptosigner?

cryptoservice/crypto_service.go
```diff
@@ -110,28 +100,23 @@ func (ccs *CryptoService) Sign(keyIDs []string, payload []byte) ([]data.Signatur
...
 
-		algorithm := privKey.Algorithm()
 		var sigAlgorithm data.SigAlgorithm
-		var sig []byte
 
-		switch algorithm {
-		case data.RSAKey:
-			sig, err = rsaSign(privKey, payload)
+		switch privKey.(type) {
+		case *data.RSAPrivateKey, *data.RSAx509PrivateKey:
```
diogomonica: There should be no such thing as an `RSAx509PrivateKey`, only `RSAx509PublicKey`, correct?

cryptoservice/crypto_service.go
```diff
@@ -110,28 +100,23 @@ func (ccs *CryptoService) Sign(keyIDs []string, payload []byte) ([]data.Signatur
...
 			sigAlgorithm = data.RSAPSSSignature
-		case data.ECDSAKey:
-			sig, err = ecdsaSign(privKey, payload)
+		case *data.ECDSAPrivateKey, *data.ECDSAx509PrivateKey:
```
diogomonica: Same here, why do we have a `ECDSAx509PrivateKey`

cryptoservice/crypto_service.go
```diff
@@ -91,18 +85,14 @@ func (ccs *CryptoService) RemoveKey(keyID string) error {
 func (ccs *CryptoService) Sign(keyIDs []string, payload []byte) ([]data.Signature, error) {
```
diogomonica: I thought `Sign` never gets called with non-canonical keyIDs.

tuf/data/keys.go
```diff
@@ -317,113 +327,226 @@ func NewED25519PublicKey(public []byte) *ED25519PublicKey {
...
-func (k ECDSAPrivateKey) Private() []byte {
+func (k privateKey) Private() []byte {
 	return k.private
 }
 
-// Private return the serialized private bytes of the key
-func (k ECDSAx509PrivateKey) Private() []byte {
-	return k.private
+// CryptoSigner returns the underlying crypto.Signer for use cases where we need the default
+// signature or public key functionality (like when we generate certificates)
```
diogomonica: I don't understand this comment. "public key functionality"

tuf/data/keys.go
```diff
@@ -317,113 +327,226 @@ func NewED25519PublicKey(public []byte) *ED25519PublicKey {
...
+	sigASN1, err := ecdsaPrivKey.Sign(rand, hashed[:], opts)
+	if err != nil {
+		return nil, err
+	}
+
+	return convertECDSASig(sigASN1, ecdsaPrivKey.Params().BitSize)
+}
+
+// Sign creates an ecdsa signature
+func (k ECDSAx509PrivateKey) Sign(rand io.Reader, msg []byte, opts crypto.SignerOpts) (signature []byte, err error) {
```
diogomonica: I don't understand why this exists. ECDSAx509PrivateKey?!

cryptoservice/certificate.go
endophage: In this instance we want the `ecdsa.PrivateKey` or `rsa.PrivateKey`, not the public key bytes as serialized for TUF, which is what `rootKey.Public()` would return.

cryptoservice/crypto_service.go
endophage: It doesn't. The translation happens in tuf/signed/sign.go

cryptoservice/crypto_service.go
endophage: I made everything very explicit so the RSAx509PrivateKey and ECDSAx509PrivateKey exist specifically as private counterparts to the x509 variants of the public keys. It's possible we don't need them, I'm not 100% certain though. I'll continue to assess but it's something that can safely be removed at a later time if we confirm they're not required.

diogomonica: but *x509PrivateKey doesn't make sense.

cyli: I don't think we currently use it anywhere - or, at least, the tests seem to pass if they're removed?

endophage: In this context it means it's an RSA private key with an x509 Public component. Have a look at the actual key definitions in tuf/data/keys.go to see the difference between RSAPrivateKey and RSAx509PrivateKey

diogomonica:
LGTM

tuf/data/keys.go
```diff
@@ -317,113 +327,162 @@ func NewED25519PublicKey(public []byte) *ED25519PublicKey {
...
 // NewECDSAPrivateKey initializes a new ECDSA private key
-func NewECDSAPrivateKey(public ECDSAPublicKey, private []byte) *ECDSAPrivateKey {
-	return &ECDSAPrivateKey{
-		ECDSAPublicKey: public,
-		private:        private,
+func NewECDSAPrivateKey(public PublicKey, private []byte) (*ECDSAPrivateKey, error) {
+	switch public.(type) {
+	case *ECDSAPublicKey, *ECDSAx509PublicKey:
+	default:
+		return nil, errors.New("Invalid public key type provided to NewRSAPrivateKey")
```
cyli: "...provided to NewECDSAPrivateKey" ?

cyli:
LGTM

PR #247, opened 2015-10-29
Remove unlocked_cryptoservice and GetRootCryptoService

cyli:
This just has the single crypto service which does signing for everything.

This unfortunately makes it so that cryptoservice has to try the key ID both with and without the gun (in case it's a root key) - wondering if the role should be passed to CryptoService.Sign, so it can validate that the roles for the key IDs match?


cryptoservice/crypto_service.go
```diff
@@ -86,16 +86,20 @@ func (ccs *CryptoService) RemoveKey(keyID string) error {
...
-
-		var privKey data.PrivateKey
-		var err error
+		var (
+			privKey data.PrivateKey
+			err     error
+			keyName = keyid
+		)
 
 		privKey, _, err = ccs.keyStore.GetKey(keyName)
```
diogomonica: We need comments here.

client/client.go
```diff
@@ -137,8 +136,14 @@ func NewNotaryRepository(baseDir, gun, baseURL string, rt http.RoundTripper,
 
 // Initialize creates a new repository by using rootKey as the root Key for the
 // TUF repository.
-func (r *NotaryRepository) Initialize(uCryptoService *cryptoservice.UnlockedCryptoService) error {
-	rootCert, err := uCryptoService.GenerateCertificate(r.gun)
+func (r *NotaryRepository) Initialize(rootKeyID string) error {
+	privKey, _, err := r.KeyStoreManager.KeyStore.GetKey(rootKeyID)
```
endophage: In the interests of cryptoservices supporting multiple keystores: `r.cryptoservice.GetKey(rootKeyID)`

cyli: Ah, good point, thanks!

client/client.go
```diff
@@ -384,21 +389,7 @@ func (r *NotaryRepository) Publish() error {
...
-		canonicalRootID, err := tufutils.CanonicalKeyID(rootKey)
-		if err != nil {
-			return err
-		}
-		rootCryptoService, err := r.KeyStoreManager.GetRootCryptoService(canonicalRootID)
-		if err != nil {
-			return err
-		}
-		root, err = r.tufRepo.SignRoot(data.DefaultExpires("root"), rootCryptoService.CryptoService)
+		root, err = r.tufRepo.SignRoot(data.DefaultExpires("root"), r.cryptoService)
```
diogomonica: Do we have a test that enforces that this check of signature was redundant?

endophage: Now that we only have the one cryptoservice, we should be able to remove the final argument from all the SignX functions in tuf.Repo. We only added it in to support unlocked crypto service and I'd love to kill it off again.

diogomonica:
LGTM

client/client.go
cyli: IRL recap:  when `tuf.Repo.SignRoot` (and then `tuf.Repo.sign`) is called, all the key IDs are validated also.  So it would fail when `SignRoot` is called instead of slightly earlier.

endophage:
Net negative LoC, UnlockedCryptoService has been shown the door, there's nothing not to love about this PR.

LGTM

PR #255, opened 2015-10-30
Don't use elliptic.P224()

mtrmac:
This curve is not available on Fedora and RHEL systems, so removing the
reference allows tests to pass there.  Vast majority of the
curve-specific work is done in the golang crypto/elliptic package, so
this does not weaken the tests noticeably.


mtrmac:
See also https://github.com/endophage/gotuf/pull/52 .

diogomonica:
I'm ok with this.

PR #256, opened 2015-10-30
Remove key ID from (notary key import-root)

mtrmac:
PR #242 has started requiring a passphrase for the imported key, and
recomputes the key ID, making the command-line argument redundant.  So,
remove it from the command line and from the KeyStoreManager API.

Also updates the comment for KeyStoreManager.ImportRootKey, and changes
(notary key import-root) to refuse unexpected arguments instead of
silently ignoring them.

### DIFF
--- a/cmd/notary/keys.go
+++ b/cmd/notary/keys.go
@@ -96,7 +96,7 @@ var cmdKeyImport = &cobra.Command{
 }
 
 var cmdKeyImportRoot = &cobra.Command{
-	Use:   "import-root [ keyID ] [ filename ]",
+	Use:   "import-root [ filename ]",
 	Short: "Imports root key.",
 	Long:  "imports a root key from a PEM file.",
 	Run:   keysImportRoot,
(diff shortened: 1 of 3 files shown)


mtrmac:
(FWIW it would have been a bit useful to me have the `import-root` behavior change noted in a commit message as well as in the PR description. It took me some time to find that discussion and satisfy myself that this change was intended.)

cyli:
Thank you for noticing and for the fix, and apologies for the inconvenience in finding the change description.  Will make sure to add more extensive comments to commit messages.  :)  LGTM.

mtrmac:
Actually, I must apologize barging in like that about commit messages just based on how other, mostly pre-github, projects have been run in the past.

Is the general expectation for notary that intent and high-level changes are discussed in the PRs and not in commit messages?

cyli:
@mtrmac I don't think [there is any specific policy](https://github.com/docker/notary/blob/master/CONTRIBUTING.md), but your suggestion would probably make it easier to grep through logs, and seems like a good practice.  

cc @diogomonica?  Add it to CONTRIBUTING.md?

diogomonica:
@mtrmac we're probably going to decide it as it goes. So far we have very few external contributors, so most of the discussions end up behind closed doors. As we start getting people from the outside contributing, we'll make sure we spend more time being extra transparent.

I think you can expect some more process around changes as soon as we release notary 0.1 (at Dockercon).

PR #257, opened 2015-10-30
Fix (notary key export-root)

mtrmac:
-c was recently taken over by --configFile; using it for --change-passphrase as well results in

panic: shorthand redefinition

So, move --change-passphrase to -p.


diogomonica:
There are a few more changes we want to do around import/export, but LGTM

PR #262, opened 2015-11-02
WIP: Removing _alias in a backwards compatible way

diogomonica:
I've been fighting this for a while. Maintaining backwards compatibility involves always trying the new style of file, and attempting to fallback to the old one in case that one fails.

This part will get cleaner as soon as we stop supporting "_alias", and I'm happy with the way the code for keeping aliases in PEMs looks.

The major problem that I can't solve is the fact that to find a key in the filesystem given only the KeyID, we have to transverse all the files until we find the right one. Additionally, to get the Alias from a KeyID (since the alias was in the filename, and is now in the file itself) means also transversing all of the file system looking for the file so we can parse it and return the alias. This is super ugly, but the only ways I see around it is:

- Maintain a flat list of keys, instead of doing `path/gun/keyid.key`
- Maintain a file mapping IDs -> locations

Both of those solutions are super ugly too.

@cyli 

Also, I can't seem to make ImportExport work.

### DIFF
--- a/trustmanager/x509utils.go
+++ b/trustmanager/x509utils.go
@@ -211,23 +211,53 @@ func GetIntermediateCerts(certs []*x509.Certificate) (intCerts []*x509.Certifica
 	return intCerts
 }
 
-// ParsePEMPrivateKey returns a data.PrivateKey from a PEM encoded private key. It
-// only supports RSA (PKCS#1) and attempts to decrypt using the passphrase, if encrypted.
-func ParsePEMPrivateKey(pemBytes []byte, passphrase string) (data.PrivateKey, error) {
-	block, _ := pem.Decode(pemBytes)
-	if block == nil {
-		return nil, errors.New("no valid private key found")
+// GetPemKeyAliasFromFile attempts to retrieve an alias from a a file that contains
+// a PEM encoded Private Key
+func GetPemKeyAliasFromFile(filename string) (string, error) {
+	b, err := ioutil.ReadFile(filename)
+	if err != nil {
+		return "", err
 	}
 
+	return GetPemKeyAlias(b), nil
+}
+
+// GetPemKeyAlias attempts to retrieve an alias from a PEM encoded Private Key
+func GetPemKeyAlias(pemBytes []byte) string {
+	block, _ := pem.Decode(pemBytes)
+	if block != nil && block.Headers != nil && block.Headers["alias"] != "" {
+		return block.Headers["alias"]
+	}
+
+	return ""
+}
+
+// ParsePEMPrivateKey returns a data.PrivateKey and an alias string from a PEM
+// encoded private key. It attempts to decrypt using the passphrase, if encrypted.
+func ParsePEMPrivateKey(pemBytes []byte, passphrase string) (data.PrivateKey, string, error) {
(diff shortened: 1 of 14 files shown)


endophage:
@diogomonica a thought I've been having on this is that even with quite a large number of keys, the amount of memory required to store all the IDs and paths will be relatively small. Does it make sense for a KeyStore to load all its keys on init and store the IDs mapped to their paths for lookup of the actual key data when required?

cyli:
@diogomonica Apologies, I think I'm misunderstanding the key ID - I thought currently the keyfilestore takes a key name, which includes the GUN as well as the SHA of the key json?  Aren't there only 2 places it could be?

```
root_keys/key_id.key
root_keys/key_id_alias.key
```

for root keys, and then for other keys:

```
tuf_keys/<gun>/key_id.key,
tuf_keys/<gun>/key_id_alias.key
```

We'd have to list keys under those subdirectories, or we can just try the roles that make sense.

And agree with @endophage's comment that we can probably just prepopulate the cache on startup?

diogomonica:
@cyli @endophage we already have a cache, but it's useless: every-time notary gets ran it gets repopulated, which means full filesystem scan anyway. The problem is that there is no storage between invocations.

We can keep a map of KeyID to alias and location, populate it on keystore creation, but that means pem decoding *every* key on disk, where currently what we do is we list *every* key, but we only pem decode one.

endophage:
@diogomonica keep in perspective that "full filesystem scan" means reading filenames only for a relatively small number of files. Even at 100 repos being signed for, it would only be 300-400 files and directories (looking at names only) that have to be walked. As a quick benchmark. I can walk (using `filepath.Walk`) the 115616 files in my home directory in 0.64 seconds. It takes longer to compile the 17 line Go program that does it than it takes to actually run the walk.

diogomonica:
@endophage if we're not worried about `filepath.Walk` the current implementation is fine.

trustmanager/x509utils.go
```diff
@@ -440,7 +477,12 @@ func EncryptPrivateKey(key data.PrivateKey, passphrase string) ([]byte, error) {
 		return nil, err
 	}
 
-	return pem.EncodeToMemory(encryptedPEMBlock), nil
+	if alias != "" && encryptedPEMBlock.Headers != nil {
```
cyli: Are the headers ever nil?  Don't the headers for an encrypted PEM block have to include the algorithm (e.g. should there be an error if the headers are nil?)

trustmanager/x509utils.go
```diff
@@ -211,23 +211,53 @@ func GetIntermediateCerts(certs []*x509.Certificate) (intCerts []*x509.Certifica
...
 	if block == nil {
-		return nil, errors.New("no valid private key found")
+		return nil, "", errors.New("no valid private key found")
+	}
+	// Attempt to extract the alias from the PEM headers
+	if block.Headers != nil {
+		alias = block.Headers["alias"]
 	}
 
+	// Parse th private key bytes, decrypting if necessary
```
cyli: Nonblocking nitpick:  "Parse the..."

diogomonica:
Functionality being added in #383

PR #267, opened 2015-11-09
Certificate rotation

mtrmac:
When initializing a repository, notary currently creates a certificate with 10-year expiration. We can't have customers set up their system with such a time bomb without knowing for sure that we will be able to give them a smooth key rollover path.

So, this PR will eventually contain a proof-of-concept implementation, sufficient at least to make sure we are not painting ourselves into a corner.

The individual improvement/refactoring commits might be worth looking at; the final WIP commit is pretty useless at this moment and something which is likely to yet change drastically, feel free to completely ignore it.

### DIFF
--- a/tuf/signed/sign.go
+++ b/tuf/signed/sign.go
@@ -20,19 +20,14 @@ import (
 	"github.com/docker/notary/tuf/utils"
 )
 
-// Sign takes a data.Signed and keys, calculates and adds at least
+// Sign takes a data.Signed and replaces data.Signature with at least
 // minPrimarySignature signatures using primaryKeys and any possible
-// signatures using optionalKeys to the data.Signed
-// N.B. All public keys for a role should be passed so that this function
-//      can correctly clean up signatures that are no longer valid.
+// signatures using optionalKeys
 func Sign(service CryptoService, s *data.Signed, primaryKeys []data.PublicKey,
 	minSignatures int, optionalKeys []data.PublicKey) error {
 
 	logrus.Debugf("sign called with %d/%d required + %d optional keys",
 		minSignatures, len(primaryKeys), len(optionalKeys))
-	signatures := make([]data.Signature, 0, len(s.Signatures)+1)
-	signingKeyIDs := make(map[string]struct{})
-	tufIDs := make(map[string]data.PublicKey)
 
 	privKeys := make(map[string]data.PrivateKey)
 
@@ -40,7 +35,6 @@ func Sign(service CryptoService, s *data.Signed, primaryKeys []data.PublicKey,
 	missingKeyIDs := []string{}
 	for _, key := range primaryKeys {
 		canonicalID, err := utils.CanonicalKeyID(key)
-		tufIDs[key.ID()] = key
 		if err != nil {
 			return err
 		}
@@ -65,7 +59,6 @@ func Sign(service CryptoService, s *data.Signed, primaryKeys []data.PublicKey,
 		if _, ok := privKeys[key.ID()]; ok {
 			continue
 		}
-		tufIDs[key.ID()] = key
 		canonicalID, err := utils.CanonicalKeyID(key)
 		if err != nil {
 			return err
@@ -78,13 +71,13 @@ func Sign(service CryptoService, s *data.Signed, primaryKeys []data.PublicKey,
 	}
 
 	// Do signing and generate list of signatures
+	signatures := make([]data.Signature, 0, len(privKeys))
 	for keyID, pk := range privKeys {
 		sig, err := pk.Sign(rand.Reader, *s.Signed, nil)
 		if err != nil {
 			logrus.Debugf("Failed to sign with key: %s. Reason: %v", keyID, err)
 			return err
 		}
-		signingKeyIDs[keyID] = struct{}{}
 		signatures = append(signatures, data.Signature{
 			KeyID:     keyID,
 			Method:    pk.SignatureAlgorithm(),
@@ -92,27 +85,6 @@ func Sign(service CryptoService, s *data.Signed, primaryKeys []data.PublicKey,
 		})
 	}
 
-	for _, sig := range s.Signatures {
-		if _, ok := signingKeyIDs[sig.KeyID]; ok {
-			// key is in the set of key IDs for which a signature has been created
-			continue
-		}
-		var (
-			k  data.PublicKey
-			ok bool
-		)
-		if k, ok = tufIDs[sig.KeyID]; !ok {
-			// key is no longer a valid signing key
-			continue
-		}
-		if err := VerifySignature(*s.Signed, sig, k); err != nil {
-			// signature is no longer valid
-			continue
-		}
-		// keep any signatures that still represent valid keys and are
-		// themselves valid
-		signatures = append(signatures, sig)
-	}
 	s.Signatures = signatures
 	return nil
 }
--- a/tuf/signed/sign.go
+++ b/tuf/signed/sign.go
@@ -20,11 +20,11 @@ import (
 	"github.com/docker/notary/tuf/utils"
 )
 
-// Sign takes a data.Signed and a key, calculated and adds the signature
+// Sign takes a data.Signed and keys, calculates and adds the signature
 // to the data.Signed
 // N.B. All public keys for a role should be passed so that this function
 //      can correctly clean up signatures that are no longer valid.
-func Sign(service CryptoService, s *data.Signed, keys ...data.PublicKey) error {
+func Sign(service CryptoService, s *data.Signed, keys []data.PublicKey) error {
 	logrus.Debugf("sign called with %d keys", len(keys))
 	signatures := make([]data.Signature, 0, len(s.Signatures)+1)
 	signingKeyIDs := make(map[string]struct{})
--- a/tuf/signed/sign.go
+++ b/tuf/signed/sign.go
@@ -19,21 +19,25 @@ import (
 	"github.com/docker/notary/tuf/utils"
 )
 
-// Sign takes a data.Signed and keys, calculates and adds at least minSignature
-// signatures to the data.Signed
+// Sign takes a data.Signed and keys, calculates and adds at least
+// minPrimarySignature signatures using primaryKeys and any possible
+// signatures using optionalKeys to the data.Signed
 // N.B. All public keys for a role should be passed so that this function
 //      can correctly clean up signatures that are no longer valid.
-func Sign(service CryptoService, s *data.Signed, keys []data.PublicKey, minSignatures int) error {
-	logrus.Debugf("sign called with %d/%d keys", minSignatures, len(keys))
+func Sign(service CryptoService, s *data.Signed, primaryKeys []data.PublicKey,
+	minSignatures int, optionalKeys []data.PublicKey) error {
+
+	logrus.Debugf("sign called with %d/%d required + %d optional keys",
+		minSignatures, len(primaryKeys), len(optionalKeys))
 	signatures := make([]data.Signature, 0, len(s.Signatures)+1)
 	signingKeyIDs := make(map[string]struct{})
 	tufIDs := make(map[string]data.PublicKey)
-	ids := make([]string, 0, len(keys))
+	ids := make([]string, 0, len(primaryKeys))
 
 	privKeys := make(map[string]data.PrivateKey)
 
 	// Get all the private key objects related to the public keys
-	for _, key := range keys {
+	for _, key := range primaryKeys {
 		canonicalID, err := utils.CanonicalKeyID(key)
 		ids = append(ids, canonicalID)
 		tufIDs[key.ID()] = key
@@ -51,6 +55,22 @@ func Sign(service CryptoService, s *data.Signed, keys []data.PublicKey, minSigna
 	if len(privKeys) < minSignatures {
 		return ErrInsufficientSignatures{FoundKeys: len(privKeys),
 			NeededKeys: minSignatures, KeyIDs: ids}
+	}
+
+	for _, key := range optionalKeys {
+		if _, ok := privKeys[key.ID()]; ok {
+			continue
+		}
+		tufIDs[key.ID()] = key
+		canonicalID, err := utils.CanonicalKeyID(key)
+		if err != nil {
+			return err
+		}
+		k, _, err := service.GetPrivateKey(canonicalID)
+		if err != nil {
+			continue
+		}
+		privKeys[key.ID()] = k
 	}
 
 	// Do signing and generate list of signatures
--- a/tuf/signed/sign.go
+++ b/tuf/signed/sign.go
@@ -13,19 +13,18 @@ package signed
 
 import (
 	"crypto/rand"
-	"fmt"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/docker/notary/tuf/data"
 	"github.com/docker/notary/tuf/utils"
 )
 
-// Sign takes a data.Signed and keys, calculates and adds the signature
-// to the data.Signed
+// Sign takes a data.Signed and keys, calculates and adds at least minSignature
+// signatures to the data.Signed
 // N.B. All public keys for a role should be passed so that this function
 //      can correctly clean up signatures that are no longer valid.
-func Sign(service CryptoService, s *data.Signed, keys []data.PublicKey) error {
-	logrus.Debugf("sign called with %d keys", len(keys))
+func Sign(service CryptoService, s *data.Signed, keys []data.PublicKey, minSignatures int) error {
+	logrus.Debugf("sign called with %d/%d keys", minSignatures, len(keys))
 	signatures := make([]data.Signature, 0, len(s.Signatures)+1)
 	signingKeyIDs := make(map[string]struct{})
 	tufIDs := make(map[string]data.PublicKey)
@@ -48,9 +47,10 @@ func Sign(service CryptoService, s *data.Signed, keys []data.PublicKey) error {
 		privKeys[key.ID()] = k
 	}
 
-	// Check to ensure we have at least one signing key
-	if len(privKeys) == 0 {
-		return ErrNoKeys{KeyIDs: ids}
+	// Check to ensure we have enough signing keys
+	if len(privKeys) < minSignatures {
+		return ErrInsufficientSignatures{FoundKeys: len(privKeys),
+			NeededKeys: minSignatures, KeyIDs: ids}
 	}
 
 	// Do signing and generate list of signatures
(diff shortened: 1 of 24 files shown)


endophage:
@mtrmac thanks for taking a crack at this. We'll probably not be able to give this PR the attention it deserves until after DockerCon. There are a lot of dependencies on supporting root key rotation, some of which you've already identified, like how to track and continue signing with the old root keys.

We intentionally built in the reading and validation side of root key rotations with tests at both the Notary and TUF levels to ensure, as you say, we aren't painting ourselves into a corner. Do you feel more tests are necessary around the consuming side of a root rotation to make your team feel confident in it? Also bear in mind that the 10 years expiration on the cert is essentially irrelevant as TUF ultimately basis the expiry of data (keys and targets) on the timestamp.json file. 

For now, one thing we'd prefer at a design level is to not add a command specifically for root key rotation. Instead a flag on the existing `notary key rotate` command would be cleaner. If one is rotating a root we might as well rotate the other keys at the same time, on the basis that there's no reason to rotate a root unless it's compromised (loss of root keys being a different scenario entirely), in which case, all keys should be considered compromised.

mtrmac:
@endophage I must admit I haven’t checked how extensive the tests are; they may well cover 90% of the problem, however without code available to create the updated root.json, there is still a slight chance that something which is possible in the tests is not possible in the deployed environment (e.g. because some information has been lost, or because there are UI implications—e.g. I didn’t originally realize that even after a known compromise it is still necessary to keep the old compromised root key around indefinitely).

The expiry in `timestamp.json` is AFAICS not much of an issue because the notary server will keep creating new versions of that file. Yes, the notary server could turn malicious and DoS the repository, but fixing or mitigating that does not make the 10-year certificate expiration go away.

On the CLI, noted, I’ll see how it turns out. This part of the code is *very* much in flux yet. Two notes, however:

* Rotating the root *key* and rotating the *certificate* are different things; rotating the certificate is necessary simply because it has an expiration date (or, well, we could make it have no expiration, but that is probably worse), even without any assumed attack. (I’m not quite certain that the difference truly matters, perhaps rotating the private key as well would make sense—OTOH it is not really possible to lose the certificate as long as the private root key is still around, so rotating only the certificate may in some sense be more benign.)
* Rotating the root private key should not happen only after a known compromise; in fact at that point it might be too late (because the attacker have published an update first and might have caused most clients to already rotate to a malicious key only the attacker owns, leaving the legitimate owner no automated way to regain control). Rotating the root private key from time to time may prevent misuse of an inadvertently disclosed private key (e.g. in backups, or to an employee which may leave in the future on bad terms) in some indeterminate future. (You are right that after rotating the root key one may just as well rotate the tagging keys, though.)

endophage:
Good point on rotating key vs. cert.

I wasn't being particularly pedantic when I used the term compromised. If you have the slightest inkling that the security of your root key hasn't been perfect, sure it makes sense to rotate more frequently, possibly just making it policy to rotate every X years on the basis that somebody has probable done something bad that hasn't been discovered in that time. The case you mention about an attacker already rotating the root is the biggest hole in the security of TUF. It largely relies on the fact that the attacker would have to compromise multiple different devices with different security profiles to achieve the ability to get the private root key _and_ publish malicious content. It's not great in theory, but in practice, a properly managed TUF service should be reasonably secure against anything but a malicious insider.

On the case of keeping old root keys around indefinitely, we've had some discussions about only keeping the last X years worth of root keys around on the basis that if somebody hasn't updated in that time, they really shouldn't expect it to work cleanly. Plenty of projects take that approach (I've been caught out by it myself when trying to update an old LTS install of Ubuntu a couple of months after they had taken down the repos for that release). What are your feelings on that approach?

mtrmac:
@endophage Re: keeping only last X years worth of keys around, at the moment I am treating this as an UI issue which can be worked out at the very last minute; certainly the rotation code should be able to deal with a completely lost access to a non-critical key, it’s a matter of whether we need/should make it easy/possible to give up on a key intentionally.

So far I can’t really see an advantage to voluntarily dropping the old keys; it’s not like the few kilobytes on disik hurt anything, and manually updating an old user would be a huge hassle.

But the meat of the question is really: what is the trust bootstrapping/maintenance design of Notary? Will there be a CA which can always allow obsolete clients to migrate to new keys = which can hijack a repository from the legitimate author? Will we go with the originally-intended TUF design, which treats bootstrapping as an one-time special event (TOFU/Trust-On-Only-Use) and then prefers to keep everything self-contained, mitigating risks by keeping root keys off-line, and having more than one root key co-signing? Or something completely different? I suspect this discussion is best held somewhere else than in a github ticket ☺

> It largely relies on the fact that the attacker would have to compromise multiple different devices with different security profiles to achieve the ability to get the private root key and publish malicious content.

AFAICS at the moment it is completely sufficient to compromise a single machine which holds the root private key and possibly credentials to push to the registry+notary; with the current implementations these two items are likely to be both located on the same on-line machine, and having them allows the attacker to both push new images to the registry and push new root.json (+ new tagging keys+all other metadata) to the notary. What am I missing?

endophage:
I think the thing you're missing is that the root key should never be left sitting online. Both our and TUF's recommendations are that it is kept offline, preferably in a safe, and only brought online for brief periods to create new repos or rotate other keys. The whole point of the key heirarchy is to have a key (the root key) that anchors trust and kept be kept offline. This currently already works in docker and notary and is how we manage our official images keys. Our root keys are on USB keys, in tamper evident bags, locked away.  When signing official images, we only need the targets and snapshots keys on our signing server.

mtrmac:
OK, that is doable — but still a single on-line device to compromise, even it must be compromised in some very small time windows. That got me confused. Thanks!

endophage:
@mtrmac sorry for dropping a sweeping change. We wanted to do a bit of a reveal for DockerCon with the YubiKey support. This is something we've been working on to address the security of bringing a root key online temporarily. Between the root key being stored in the YubiKey, and a physical touch being required to sign, it greatly increases the challenge to steal the private root key or use it to sign something the user doesn't want signed.

In the not too distant future we want to move all the signing keys onto hardware (albeit different YubiKeys for the root, vs targets+snapshot keys).

mtrmac:
FWIW I think the back-end part is close enough to done, and I would appreciate a review, especially if you have any comments on https://github.com/mtrmac/notary/commit/014e9768b530b4905a0e91d594105e325c8fe474 . (Please ignore the WIP commit for now.)

I’d be also happy to split any part of this into separate PRs if you want to deal with any subset independently.

tuf/tuf.go
```diff
@@ -143,13 +143,16 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
...
+	// Remove keys no longer in use by any roles, except for root keys.
+	// Root private keys must be kept in tr.cryptoService to be able to sign
+	// for rotation, and root certificates must be kept in tr.Root.SignedKeys
+	// because we are not necessarily storing them elsewhere (tuf.Repo does not
+	// depend on certs.Manager, that is an upper layer), and without storing
+	// the certificates in their x509 form we are not able to do the
+	// util.CanonicalKeyID conversion.
+	if role != data.CanonicalRootRole {
+		for k := range toDelete {
+			delete(tr.Root.Signed.Keys, k)
```
endophage: What am I missing? We definitely need/want to delete old root keys from the `Keys` map in the root data. These public keys must be loaded from a previous root file the client is already holding.

mtrmac: The *client* verifying signatures is holding a previous `root.json` file which contains the old root X509 public keys in `Signed.Role["root"]` (and which we want the client to migrate away from).

The *author* signing a new version the `root.json` may be completely done with that old X509 public key (for all we know, lacking more policy, it may now be several years after the author has migrated to a new trusted key from the old key we are talking about, or several generations of root keys after the old key, or the old root key may even known to be compromised)—but we still need to sign with the old key so that clients can rotate away from it. And in addition to just signing, we also need to store the X509 public key *somewhere* (a certificate can’t be regenerated from the raw non-X509 private key pair if we just know the ID=hash of the certificate).

For authors, we haven’t had designated a place to store the X509 public keys so far:
* It is not the `certs.Manager`, because the layering does not work well (as described in the comment), and because `certs.Manager` is used to manage the trusted certificates, and this old certificate may be known compromised and therefore not trusted = not suitable for `certs.Manager`.
* It is not a `trustmanager.KeyStore` because historically we haven’t been storing the generated X509 root public keys in there, so starting to use this location would require a migration step (though it could otherwise be a plausible candidate).

`root.json`’s `Signed.Keys` is actually the only place where the X509 public keys are currently being stored, so it seems easiest to designate this as the right place, and the implementation amounts to mostly the above hunk not deleting old keys which are already stored in there.

Also, in addition to just storing the X509 public keys, we need to be able to collect the set of X509 public keys which should be used to sign the `root.json`. Note that this is *not* `Signed.Roles["root"]` which contains keys trusted *for the future*, and should exclude the old known-compromised key. We could invent a special place to store the list of `root.json`-signing key IDs, but actually simply using the key IDs from the previous signature of `root.json` + the future trusted keys from `Signed.Roles["root"]` works just fine. Then the author’s copy of `root.json` becomes the canonical storage place for the root signing key IDs, so it seems natural to store the X509 public keys in the same place as well.

endophage: I would greatly prefer to use the previous root.json's signature list to decide which old keys to sign with. Leaving old root keys in the list creates a special case. Anyone looking at the root.json has to make the assumption or know already that orphaned keys entries are the special case of old root keys.

Using that model, ceasing to sign with an old key (i.e. because it is years old and considered no longer relevant to rotations) is as simple as deleting if from the keystore, rather than needing an additional command to remove the orphaned key from the root.json as well.

mtrmac: > I would greatly prefer to use the previous root.json's signature list to decide which old keys to sign with.

Yes, that’s what this PR is already doing (`oldKeyIDs` in `Repo.SignRoot`). But the old signature list only contains X509-based key IDs, which are sufficient neither to regenerate the X509 certificates nor to even locate the non-X509 private key which should be used for signing.

> Using that model, ceasing to sign with an old key (i.e. because it is years old and considered no longer relevant to rotations) is as simple as deleting if from the keystore

This also already works that way with this PR: The `oldKeyIDs` mentioned above eventually propagate to `signed.Sign`’s `optionalKeys`; if the underlying private key for any of the `optionalKeys` can not be loaded, this key ID will be silently skipped, and none of the future signatures of that `root.json` will use that key any more. The related X.509 certificate will be kept in `Signed.Keys` forever, but having unused public keys in there does not hurt anything AFAICS.

endophage: Yep, I see what you're saying now. Having old (unused) certs in the root is just cruft I'd like not to have but we need the certs somewhere :-/ I guess when we add CLI commands in general to add and remove keys from root.json it can handle these orphaned keys.

mtrmac:
> Can one of the admins verify this patch?

While I would appreciate any review, I am working on more tests right now, and the UI part will happen Real Soon Now, so if $you plan to do only one review, I’d prefer if you could wait a bit.

mtrmac:
This is now feature-complete and ready for review.

tuf/signed/sign.go
```diff
@@ -13,73 +13,76 @@ package signed
...
+			return err
 		}
 		privKeys[key.ID()] = k
 	}
 
-	// Check to ensure we have at least one signing key
-	if len(privKeys) == 0 {
-		return ErrNoKeys{KeyIDs: ids}
+	// Check to ensure we have enough signing keys
+	if len(privKeys) < minSignatures {
```
endophage: A user cannot be expected to reasonably be holding more than 1 current key for a given role. Holding more than one key defeats the whole point of thresholds. That's why my original check at the end was simply that at least 1 signature had been generated. If a user is asking to sign, they typically expect to generate exactly 1 signature for a role, not `Threshold` signatures.

tuf/signed/sign.go
```diff
@@ -13,73 +13,76 @@ package signed
...
-	// Check we produced at least on signature
-	if len(signatures) < 1 {
-		return ErrInsufficientSignatures{
-			Name: fmt.Sprintf(
-				"cryptoservice failed to produce any signatures for keys with IDs: %v",
-				ids),
-		}
-	}
-
-	for _, sig := range s.Signatures {
```
endophage: I've been working on an update to this little loop in #526 to clean up any signatures that are no longer valid. Given the expected behaviour/use case of thresholds, it's incorrect to simply delete all the other signatures.

tuf/signed/sign.go
```diff
@@ -13,73 +13,76 @@ package signed
...
 		canonicalID, err := utils.CanonicalKeyID(key)
-		ids = append(ids, canonicalID)
 		if err != nil {
-			continue
+			return err
 		}
 		k, _, err := service.GetPrivateKey(canonicalID)
 		if err != nil {
-			continue
+			if _, ok := err.(trustmanager.ErrKeyNotFound); ok {
```
endophage: I'd accept any error here. Whether it's actually missing or the file is corrupted, it's worth messaging the key ID clearly. For the same reason, it should always continue, it should not preemptively return error. 

Absence of a key (for any reason) is not sufficient reason to fail signing entirely.

tuf/signed/sign.go
```diff
@@ -13,73 +13,76 @@ package signed
...
+			NeededKeys: minSignatures, MissingKeyIDs: missingKeyIDs}
+	}
+
+	for _, key := range optionalKeys {
+		if _, ok := privKeys[key.ID()]; ok {
+			continue
+		}
+		canonicalID, err := utils.CanonicalKeyID(key)
+		if err != nil {
+			return err
```
endophage: Particularly because this is an optional key, this should continue.

tuf/signed/sign.go
mtrmac: We are completely unprepared for true multi-person signing in any case: for that to work, there would have to be a mechanism for the first user to sign an updated file, and then to transfer the half-signed file to the other user, who would sign it again and then publish. We don’t have any file transfer mechanism other than the server (which refuses, or should refuse, half-signed files, and they should not be visible to the public anyway), and [incrementing the version = breaking old signatures, signing, and publishing] is linked into a single `NotaryRepository.Publish()` so there is not even a way to do this manually at the moment.

All of the code would have to be restructured to support thresholds anyway; this is at worst irrelevant to such a restructuring.

(Perhaps this is actually moving towards the easiest way to _implement_ thresholds: We could create a new `RemoteCryptoService` which transfers blobs for signing to the other user. That would allow us to implement multiple signatures without touching the `NotaryRepository.Publish` workflow at all — and in such a system, we do want to fail if we can not collect enough signatures, just like this PR is doing.)

tuf/signed/sign.go
mtrmac: _Every_ non-test caller of `signed.Sign` (or `tuf.Repo.Sign` which calls it) breaks the signature by either incrementing `Version` or creating new data from scratch. So the check in #526 currently simplifies to
> signatures := make([]data.Signature, 0, len(privKeys))

as this PR is doing.

Re: thresholds, see above.

tuf/signed/sign.go
mtrmac: > Absence of a key (for any reason) is not sufficient reason to fail signing entirely.
_Absence_ of a key, no.

For a corrupted key or key store, doesn’t it make more sense to fail early (e.g. while recent backups are still available)?

We could have `ErrInsufficientSignatures` collect a map of key ID → failure reason, but that’s a fairly non-obvious error reporting model: Are UIs prepared to deal with a kilobyte of text if 5 keys fail? Can we even make this readable without newlines? And all of this for a hopefully very rare storage or HSM failure.

tuf/signed/sign.go
mtrmac: This does continue if `service.GetPrivateKey` fails (and I am a bit worried about that as well, TBH).

The only way `utils.CanonicalKeyID` can fail here is if the key has an unknown algorithm, or if PEM parsing fails. These things pretty much by definition can’t happen for our `optionalKeys`, which are previously successfully used root keys (probably including signature validation in this very client); if `utils.CanonicalKeyID` fails here, something is very wrong and we would want to know.

tuf/signed/sign.go
endophage: The thinking was that Publish would largely work the same but the server would accept partially signed files. Those files would not be snapshotted/timestamped in to the repo, and there would be some new mechanism for clients to request partially signed files that they could add a signature to.

As such, a current client could be the initiator of the first push of partially signed content.

tuf/signed/sign.go
endophage: That seems like a poor reason for Sign to not do the right thing. Sure we could add it back in later, but why not just do it right the first time?

endophage:
I get what you're reiterating in your comments about what the client does now. However we should be implementing code that moves us towards completing the TUF spec, not things that take us further away from it. It's highly unlikely that a massive PR or collection of PRs lands together that completely finishes implementing the spec, so we have to do it piecemeal.

tuf/signed/sign.go
endophage: If an error here is a failure case, then these keys aren't optional.

tuf/signed/sign.go
mtrmac: Anyway, such a multi-person signing can be easily supported by setting `minSignatures` to 1 instead of the role’s threshold. The API as implemented in this patch will do the right thing when a bigger threshold is somehow set up (and you are right, this likely means failing—but failing is the right thing to do right now), and it does not preclude future multi-signatures.

tuf/signed/sign.go
mtrmac: I still disagree, but fair enough. Do you want me to submit a version with a `signed.AppendSignature` which ~preserves the current appending semantics, and a `signed.Sign` helper used by the current callers which don’t want the appending semantics?

tuf/signed/sign.go
mtrmac: Would renaming the parameter to something like `oldRoleKeys` help, then? Or do you think that failures signing with old keys must never prevent signing with the currently trusted keys?

mtrmac:
Rebased again. Is there any way I can help move this forward?

cyli:
@mtrmac Thank you so much for all your work on this and apologies for the delay - we were working on some other changes that were a priority, which has unfortunately created a massive merge conflict with this PR.  I'm working on resolving that in another branch on my fork currently, since we were the ones that caused the massive conflict.

I'm not sure how you would like to proceed afterward - I would be happy to give you write access to that fork so you can continue working on changes, or I can address any outstanding issues as per your and @endophage's comments.

tuf/tuf.go
cyli: As @endophage suggested, we can add old root keys to a non-spec role in root.json.

cyli:
Superceded by #648, which has been merged, so closing.

PR #278, opened 2015-11-13
More defensive coding around listing our keys in the yubikey.

cyli:
If there are no attributes matching the pkcs11 template we specified, then we should bail and move onto the next object.


cyli:
Tests pass when running with a yubikey.  I can't reproduce the garbage data that seemed to cause this that one time.

diogomonica:
LGTM

PR #279, opened 2015-11-13
Pretty print output of `notary key list`.

### LINKED ISSUE #273: Pretty-print keys when listing keys
So that they are are shown on the command line in fixed-width columns, similar to what is displayed when listing images or containers in docker.

cyli:
The output is in sorted order by gun, by role, by key ID, and by location, and looks like:

```
    ROLE     GUN                                 KEY ID                                        LOCATION          
-----------------------------------------------------------------------------------------------------------------
  root              4e7621b361d21344eaa4d4f0dff24cc3db4eb305dbe7f5009b3592e16908b0ad   yubikey                   
  root              95e4c4e2c65b5b4e088c64a284f89c505eee60e0c741c7e1788bb2239c8d1173   file (/tmp/xyzf/private)  
  snapshot   repo   d7897f2e74403c9c4939dd8a975c61ac1031db5511ec5f31d06b6ed4772643ee   file (/tmp/xyzf/private)  
  targets    repo   7cc91433422c0f3f1a2b3973955625c907ac79ef2621d4af9e329217c634a79b   file (/tmp/xyzf/private) 
```

If there are no keys, the message is:
```
No signing keys found.
```

Fixes #273.


diogomonica:
Wow, this looks great!

cmd/notary/integration_test.go
```diff
@@ -162,23 +161,50 @@ func splitLines(chunk string) []string {
...
-			strings.TrimPrefix(strings.TrimSpace(part), "# Root keys:"))
-		sort.Strings(fixed[i])
+	var (
+		rootMap    = make(map[string]bool)
+		nonrootMap = make(map[string]bool)
+		root       []string
+		nonroot    []string
+	)
+	// first two lines are header
+	for _, line := range lines[2:] {
```
diogomonica: Small nit, but what if output is empty, and lines is empty too? Wouldn't lines[2:] fail?

cmd/notary/keys.go
```diff
@@ -88,51 +91,117 @@ var cmdKeyImportRoot = &cobra.Command{
...
+
+	table := tablewriter.NewWriter(writer)
+	table.SetHeader([]string{"ROLE", "GUN", "KEY ID", "LOCATION"})
+	table.SetBorder(false)
+	table.SetColumnSeparator(" ")
+	table.SetAlignment(tablewriter.ALIGN_LEFT)
+	table.SetCenterSeparator("-")
+	table.SetAutoWrapText(false)
+
+	for _, oneKeyInfo := range info {
```
endophage: We should have a "no signing keys found" message or similar should info be empty.

endophage:
Assuming a message is added about no keys being found (or if I've just missed it and it's in there already) LGTM.

cmd/notary/keys.go
cyli: Do you mean instead of the table header, or under it?  E.g:

````
ROLE  GUN ...
-------------
      (no signing keys found)
```

Or just 
```
No signing keys found.
```

?

cmd/notary/integration_test.go
cyli: Currently, it just prints out the headers and nothing else like docker does (headers plus the horizontal line).  But as per discussion with @endophage, we should probably print out some kind of message about there being no keys.

cmd/notary/keys.go
cyli: No headers seems neater and easier to test, so I'm going to go with "No signing keys found."

cyli: Fixed in https://github.com/docker/notary/commit/eb9de9f0e89844c60a5ea8e0123656cd02c4d470

cmd/notary/integration_test.go
cyli: That message was added in https://github.com/docker/notary/commit/eb9de9f0e89844c60a5ea8e0123656cd02c4d470, and the test fixed to check for the message before moving on.

PR #281, opened 2015-11-13
Log whether a pkcs11 library was found and if it was loadable.

cyli:

This unfortunately prints out every time any operation is done on the Yubikey, producing a lot of log output, but perhaps that is better because an operation might fail at any given time.

Output if no Yubikey:
`DEBU[0000] Failed to initialize PKCS11 environment: loaded library /usr/local/lib/libykcs11.dylib, but no HSM slots found 
`

If there is a Yubikey:
`DEBU[0000] Initialized PKCS11 library /usr/local/lib/libykcs11.dylib and started HSM session`

Fixes #281.


endophage:
I'm ok with the output for experimental. LGTM

PR #282, opened 2015-11-13
Fixes client to report problems contacting the remote server.

### LINKED ISSUE #277: Change error on 5XX to not be "no trust data available"
DEBU[0000] received HTTP status 503 when requesting root. 
* fatal: no trust data available

cyli:
Currently, when listing, publishing, or getting a particular target, if the remote server errors, the client attempts to load it from a local cache.  However, if there is no local cache, it just returns Metadata Not Found for listing and getting.  Have it report the remote the original remote error instead of Metadata Not Found locally.

Fixes #277.


endophage:
Very nice. LGTM.

PR #285, opened 2015-11-14
Add an interactive command to delete a key from any keystore.

cyli:
This lists any matching keys, and requires the user to pick which one
to choose, if there is more than 1 matching key.  Also requires the
user to confirm before deleting.

Example output:

screen shot 2015-11-14 at 2 39 50 am


cmd/notary/keys.go
```diff
@@ -91,6 +94,13 @@ var cmdKeyImportRoot = &cobra.Command{
 	Run:   keysImportRoot,
 }
 
+var cmdKeyRemoveKey = &cobra.Command{
```
endophage: Can we just do `cmdRemoveKey` or `cmdKeyRemove`? `KeyRemoveKey` just feels so redundant.

cmd/notary/keys.go
```diff
@@ -411,6 +421,101 @@ func keysRotate(cmd *cobra.Command, args []string) {
...
+	if err != nil {
+		return err
+	}
+
+	fmt.Fprintf(out, "\nDeleted %s.\n", keyDescription)
+	return nil
+}
+
+// keysRemoveKey deletes a private key based on ID
+func keysRemoveKey(cmd *cobra.Command, args []string) {
```
endophage: I'd call this just `removeKey`

cmd/notary/keys.go
```diff
@@ -411,6 +421,101 @@ func keysRotate(cmd *cobra.Command, args []string) {
 	}
 }
 
+func removeKey(keyStores []trustmanager.KeyStore, keyID string,
```
endophage: And I'd call this `removeKeyUX` or `removeKeyInteractivity`

PR #289, opened 2015-11-14
pretty printing targets

endophage:
Uses the same tablewriter we're using for keys.

### DIFF
--- a/cmd/notary/tuf.go
+++ b/cmd/notary/tuf.go
@@ -3,13 +3,16 @@ package main
 import (
 	"bufio"
 	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net"
 	"net/http"
 	"net/url"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


cmd/notary/tuf.go
```diff
@@ -447,3 +448,38 @@ func getRemoteTrustServer() string {
...
+		writer.Write([]byte("No targets present in this repository.\n"))
+	}
+
+	sort.Stable(targetsSorter(ts))
+
+	table := tablewriter.NewWriter(writer)
+	table.SetHeader([]string{"Name", "Digest", "Size (bytes)"})
+	table.SetBorder(false)
+	table.SetColumnSeparator(" ")
+	table.SetAlignment(tablewriter.ALIGN_LEFT)
```
cyli: Non-blocking - I vaguely remember a slack discussion about whether this should be centered?  If so, should this line be removed (IIRC tablewriter is by default centered, but could also specify a center alignment)

cyli:
LGTM!

cmd/notary/tuf.go
endophage: We realized that was going to make the tag names look really bad so didn't do it in the end :-/

PR #290, opened 2015-11-14
Make vet target now also checks for filenames with _test_ in the middle.

cyli:
Since if they are generic test helpers that should be exported,
they should probably go into their own package.  If they are used
during testing, they should end in _test.


diogomonica:
We should probably be more strict and just check the existence of the word test in the file.

endophage:
LGTM

Issue #275, opened 2015-11-12
Client docs


cyli:
How to use the notary command line client to manage repositories and rotate signing keys.

cyli:
Sorry, this is a duplicate of #187.  Closing in favor of that.

PR #291, opened 2015-11-15
Server and signer docs

cyli:
- [x] Purpose of the server
- [x] Attack against the server
- [x] Document server config file
- [x] storage and signer backend options
- [x] TLS certificates
- [x] Script showing how to generate TLS certificates
- [x] health and stat endpoints

Note this should probably be merged in after #293


opensslCertGen.sh
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env bash
+
+CLIENT_USAGE="Generate a self-signed client cert and key to be used in mutual TLS.\n${0} client [-o <output name>]"
+ROOT_USAGE="Generate CA root key and certificate which can only be used to sign leaf cert and not other CAs.\n${0} root [-o <output name>]"
+SERVER_USAGE="Generate a server cert key and certificate.\n${0} root -n <common name> [-o <output name>] [-r <root key if don't want it self-signed>] [-a <subjectAltName>]...\n\nExample:\n${0} server -o servertls -r rootCA.key notary-server DNS:notaryserver DNS:notary_server IP:127.0.0.1"
```
endophage: I'm guessing it should be `server -n <common name> ...`

endophage: Is `output name` a directory?.

Example doesn't contain `-a` and should the ellipsis be inside the `[-a ]` to indicate multiple values passed to that flag?

opensslCertGen.sh
```diff
@@ -0,0 +1,135 @@
...
+
+if [[ -z "${1}" ]]; then
+    printf "${ROOT_USAGE}\n\n${CLIENT_USAGE}\n\n${SERVER_USAGE}\n\n"
+    exit 1
+fi
+
+OPENSSLCNF=
+for path in /etc/openssl/openssl.cnf /usr/local/etc/openssl/openssl.cnf; do
```
endophage: `/etc/ssl/openssl.cnf` is also a conventional linux location.

endophage:
I'd probably be more opinionated and always generate a self signed cert. Users wanting more security than that should be going and getting certs from a public CA, or will already have their own internal CA.

diogomonica:
I refactored some server stuff in #293

cyli:
@diogomonica Awesome, thanks!  I'll just document the config options more thoroughly here then, I guess, and the HTTP API.

docs/notary-server-config.md
```diff
@@ -0,0 +1,160 @@
...
+
+```json
+"server": {
+	"addr": ":4443",
+	"tls_key_file": "./fixtures/notary-server.key",
+	"tls_cert_file": "./fixtures/notary-server.crt"
+}
+```
+
+- `addr` - **required** - the TCP address (IP and port) to listen on.  Examples:
```
cyli: TODO: document auth config

docs/notary-server-config.md
```diff
@@ -0,0 +1,160 @@
...
+
+The reporting section contains any configuration for reporting errors, etc. to
+services via [logrus hooks](https://github.com/Sirupsen/logrus).  Currently the
+only supported services is [Bugsnag](https://bugsnag.com).  (See
+[bugsnag-go](https://github.com/bugsnag/bugsnag-go/) for more information about
+configuration.
+
+```json
+"reporting": {
+	"bugsnag": "yes",
```
cyli: Should we make this a subtree instead?  e.g.:

```json
"reporting" : {
    "bugsnag": {
    ...
    }
}
```

Does it need to remain backwards compatible?

docs/notary-server-config.md
```diff
@@ -0,0 +1,160 @@
...
+
+- `tls_key_file` and `tls_cert_file` - **optional**, but must be provided together
+	if provided at all.  These specify the private key and TLS cert,
+	respectively, to use for HTTPS.  If not provided, the server will use HTTP
+	instead.
+
+## `trust service` section (optional)
+
+The trust service section must be included in order to specify a remote trust
+service.  If it is left out, a local in-memory trust service will be used.
```
cyli: Not quite true yet - see https://github.com/docker/notary/issues/298

docs/notary-server-config.md
```diff
@@ -0,0 +1,161 @@
...
+
+- `tls_key_file` and `tls_cert_file` - **optional**, but must be provided together
+	if provided at all.  These specify the private key and TLS cert,
+	respectively, to use for HTTPS.  If not provided, the server will use HTTP
+	instead.
+
+## `trust service` section (optional)
+
+The trust service section must be included in order to specify a remote trust
+service.  If it is left out, a local in-memory ED25519 trust service will be
```
cyli: Not quite true yet - see #298

cyli: See https://github.com/docker/notary/pull/299.

docs/notary-server-config.md
```diff
@@ -0,0 +1,353 @@
...
+}
+```
+<table>
+	<tr>
+		<th>Parameter</th>
+		<th>Required</th>
+		<th>Description</th>
+	</tr>
+	<tr>
+		<td valign="top"><code>addr</code></td>
```
cyli: https://github.com/docker/notary/pull/304 changes this to `http_addr`

docs/notary-server-config.md
```diff
@@ -0,0 +1,353 @@
...
+			</ul>
+		</td>
+	</tr>
+	<tr>
+		<td valign="top"><code>tls_key_file</code></td>
+		<td valign="top">no</td>
+		<td valign="top">Specifies the private key to use for HTTPS.  Must be
+			provided together with <code>tls_cert_file</code>, or not at all.
+			If neither are provided, the server will use HTTP instead of HTTPS.
+			The path is relative to the current working directory where
```
cyli: This is changed in https://github.com/docker/notary/pull/304 to be relative to the configuration file path, here and in `tls_cert_file`, `trust_server.tls_client_cert`, `trust_server.tls_client_key`, and `trust_server.tls_ca_file`.

docs/notary-server-config.md
cyli: #304 Updates this to be 
```
"reporting" : {
    "bugsnag": {
    ...
    }
}
```
instead.

docs/notary-server.md
```diff
@@ -0,0 +1,154 @@
...
+
+If you would like to specify a different `db_url`, the full path from the top
+of the configuration tree is `storage -> db_url`, so the environment variable
+to set would be `NOTARY_SERVER_STORAGE_DB_URL`.
+
+Note that you cannot override an intermediate level name.  Setting
+`NOTARY_SERVER_STORAGE=""` will not disable the MySQL storage.  Each leaf
+parameter value must be set indepedently.
+
+#### Running a Docker image
```
cyli: Note: I'm writing about this as if the official docker image already exists - not sure what the name will be, so this may need to be fixed.

docs/notary-signer-config.md
```diff
@@ -0,0 +1,224 @@
...
+# Notary Signer Configuration File
+
+An example (full) server configuration file.
+
+```json
+{
+	"server": {
+		"http_addr": ":4444",
+		"grpc_addr": ":7899",
+		"cert_file": "./fixtures/notary-signer.crt",
```
cyli: #304 Makes this `tls_cert_file` and `tls_key_file`, to be more consistent with notary-server.

docs/notary-signer-config.md
```diff
@@ -0,0 +1,224 @@
...
+
+The logging section sets the log level of the server.  If not provided, or if
+any part of this section is invalid, the server defaults to an ERROR logging
+level.
+
+Example:
+
+```json
+"logging": {
+	"level": 2
```
cyli: #304 Fixes this to be a string, like in notary-server

cyli:
I also didn't document the environment variables
```
ENV NOTARY_SIGNER_DEFAULT_ALIAS="timestamp_1"
ENV NOTARY_SIGNER_TIMESTAMP_1="testpassword"
```

I'm wondering if we can remove them, and maybe remove the passphrase retriever from notary-signer?  That seemed to have been for HSM stuff.  Also, the signer is still based on the golang-softhsm image.  Do we want to keep that?

docs/notary-server-config.md
```diff
@@ -0,0 +1,355 @@
...
+			</ul>
+		</td>
+	</tr>
+	<tr>
+		<td valign="top"><code>tls_key_file</code></td>
+		<td valign="top">no</td>
+		<td valign="top">The path to the private key to use for
+			HTTPS.  Must be provided together with <code>tls_cert_file</code>,
+			or not at all. If neither are provided, the server will use HTTP
+			instead of HTTPS. The path is relative to the directory where
```
cyli: This changes in https://github.com/docker/notary/pull/304 to be relative to the configuration file path, here and in `tls_cert_file`, `trust_server.tls_client_cert`, `trust_server.tls_client_key`, and `trust_server.tls_ca_file`.

docs/notary-server-config.md
```diff
@@ -0,0 +1,355 @@
...
+	},
+	"logging": {
+		"level": "debug"
+	},
+	"storage": {
+		"backend": "mysql",
+		"db_url": "dockercondemo:dockercondemo@tcp(notarymysql:3306)/dockercondemo"
+	},
+	"reporting": {
+		"bugsnag": "yes",
```
cyli: #304 Updates this to be 
```
"reporting" : {
    "bugsnag": {
    ...
    }
}
```
instead, and also adds bugsnag support to signer.

docs/notary-server-config.md
```diff
@@ -0,0 +1,355 @@
...
+			the registry token configuration documentation</a>
+			for the parameter details.</td>
+	</tr>
+</table>
+
+## `trust service` section (optional but recommended)
+
+This section is required to specify a remote trust service, such as
+[Notary Signer](notary-signer.md).  If it is left out or invalid, a local
+in-memory ED25519 trust service will be used instead.
```
cyli: Sort of - see https://github.com/docker/notary/pull/299.

cyli:
@endophage @NathanMcCauley (and @ d i o g o m o n i c a when he gets back) PTAL?

moxiegirl:
@cyli I'll make to review by Monday.  Thank you for the ping @cyli (and cool avatar).

cyli:
@moxiegirl Thank you!  No hurry :)  There will be more documentation in #275, but still working on that.

docs/notary-server.md
```diff
@@ -0,0 +1,188 @@
...
+Get the official Docker image, which comes with [some sane defaults](
+https://github.com/docker/notary/blob/master/fixtures/server-config-local.json),
+which include a remote trust service but local in-memory backing store.
+
+You can override the default configuration with environment variables.
+For example, if you wanted to run it with just a local signing service instead
+(not recommended for production):
+
+```
+$ docker pull docker.io/docker/notary-server
```
diogomonica: this doesn't work

docs/notary-server.md
```diff
@@ -0,0 +1,188 @@
...
+	-e NOTARY_SERVER_TRUST_SERVICE_TYPE=local
+	notary-server
+```
+
+Alternately, you can run the image with your own configuration file entirely.
+You just need to mount your configuration directory, and then pass the path to
+that configuration file as an argument to the `docker run` command:
+
+```
+$ docker run -p "4443:4443" -v /path/to/config/dir/on/host:/path/in/container \
```
diogomonica: I think it's ok to be prescriptive of a path in container here. /etc/docker/notary-server/config.json maybe?

docs/notary-signer-config.md
```diff
@@ -0,0 +1,249 @@
...
+		<td valign="top">yes if not <code>memory</code></td>
+		<td valign="top">This parameter specifies the alias of the current
+			password used to encrypt the private keys in the DB.  All new
+			private keys will be encrypted using this password, which
+			must also be provided as the environment variable
+			<code>NOTARY_SIGNER_&lt;default_alias_value&gt;</code>.</td>
+	</tr>
+</table>
+
+**Required environment variable:**
```
diogomonica: I think the default alias needs a better explanation (in order to support key rotation, keys get created with a field that says what password they have been encrypted with, bla bla)

docs/notary-signer-config.md
```diff
@@ -0,0 +1,249 @@
...
+<code>NOTARY_SIGNER_&lt;default_alias_value&gt;</code>
+
+**Optional environment variables:**
```
diogomonica: I don't think this is true. I looked at the code and I can't see anywhere where we load multiple env variables for old keys.

The current passphraseRetriever in main.go (signer is):

```
func passphraseRetriever(keyName, alias string, createNew bool, attempts int) (passphrase string, giveup bool, err error) {
	passphrase = mainViper.GetString(strings.ToUpper(alias))

	if passphrase == "" {
		return "", false, errors.New("expected env variable to not be empty: " + alias)
	}

	return passphrase, false, nil
}
```

docs/notary-signer-config.md
```diff
@@ -0,0 +1,249 @@
...
+		"tls_key_file": "./fixtures/notary-signer.key",
+		"client_ca_file": "./fixtures/notary-server.crt"
+	},
+	"logging": {
+		"level": 2
+	},
+	"storage": {
+		"backend": "mysql",
+		"db_url": "user:pass@tcp(notarymysql:3306)/databasename?parseTime=true",
+		"default_alias": "password1"
```
diogomonica: I think this example should be "passwordalias1"

docs/notary-server.md
cyli: Yeah official images aren't up yet since automated builds expect dockerfiles to be just `/Dockerfile` currently.  Waiting for some changes to hub.

docs/notary-signer-config.md
cyli: That's true - but the alias gets passed in.  So if there are still keys in the db with the old alias, [passphraseRetriever is called with that old alias](https://github.com/docker/notary/blob/master/signer/keydbstore/keydbstore.go#L124).

[New keys are created with the default alias](https://github.com/docker/notary/blob/master/signer/keydbstore/keydbstore.go#L73).

diogomonica: Makes sense.

diogomonica:
LGTM

diogomonica:
@cyli this seems to be stuck on codecov. How do we kick it?

cyli:
@diogomonica Sorry, I thought I had rebased, but it was based on an old version of master that was not reporting to codecov.  Hopefully this should fix it.

docs/notary-server.md
```diff
@@ -0,0 +1,189 @@
+<!--[metadata]>
```
moxiegirl: @cyli This file could be better. I started rewrites but ran into a lot of questions. So, I"ll post what I have with the questions. I think we should sit down and talk it out.  I think it is important to provide an architecture diagram and also some kind of diagram that shows the flow of information among the component parts.  If you want to sketch these I'm happy to formalize them.

Often, it is not clear what is pluggable and what is not in the architecture. We need to make that clear. Also, the installation, configuration, and location on a network of the Notary server not clear.  I suspect I only need to deal with any of this if I'm standing up a private repository --- if that's right, we need to say this up front.

I'll attach my rewrite so far below and we can discuss it next week. 

-------



# Notary server architecture

The Notary Server secures the interaction between a client and a  Docker registry. This page describes the components in the Notary server that secure these communications.

## Overview of Notary interactions

The Notary Server stores and updates a repository's signed TUF (The Update Framework) metadata files. TUF support Notary's system for managing image security. Metadata files are generated and signed by the client at upload time. These files include root, snapshot, and targets files.  

**REVIEWER**: Are the clients uploading images as well with the metadata --- I think so -- any problem with mention that here?  What kind of conflicts can happen? Are conflicts possible the first time an image is uploaded? Is the very first image upload different than the subsequent uploads?  Is the Notary Signer pluggable? The original language implied it.

When clients upload images they also upload these metadata files. The Notary
server checks the metadata files for conflicts and then verifies both the
signatures and keys. If the checks pass, the server generates and signs a
timestamp metadata file for the repository. The timestamp metadata file
certifies that the clients' uploaded files are the most recent for that
repository.

A remote key storage/signing service makes the timestamp creation and signing
possible. You'll learn more about the key storage/signing service later in this
document.


### Authentication

Notary Server supports authentication from clients using JSON Web Tokens (JWT).
JWT is an open, industry standard method for representing claims securely
between two parties.  JWT requires an authorization server to manage access
controls. A certificate bundle from this authorization server contains the
public key the server uses to sign tokens. A Notary Server is configured to trust signatures from this authorization server.

**REVIEWER**: Under what circumstances would Notary Server have token authentication *disabled*?  If a client makes a request and is redirected to JWT, do they have reissues the first request?

If token authentication is enabled on Notary Server, then any client that does
not have a token is redirected to the JWT authorization server. The client must
log in to the authorization server, obtain a token, and present that token to
Notary Server on future requests.

For more information about **XXXXX**, please see the docs for [Docker Registry
v2 authentication](
https://github.com/docker/distribution/blob/master/docs/spec/auth/token.md)

### Server storage

Notary Server uses MySQL as a backend for storing the timestamp public keys and
the TUF metadata for each repository.  It relies on a signing service to store
the private keys.

**REVIEWER**: Is the Server storage backend configurable?

### Signing service

**REVIEWER**: Which private keys are

You can store the private keys on the XXXX server or you can configure the
Notary Server to use a separate, remote signing service. Using a remote signing
service is preferred over storing the keys on the server itself. By default,
Docker users the Notary Signer signing service.

This signing service generates and stores the timestamp private keys and
manages signing for the Notary server. Notary Signer supports mutual authentication. This mean you need to generate generate client
certificates for your deployment of Notary Server.  These client certificates **are not CAs**. Were they CAs, a compromised Notary server could sign any number of other client certs.

To see how to generate client SSL certs with basic constraints using OpenSSL, see [this OpenSSL certificate generation script](opensslCertGen.sh) .

## How to configure and run Notary Server

You use a JSON configuration file to configure Notary Server. In this file you configure:

* the server's address and signatures
* the signing service (optional)
* storage backend
* authentication service
* logging and reporting

A full configuration file looks like the following:


```json
{
	"server": {
		"http_addr": ":4443",
		"tls_key_file": "./fixtures/notary-server.key",
		"tls_cert_file": "./fixtures/notary-server.crt",
	},
	"trust_service": {
		"type": "remote",
		"hostname": "notarysigner",
		"port": "7899",
		"key_algorithm": "ecdsa",
		"tls_ca_file": "./fixtures/root-ca.crt",
		"tls_client_cert": "./fixtures/notary-server.crt",
		"tls_client_key": "./fixtures/notary-server.key"
	},
	"storage": {
		"backend": "mysql",
		"db_url": "user:pass@tcp(notarymysql:3306)/databasename?parseTime=true"
	},
	"auth": {
		"type": "token",
		"options": {
			"realm": "https://auth.docker.io/token",
			"service": "notary-server",
			"issuer": "auth.docker.io",
			"rootcertbundle": "/path/to/auth.docker.io/cert"
		}
	},
	"logging": {
		"level": "debug"
	},
	"reporting": {
		"bugsnag": {
			"api_key": "c9d60ae4c7e70c4b6c4ebd3e8056d2b8",
			"release_stage": "production"
		}
	}
}
```
**REVIEWER**: Where do I do this override on the client end? Why would I override the configuration?

 Once the Notary server is configured, you can override the configuration
parameters by setting environment variables in the form `NOTARY_SERVER_var`.
Where `var` is the ALL-CAPS, `"_"`(underscore) delimited full path of the configuration's JSON keys.  For instance, consider the storage URL of
the Notary Server configuration:

```json
"storage": {
	"backend": "mysql",
	"db_url": "dockercondemo:dockercondemo@tcp(notary-mysql)/dockercondemo"
}
```

The full JSON path of the keys is `storage -> db_url`. So the environment
variable you'd need to set would be `NOTARY_SERVER_STORAGE_DB_URL`.  For example, if running the binary:

```
$ export NOTARY_SERVER_STORAGE_DB_URL=myuser:mypass@tcp(my-db)/dbname?parseTime=true
$ NOTARY_SERVER_LOGGING_LEVEL=info notary-server -config /path/to/config.json
```

You can only override keys whose values are strings or numbers. You cannot
override a key whose value is another map. For instance, setting
`NOTARY_SERVER_STORAGE='{"storage": {"backend": "memory"}}'` does not set
in-memory storage.  It fails to parse.  

## Running Notary Server

**Reviewer**: There is something missing here. Who runs Notary Server? What kind of server does it need if any. Does Docker need to be running (looks like yes) Does it need to be replicated -- what happens if it goes down?

You run Notary Server from ####. The Notary Server is a Docker image you run on a server in your network.  

- `-config=<config file>` - The JSON configuration file.

- `-debug` - Passing this flag enables the debugging server on `localhost:8080`.
	The debugging server provides [pprof](https://golang.org/pkg/net/http/pprof/)
	and [expvar](https://golang.org/pkg/expvar/) endpoints.


Get the official Docker image, which comes with [some sane defaults](
https://github.com/docker/notary/blob/master/fixtures/server-config-local.json),
which include a remote trust service but not (?) local in-memory backing store.

You can override the default configuration with environment variables.
For example, if you wanted to run it with just a local signing service instead
(not recommended for production):

```
$ docker pull docker.io/docker/notary-server
$ docker run -p "4443:4443" \
	-e NOTARY_SERVER_TRUST_SERVICE_TYPE=local
	notary-server
```

Alternately, you can run the image with your own configuration file entirely.
You just need to mount your configuration directory, and then pass the path to
that configuration file as an argument to the `docker run` command:

```
$ docker run -p "4443:4443" \
	-v /path/to/config/dir/on/host:/etc/docker/notary-server/ \
	notary-server -config=/etc/docker/notary-server/config.json
```

You can also pass the `-debug` flag to the container in addition to the
configuration file, but the debug server port is not exposed by the container.
In order to view the debug endpoints, you will have to `docker exec` into
your container.

### What happens if the server is compromised?

The server does not hold any keys for repositories, except the for timestamp
keys if you are using a local signing service, so the attacker cannot modify
the root, targets, or snapshots metadata.

If you are using a signer service, an attacker cannot get access to the
timestamp key either. They can use the server's credentials to get the signer
service to sign arbitrary data, such as an empty timestamp,
an invalid timestamp, or an old timestamp.

However, TOFU (trust on first use) would prevent the attacker from tricking
existing clients for existing repositories to download arbitrary data.
They would need the original root/target/snapshots keys to do that. The
attacker could only, by signing bad timestamps, prevent the such a user from
seeing any updated metadata.

The attacker can also make all new keys, and simply replace the repository
metadata with metadata signed with these new keys.  New clients who have not
seen the repository before will trust this bad data, but older clients will
know that something is wrong.

### Ops features

Notary Server provides the following features for operational friendliness:

1. A health endpoint at `/_notary_server/health` which returns 200 and a
	body of `{}` if the server is healthy, and a 500 with a map of
	failed services if the server cannot access its storage backend.

	If it cannot contact the signing service, an error will be logged but the
	service will still be considered healthy, because it can still serve
	existing metadata.  It cannot accept updates, so the service is degraded.

1. A [Bugsnag](https://bugsnag.com) hook for error logs, if a Bugsnag
	configuration is provided.

1. A [prometheus](http://prometheus.io/) endpoint at `/_notary_server/metrics`
	which provides HTTP stats.


## Related information

* TUF metadata files](
https://github.com/theupdateframework/tuf/blob/develop/docs/tuf-spec.txt#L348)
* [Notary Signer](notary-signer.md)
* [JWT](http://jwt.io/)
[Docker Registry v2 authentication](
https://github.com/docker/distribution/blob/master/docs/spec/auth/token.md)
As an example, please see [this script](opensslCertGen.sh) to see how to
generate client SSL certs with basic constraints using OpenSSL.
Please see the
[Notary Server configuration document](notary-server-config.md)
for more details about the format of the configuration file.

cyli:
As per discussion with @moxiegirl, we're going to merge this one, and continue addressing doc shortfalls in #293

PR #292, opened 2015-11-15
The NotarySigner cryptoservice now implements GetPrivateKey.

cyli:
Previously, because it's a CryptoService wrapper around a remote signer
service, it returned nil all the time.  Now, because signing is done
via private key more than CryptoService, it has to return a PrivateKey.

The key doesn't have private bytes, but can be used for signing.


signer/remote/signer_trust.go
```diff
@@ -22,6 +24,80 @@ type checkableConnectionState interface {
 	State() grpc.ConnectivityState
 }
 
+// RemotePrivateKey is a key that is on a remote service, so no bytes are
+// availabble
+type RemotePrivateKey struct {
+	data.PublicKey // this will be a generic data.PrivateKey with empty bytes
```
endophage: Comment doesn't make sense.

endophage:
Post merge LGTM

HuKeping:
what is the `private key` here mean? is it the `root` key?

cyli:
@HuKeping The interface specifies a GetPrivateKey function which returns the private key bytes of an RSA, ECDSA, or ED25519 key, as opposed to the public bytes.  In the case of NotarySigner, this would be the private key of the timestamp key.

endophage:
@HuKeping  We should make clear, in this specific context it doesn't return the actual private bytes. The `PrivateKey` interface has been updated to have a `Sign` method. This allowed us to better segment and abstract signing code for different algorithms, reducing the number of `switch` statements we had.

The `PrivateKey` object returned by `NotarySigner` implements the `Sign` method and internally calls back to the signing service. The private key bytes never make their way to the notary server.

With that in mind, I want to make the `PrivateKey.Private` return `([]byte, error)` so we can be explicit about when it is being called incorrectly (i.e. when it is not permissible to return the private bytes)

PR #294, opened 2015-11-15
CryptoService.Sign is now dead code. Remove it and update tests

endophage:

### DIFF
--- a/cryptoservice/crypto_service_test.go
+++ b/cryptoservice/crypto_service_test.go
@@ -127,13 +127,11 @@ func (c CryptoServiceTester) TestSignNoMatchingKeys(t *testing.T) {
 	assert.Error(t, err, c.errorMsg("Should not have found private key"))
 }
 
-// If there are multiple keystores, even if all of them have the same key,
-// only one signature is returned.
-func (c CryptoServiceTester) TestSignWhenMultipleKeystores(t *testing.T) {
+// Test GetPrivateKey succeeds when multiple keystores have the same key
+func (c CryptoServiceTester) TestGetPrivateKeyMultipleKeystores(t *testing.T) {
 	cryptoService := c.cryptoServiceFactory()
 	cryptoService.keyStores = append(cryptoService.keyStores,
 		trustmanager.NewKeyMemoryStore(passphraseRetriever))
-	content := []byte("this is a secret")
 
 	privKey, err := trustmanager.GenerateECDSAKey(rand.Reader)
 	assert.NoError(t, err, c.errorMsg("error creating key"))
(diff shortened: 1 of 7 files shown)


endophage:
@cyli if you want to take a look at this now that would be great.

cyli:
@endophage Will do - I think there's a merge conflict though.

tuf/signed/interface.go
```diff
@@ -47,7 +39,6 @@ type KeyService interface {
 // CryptoService defines a unified Signing and Key Service as this
```
cyli: This comment is no longer true.  Also, not necessary in this PR, but should `CryptoService` be eliminated in favor of just `KeyService`, or do we anticipate other uses for `CryptoService`?

cryptoservice/crypto_service_test.go
```diff
@@ -138,13 +139,16 @@ func (c CryptoServiceTester) TestSignWhenMultipleKeystores(t *testing.T) {
 	assert.NoError(t, err, c.errorMsg("error creating key"))
 
 	for _, store := range cryptoService.keyStores {
-		err := store.AddKey(privKey.ID(), "root", privKey)
```
cyli: Non-blocking - this test seems to now actually be testing that `GetPrivateKey` returns the first key it finds that matches.  The signing part no longer seems relevant, since the previous test tests that one can sign with a private key.  Would it make sense to change this to `TestGetPrivateKeyWithMultipleKeystores` instead and skip the signing part?

cyli:
@endophage Other than the above nitpicks and a merge conflict, this LGTM

tuf/signed/interface.go
endophage: Yeah, good call. Would definitely make sense to just replace CryptoService with KeyService. That touches a lot of things so I'll do that in a separate PR.

cryptoservice/crypto_service_test.go
endophage: Done!

cryptoservice/crypto_service_test.go
```diff
@@ -115,20 +118,18 @@ func (c CryptoServiceTester) TestSignWithKey(t *testing.T) {
...
-	signatures, err := cryptoService.Sign([]string{privKey.ID()}, content)
-	assert.NoError(t, err, c.errorMsg("signing failed"))
-	assert.Len(t, signatures, 0, c.errorMsg("wrong number of signatures"))
+	// Test Sign
+	_, _, err = cryptoService.GetPrivateKey(privKey.ID())
+	assert.Error(t, err, c.errorMsg("Should not have found private key"))
 }
 
 // If there are multiple keystores, even if all of them have the same key,
 // only one signature is returned.
```
cyli: Oops sorry, forgot about this comment.  Could we please change this to say "GetPrivateKey succeeds even if there are multiple keystores with the same key" or somesuch?

PR #295, opened 2015-11-15
Set the default output for all cobra commands to be STDOUT

cyli:


diogomonica:
Shouldn't we update our tests to make sure we're taking output from STDOUT? How did the tests worked in the first place?

cyli:
@diogomonica In order to read the input, we set the output of the command to be a buffer.  So the tests don't actually test whether the real command is stdout or stderr

diogomonica:
Got it. LGTM

endophage:
LGTM

PR #296, opened 2015-11-15
headers were still printing when no targets were found

endophage:


diogomonica:
LGTM

PR #299, opened 2015-11-17
Fixed bug parsing trust service info in notary server.

### LINKED ISSUE #298: keyAlgorithm from notary-server should be ignored if local trust service
If the trust service is not remote, we still allow `keyAlgorithm` to be arbitrarily set by the config.  However, the local trust service is an ED25519 crypto service.  Any algorithm other than "ed25519" will cause notary-server to 500.

We should probably only parse the keyAlgorithm if it's a remote service.

cyli:
Previously, if it wasn't a remote service, the config parser was
still setting the key algorithm to be whatever was configured.
Now, if we are using a local trust service, the algorithm is always
ED25519.

Also broke the trust parsing into its own function for testing.

Fixes #298.

### DIFF
--- a/cmd/notary-server/main.go
+++ b/cmd/notary-server/main.go
@@ -119,12 +119,13 @@ func getStore(configuration *viper.Viper, allowedBackends []string) (
 	return store, nil
 }
 
+type signerFactory func(hostname, port string, tlsConfig *tls.Config) *client.NotarySigner
+type healthRegister func(name string, checkFunc func() error, duration time.Duration)
+
 // parses the configuration and determines which trust service and key algorithm
 // to return
-func getTrustService(configuration *viper.Viper,
-	signerFactory func(string, string, *tls.Config) *client.NotarySigner,
-	healthRegister func(string, func() error, time.Duration)) (
-	signed.CryptoService, string, error) {
+func getTrustService(configuration *viper.Viper, sFactory signerFactory,
+	hRegister healthRegister) (signed.CryptoService, string, error) {
 
 	switch configuration.GetString("trust_service.type") {
 	case "local":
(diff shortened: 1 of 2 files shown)


mtrmac:
:+1: in general, clearly an improvement.  The pedantic me would be happier with rejecting `"keyAlgorithm"` values different from `data.ED25519Key` in the `"local"` case, instead of silently ignoring different values, but this is in a testing-only scenario, so good enough I suppose.

cyli:
@mrmac - I figure for testing/debugging folks might want to have to override as little as possible from the production config, which is why I wanted to ignore all the other parameters, so they'd have to change as little as possible back.  But I can see that it is a little strange to not get the `keyAlgorithm` you asked for.  Perhaps if I explicitly log a message stating that because a `"local?` type service was selected, that all the other parameters are being ignored?

mtrmac:
@cyli, overriding as little as possible does make some sense. Really, this is clearly an improvement whatever happens with `"keyAlgorithm"`, so don’t let my bikeshedding over the unchanged aspects stop you ☺

cmd/notary-server/main.go
```diff
@@ -94,6 +95,59 @@ func grpcTLS(configuration *viper.Viper) (*tls.Config, error) {
 	return tlsConfig, nil
 }
 
+// parses the configuration and determines which trust service and key algorithm
+// to return
+func getTrustService(configuration *viper.Viper,
+	signerFactory func(string, string, *tls.Config) *client.NotarySigner,
```
endophage: Declare a `signerFactory` type and provide names for the arguments. Were I writing my own signer, it's not clear from looking at this function signature what the first 2 arguments are.

```
type signerFactory func(host string, port string, tlsConf *tls.Config) *client.NotarySigner
```

endophage:
Pending tests LGTM

PR #300, opened 2015-11-18
Server check that the root.json's timestamp key ID is valid.

### LINKED ISSUE #297: Can push invalid timestamp key in root.json
Server validation needs to check that the timestamp key inside the root.json matches the timestamp key we have in the database. Currently I can push (and have accepted) a root.json for a repo that doesn't have a timestamp key. This causes pull to fail because a timestamp key is generated and used to sign a new timestamp.json that doesn't validate against the root.json

Careful: It should not create a timestamp key if one doesn't exist!

cyli:
If the client sends a root.json with an invalid timestamp key ID,
possibly because they are pushing an existing repo to a new server,
then the server should reject the update.

Fixes #297

### DIFF
--- a/server/handlers/validation.go
+++ b/server/handlers/validation.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/docker/notary/tuf"
 	"github.com/docker/notary/tuf/data"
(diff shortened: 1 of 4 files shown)


mtrmac:
:+1: other than the three comments above.

mtrmac:
:+1:

endophage:
LGTM

Saw the comments about thresholds on root, etc... There's probably no strong use cases for thresholds on timestamp. However there is a very strong use case for thresholds on root as it could be used to prevent one bad actor taking over a repository by distributing multiple root keys among a group of trusted admins and requiring a quorum of signers to make top level key changes.

Once we work out a good way to limit the complexity (for users) that bleeds out of having thresholds, we should absolutely support them at the root level.

PR #304, opened 2015-11-20
Make the server/signer configurations more similar

cyli:
This factors out some config file parsing utilities common to server and signer.  This makes the config files look a lot more similar, but it also adds a few changes:

1.  Both server/signer server TLS have the following parameters now:
    - `tls_cert_file` (signer was previously `cert_file`)
    - `tls_key_file` (signer was previously `key_file`)

2. Both the server/signer have the following http parameters now:
   - `http_addr` (server was previously `addr`)

3. Both the server/signer have the following logging parameters now:
   - `debug` as a string value (signer was previously an int)

4. Both server/signer now support Bugsnag reporting (previously the signer did not)

5. The default alias for the signer DB store are also available as config file parameters under "storage" now.  The password remains the same, because I'm not sure how best to deal with that.

6. The tls files are now parsed as paths relative to the config file, if they aren't absolute paths or empty.

7. Both the server and signer have mandatory storage sections now, and both also now support a memory backend (previously, server's storage section was optional, and signer's storage section only supported mysql)

Apologies for the massive change.  Tested this with `docker-compose up` and initting/adding/publishing/listing a repo (but not the bugsnag part or alias change).

Fixes #304.

### DIFF
--- a/Dockerfile.server
+++ b/Dockerfile.server
@@ -20,4 +20,4 @@ RUN go install \
     ${NOTARYPKG}/cmd/notary-server
 
 ENTRYPOINT [ "notary-server" ]
-CMD [ "-config", "cmd/notary-server/config.json" ]
+CMD [ "-config", "fixtures/server-config.json" ]
--- a/Dockerfile.server
+++ b/Dockerfile.server
@@ -20,4 +20,4 @@ RUN go install \
     ${NOTARYPKG}/cmd/notary-server
 
 ENTRYPOINT [ "notary-server" ]
-CMD [ "-config", "fixtures/server-config.json" ]
+CMD [ "-config=fixtures/server-config-local.json" ]
--- a/cmd/notary-signer/main.go
+++ b/cmd/notary-signer/main.go
@@ -73,24 +73,17 @@ func passphraseRetriever(keyName, alias string, createNew bool, attempts int) (p
 	return passphrase, false, nil
 }
 
-// parses and sets up the TLS for the signer http + grpc server
-func signerTLS(configuration *viper.Viper, printUsage bool) (*tls.Config, error) {
-	certFile := configuration.GetString("server.cert_file")
-	keyFile := configuration.GetString("server.key_file")
-	if certFile == "" || keyFile == "" {
+// validates TLS configuration options and sets up the TLS for the signer
+// http + grpc server
+func signerTLS(tlsConfig *utils.ServerTLSOpts, printUsage bool) (*tls.Config, error) {
+	if tlsConfig.ServerCertFile == "" || tlsConfig.ServerKeyFile == "" {
 		if printUsage {
 			usage()
 		}
 		return nil, fmt.Errorf("Certificate and key are mandatory")
 	}
 
-	clientCAFile := configuration.GetString("server.client_ca_file")
-	tlsConfig, err := utils.ConfigureServerTLS(&utils.ServerTLSOpts{
-		ServerCertFile:    certFile,
-		ServerKeyFile:     keyFile,
-		RequireClientAuth: clientCAFile != "",
-		ClientCAFile:      clientCAFile,
-	})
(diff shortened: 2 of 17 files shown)


fixtures/signer-config.json
```diff
@@ -0,0 +1,16 @@
...
+		"tls_cert_file": "./notary-signer.crt",
+		"tls_key_file": "./notary-signer.key",
+		"client_ca_file": "./notary-server.crt"
+	},
+	"logging": {
+		"level": "debug"
+	},
+	"storage": {
+		"backend": "mysql",
+		"db_url": "dockercondemo:dockercondemo@tcp(notarymysql:3306)/dockercondemo"
```
endophage: Lets finally update the username/password :-P

It just needs to match whatever we have set here: https://github.com/docker/notary/blob/master/notarymysql/start

For consistency with a default mysql development installation, let's set `root@tcp(notarymysql:3306)/notary` (by default mysql only listens on localhost which is why it doesn't complain too much about the root user with no password during installation).

cyli: Actually if we want to make this stand-aloneable for the official images, should we provide a separate, standalone config that uses the in-memory keystore instead?

cyli: Probably in a different PR though.

endophage: Would be good to talk to the registry guys to see how people use their images. Want to make sure nobody is going to spin it up in memory and expect that to be a healthy way to deploy production. I didn't comment because you've maintained the existing behaviour, but we may want to consider making the in memory deployment be explicit in the configuration rather than being the fallback/default when nothing is set.

cyli: Yeah - I wouldn't want people to download the image and run it as the default, but on the other hand maybe it's not friendly if people download the image but it doesn't work unless they override the right configuration parameters?

cyli: Possibly maybe enable sqlite3, so that if they don't run the container with `--rm` they can copy the sqlite file out, dump the DB, and migrate to a real DB?

cyli: In an attempt to not increase the size of this PR, did the mysql thing here:  https://github.com/docker/notary/pull/314

Dockerfile.server
```diff
@@ -20,4 +20,4 @@ RUN go install \
     ${NOTARYPKG}/cmd/notary-server
 
 ENTRYPOINT [ "notary-server" ]
-CMD [ "-config", "cmd/notary-server/config.json" ]
+CMD [ "-config", "fixtures/server-config.json" ]
```
endophage: Are we able to make this the same as the Dockerfile.signer? `"-config=fixtures/server-config.json"` (I actually prefer this syntax to the list with 2 strings, it makes it obvious that `-config` isn't just a flag, it accepts a parameter)

endophage:
LGTM. Definitely want somebody else to look at this one.

cmd/notary-signer/main.go
```diff
@@ -73,28 +67,113 @@ func passphraseRetriever(keyName, alias string, createNew bool, attempts int) (p
...
+		logrus.Debug("Default Alias: ", defaultAlias)
+
+		dbSQL, err := sql.Open(storeConfig.Backend, storeConfig.Source)
+		if err != nil {
+			return nil, fmt.Errorf("failed to open the %s database: %s, %v",
+				storeConfig.Backend, storeConfig.Source, err)
+		}
+		logrus.Debugf("Using %s DB: %s", storeConfig.Backend, storeConfig.Source)
+
+		keyStore, err := keydbstore.NewKeyDBStore(
```
endophage: `:=` is going to create a keyStore variable scoped within this `else`. I think you just want `=`

cyli: Wow, great catch!  Sorry the tests didn't catch it - fixed and added more tests.

endophage:
Getting the key seems like a sufficient test to catch that bug.

LGTM.

PR #305, opened 2015-11-20
Convert to use inverted build tags for pkcs11

runcom:
ping @jfrazelle 
Fix https://github.com/docker/docker/pull/17993

Signed-off-by: Antonio Murdaca <runcom@redhat.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ GO_EXC = go
 NOTARYDIR := /go/src/github.com/docker/notary
 
 # check to be sure pkcs11 lib is always imported with a build tag
-GO_LIST_PKCS11 := $(shell go list -e -f '{{join .Deps "\n"}}' ./... | xargs go list -e -f '{{if not .Standard}}{{.ImportPath}}{{end}}' | grep -q pkcs11)
+GO_LIST_PKCS11 := $(shell go list -e -f '{{join .Deps "\n"}}' ./... | xargs go list -e -f '{{if not .Standard}}{{.ImportPath}}{{end}}' | grep -q '!exclude_pkcs11')
 ifeq ($(GO_LIST_PKCS11),)
 $(info pkcs11 import was not found anywhere without a build tag, yay)
 else
@@ -85,7 +85,8 @@ fmt:
 
 lint:
 	@echo "+ $@"
-	@test -z "$$(golint ./... | grep -v .pb. | grep -v Godeps/_workspace/src/ | tee /dev/stderr)"
+	# TODO: remove trustmanager/yubikey/ from ignored once linted
+	@test -z "$$(golint ./... | grep -v .pb. | grep -v Godeps/_workspace/src/ | grep -v trustmanager/yubikey/ | tee /dev/stderr)"
 
 build: go_version
 	@echo "+ $@"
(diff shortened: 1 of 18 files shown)


Makefile
```diff
@@ -86,7 +85,8 @@ fmt:
 
 lint:
 	@echo "+ $@"
-	@test -z "$$(golint ./... | grep -v .pb. | grep -v Godeps/_workspace/src/ | tee /dev/stderr)"
+	# TODO: remove trustmanager/yubikey/ from ignored once linted
```
runcom: just didn't want to overload this PR (plus I'm not so confident in documenting that right now)

runcom:
This way we should add `exclude_pkcs11` in stable docker build and nothing in experimental

endophage:
Just curious, is there a docker/docker preference for excluding features rather than including? In the past I've generally preferred that the minimal command (i.e. no build tags) should build/run stable.

Makefile
jfrazelle: No no can you do in this PR I added this as a safety

jfrazelle: Oh nm well idk I have a bad feeling it's going to cause something bad eventually

runcom: alright, I hope I'll get everything right then..

runcom: btw, before this PR this pkg wasn't linted at all :/

jfrazelle: My bad I thought this was my test if there was pkcs11 in a non build tag

jfrazelle: But regarding that... It should probably be updated to include the new tag sorry

runcom: @jfrazelle  you mean here right? https://github.com/docker/notary/pull/305/files#diff-b67911656ef5d18c4ae36cb6741b7965L23

jfrazelle: But ya sorry do whatever w the lint hahaha not my project :P

jfrazelle: It's this https://github.com/runcom/notary/blob/invert-pkcs11-buildtag/Makefile#L21

runcom: yeah I meant that but link doesn't point there :/, I'll update that to be `!exclude_pkcs11`

jfrazelle: Awesome thanks and sorry for the confusion

runcom: Updated as per your comment @jfrazelle

runcom:
@endophage this is because pkcs11 is coming with experimental for now, usually it should be like you said I think

endophage:
I'm not following. I'd expect experimental to be stable + enabling build tags.

runcom:
> Just curious, is there a docker/docker preference for excluding features rather than including? In the past I've generally preferred that the minimal command (i.e. no build tags) should build/run stable.

The minimal command will continue to build/run stable as soon as pkcs11 will get in stable.

Right now the issue is that there's no way with Docker+experimental to opt-out pkcs11 w/o breaking builds (https://github.com/docker/docker/pull/17993#issuecomment-156799314)
Experimental will be stable+enabled **experimental** build tag and for now we well exclude pkcs11 from stable (@jfrazelle can you confirm I haven't totally misunderstood?)
To the first question docker/docker has `exclude_*` tags used in the graphdrivers for instance (i.e. opt-out overlay or aufs), I don't know if it's a preference tho.
Does all of this feel wrong btw?

endophage:
Based on the docker/docker PR you've linked, it seems like you're already providing quite the litany of build flags so rather than setting the `DOCKER_EXPERIMENTAL` env var, would it not simply make more sense to just directly provide the `experimental` build tag, which would also give you the flexibility to not provide the pkcs11 tag?

runcom:
@endophage you're right, setting just `experimental` works fine and solves my issue...thanks, I'll close both

PR #308, opened 2015-11-20
Pretty-print certificates from the notary CLI command `notary cert list`

cyli:

Output looks like:

```
  GUN                FINGERPRINT OF TRUSTED ROOT CERTIFICATE                EXPIRES IN  
----------------------------------------------------------------------------------------
  repo   0bb218afd3e64cf29622948cc1b5e5a7b0b306e2cbe0617f5092fc3dc89927ec   3649 days   
```

If there are no certs:

```
No trusted root certificates present.
```


endophage:
I feel like we should move all our pretty printing functions into their own file within the package... thoughts?

cyli:
I have no strong opinion, but would be happy to do so :)

endophage:
I think it would be good. We're in the region of 150 loc of pretty printing code. Seems like a sizable enough chunk of cohesive functionality that it deserves its own file.

cmd/notary/prettyprint_test.go
```diff
@@ -0,0 +1,269 @@
...
+		{role: data.CanonicalRootRole, gun: "", keyID: "a", location: "j"},
+		{role: data.CanonicalRootRole, gun: "", keyID: "z", location: "z"},
+		{role: "a", gun: "a", keyID: "a", location: "y"},
+		{role: "b", gun: "a", keyID: "a", location: "y"},
+		{role: "b", gun: "a", keyID: "b", location: "y"},
+		{role: "b", gun: "a", keyID: "b", location: "z"},
+		{role: "a", gun: "b", keyID: "a", location: "z"},
+	}
+	jumbled := make([]keyInfo, len(expected))
+	// randomish indices
```
endophage: Did you use a d8? :-P

endophage:
LGTM

cmd/notary/prettyprint_test.go
cyli: Heh, no, hence random-ish.  :)

PR #309, opened 2015-11-21
Renamed keystoremanager to certs, and KeyStoreManager to Manager.

cyli:
Since it no longer depends upon KeyStore, nor does it manipulate keys
in any way.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -11,9 +11,9 @@ import (
 	"path/filepath"
 
 	"github.com/Sirupsen/logrus"
+	"github.com/docker/notary/certs"
 	"github.com/docker/notary/client/changelist"
 	"github.com/docker/notary/cryptoservice"
-	"github.com/docker/notary/keystoremanager"
 	"github.com/docker/notary/trustmanager"
 	"github.com/docker/notary/tuf"
 	tufclient "github.com/docker/notary/tuf/client"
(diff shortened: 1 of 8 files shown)


endophage:
Just to be even more terse, what do you think of calling the package `certs` and the struct simple `Manager`? I'm happy to make that change if you agree.

cyli:
@endophage that'd work :)

client/client.go
```diff
@@ -63,15 +63,15 @@ var ErrRepositoryNotExist = errors.New("repository does not exist")
...
-	KeyStoreManager *keystoremanager.KeyStoreManager
+	baseDir       string
+	gun           string
+	baseURL       string
+	tufRepoPath   string
+	fileStore     store.MetadataStore
+	CryptoService signed.CryptoService
+	tufRepo       *tuf.Repo
+	roundTrip     http.RoundTripper
+	Manager       *certs.Manager
```
endophage: Probably good to keep the field name as `CertManager` (sorry, I know you had that before my suggestion). Just reading further down, `Manager` is too generic for a field name, without the preceding `certs` package name it's a little obscure.

cyli: No worries!  That makes sense - thanks!  Should have thought of that.

endophage:
LGTM

PR #311, opened 2015-11-21
Fix error handling on invalid root passphrase

mtrmac:
When the user insists on an invalid passphrase (or aborts the operation), `CryptoService.GetPrivateKey` will try the correct root location first, correctly failing, and then try to look for the root key in the $gun subdirectory, and so will return the last error, a confusing
> open $path: no such file or directory

So, recognize the passphrase-related errors and fail with them directly.

Also remove misleading passphrase-related error handling in `NotaryRepository.Initialize`:
1. It is on a path where those errors can never happen
2. The specific error handling would silently ignore the error, which can’t be right anyway.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -209,11 +209,7 @@ func (r *NotaryRepository) Initialize(rootKeyID string) error {
 	err = r.tufRepo.InitRoot(false)
 	if err != nil {
 		logrus.Debug("Error on InitRoot: ", err.Error())
-		switch err.(type) {
-		case signed.ErrInsufficientSignatures, trustmanager.ErrPasswordInvalid:
-		default:
-			return err
-		}
+		return err
 	}
 	err = r.tufRepo.InitTargets()
 	if err != nil {
--- a/cryptoservice/crypto_service.go
+++ b/cryptoservice/crypto_service.go
@@ -82,10 +82,15 @@ func (cs *CryptoService) GetPrivateKey(keyID string) (k data.PrivateKey, role st
 	for _, ks := range cs.keyStores {
 		for _, keyPath := range keyPaths {
 			k, role, err = ks.GetKey(keyPath)
-			if err != nil {
+			if err == nil {
+				return
+			}
+			switch err.(type) {
+			case trustmanager.ErrPasswordInvalid, trustmanager.ErrAttemptsExceeded:
+				return
+			default:
 				continue
 			}
-			return
 		}
 	}
 	return // returns whatever the final values were
--- a/cryptoservice/crypto_service_test.go
+++ b/cryptoservice/crypto_service_test.go
@@ -23,9 +23,13 @@ var algoToSigType = map[string]data.SigAlgorithm{
 var passphraseRetriever = func(string, string, bool, int) (string, bool, error) { return "", false, nil }
 
 type CryptoServiceTester struct {
-	cryptoServiceFactory func() *CryptoService
-	role                 string
-	keyAlgo              string
+	role    string
+	keyAlgo string
+	gun     string
+}
+
+func (c CryptoServiceTester) cryptoServiceFactory() *CryptoService {
+	return NewCryptoService(c.gun, trustmanager.NewKeyMemoryStore(passphraseRetriever))
 }
 
 // asserts that created key exists
@@ -251,10 +255,6 @@ func (c CryptoServiceTester) errorMsg(message string, args ...interface{}) strin
 }
 
 func testCryptoService(t *testing.T, gun string) {
-	getTestingCryptoService := func() *CryptoService {
-		return NewCryptoService(
-			gun, trustmanager.NewKeyMemoryStore(passphraseRetriever))
-	}
 	roles := []string{
 		data.CanonicalRootRole,
 		data.CanonicalTargetsRole,
@@ -265,9 +265,9 @@ func testCryptoService(t *testing.T, gun string) {
 	for _, role := range roles {
 		for algo := range algoToSigType {
 			cst := CryptoServiceTester{
-				cryptoServiceFactory: getTestingCryptoService,
-				role:                 role,
-				keyAlgo:              algo,
+				role:    role,
+				keyAlgo: algo,
+				gun:     gun,
 			}
 			cst.TestCreateAndGetKey(t)
 			cst.TestCreateAndGetWhenMultipleKeystores(t)
--- a/cryptoservice/crypto_service_test.go
+++ b/cryptoservice/crypto_service_test.go
@@ -3,12 +3,15 @@ package cryptoservice
 import (
 	"crypto/rand"
 	"fmt"
+	"io/ioutil"
+	"os"
 	"path/filepath"
 	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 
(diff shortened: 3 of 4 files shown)


client/client.go
```diff
@@ -172,11 +172,7 @@ func (r *NotaryRepository) Initialize(rootKeyID string) error {
 	err = r.tufRepo.InitRoot(false)
 	if err != nil {
 		logrus.Debug("Error on InitRoot: ", err.Error())
-		switch err.(type) {
-		case signed.ErrInsufficientSignatures, trustmanager.ErrPasswordInvalid:
-		default:
-			return err
-		}
+		return err
```
endophage: This changes behaviour, I'm trying to remember why we permitted `ErrInsufficientSignatures` to continue (and later added `ErrPasswordInvalid`). Prior to the switch it was a very explicit type check. I'll have to dig back into the history and see if I can work out what the intent was.

mtrmac: I am not sure what the original intent was either; but with the current code the patch can not change behavior because the only kind of error `InitRoot` can return is `data.ErrInvalidRole`.

(It is the `CryptoService.GetPrivateKey` called at the very start of `NotaryRepository.Initialize` which fails if the user can’t provide the correct passphrase; for that kind of failure the code doesn’t get this far anyway.)

endophage: It's possible I had `InitRoot` attempting to sign the root at some point in the past. I don't remember it doing that though, and it wouldn't be OK to continue if the signing had failed.

cyli:
Should a test be added for CryptoService.GetPrivateKey to ensure the right error gets propagated?

endophage:
That would be good.

mtrmac:
Thanks for the reviews! Updated:
* Added tests for `CryptoService.GetPrivateKey` directly.
* Split the conceptually separate `NotaryRepository.Initialize` change into a separate commit.
* Added a slightly-related cleanup to `CryptoService` tests.

cryptoservice/crypto_service_test.go
```diff
@@ -147,6 +157,61 @@ func (c CryptoServiceTester) TestGetPrivateKeyMultipleKeystores(t *testing.T) {
...
+	store, err := trustmanager.NewKeyFileStore(tempBaseDir, retriever)
+	assert.NoError(t, err)
+	cryptoService := NewCryptoService(c.gun, store)
+	pubKey, err := cryptoService.Create(c.role, c.keyAlgo)
+	assert.NoError(t, err, "error generating key: %s", err)
+
+	// trustmanager.KeyFileStore and trustmanager.KeyMemoryStore both cache the unlocked
+	// private key, so to test private key unlocking we need a new instance using the
+	// same underlying storage; this also makes trustmanager.KeyMemoryStore (and
+	// c.cryptoServiceFactory()) unsuitable.
```
cyli: +1 for the explanation, thank you!

cyli:
Looks like it needs a rebase, but LGTM.  Thank you for the fix and great testing refactor!

cryptoservice/crypto_service.go
```diff
@@ -82,10 +82,15 @@ func (cs *CryptoService) GetPrivateKey(keyID string) (k data.PrivateKey, role st
 	for _, ks := range cs.keyStores {
 		for _, keyPath := range keyPaths {
 			k, role, err = ks.GetKey(keyPath)
-			if err != nil {
+			if err == nil {
+				return
+			}
+			switch err.(type) {
+			case trustmanager.ErrPasswordInvalid, trustmanager.ErrAttemptsExceeded:
```
endophage: I'm not sure how I feel about this. It's now possible for a key to be in multiple key stores (yubikey, and filestore) with different "passwords" (pin for the yubikey and password for key on disk). Failing to remember the yubikey pin doesn't preclude correctly remembering the password for the file on disk (although I think the passwords are requested at different times, yubikey pin is requested at signing time, file password is requested at `GetPrivateKey` time I think). 

I agree that the error messaging in the current case is pretty poor and mostly this should be solveable by pulling the yubikey, but I don't know if the general case of the private key being stored differently in distinct key stores is going to come up again.

@cyli thoughts?

cyli: Given that the keyStores are stored and iterated over in order of usage preference, I thought might failing on the first key would make sense?  I was mentally mapping this to something like multiple 2FA devices, where it asks you for your preferred 2nd factor token, and only if you explicitly say "I don't have that factor, please use a different one" does it ask you for another.

We also explicitly recommend the user remove their on-disk key once the key has already been imported into the yubikey, so this seems reasonable unless we explicitly decide to support/encourage the general case the same key in multiple keystores?

cyli: Although I agree this is probably not intuitive to the user - we don't have any messaging about what the expected behavior is.  This is probably not the correct place to do this, but maybe we could detect whether a key exists in multiple places, and:

1. If we want the first keystore to win, say when asking for the password which keystore we're using and suggest that the user needs to do something differently if they want to use a key from a different keystore

2. List all the places the key exists, and ask the user which key they want to use so they know which credential to supply?

endophage: @mtrmac for the record, I'm not opposed to this logic as you have it, just want to make sure we've considered the experience and are aware of how it will act.

mtrmac: > (although I think the passwords are requested at different times, yubikey pin is requested at signing time, file password is requested at GetPrivateKey time I think)

AFAICS that is indeed the case.

mtrmac: To the more general point:

The UI for listing all available key stores, letting the user chose between them, telling the user what store will be used in what order, etc., with its formatting, sorting, possibly interactivity, really does not belong in the middle-of-the-call-stack `CryptoService.GetPrivateKey`; it can not even be done in there at all because `CryptoService` does not have any UI channel available (not even `passphrase.Retriever`, which would not be sufficient anyway¹).

If selection between key stores redundantly storing the same key, or even telling the user what the options and impact are, should have any UI, that UI would need to exist somewhere near the top level, and this would impact the whole call stack by having to use (`KeyService`, key ID) as an identifier instead of a key ID stand-alone and having `GetPrivateKey` automatically figure out which `KeyService` the user meant.

(Even if `GetPrivateKey` could always guess right and we only wanted to add `GetWhichServiceWillBeUsedByGetPrivateKey` to allow the top level to have a nicer UI, we should still pass an explicit `KeyService` back to `GetPrivateKey` because the guessing in `GetPrivateKey` depends on changing factors like the user being able to enter the right passphrase or possible storage/ YubiKey hardware failures, so the answer could change between `GetWhichService…` and `GetPrivateKey`.)

¹ (FWIW I would advise against PAM-like (give the user a message, prompt the user for an answer) “UI abstraction” callback: it is inherently untyped/losing information and it makes it impossible to build nice/intelligent UIs by enforcing a sequence-of-those-interactions paradigm.)

endophage:
LGTM

Please rebase against master then we'll be able to merge.

mtrmac:
Rebased. codecov results are at https://codecov.io/github/mtrmac/notary?ref=e19e7fc44d6b8428ae3eee965d70b93ca4c5a8aa , I’m not sure why github says “Waiting for status to be reported”.

PR #312, opened 2015-11-23
Cert expiration

mtrmac:
This is a small subset of #267 , but mergeable and useful stand-alone:

Add an explicit `time.Time` parameter to `cryptoservice.GenerateCertificate` and `trustmanager.NewCertificate`.  This will allow writing tests of certificate expiry by generating appropriate certificates.

Then, tighten `TestNewCertificate` to use equality comparisons and fix a revealed bug in `trustmanager.NewCertificate`, which was claiming to expire certificates in 10 years but did not account for leap days in the calculation.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -145,7 +145,7 @@ func (r *NotaryRepository) Initialize(rootKeyID string) error {
 
 	// Hard-coded policy: the generated certificate expires in 10 years.
 	startTime := time.Now()
-	rootCert, err := cryptoservice.GenerateCertificate(privKey, r.gun, startTime, startTime.Add(time.Hour*24*365*10))
+	rootCert, err := cryptoservice.GenerateCertificate(privKey, r.gun, startTime, startTime.AddDate(10, 0, 0))
 
 	if err != nil {
 		return err
--- a/client/client.go
+++ b/client/client.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/docker/notary/certs"
@@ -142,7 +143,9 @@ func (r *NotaryRepository) Initialize(rootKeyID string) error {
 		return err
 	}
 
-	rootCert, err := cryptoservice.GenerateCertificate(privKey, r.gun)
+	// Hard-coded policy: the generated certificate expires in 10 years.
+	startTime := time.Now()
+	rootCert, err := cryptoservice.GenerateCertificate(privKey, r.gun, startTime, startTime.Add(time.Hour*24*365*10))
 
 	if err != nil {
 		return err
--- a/certs/certmanager_test.go
+++ b/certs/certmanager_test.go
@@ -242,7 +242,7 @@ func filestoreWithTwoCerts(t *testing.T, gun, keyAlg string) (
 		key, _, err := fileKeyStore.GetKey(pubKey.ID())
 		assert.NoError(t, err)
 
-		cert, err := cryptoservice.GenerateCertificate(key, gun)
+		cert, err := cryptoservice.GenerateTestingCertificate(key.CryptoSigner(), gun)
 		assert.NoError(t, err)
 
 		certificates[i] = cert
--- a/cmd/notary/prettyprint_test.go
+++ b/cmd/notary/prettyprint_test.go
@@ -2,8 +2,6 @@ package main
 
 import (
 	"bytes"
-	"crypto/ecdsa"
-	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/x509"
 	"encoding/hex"
@@ -16,6 +14,7 @@ import (
 	"time"
 
 	"github.com/docker/notary/client"
+	"github.com/docker/notary/cryptoservice"
 	"github.com/docker/notary/passphrase"
 	"github.com/docker/notary/trustmanager"
 	"github.com/docker/notary/tuf/data"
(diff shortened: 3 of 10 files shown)


endophage:
Does it make more sense to have `duration` (of type `time.Duration`) rather than `startTime`? Is there any use case for being able to set `startTime` to anything other than `time.Now()`?

mtrmac:
Setting `startTime` (or perhaps rather, indirectly, the computed `notAfter` value) will be useful for writing tests of certificate expiration. Fixing `startTime` == `notBefore` at `time.Now()` and having duration configurable would not allow us to generate an already expired certificate.

Supplying a fixed point in time instead of relying on `time.Now()` also allows writing exact exact equality tests, as the second commit is already doing.

It might make sense to have the callers supply both `startTime` and `endTime`, i.e. to move the hard-coded 10-year expiration out of `trustmanager.NewCertificate` to an upper-level layer, but I don’t think I will need that for writing tests and at the moment I don’t see any other immediate need; this can of course be done later if a need arises.

mtrmac:
(I was also considering using package-private functions which use explicit parameters, and having the public functions implicitly use `time.Now()`; but this was already impossible with the way certificate generation is split between `cryptoservice` and `trustmanager` now, and I expect this will be necessary also for calls from `client`: If I add a `NotaryRepository.initializeAsOfStartTime` used only in tests and as an underlying implementatino of the current `NotaryRepository.Initialize`, `cryptoservice.GenerateCertificate` with an explicit time parameter will need to be a package-public function; so that is what the patch is doing.)

cyli:
This would definitely be useful for tests, but as for a usecase where `endTime` would be useful to set, https://github.com/docker/notary/pull/308 has tests which test certificate expiry.  Granted it's just a UI thing, and not particularly functional, but having this take `endTime` would mean I could delete the generate certificate function in those tests :)

endophage:
@mtrmac I'm a fan of the package-private functions with explicit params idea. `trustmanager` is predominantly about key storage (and should probably get renamed at some point), yet x509utils.go became a dumping ground for a lot of stuff. It would be great (in another PR) to reorganize that such that `cryptoservice` doesn't depend on `trustmanager` for certificate generation.

mtrmac:
OK, I think this is a clean approach:

* `trustmanager.NewCertificate` now has a mandatory `startTime` and `endTime` parameters. It needs to remain a public function because it is accessed from `cryptoservice.GenerateCertificate` and `trustmanager/yubikey.addECDSAKey`. (So, moving it from `trustmanager` to `cryptoservice` would not help.) Most other users don't care because this migrates them away from using this function directly.
* `cryptoservice.GenerateCertificate` now has mandatory `startTime` and `endTime` parameters: such policy belongs into upper levels, and has been moved accordingly. This remains a public function.
* A new `cryptoservice.GenerateTestingCertificate` is available for tests who don’t care about the times, which keeps tests short. Public due to how certificate generation is spread among the various packages’ tests. Also, tests now use this instead of calling `trustmanager.NewCertificate` and open-coding a near-equivalent of `cryptoservice.Generate`*`Certificate`.
* None of the above should be changing behavior; the only behavior change is the fix of certificate computation expiration, from 3650 days to 10 years including leap days.

(And yes, this does simplify `cmd/notary/prettyprint_test.go`.)

endophage:
I haven't actually tried it but I think you can put test utility functions in an `_test.go` file and other tests in other packages will still be able to access them via the normal imports. I saw something about that a few weeks back and have been meaning to try it so I can get rid of the testutils package.

endophage:
@cyli tells me this doesn't work. I'll have to dig up wherever it was I read that and work out what they meant because I clearly misunderstood.

client/client.go
```diff
@@ -105,7 +106,9 @@ func (r *NotaryRepository) Initialize(rootKeyID string) error {
 		return err
 	}
 
-	rootCert, err := cryptoservice.GenerateCertificate(privKey, r.gun)
+	// Hard-coded policy: the generated certificate expires in 10 years.
+	startTime := time.Now()
+	rootCert, err := cryptoservice.GenerateCertificate(privKey, r.gun, startTime, startTime.AddDate(10, 0, 0))
```
cyli: I wonder if it might be easier to find this policy if we used `Time.Add(Duration)`, and specified the duration as a `const` near the top of the file?

cyli: Or even as a parameter to NotaryRepository, so that people who use this as a library can specify different policies?

mtrmac: The ambition of this PR is to make functions available for testing; making the hard-coded policy more visible is mostly a side effect. Making the policy configurable should IMHO go into a separate PR, if it needs to be configurable at all (I don’t currently see a really strong reason for it to be configurable, OTOH I haven’t been thinking about it much.)

(Also, `Time.Add(hours)` does not account for leap years the way `Time.AddDate` does, see 856f9021c8f08e9e6c7c7d43b9cec64d531a3605 . That’s certainly not a showstopper but it does lead to non-intuitive and surprising results.)

cyli: Ah ok.  That makes sense, thanks!  :)

cyli:
LGTM!

endophage:
@mtrmac could you please rebase against master (to get the new circle.yml with codecov configuration)? That will trigger a new CircleCI build which should cause the codecov check to complete.

Pending that, LGTM

mtrmac:
Rebased. codecov results are at https://codecov.io/github/mtrmac/notary?ref=06e58c1d11ba646f8d3181fa1ec5e9524a609289 , not sure why github says “ Waiting for status to be reported”.

endophage:
I noticed it's building under your own circleci account, there's probably something wrong with how the codecov results are being looked up. @cyli could you take a look?

cyli:
@endophage I think probably codecov on the docker account marks all PRs in docker/notary as waiting, because it expects some coverage to be sent in.  And it was, but not to the docker account, as far as it knows, no coverage was ever reported.

cyli:
@mtrmac @endophage According to https://circleci.com/docs/fork-pr-builds:  "If otherdev/yourproject is explicitly configured to build on CircleCI in its own right, then we will not run builds in the yourorg/yourproject context for PRs from the fork...."

(since it's building on @mtrmac's CircleCi account, it's getting pushed to @mtrmac's codecov account as well, and not to docker's.)

I'm not sure how to make CircleCI switch over without disabling CircleCI for your fork.

PR #314, opened 2015-11-23
Just connect to mysql as root, rather than as dockerdemo user.

cyli:
This way no password is needed.

As per https://github.com/docker/notary/pull/304#discussion_r45666968.


HuKeping:
ye, it's more than just for the dockercon, glad to see this change ;)

endophage:
LGTM

PR #318, opened 2015-11-30
Re-enable pushing coverage reports to codecov.io

cyli:

~~This needs someone with admin access to docker org needs to grant access to codecov and enable this repo in codecov.io.~~ Done

Also, [codecov.io](https://codecov.io), like [coveralls](https://coveralls.io), does line coverage rather than statement coverage, but this seems better than no coverage check.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ _space := $(empty) $(empty)
 
 # go cover test variables
 COVERDIR=.cover
-COVERPROFILE=$(COVERDIR)/cover.out
+COVERPROFILE?=$(COVERDIR)/cover.out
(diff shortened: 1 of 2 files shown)


endophage:
LGTM. Disappointed with how low our coverage is at the moment but it seems like a lot of it is error cases we can't always trigger (like how do I make `json.Marshal` return an error?)

cyli:
@endophage Statement coverage is higher (more around 64%), as opposed to line coverage.  You can see statement coverage by looking at the circleci details, the `gvm use stable && make covmerge` section.  But yeah, error coverage is harder.  :|

PR #319, opened 2015-12-02
initial database migration state

endophage:
Adds a super simple migrations solution to keep a handle on database changes. The initial files for the server and signer are current for the master branch.

cc @vikstrous @cyli @NathanMcCauley 

### DIFF
--- a/migrations/README.md
+++ b/migrations/README.md
@@ -1,0 +1,8 @@
+# Database Migrations
+
+This directory contains database migrations for the server and signer. They
+are being managed using [this tool](https://github.com/mattes/migrate).
+Within each of the server and signer directories are directories for different
+database backends. Notary server and signer use GORM and are therefore 
+capable of running on a number of different databases, however migrations
+may contain syntax specific to one backend.
--- a/migrations/server/mysql/0001_initial.down.sql
+++ b/migrations/server/mysql/0001_initial.down.sql
@@ -1,0 +1,3 @@
+DROP TABLE `timestamp_keys`;
+
+DROP TABLE `tuf_files`;
--- a/migrations/server/mysql/0001_initial.up.sql
+++ b/migrations/server/mysql/0001_initial.up.sql
@@ -1,0 +1,24 @@
+CREATE TABLE `timestamp_keys` (
+	  `id` int(11) NOT NULL AUTO_INCREMENT,
+	  `created_at` timestamp NULL DEFAULT NULL,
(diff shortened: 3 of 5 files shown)


cyli:
Are we counting right now as the initial migration, as opposed to the sql before we gormized everything?

thaJeztah:
Hm, a note of warning, because I wonder if that migration plugin [actually does what it says](https://github.com/mattes/migrate/tree/master/driver/mysql);

> Runs migrations in transcations. That means that if a migration failes, it will be safely rolled back.

Because MySQL has *no support for transactions* for DDL (schema changes); https://dev.mysql.com/doc/refman/5.7/en/cannot-roll-back.html (one of the reasons I hate MySQL :innocent: )

endophage:
@cyli yes. Current as deployed with latest code yesterday is the initial migration from an empty DB.

@thaJeztah yeah, @cyli mentioned that to me. However, the benefit of this migration tool is that its data format is just a bunch of sql files that can be run manually if necessary. For MySQL the transaction problem is one we have to deal with regardless of our migration tool and this particular tool just felt extremely minimal and easy to move away from (or fork and maintain ourselves) if necessary.

thaJeztah:
@endophage yup, it looked quite simple so from that PoV, all fine with me. Just wanted to warn for possible pitfalls, I've been bit by them on many occasions. Perhaps a more solid database someday? (e.g. PostgreSQL)

cyli:
Non-blocking thoughts, since I don't know how to do these yet:

1.  We should probably have some kind of integration tests for the DB migrations - for this set, probably just run them, and then a basic check that our data models can read/write from them?  In the future, we might want to add test fixtures such that we:

    - apply one migration
    - add some data
    - apply another migration
    ...
    - final migration
    - tests ensure that it can read/write new and old data

2. Should our notarymysql container use apply these migrations instead of the initialization sql?  It'd be nice to have all the sql in one place.  On the other hand, since we're building a new container, there's no reason to have to run all the migrations and it might be nice to have a single sql file that reflects the current DB schema.

endophage:
@cyli I've just filed a kind of related issue about notarymysql. If we modify `notarymysql/start` I think that might be possible. https://github.com/docker/notary/issues/326

cyli:
@endophage This LGTM (modulo) if you want to do the mysql stuff in a different branch.

endophage:
Yeah, the MySQL container stuff will definitely be in a different branch

PR #320, opened 2015-12-02
Tiny refactor

HuKeping:
It's no need to keep the `err` after asserting it should be `nil`, and
we can merge these two logs into one I suppose.


mtrmac:
:+1:

cyli:
LGTM, thank you for your contribution!

PR #321, opened 2015-12-02
Refactor ParseServerTLS

HuKeping:
If I understand correctly, this function is used to generate a instance
of structure `ServerTLSOpts` if TLS is required and both of the certificate
and key are provided. Otherwise return nil.

So why should we check whether the one of certificate or key is empty
even if the TLS is not required. Will it miss something if just return
nil if the TLS is not required.

### DIFF
--- a/utils/configuration.go
+++ b/utils/configuration.go
@@ -51,23 +51,16 @@ func ParseServerTLS(configuration *viper.Viper, tlsRequired bool) (*ServerTLSOpt
 		ServerKeyFile:  GetPathRelativeToConfig(configuration, "server.tls_key_file"),
 		ClientCAFile:   GetPathRelativeToConfig(configuration, "server.client_ca_file"),
 	}
-
 	cert, key := tlsOpts.ServerCertFile, tlsOpts.ServerKeyFile
-	if tlsRequired {
-		if cert == "" || key == "" {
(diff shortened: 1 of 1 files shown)


HuKeping:
Ths `CI` should be failed since it was designed so.

mtrmac:
`!tlsRequired` does not mean “TLS forbidden”, it merely means “the server can be configured not to use TLS”. This change would make TLS use in such sitautions (in particular having the server serve over TLS) impossible.

Doesn’t the optional token authentication mechanism rely on confidentiality of the tokens? If so, inability to use TLS would prevent secure use of token authentication.

endophage:
@HuKeping if you have a look at where `ParseServerTLS` is used, the `tlsRequired` flag indicates whether to be strict about requiring it. As @mtrmac noted, `!tlsRequired` doesn't mean "never use TLS" it means "TLS is optional".

Currently, we _require_ (i.e. set `tlsRequired` to `true`) that notary signer has TLS enabled for all communications, but notary server does not require TLS to be configured (it/s optional, `tlsRequired` set to `false`) for those users that want to do things like load balancer TLS termination.

Closing as current implementation is the intended behaviour.

HuKeping:
Thanks for the explanation, much more clear now :)

PR #323, opened 2015-12-02
when using a mysql database, need to set parseTime=True in connection…

endophage:
… string query params


cyli:
LGTM!

PR #325, opened 2015-12-02
Added check to skip db import if exists

mbentley:
Signed-off-by: Matt Bentley <matt@docker.com>


endophage:
Thanks! LGTM.

This allows users to use volumes with notarymysql (if they want to) with the compose file and it won't clobber the database already created in the volume each time a notarymysql container is started.

cyli:
LGTM!

PR #327, opened 2015-12-02
unify the timestamp handler with the handler for getting other tuf metadata

endophage:
Rather than further fragmenting the code as we add support for server side snapshotting, I'd rather unify the handler that serves requests to retrieve TUF metadata. Conceptually I think it will make some of the logic a little clearer.


cyli:
LGTM

PR #329, opened 2015-12-03
Generalize server keys

endophage:
Unfortunately the database is still called "timestamp_keys" and I'm not sure how to fix that cleanly. Everything else has been generalized though so that one piece of ugly legacy hangover is highly contained to the database migration files, and the `Key.TableName` method.

### DIFF
--- a/migrations/server/mysql/0002_role_on_keys.up.sql
+++ b/migrations/server/mysql/0002_role_on_keys.up.sql
@@ -1,0 +1,1 @@
+ALTER TABLE `timestamp_keys` ADD COLUMN `role` VARCHAR(255) NOT NULL, DROP KEY `gun`, ADD UNIQUE KEY `gun_role` (`gun`, `role`);
--- a/migrations/server/mysql/0002_role_on_keys.up.sql
+++ b/migrations/server/mysql/0002_role_on_keys.up.sql
@@ -1,1 +1,3 @@
 ALTER TABLE `timestamp_keys` ADD COLUMN `role` VARCHAR(255) NOT NULL, DROP KEY `gun`, ADD UNIQUE KEY `gun_role` (`gun`, `role`);
+
+UPDATE `timestamp_keys` SET `role`="timestamp";
--- a/server/storage/database_test.go
+++ b/server/storage/database_test.go
@@ -39,12 +39,12 @@ func SetUpSQLite(t *testing.T, dbDir string) (*gorm.DB, *SQLStorage) {
 	err = CreateTUFTable(dbStore.DB)
 	assert.NoError(t, err)
 
-	err = CreateTimestampTable(dbStore.DB)
+	err = CreateKeyTable(dbStore.DB)
 	assert.NoError(t, err)
 
 	// verify that the tables are empty
 	var count int
-	for _, model := range [2]interface{}{&TUFFile{}, &TimestampKey{}} {
+	for _, model := range [2]interface{}{&TUFFile{}, &Key{}} {
 		query := dbStore.DB.Model(model).Count(&count)
 		assert.NoError(t, query.Error)
 		assert.Equal(t, 0, count)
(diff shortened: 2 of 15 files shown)


cyli:
@endophage I think MySQL supports atomic renaming of tables:  https://dev.mysql.com/doc/refman/5.7/en/rename-table.html

So we can add a migration for that, if you like.

migrations/server/mysql/0002_role_on_keys.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE `timestamp_keys` ADD COLUMN `role` VARCHAR(255) NOT NULL, DROP KEY `gun`, ADD UNIQUE KEY `gun_role` (`gun`, `role`);
```
cyli: I don't know if this will work, but we probably also want to add a DEFAULT 'timestamp' to the alter table, else all our existing rows will have an empty role.  We could, in the code, assume that an empty role means "timestamp role" but it'd be awesome to do this in the migration, if it works.

cyli:
A couple more test cases involving the same gun with a different role would be awesome (maybe even inserting the same key for multiple roles?)

migrations/server/mysql/0002_role_on_keys.up.sql
endophage: Good catch! I don't think we want to set it as `DEFAULT "timestamp"` as that would be a long term change for a point in time fix. Instead I'll add another line to the migration to set the value to "timestamp" for all existing records. This might create a small race condition between running the migration and updating the containers, but we can solve that by further manually running an update to set all records with a `NULL` role to "timestamp".

endophage:
@cyli I figured I'd add those tests as part of the actual snapshot management stuff so they're coming, just in a different PR.

As far as renaming the table, it's definitely possible, it's just impossible to synchronize with a code push, and even more scary, if something goes wrong and we say, have to restore the DB from a backup, we've suddenly caused a significant outage.

cyli:
@endophage Oh I meant specifically tests for the storage driver that tests the unique key comprising the gun and role.  (e.g. that same role but different gun and vice versa can be successfully added to the DB, but the same gun and same role, even if the keys are different, cannot be added to the DB).

Also, similar DB changes should probably be added to either notarymysql `initial.sql` or `migrate.sql`, else `notarymysql` will be in a broken state after this PR - I know that we were going to fix our mysql stuff anyway, I just wasn't sure whether it's ok to leave it broken in the meantime?

endophage:
@cyli last 2 commits update migrate.sql and add tests respectively.

server/storage/database_test.go
```diff
@@ -257,53 +258,126 @@ func TestSQLDelete(t *testing.T) {
...
+
+func TestSQLSetKeySameRoleGun(t *testing.T) {
+	tempBaseDir, err := ioutil.TempDir("", "notary-test-")
+	_, dbStore := SetUpSQLite(t, tempBaseDir)
+	defer os.RemoveAll(tempBaseDir)
+
+	err = dbStore.SetKey("testGUN", data.CanonicalTimestampRole, "testCipher", []byte("1"))
+	assert.NoError(t, err, "Inserting timestamp into empty DB should succeed")
+
+	err = dbStore.SetKey("testGUN", data.CanonicalTimestampRole, "testCipher", []byte("1"))
```
cyli: Thanks!  Apologies, I should have added a test for this one before.  :)  Could we change this to `[]byte("2")`?  If it's `[]byte("1")`, then the uniqueness constraint might be on the combination of "gun", "role", and "key", whereas we want to test that the uniqueness constraint is on "gun" and "role" only (you can't add another key for the same gun and role)

cyli:
Other than the above nitpick, LGTM!

server/storage/database_test.go
endophage: Ah, yep, makes sense, will change.

endophage:
@cyli good call on that test update. I think gorm may not set up multicolumn unique indexes properly... (possibly just in sqlite). It wasn't erroring when I tried to insert a duplicate gun+role but a different key/algorithm. Small code update along with test fix in the last commit.

PR #331, opened 2015-12-03
Fix bug when parsing the storage

HuKeping:
Prior to this patch, everything seems fine because there are only two
backends that passed to `ParseStorage` via `allowedBackends`.

But it is possible that in the future, someone would add a new storage
backend to Notary and unfortunately it doesn't need the `storage.db_url`.
If that happened, the prior logical will return error which is wrong.

### DIFF
--- a/utils/configuration.go
+++ b/utils/configuration.go
@@ -110,10 +110,15 @@ func ParseStorage(configuration *viper.Viper, allowedBackends []string) (*Storag
 	if store.Backend == MemoryBackend {
 		return &Storage{Backend: MemoryBackend}, nil
 	}
-	if store.Source == "" {
+	if store.Backend == MySQLBackend && store.Source == "" {
 		return nil, fmt.Errorf(
 			"must provide a non-empty database source for %s", store.Backend)
 	}
+	if store.Backend == SqliteBackend && store.Source == "" {
+		return nil, fmt.Errorf(
+			"must provide a non-empty database source for %s", store.Backend)
+	}
+
 	return &store, nil
 }
 


HuKeping:
Not sure if this is intended https://github.com/docker/notary/blob/master/cmd/notary-signer/main.go#L226

Because we have 3 backend there https://github.com/docker/notary/blob/master/utils/configuration.go#L17-L20

but only use two as `allowedBackends`

cyli:
@HuKeping Yep, that is intended - the sqlite backend is just use for our testing purposes.  I'm not sure that makes sense as a backend to run in production (although maybe, for small cases) - for other testing purposes, the memory backend should be sufficient.

We definitely want to support more DB backends, like Postgres, which all require `db_urls`, in which case if this patch is merged the parsing code would have to be changed for each, but which could just be added to the allowed backends without this change.

You are correct in that if someone in the future wants to add a backend that does not take a `db_url`, that they would have to come change this code - I cannot currently think of any more backends that do not take some kind of argument though, so this seems like a less likely case than adding more backends which do require an argument.

I don't feel terribly strongly though.  Are there more backends you can think of that do not require a `db_url`?

endophage:
`ParseStorage` is essentially input validation so adding additional backends that have differing configuration formats would be expected to necessitate changes in `ParseStorage`. As with any input validation, if the range of valid inputs changes, the validation has to be updated.

Currently, while 3 potential backends are supported, 2 of them (mysql and sqlite3) have identical configuration formats, so there is no need to distinguish between them (or any other database backends we may add in the future, i.e. postgreSQL) for the purpose of input validation. If a memory backend is detected the function returns early, if it's any other type of backend, the `db_url` stored in `store.Source` is required.

I think until we have a reason to add more verbosity to `ParseStorage` the current code is sufficient. Do you agree?

HuKeping:
Agree, if someday a new backend was added which does not take a `db_url`, we can update the code then. But until now the current code is sufficient.

Closing this :)

PR #332, opened 2015-12-03
Server snapshot

endophage:
This builds on top of #329 to add optional server side snapshotting.

### DIFF
--- a/server/handlers/validation.go
+++ b/server/handlers/validation.go
@@ -7,68 +7,25 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/Sirupsen/logrus"
+
+	"github.com/docker/notary/server/storage"
 	"github.com/docker/notary/tuf"
 	"github.com/docker/notary/tuf/data"
-
-	"github.com/Sirupsen/logrus"
-	"github.com/docker/notary/server/storage"
 	"github.com/docker/notary/tuf/keys"
 	"github.com/docker/notary/tuf/signed"
 	"github.com/docker/notary/tuf/utils"
 )
 
-// ErrValidation represents a general validation error
-type ErrValidation struct {
-	msg string
-}
-
-func (err ErrValidation) Error() string {
-	return fmt.Sprintf("An error occurred during validation: %s", err.msg)
-}
-
-// ErrBadHierarchy represents a missing snapshot at this current time.
-// When delegations are implemented it will also represent a missing
-// delegation parent
-type ErrBadHierarchy struct {
-	msg string
-}
-
-func (err ErrBadHierarchy) Error() string {
-	return fmt.Sprintf("Hierarchy of updates in incorrect: %s", err.msg)
-}
-
-// ErrBadRoot represents a failure validating the root
-type ErrBadRoot struct {
-	msg string
-}
-
-func (err ErrBadRoot) Error() string {
-	return fmt.Sprintf("The root being updated is invalid: %s", err.msg)
-}
-
-// ErrBadTargets represents a failure to validate a targets (incl delegations)
-type ErrBadTargets struct {
-	msg string
-}
-
-func (err ErrBadTargets) Error() string {
-	return fmt.Sprintf("The targets being updated is invalid: %s", err.msg)
-}
-
-// ErrBadSnapshot represents a failure to validate the snapshot
-type ErrBadSnapshot struct {
-	msg string
-}
-
-func (err ErrBadSnapshot) Error() string {
-	return fmt.Sprintf("The snapshot being updated is invalid: %s", err.msg)
-}
-
 // validateUpload checks that the updates being pushed
 // are semantically correct and the signatures are correct
-func validateUpdate(gun string, updates []storage.MetaUpdate, store storage.MetaStore) error {
+// A list of possibly modified updates are returned if all
+// validation was successful. This allows the snapshot to be
+// created and added if snapshotting has been delegated to the
+// server
+func validateUpdate(cs signed.CryptoService, gun string, updates []storage.MetaUpdate, store storage.MetaStore) ([]storage.MetaUpdate, error) {
 	kdb := keys.NewDB()
-	repo := tuf.NewRepo(kdb, nil)
+	repo := tuf.NewRepo(kdb, cs)
 	rootRole := data.RoleName(data.CanonicalRootRole)
 	targetsRole := data.RoleName(data.CanonicalTargetsRole)
 	snapshotRole := data.RoleName(data.CanonicalSnapshotRole)
@@ -78,11 +35,6 @@ func validateUpdate(gun string, updates []storage.MetaUpdate, store storage.Meta
 	for _, v := range updates {
 		roles[v.Role] = v
 	}
-	if err := hierarchyOK(roles); err != nil {
-		logrus.Error("ErrBadHierarchy: ", err.Error())
-		return ErrBadHierarchy{msg: err.Error()}
-	}
-	logrus.Debug("Successfully validated hierarchy")
 
 	var root *data.SignedRoot
 	oldRootJSON, err := store.GetCurrent(gun, rootRole)
@@ -90,33 +42,33 @@ func validateUpdate(gun string, updates []storage.MetaUpdate, store storage.Meta
 		// problem with storage. No expectation we can
 		// write if we can't read so bail.
 		logrus.Error("error reading previous root: ", err.Error())
-		return err
+		return nil, err
 	}
 	if rootUpdate, ok := roles[rootRole]; ok {
 		// if root is present, validate its integrity, possibly
 		// against a previous root
 		if root, err = validateRoot(gun, oldRootJSON, rootUpdate.Data, store); err != nil {
 			logrus.Error("ErrBadRoot: ", err.Error())
-			return ErrBadRoot{msg: err.Error()}
+			return nil, ErrBadRoot{msg: err.Error()}
 		}
 
 		// setting root will update keys db
 		if err = repo.SetRoot(root); err != nil {
 			logrus.Error("ErrValidation: ", err.Error())
-			return ErrValidation{msg: err.Error()}
+			return nil, ErrValidation{msg: err.Error()}
 		}
 		logrus.Debug("Successfully validated root")
 	} else {
 		if oldRootJSON == nil {
-			return ErrValidation{msg: "no pre-existing root and no root provided in update."}
+			return nil, ErrValidation{msg: "no pre-existing root and no root provided in update."}
 		}
 		parsedOldRoot := &data.SignedRoot{}
 		if err := json.Unmarshal(oldRootJSON, parsedOldRoot); err != nil {
-			return ErrValidation{msg: "pre-existing root is corrupted and no root provided in update."}
+			return nil, ErrValidation{msg: "pre-existing root is corrupted and no root provided in update."}
 		}
 		if err = repo.SetRoot(parsedOldRoot); err != nil {
 			logrus.Error("ErrValidation: ", err.Error())
-			return ErrValidation{msg: err.Error()}
+			return nil, ErrValidation{msg: err.Error()}
 		}
 	}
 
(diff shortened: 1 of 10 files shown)


server/handlers/validation.go
```diff
@@ -125,34 +77,150 @@ func validateUpdate(gun string, updates []storage.MetaUpdate, store storage.Meta
...
+		logrus.Debug("Successfully validated snapshot")
+	} else {
+		// Check:
+		//	 - we have a snapshot key
+		//   - it matches a snapshot key signed into the root.json
+		// Then:
+		//   - generate a new snapshot
+		//   - add it to the updates
+		//
+		// TODO: Assumptions have been made about what is loaded into the repo.
```
cyli: The root seems to have to be loaded, right?  Since if `SetRoot` above fails, this function will bail?

Also, is a `snapshots.json` that with no entry for targets metadata valid?  Previously, the hierarchy valid only checked for the presence of the snapshots file, and not the update file.  Maybe no targets metadata could just mean there are no targets yet?

server/handlers/validation.go
```diff
@@ -125,34 +77,150 @@ func validateUpdate(gun string, updates []storage.MetaUpdate, store storage.Meta
...
+		}
+
+		if err := validateSnapshot(snapshotRole, oldSnap, roles[snapshotRole], roles, kdb); err != nil {
+			logrus.Error("ErrBadSnapshot: ", err.Error())
+			return nil, ErrBadSnapshot{msg: err.Error()}
+		}
+		logrus.Debug("Successfully validated snapshot")
+	} else {
+		// Check:
+		//	 - we have a snapshot key
```
cyli: Nitpick: this bullet point is over-indented

server/handlers/validation.go
```diff
@@ -125,34 +77,150 @@ func validateUpdate(gun string, updates []storage.MetaUpdate, store storage.Meta
...
+		updates = append(updates, *update)
 	}
+	return updates, nil
+}
 
-	if err := validateSnapshot(snapshotRole, oldSnap, roles[snapshotRole], roles, kdb); err != nil {
-		logrus.Error("ErrBadSnapshot: ", err.Error())
-		return ErrBadSnapshot{msg: err.Error()}
+func prepRepo(gun string, repo *tuf.Repo, store storage.MetaStore) error {
+	if repo.Root == nil {
```
cyli: Isn't the root loaded by [the root checking code in `validateUpdate`](https://github.com/docker/notary/pull/332/files#diff-a79ef7cee1c3c13ab34d284bf98f2fa3L95)?  Is it possible to get to here without having called `SetRoot`?

server/handlers/validation.go
```diff
@@ -125,34 +77,150 @@ func validateUpdate(gun string, updates []storage.MetaUpdate, store storage.Meta
...
+		}
+		root := &data.SignedRoot{}
+		err = json.Unmarshal(rootJSON, root)
+		if err != nil {
+			return fmt.Errorf("could not load repo for snapshot generation: %v", err)
+		}
+		repo.SetRoot(root)
+	}
+	if repo.Targets[data.CanonicalTargetsRole] == nil {
+		targetsJSON, err := store.GetCurrent(gun, data.CanonicalTargetsRole)
```
cyli: Non-blocking - we seem to be getting the current role, unmarshalling it into a `data.Signed*` a lot.  I was wondering if it might be helpful to abstract these out to helper functions like `GetCurrentSignedt*`?

server/handlers/validation.go
```diff
@@ -125,34 +77,150 @@ func validateUpdate(gun string, updates []storage.MetaUpdate, store storage.Meta
...
+	}
+	var sn *data.SignedSnapshot
+	if currentJSON != nil {
+		sn = new(data.SignedSnapshot)
+		err := json.Unmarshal(currentJSON, sn)
+		if err != nil {
+			return nil, fmt.Errorf("could not retrieve previous snapshot: %v", err)
+		}
+	} else {
+		root, err := repo.Root.ToSigned()
```
cyli: Would `repo.InitSnapshot()` work here?  I think the code is basically the same.

server/handlers/validation.go
```diff
@@ -125,34 +77,150 @@ func validateUpdate(gun string, updates []storage.MetaUpdate, store storage.Meta
...
 }
 
+func generateSnapshot(gun string, kdb *keys.KeyDB, repo *tuf.Repo, store storage.MetaStore) (*storage.MetaUpdate, error) {
+	role := kdb.GetRole(data.RoleName(data.CanonicalSnapshotRole))
+	if role == nil {
+		return nil, ErrBadRoot{msg: "root did not include snapshot role"}
+	}
+
+	algo, keyBytes, err := store.GetKey(gun, data.CanonicalSnapshotRole)
+	foundK := data.NewPublicKey(algo, keyBytes)
```
cyli: Should there be a check for if there was an error getting the key?

server/handlers/validation.go
```diff
@@ -125,34 +77,150 @@ func validateUpdate(gun string, updates []storage.MetaUpdate, store storage.Meta
...
+		if _, ok := err.(*storage.ErrNotFound); !ok {
+			return nil, fmt.Errorf("could not retrieve previous snapshot: %v", err)
+		}
+	}
+	var sn *data.SignedSnapshot
+	if currentJSON != nil {
+		sn = new(data.SignedSnapshot)
+		err := json.Unmarshal(currentJSON, sn)
+		if err != nil {
+			return nil, fmt.Errorf("could not retrieve previous snapshot: %v", err)
```
cyli: Non-blocking - clearly something is wrong if the previous snapshot is corrupt, but is this recoverable by just synthesizing a new snapshot?

server/handlers/validation_test.go
```diff
@@ -224,16 +226,226 @@ func TestValidateSnapshotMissing(t *testing.T) {
...
+
+	store.UpdateCurrent("testGUN", root)
+	store.UpdateCurrent("testGUN", targets)
+
+	copyTimestampKey(t, kdb, store, "testGUN")
+	updates, err = validateUpdate(cs, "testGUN", updates, store)
+	assert.NoError(t, err)
+}
+
+func TestPreRepoLoadRootTargets(t *testing.T) {
```
cyli: Nitpick:  "TestPrepRepo*" for all of these tests :)

cyli:
I think there might be a bug in the way we generate the code coverage:  `server/snapshot/snapshot.go` is not covered, but I guess because nothing ever calls it no coverage report was produced?  I filed https://github.com/docker/notary/issues/337 for it.  

But it would be awesome to add some test coverage for that :D

server/handlers/validation.go
endophage: Technically having no targets file in the snapshot is valid, but the way we've set everything up, we always create and publish at least an empty targets file.

I think you're right that the root will always have been loaded, we get it from the update or the storage to ensure the other keys are all OK. In case anything about that changes, I also load it in `prepRepo`. I should have deleted that TODO as I `prepRepo` to ensure no assumptions needed to be made.

server/handlers/validation.go
endophage: Looks like I had an errant tab.

server/handlers/validation.go
endophage: You are 100% correct, I'm just being extra cautious and the nil check is cheap.

server/handlers/validation.go
endophage: Yes! Good catch!

server/handlers/validation.go
endophage: yep

server/handlers/validation.go
endophage: It's indeterminate whether it's recoverable. Clients will reject the new snapshot if its version is not greater than the previous one. If we can't parse the previous one, we don't know what the version of the new one should be.

We have some yet to be done work around resetting the version counter for a role when its key gets rotated which would potentially make this a recoverable state, but for now, I don't think there's much we can do short of manual resolution.

server/handlers/validation.go
endophage: Yeah possibly. Should be a different PR probably because I think it will show a lot of small scattered diffs which will make this harder to follow.

endophage:
Yeah, was planning to add coverage for snapshot.go.

server/handlers/validation.go
cyli: Ah, ok, I had forgotten about the version number, thanks!

server/handlers/validation.go
cyli: Sounds good

server/snapshot/snapshot.go
```diff
@@ -0,0 +1,161 @@
...
+
+	"github.com/docker/notary/server/storage"
+	"github.com/docker/notary/tuf/data"
+	"github.com/docker/notary/tuf/signed"
+)
+
+// GetOrCreateSnapshotKey either creates a new snapshot key, or returns
+// the existing one. Only the PublicKey is returned. The private part
+// is held by the CryptoService.
+func GetOrCreateSnapshotKey(gun string, store storage.KeyStore, crypto signed.CryptoService, createAlgorithm string) (data.PublicKey, error) {
```
cyli: I'm not sure where it'd go, but could this be unified with `GetOrCreateTimestampKey`?  The code seems the same except for the role.

endophage: Yeah, agreed. I basically copy/pasted the entire timestamp package and modified it to be snapshot. I wasn't sure initially how much difference there would be... Will probably fix once end to end PoC for server snapshotting is complete. I figure at that point there will be a bunch of refactoring we want to do.

server/snapshot/snapshot_test.go
```diff
@@ -0,0 +1,209 @@
...
+	assert.NotNil(t, k, "Key should not be nil")
+
+	k2, err := GetOrCreateSnapshotKey("gun", store, crypto, data.ED25519Key)
+
+	assert.Nil(t, err, "Expected nil error")
+
+	// trying to get the same key again should return the same value
+	assert.Equal(t, k, k2, "Did not receive same key when attempting to recreate.")
+	assert.NotNil(t, k2, "Key should not be nil")
+}
```
cyli: Do we want to ensure that the key has been added to both the store and cryptoservice here?

server/snapshot/snapshot_test.go
```diff
@@ -0,0 +1,209 @@
...
```
cyli: Do we want to ensure that the key has been added to both the store where it wasn't there previously?

server/snapshot/snapshot_test.go
```diff
@@ -0,0 +1,209 @@
...
+		},
+	}
+	assert.False(t, snapshotExpired(sn), "Snapshot should NOT have expired")
+}
+
+func TestGetSnapshotKeyCreate(t *testing.T) {
+	store := storage.NewMemStorage()
+	crypto := signed.NewEd25519()
+	k, err := GetOrCreateSnapshotKey("gun", store, crypto, data.ED25519Key)
+	assert.Nil(t, err, "Expected nil error")
```
cyli: Non-blocking:  Should we be `assert.NoError` here?

server/snapshot/snapshot_test.go
```diff
@@ -0,0 +1,209 @@
...
+		return ks.k.Algorithm(), ks.k.Public(), nil
+	}
+	return "", nil, &storage.ErrNoKey{}
+}
+
+func (ks keyStore) SetKey(gun, role, algorithm string, public []byte) error {
+	return &storage.ErrKeyExists{}
+}
+
+func TestGetSnapshotKeyExistsOnSet(t *testing.T) {
```
cyli: Nice test!  You hit all the weird error cases :D  Could we add a docstring to this test to the effect of this testing a race condition where the key didn't exist, but when adding the key we found out it did exist, or something?

server/snapshot/snapshot_test.go
```diff
@@ -0,0 +1,209 @@
...
+	_, err = GetOrCreateSnapshot("gun", store, crypto)
+	assert.NoError(t, err)
+
+	// test when db role data is expired
+	store.UpdateCurrent("gun", storage.MetaUpdate{Role: "root", Version: 1, Data: []byte{3}})
+	_, err = GetOrCreateSnapshot("gun", store, crypto)
+	assert.NoError(t, err)
+}
+
+func TestGetSnapshotCurrExpired(t *testing.T) {
```
cyli: +1 awesome for testing that this makes use of `snapshotExpired` :D

Could we also add a test for when the content has expired?

server/snapshot/snapshot_test.go
endophage: I don't think so. Notary server only deals in public keys and shouldn't have a view into the cryptoservice. Tests on `CryptoService.Create` should be what guarantees that keys get added to cryptoservices.

server/snapshot/snapshot_test.go
endophage: It would be more intuitive.

server/snapshot/snapshot_test.go
endophage: Yeah, I did a tiny refactor on `storage.MetaStore` to create a tiny `KeyStore` interface I could implement to let me hit all the errors in `GetOrCreateKey`. Once the snapshot and timestamp versions are unified it'll be great.

server/snapshot/snapshot_test.go
endophage: See line 177 just above :-)

server/snapshot/snapshot.go
```diff
@@ -0,0 +1,161 @@
...
+// so that we don't have to run through all this again a second time.
+func contentExpired(gun string, sn *data.SignedSnapshot, store storage.MetaStore) bool {
+	expired := false
+	updatedMeta := make(data.Files)
+	for role, meta := range sn.Signed.Meta {
+		curr, err := store.GetCurrent(gun, role)
+		if err != nil {
+			return false
+		}
+		roleExpired, newHash := roleExpired(curr, meta)
```
cyli: Non-blocking:  It might be less confusing to rename the `roleExpired` variable here, as there is also a `roleExpired `function.

server/snapshot/snapshot_test.go
cyli: Right, but we want to make sure `CryptoService.Create` gets called.  `GetOrCreateSnapshotKey` just creating a key and not persisting it would also pass this test,  or just persisting it to the store, or just calling `CryptoService.Create` and not adding to the store.

server/snapshot/snapshot_test.go
cyli: Oh oops, missed that!  Awesome, thanks!

server/snapshot/snapshot_test.go
endophage: Understood, but the only way `GetOrCreate` creates a key is by calling `CryptoService.Create`. Calling `GetOrCreate` twice ensures was added to the store on the first call.

cyli: Ok, that's fair enough, especially since the tests for `GetOrCreateSnapshot` also test signing, which can't be done without the key being in the crypto service.  It might be useful to check that `CryptoService.Create` isn't called twice, and getting the key from the store is just masking the second create, but as that doesn't actually hurt anything if the code did that, it makes sense not to check it.

cyli:
2 non-blocking nitpicks:  https://github.com/docker/notary/pull/332#discussion_r46746874 and https://github.com/docker/notary/pull/332#discussion_r46744670, which can be addressed in a separate PR if necessary.  Otherwise LGTM!  :D

server/handlers/validation.go
mtrmac: The nil check is cheap, sure; the risk of that untested/unlikely-to-be-too-maintained code unexpectedly becoming “activated” by a harmless refactoring not so. I would also feel more comfortable with a
`if repo.Root == nil { return SomeKindOfInternalError(); }`
Or do you perhaps already have plans to use this somewhere else?

server/snapshot/snapshot.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	sn := &data.SignedSnapshot{}
+	if d != nil {
+		err := json.Unmarshal(d, sn)
+		if err != nil {
+			logrus.Error("Failed to unmarshal existing snapshot")
+			return nil, err
+		}
+		if !snapshotExpired(sn) && !contentExpired(gun, sn, store) {
```
mtrmac: `if snapshotExpired(sn)`, `contentExpired` will not be called and `sn.Signed.Meta` will not be updated.

server/snapshot/snapshot.go
```diff
@@ -0,0 +1,161 @@
...
+				Length: int64(len(curr)),
+				Hashes: data.Hashes{
+					"sha256": newHash,
+				},
+			}
+		}
+		expired = expired || roleExp
+	}
+	if expired {
+		sn.Signed.Meta = updatedMeta
```
mtrmac: `updatedMeta` AFAICS does not contain metadata of any non-expired roles.

mtrmac: (I appreciate that computing the hash only once is clever, but this seems like a place where not being clever and calling some kind of canonical/non-repeated implementation in `tuf.Repo` might be worthwhile.)

And more generally, how can this function ever return `true`? AFAICS if the snapshot key is on the server (i.e. not available to the clients), the clients will not be able to send a snapshot during any update, so the server will auto-generate the snapshot at that time, and update all of them within a transaction. So AFAICS this could only trigger with a client which defines two snapshot keys, one kept locally and one on the server. What scenario am I missing?

endophage: This is aiming to cover the case that a user has not pushed an update in a sufficiently long period of time such that the current snapshot has expired. As you note a snapshot is generated on upload (if the server holds the key), so the roles contained in the most recent are assumed to be the current set of roles that should be included in the updated snapshot. All we're doing in this function is checking if they're all up to date (which they always should be due to updating the snapshot at upload time).

endophage: On returning true: `expired = expired || roleExp`

If any individual role has expired, the value of expired will become `true` and remain `true`

endophage: You're right about updatedMeta not containing unexpired roles.

server/handlers/validation.go
endophage: @mtrmac if you install the codecov plugin this function has 100% test coverage. Raising an error here could leave a person in a state where they are unable to push a valid update. This better isolates the snapshot update from changes in other parts of the validation code rather than making assumptions about what the rest of the validation code is doing and will continue to do.

server/snapshot/snapshot.go
mtrmac: > All we're doing in this function is checking if they're all up to date (which they always should be

In that case I would expect the code to fail with an internal error if the verification fails, instead of trying to fix things up. But, \*shrug\*, if the code already exists…

endophage: I thought about simply resigning the existing snapshot with an incremented version and expiry time. I'm not opposed to simplifying down to that. I wrote this part of the code before the validations so I didn't have an entirely clear picture of how it was all going to come together. After I wrote the validations code leaving this in pace was pretty much what you said "if the code already exists..."

PR #334, opened 2015-12-04
Refactor the client TestInitRepo test into reusable helper functions.

cyli:
Also, eliminate the timestamp JSON constant and just generate a new
one for the tests.

The client test now also uses KeyFileStore and certs.Manager to
verify the keys and certs on disk, rather than directly manipulating
the files themselves.  This way, if the exact implementation of
KeyFileStore or certs.Manager changes, this test won't fail so long
as KeyFileStore and certs.Manager are self-consistent.


endophage:
LGTM

PR #335, opened 2015-12-04
Client supports server snapshot

cyli:
This adds support to `NotaryRepository.Initialize` and `NotaryRepository.Publish` for server signing snapshots


endophage:
Taking a look at the `serverManagesSnapshot` stuff. I think this is going to need to be determined more dynamically. I figured there would be a command that creates a change object to rotate the snapshot key to one generated on the server (essentially pushes snapshot responsibility to the server), then on publish we just wouldn't push a snapshot if we didn't have the key, and let the server tell us whether that represents a valid update.

cyli:
@endpohage:  Good point, thanks!  I moved that to the Initialize function instead.  Publish just tries to do its best to sign a snapshot, and then let's the server accept or not.

cyli:
@endophage @diogomonica Probably ready for another look?

client/client.go
```diff
@@ -127,44 +145,33 @@ func (r *NotaryRepository) Initialize(rootKeyID string) error {
...
+			// This key is generated by the remote server.
+			key, err = getRemoteKey(r.baseURL, r.gun, role, r.roundTrip)
+			if err != nil {
+				return err
+			}
+			logrus.Debugf("got remote %s %s key with keyID: %s",
+				role, key.Algorithm(), key.ID())
+		} else {
+			// This is currently hardcoding the keys to ECDSA.
+			key, err = r.CryptoService.Create(role, data.ECDSAKey)
```
endophage: Wondering if we should explicitly create all the local keys first so that if there's an error there we haven't unnecessarily created keys on the server?

client/client.go
```diff
@@ -347,25 +351,50 @@ func (r *NotaryRepository) Publish() error {
...
 	// check if our root file is nearing expiry. Resign if it is.
 	if nearExpiry(r.tufRepo.Root) || r.tufRepo.Root.Dirty {
-		if err != nil {
-			return err
-		}
 		root, err = r.tufRepo.SignRoot(data.DefaultExpires("root"))
+		updateRoot = true
+	}
+
+	if updateRoot {
```
endophage: Because you've improved the code layout I think you can just delete this check and move the inner block inside the preceding `if`, deleting `updateRoot = true` at the same time.

client/client.go
```diff
@@ -373,28 +402,16 @@ func (r *NotaryRepository) Publish() error {
...
-		update["root"] = rootJSON
-	}
-	update["targets"] = targetsJSON
-	update["snapshot"] = snapshotJSON
 	err = remote.SetMultiMeta(update)
 	if err != nil {
+		// TODO: this isn't exactly right, since there could be lots of
+		// reasons a request 400'ed.  Need better error translation from HTTP
+		// status codes maybe back to the server errors?
+		if _, ok := err.(store.ErrInvalidOperation); ok && clientCantSignSnapshot {
```
endophage: Given the range of possible errors, it seems really bad to make this specific assumption. We should rely on the server to return a sensible error (which is something we definitely need to work on).

cyli: Hmm... the server would only be able to send something like BadHierarchy, because it doesn't know if it's supposed to have the key and just doesn't.  And Publish doesn't know if it's supposed to be the signer either, because that is missing information that is inferred - shouldn't it really be the same error as the user just missing the key (e.g. `ErrNoKeys`)?

client/client.go
cyli: Sure!  :)

client/client.go
cyli: I wanted to :|  `updateRoot` is also set up near the top of the function - if there was an error reaching the trust server or if the trust server had no data on the repo, then we bootstrap from local repo files and also want to update the root, because we're not sure if the server has it.

cyli: No wait, I think I got it.  :D  Thanks for pointing this out - I'd forgotten to revisit it!

client/client.go
cyli: Ok, I'm just going to return the server-propogated error here.  #351 tries to add some niceties for propagating validation errors, and we can see where that goes (if anywhere).

client/client.go
```diff
@@ -137,15 +137,34 @@ func NewTarget(targetName string, targetPath string) (*Target, error) {
...
 // Initialize creates a new repository by using rootKey as the root Key for the
 // TUF repository.
-func (r *NotaryRepository) Initialize(rootKeyID string) error {
+func (r *NotaryRepository) Initialize(rootKeyID string, serverManagedRoles ...string) error {
 	privKey, _, err := r.CryptoService.GetPrivateKey(rootKeyID)
 	if err != nil {
 		return err
 	}
 
+	// currently we only support server managing snapshots, and nothing else
```
diogomonica: snapshots and timestamps, no?

client/client.go
```diff
@@ -137,15 +137,34 @@ func NewTarget(targetName string, targetPath string) (*Target, error) {
...
+	rolesAreManaged := map[string]bool{
```
diogomonica: what about `managedRoles`

client/client.go
```diff
@@ -167,44 +186,40 @@ func (r *NotaryRepository) Initialize(rootKeyID string) error {
...
-
-	kdb.AddKey(rootKey)
-	kdb.AddKey(targetsKey)
-	kdb.AddKey(snapshotKey)
-	kdb.AddKey(timestampKey)
-
-	err = initRoles(kdb, rootKey, targetsKey, snapshotKey, timestampKey)
-	if err != nil {
-		return err
+	// we want to create all the local keys first so we don't have to
```
diogomonica: Talked in person, we should refactor

client/client.go
```diff
@@ -337,22 +352,17 @@ func (r *NotaryRepository) Publish() error {
...
 		if _, ok := err.(store.ErrMetaNotFound); ok {
 			// if the remote store return a 404 (translated into ErrMetaNotFound),
-			// the repo hasn't been initialized yet. Attempt to load it from disk.
+			// there is no trust data for yet. Attempt to load it from disk.
 			err := r.bootstrapRepo()
 			if err != nil {
-				// Repo hasn't been initialized, It must be initialized before
-				// it can be published. Return an error and let caller determine
-				// what it wants to do.
-				logrus.Debug(err.Error())
```
diogomonica: Why are we removing the debug?

client/client.go
```diff
@@ -383,25 +393,45 @@ func (r *NotaryRepository) Publish() error {
 		return err
 	}
 
+	// these are the tuf files we will need to update, serialized as JSON before
+	// we send anything to remote
+	update := make(map[string][]byte)
```
diogomonica: what about updatedFiles

client/client.go
```diff
@@ -383,25 +393,45 @@ func (r *NotaryRepository) Publish() error {
...
 			return err
 		}
-		updateRoot = true
+		update[data.CanonicalRootRole] = rootJSON
 	}
-	// we will always resign targets and snapshots
-	targets, err := r.tufRepo.SignTargets("targets", data.DefaultExpires("targets"))
+
+	// we will always resign targets, and snapshots if we have the snapshots key
+	targetsJSON, err := serializeCanonicalRole(r.tufRepo, data.CanonicalTargetsRole)
```
diogomonica: We're getting targetsJSON as the output of `serializeCanonicalRole` when 10 lines up we're getting rootJSON, as the output of a direct call to `json.Marshal(root)` where `root` comes from a direct `r.tufRepo.SignRoot`. We should have a single way of getting the resigned files for all roles instead.

client/client_test.go
```diff
@@ -98,26 +115,104 @@ func initializeRepo(t *testing.T, rootType, tempBaseDir, gun, url string) (
...
+	// Temporary directory where test files will be created
+	tempBaseDir, err := ioutil.TempDir("/tmp", "notary-test-")
+	assert.NoError(t, err, "failed to create a temporary directory")
+	defer os.RemoveAll(tempBaseDir)
+
+	repo, rootPubKeyID := createRepoAndKey(
+		t, data.ECDSAKey, tempBaseDir, "docker.com/notary", "http://localhost")
+	err = repo.Initialize(rootPubKeyID, data.CanonicalRootRole)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(),
```
endophage: Seems like we should create an error type for this case of attempting to have the server manage an unsupported key.

client/client.go
```diff
@@ -383,25 +393,45 @@ func (r *NotaryRepository) Publish() error {
...
+	// we send anything to remote
+	update := make(map[string][]byte)
+
 	// check if our root file is nearing expiry. Resign if it is.
-	if nearExpiry(r.tufRepo.Root) || r.tufRepo.Root.Dirty {
-		if err != nil {
-			return err
-		}
+	if nearExpiry(r.tufRepo.Root) || r.tufRepo.Root.Dirty || updateRoot {
 		root, err = r.tufRepo.SignRoot(data.DefaultExpires("root"))
```
diogomonica: should this be `serializeCanonicalRole` ?

client/client.go
diogomonica: The ecomment says "if we have the snapshots key". How do we know if we have the snapshots key? After this comment I expected an if condition that gated the call to `serializeCanonicalRole`. Like:

```
if SnapshotsKey {
	targetsJSON, err := serializeCanonicalRole(r.tufRepo, data.CanonicalTargetsRole)
}
```

client/client.go
```diff
@@ -383,25 +393,45 @@ func (r *NotaryRepository) Publish() error {
...
-		return err
+	update[data.CanonicalTargetsRole] = targetsJSON
+
+	// do not update the snapshot role if we do not have the snapshot key or
+	// any snapshot data.  There might not be any snapshot data the repo was
+	// initialized with the snapshot signing role delegated to the server.
+	// The repo might have snapshot data, because it was requested from
+	// the server by listing, but not have the snapshot key, so signing will
+	// fail.
+	if r.tufRepo.Snapshot != nil {
```
diogomonica: the comment says "do not update if we do not have the key", yet the condition is `if r.tufRepo.Snapshot != nil`

client/client.go
diogomonica: Now that I read it again it seems the comment is just out of place?

client/client.go
```diff
@@ -383,25 +393,45 @@ func (r *NotaryRepository) Publish() error {
...
+		snapshotJSON, err := serializeCanonicalRole(
+			r.tufRepo, data.CanonicalSnapshotRole)
+		if err == nil { // we have the key - snapshot signed, let's update it
+			update[data.CanonicalSnapshotRole] = snapshotJSON
+		} else if _, ok := err.(signed.ErrNoKeys); ok {
+			logrus.Debugf("Client does not have the key to sign snapshot. " +
```
diogomonica: why the " + "?
also assuming the server should sign the snapshot seems like a good way to get in an inconsistent state. Is there a way at this point in the code of understanding if the server should have it?

client/client.go
```diff
@@ -383,25 +393,45 @@ func (r *NotaryRepository) Publish() error {
...
+				"Assuming that server should sign the snapshot.")
+		} else {
+			return err
```
diogomonica: Good place to add a debug line.

client/client.go
```diff
@@ -471,30 +486,30 @@ func (r *NotaryRepository) bootstrapRepo() error {
...
+
 	snapshotJSON, err := r.fileStore.GetMeta("snapshot", 0)
-	if err != nil {
-		return err
-	}
-	snapshot := &data.SignedSnapshot{}
-	err = json.Unmarshal(snapshotJSON, snapshot)
-	if err != nil {
-		return err
+	if err == nil {
```
diogomonica: Are all errors safe to ignore? is there a specific error that we're looking for here?

client/client.go
```diff
@@ -512,27 +527,22 @@ func (r *NotaryRepository) saveMetadata() error {
 		targetsToSave[t] = targetsJSON
 	}
 
-	signedSnapshot, err := r.tufRepo.SignSnapshot(data.DefaultExpires("snapshot"))
-	if err != nil {
-		return err
+	for role, blob := range targetsToSave {
+		parentDir := filepath.Dir(role)
+		os.MkdirAll(parentDir, 0755)
+		r.fileStore.SetMeta(role, blob)
```
diogomonica: should setMeta ensure the existence of the parent directories?

client/client_test.go
```diff
@@ -88,7 +88,24 @@ func errorTestServer(t *testing.T, errorCode int) *httptest.Server {
 	return server
 }
 
-func initializeRepo(t *testing.T, rootType, tempBaseDir, gun, url string) (
+func initializeRepo(t *testing.T, rootType, tempBaseDir, gun, url string,
```
diogomonica: shouldn't this be generic for targets too?

client/helpers.go
```diff
@@ -112,35 +113,56 @@ func nearExpiry(r *data.SignedRoot) bool {
...
+// Fetches a public key from a remote store, given a gun and role
+func getRemoteKey(url, gun, role string, rt http.RoundTripper) (data.PublicKey, error) {
+
+	remote, err := getRemoteStore(url, gun, rt)
 	if err != nil {
-		return err
+		return nil, err
 	}
-	snapshotRole, err := data.NewRole("snapshot", 1, []string{snapshotKey.ID()}, nil, nil)
+	rawKey, err := remote.GetKey(role)
```
diogomonica: `rawPubKey`

client/client.go
cyli: My intent was to remove the extra logic - the `Debug` was removed by mistake.

client/client.go
cyli: The comment is just out of place, sorry.  :)  And yes, we should be using `serializeCanonicalRole` here too.

client/client.go
cyli: "or any snapshot data"

client/client.go
cyli: Unless we update the TUF files to add extra state about who has the key, all we can do is know that we do not have the key, not that the server is supposed to have it.  We can't tell whether we just lost it or whether we're not supposed to have it.

cyli: You bring up a good point though - if we designed the server as the signer, then we don't generate a snapshot file.  If we then change it again so that we have the key, no snapshot file can be generated, so we might not be able to switch back without having first published.  I have tests for switching back and forth in #348 (https://github.com/docker/notary/pull/348/files#diff-fce84029d346504e44323c30d21a3808R933) but not for publishing.  Will add that.

The "+" was to wrap a long text line.

client/client.go
cyli: I wanted to move forward even if the file does not exist, specifically, which as per above, may not be correct anymore.

client/client.go
diogomonica: Please change the order (to match what we are actually checking)

cyli: This whole thing may change due to https://github.com/docker/notary/pull/335#discussion_r47299482

client/client.go
diogomonica: TUF spec supports additional data to be set on keys. Maybe we should take advantage of that?  /cc @endophage

client/client.go
cyli: Fixed in https://github.com/cyli/notary/commit/aaf45a9ccec53bbcfd9c88b5032f32bf8c64cfcd

client/client.go
cyli: That seems to make sense?  I've filed issue #355 for it, since it should probably be addressed in a separate issue if so.

diogomonica:
LGTM

client/client.go
```diff
@@ -53,6 +53,17 @@ type ErrExpired struct {
 	signed.ErrExpired
 }
 
+// ErrInvalidRemoteKeyType is returned when the server is requested to manage
+// an unsupported key type
+type ErrInvalidRemoteKeyType struct {
```
endophage: "key type" is overloaded as it's already used to signify the algorithm for a key. This should probably be `ErrInvalidRemoteRole`

client/client.go
endophage: :+1:

tuf/store/filestore.go
```diff
@@ -43,6 +43,10 @@ type FilesystemStore struct {
 func (f *FilesystemStore) GetMeta(name string, size int64) ([]byte, error) {
 	fileName := fmt.Sprintf("%s.%s", name, f.metaExtension)
 	path := filepath.Join(f.metaDir, fileName)
+	if _, err := os.Stat(path); os.IsNotExist(err) {
```
endophage: No need to do the `Stat`, can just check the `err` from `ioutils.ReadFile` to see if it's `IsNotExists`

cyli: Oh nice!  For some reason I thought that only worked on the result of `os.Stat`, thanks!

diogomonica: But if we were talking about large files, `os.Stat` would be a lot more efficient than waiting for `ReadFile` to return.

endophage:
LGTM!

PR #338, opened 2015-12-04
KeyDBStore refactor so that it just directly takes the DB arguments.

### LINKED ISSUE #336: An obscure error omit
It seems the `err` will always `nil`

https://github.com/docker/notary/blob/master/cmd/notary-signer/main.go#L103-L107

since we ignore error message here

https://github.com/docker/notary/blob/master/signer/keydbstore/keydbstore.go#L53

cyli:
Rather than create an SQL DB, then create a gorm BD using the SQL
DB.  Also split the Create/Get test into two tests.

This also fixes #336

### DIFF
--- a/cmd/notary-signer/main.go
+++ b/cmd/notary-signer/main.go
@@ -4,7 +4,6 @@ package main
 
 import (
 	"crypto/tls"
-	"database/sql"
 	"errors"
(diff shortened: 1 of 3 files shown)


endophage:
This makes sense. LGTM.

PR #339, opened 2015-12-04
Get Snapshot Key Handler

cyli:
This adds a handler for getting the snapshot key.

Based on #332


endophage:
This is looking good so far. The test coverage should be fixed once #332 gets merged and this is rebased on top

endophage:
LGTM.

After this is merged I'll refactor `GetOrCreateTimestampKey` and `GetOrCreateSnapshotKey` into a single common function.

PR #341, opened 2015-12-07
Add a function to parse viper

HuKeping:
None of these `filename`, `ext` or `configPath` should be in `main`,
they are all just for creating a instance of Viper and then nothing.
Do it in a separate function will make the function `main` more readable.

### DIFF
--- a/utils/configuration_test.go
+++ b/utils/configuration_test.go
@@ -335,3 +335,28 @@ func TestParseTLSWithEnvironmentVariables(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, expected, *tlsOpts)
 }
+
+func TestParseViperWithInvalidFile(t *testing.T) {
+	v := viper.New()
+	SetupViper(v, envPrefix)
+
+	err := ParseViper(v, "Chronicle_Of_Dark_Secrets.json")
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "Could not read config")
+}
+
+func TestParseViperWithValidFile(t *testing.T) {
+	file, err := os.Create("/tmp/Chronicle_Of_Dark_Secrets.json")
+	assert.NoError(t, err)
+	defer os.Remove(file.Name())
+
+	file.WriteString(`{"logging": {"level": "debug"}}`)
+
+	v := viper.New()
+	SetupViper(v, envPrefix)
+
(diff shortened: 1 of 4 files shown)


cyli:
@HuKeping Thanks for factoring this out!  It's great to have it be its own function!   Two nitpicks:

1.  It'd be awesome if `notary-server/main.go` could also make use of this function, since it has similar code
2.  Would you mind adding a test for this function?  Since viper [does not seem to expose the config name/extension/path](https://github.com/spf13/viper/blob/master/viper.go#L140), maybe just the following cases:
   - read success:  create temporary json file with one parameter, verify that there is no error reading that config file and getting the single parameter
   - read failure: pass a config file that doesn't exist?

That should hopefully fix the code coverage check.  :)

endophage:
Yep, add tests and we'll be happy to merge this. Additionally refactor notary-server/main.go to use it and we'll be extra excited to merge :-)

HuKeping:
Tests are ready and also add the change on server side.

utils/configuration_test.go
```diff
@@ -335,3 +335,30 @@ func TestParseTLSWithEnvironmentVariables(t *testing.T) {
...
+	file.WriteString(`{"logging": {"level": "debug"}}`)
+
+	v := viper.New()
+	SetupViper(v, envPrefix)
+
+	err = ParseViper(v, "/tmp/Chronicle_Of_Dark_Secrets.json")
+	assert.NoError(t, err)
+
+	if lvl := v.GetString("logging.level"); lvl != "debug" {
+		t.Fatalf("Expect debug, got %v", lvl)
```
cyli: Non-blocking, since we don't have a coding standard for tests yet:  we are trying to convert instances of `t.Fatalf` to use `assert` now.  :)  So this would probably be something like `assert.Equal(t, "debug", v.GetString("logging.level"))`.

cyli:
Thank you for your contribution!  :)

utils/configuration_test.go
endophage: If there were other `t.Fatal` calls already in this file I'd be fine with leaving this and cleaning them all up in one go. However this has introduced the one and only `t.Fatal` call so for consistency, please update to @cyli's suggested fix.

endophage:
LGTM. Thanks!

Issue #347, opened 2015-12-09
Provide endpoint to rotate server-managed keys


cyli:
Such as snapshot and timestamp.  Not doing this means that if the user gives the command to rotate a key (particularly snapshot) and the key is already remotely stored, then no key rotation occurs.

This requires the user to sign the new `root.json`, so the server cannot just do this.

Proposal:

1. Remove unique gun/role constraint from server public keys table
1. Add column that specifies whether it's current or new
1. When user requests to rotate a timestamp for example, a new timestamp key is generated for the gun, but with `current=false`
1. Whenever a user requests to rotate the timestamp again, that same non-current timestamp key is returned.
1. Whenever a root.json is pushed, check whether we have the key in the DB for all keys.  If we do, and it's not current, promote it to current.  I think we can just then delete any key in the DB for the gun that is not in the root.json, so long as we are sure of all our server validations.

riyazdf:
Closed with #889

PR #348, opened 2015-12-09
Client rotate snapshot key to server

cyli:
This is based on #335, and adds the command line utility to rotate a single key (and optionally tell the remote server to manage it).

This involves a fairly large refactor of `cmd/notary/keys.go` in order to remove some of the global mutable state, so that the functions are testable.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -653,11 +653,7 @@ func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool) error {
 		return err
 	}
 
-	err = r.rootFileKeyChange(role, changelist.ActionCreate, pubKey)
-	if err != nil {
-		return err
-	}
-	return nil
+	return r.rootFileKeyChange(role, changelist.ActionCreate, pubKey)
 }
 
 func (r *NotaryRepository) rootFileKeyChange(role, action string, key data.PublicKey) error {
--- a/client/client.go
+++ b/client/client.go
@@ -628,19 +628,34 @@ func (r *NotaryRepository) bootstrapClient() (*tufclient.Client, error) {
 	), nil
 }
 
-// RotateKeys removes all existing keys associated with role and adds
-// the keys specified by keyIDs to the role. These changes are staged
-// in a changelist until publish is called.
-func (r *NotaryRepository) RotateKeys() error {
-	for _, role := range []string{"targets", "snapshot"} {
-		key, err := r.CryptoService.Create(role, data.ECDSAKey)
-		if err != nil {
-			return err
-		}
-		err = r.rootFileKeyChange(role, changelist.ActionCreate, key)
-		if err != nil {
-			return err
-		}
+// RotateKey removes all existing keys associated with the role, and either
+// creates and adds one new key or delegates managing the key to the server.
+// These changes are staged in a changelist until publish is called.
+func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool) error {
+	if role == data.CanonicalRootRole || role == data.CanonicalTimestampRole {
+		return fmt.Errorf(
+			"notary does not currently support rotating the %s key", role)
+	}
+	if serverManagesKey && role == data.CanonicalTargetsRole {
+		return ErrInvalidRemoteRole{Role: data.CanonicalTargetsRole}
+	}
+
+	var (
+		pubKey data.PublicKey
+		err    error
+	)
+	if serverManagesKey {
+		pubKey, err = getRemoteKey(r.baseURL, r.gun, role, r.roundTrip)
+	} else {
+		pubKey, err = r.CryptoService.Create(role, data.ECDSAKey)
+	}
+	if err != nil {
+		return err
+	}
+
+	err = r.rootFileKeyChange(role, changelist.ActionCreate, pubKey)
+	if err != nil {
+		return err
 	}
 	return nil
 }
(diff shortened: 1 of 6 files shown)


diogomonica:
rebase needed.

cyli:
Will rebase once #355 is merged :)

cyli:
@diogomonica @endophage PTAL?  :)

client/client.go
```diff
@@ -628,19 +628,34 @@ func (r *NotaryRepository) bootstrapClient() (*tufclient.Client, error) {
...
+	if serverManagesKey {
+		pubKey, err = getRemoteKey(r.baseURL, r.gun, role, r.roundTrip)
+	} else {
+		pubKey, err = r.CryptoService.Create(role, data.ECDSAKey)
+	}
+	if err != nil {
+		return err
+	}
+
+	err = r.rootFileKeyChange(role, changelist.ActionCreate, pubKey)
```
endophage: `return r.rootFileKeyChange(role, changelist.ActionCreate, pubKey)`

endophage:
Really like the implementation!

Other than the one cleanliness comment, LGTM.

client/client.go
diogomonica: shouldn't a rotation also involve a remove (of the old key?)

diogomonica:
LGTM. Good refactor. Integration tests ftw!

client/client.go
cyli: Only when published - this just creates a change file.

Removing old keys happens in https://github.com/docker/notary/blob/master/tuf/tuf.go#L117, which gets called from https://github.com/docker/notary/blob/master/tuf/tuf.go#L107, which gets called from https://github.com/docker/notary/blob/master/client/helpers.go#L92, which gets called by https://github.com/docker/notary/blob/master/client/helpers.go#L81, which gets called by https://github.com/docker/notary/blob/master/client/helpers.go#L29, which gets called by [NotaryRepository.Publish](https://github.com/docker/notary/blob/master/client/client.go#L406)

diogomonica:
```
➜  notary git:(client-rotate-snapshot-key-to-server) bin/notary -c cmd/notary/config.json key rotate diogomonica -t snapshots -r
Usage:
  notary key rotate [ GUN ] [flags]

Flags:
  -t, --key-type="": Key type to rotate.  Supported values: "targets", "snapshots". If not provided, both targets and snapshot keys will be rotated, and the new keys will be locally generated and stored.
  -r, --server-managed[=false]: Signing and key management will be handled by the remote server. (no key will be generated or stored locally) Can only be used in conjunction with --key-type.

Global Flags:
  -c, --configFile="": Path to the configuration file to use
  -s, --server="": Remote trust server location
  -d, --trustDir="": Directory where the trust data is persisted to
  -v, --verbose[=false]: Verbose output
* fatal: key rotation not supported for snapshots keys
```

This stems from the confusion of `snapshots` vs `snapshot`. We should change the `-t` flag's description.

diogomonica:
Everything works as intended on the cmd-line, good work. One thing that I notices is that our earlier decision to hide the snapshot and targets passphrases behind the same cmd-line input request is becoming increasingly confusing as we allow individual key rotation and remote delegation.

Maybe we can make snapshots delegation the default and decouple the passphraseRetriever request? Not relevant for this PR, though.

LGTM after fix of the key-type text.

diogomonica:
It seems that remote rotation doesn't actually rotate they key.

I can rotate to a local key transparently, everything works. The first time I rotate remote, a `snapshot` key indeed does get created, but future rotations with `-r` don't actually rotate the remote key.

Not sure if this got discussed and works as intended, but my MySQL shows no new keys being added:

```
mysql> select gun, cipher, role from timestamp_keys;
+-------------+--------+-----------+
| gun         | cipher | role      |
+-------------+--------+-----------+
| diogomonica | ecdsa  | timestamp |
| diogomonica | ecdsa  | snapshot  |
+-------------+--------+-----------+
2 rows in set (0.00 sec)
```

The ID `b0972` is consistent across rotations with `-t snapshot -r`

```
mysql> select key_id, algorithm, passphrase_alias from private_keys;
+------------------------------------------------------------------+-----------+------------------+
| key_id                                                           | algorithm | passphrase_alias |
+------------------------------------------------------------------+-----------+------------------+
| 29545a660fd79aefe8fbbe0f67bfd770cb9d7542044732e1adf5ce09688a485a | ecdsa     | timestamp_1      |
| b0972b77038c95c92234e027237b356915a15253103a7ac674c101cee677fd31 | ecdsa     | timestamp_1      |
+------------------------------------------------------------------+-----------+------------------+
2 rows in set (0.00 sec)
```

Nitpick1: The table with the keys is called `timestamp_keys`, but no longer only holds timestamp keys.
Nitpick2: The passphrase alias is "timestamp_1" where it should probably be something more generic (since we now host other keys other than timestamps).

cyli:
@diogomonica It's not exactly intended, but it's a known issue:  https://github.com/docker/notary/issues/347

We need to have an endpoint that does remote rotation - and create a new key instead of just getting the existing key (so there are 2 keys for a role/gun), and we need a way to clean up the old keys once the rotation 'takes'.

cyli:
@diogomonica wrt to `timestamp_keys`, we haven't renamed it yet because:  https://github.com/docker/notary/pull/329#issuecomment-161513182.

cyli:
@diogomonica wrt:
> Maybe we can make snapshots delegation the default and decouple the passphraseRetriever request? Not relevant for this PR, though.

Isn't that correct though?  It's asking for the targets file because when we upload, we always resign upload the targets file.  Even with local snapshot keys, it does not ask for the snapshots password on publish.

endophage:
@diogomonica our intent so far is to update Docker to have the server sign the snapshot by default.

cyli:
@diogomonica Nice catch on the key-type usage mismatch!

PR #349, opened 2015-12-09
fixing bugs raised by @mtrmac

endophage:
cc @mtrmac (not sure if the title inclusion will ping)

c.f. https://github.com/docker/notary/pull/332

### DIFF
--- a/server/handlers/validation.go
+++ b/server/handlers/validation.go
@@ -81,9 +81,21 @@ func validateUpdate(cs signed.CryptoService, gun string, updates []storage.MetaU
 			return nil, validation.ErrBadTargets{Msg: err.Error()}
 		}
 		repo.SetTargets(targetsRole, t)
+	} else {
+		targetsJSON, err := store.GetCurrent(gun, data.CanonicalTargetsRole)
+		if err != nil {
+			return nil, validation.ErrValidation{Msg: err.Error()}
+		}
+		targets := &data.SignedTargets{}
+		err = json.Unmarshal(targetsJSON, targets)
+		if err != nil {
+			return nil, validation.ErrValidation{Msg: err.Error()}
+		}
+		repo.SetTargets(data.CanonicalTargetsRole, targets)
 	}
 	logrus.Debug("Successfully validated targets")
 
+	// At this point, root and targets must have been loaded into the repo
 	if _, ok := roles[snapshotRole]; ok {
 		var oldSnap *data.SignedSnapshot
 		oldSnapJSON, err := store.GetCurrent(gun, snapshotRole)
@@ -111,10 +123,6 @@ func validateUpdate(cs signed.CryptoService, gun string, updates []storage.MetaU
 		// Then:
 		//   - generate a new snapshot
 		//   - add it to the updates
-		err := prepRepo(gun, repo, store)
-		if err != nil {
-			return nil, err
-		}
 		update, err := generateSnapshot(gun, kdb, repo, store)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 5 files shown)


mtrmac:
:+1:, thanks for fixing this so quickly!

endophage:
@mtrmac we're trying to get server snapshots in ASAP so we can start working on delegations but it's coming in as a bunch of separate PRs so we're expecting to find bugs up until we have it all assembled and tested end to end. We really appreciate your review!

endophage:
@mtrmac does this make sense: https://github.com/docker/notary/pull/332#discussion_r47135332 ???

mtrmac:
(re: the `testRepo` `repo.Root == nil` path): “On principle” I don’t particularly like this, code should rather fail closed, and code with clear assumptions and clear expected paths through makes it easier to figure out what is going on.

OTOH, *shrug* It is 10 fairly simple lines with no policy, and honestly, I am pressed enough for time necessary to add useful features that I am not going to submit a PR dropping this code, this December at least, so I am by definition OK with it staying in ;-)

mtrmac:
… hum, actually, the argument that this path should stand alone from assumptions about what `verifyUpdate` does elsewhere could be compelling; though it would be much more compelling if the two functions did not share the `repo` variable. Anyway, I am not submitting a PR.

endophage:
I disagree with you on whether the `repo.Root == nil` is a failure case in `prepRepo` (a failure case should by definition not be recoverable). However, I'm reassessing the code and thinking `validateUpdate` could be restructured to be a little less scattered which would also allow removing `prepRepo`.

mtrmac:
Sorry, I was a bit vague in the “fail closed” reference and I don’t particularly want to argue the detailed semantics of that.

I do very much like the restructuring commit above and how it conceptually splits `validateUpdate` into clear one-role-at-a-time sections.

endophage:
Yeah, it's very much opinion and semantics. If you feel there was a security vulnerability with that recovery case I'd like to know for my own education. Otherwise I'm very much of the opinion we should look to provide the best user experience wherever possible, which means recovering when it doesn't compromise security.

I wasn't entirely happy with `prepRepo` when I put it in but I also didn't want to lose a day trying to make it perfect. Having had time to let it simmer in the back of my mind, when you raised it I had a firmer idea about how to improve it. 

We want to look at using the server/handlers/validation.go code in the client and unifying all the validation. From this state, I'd like to spend some time actually breaking out `validateUpdate` into multiple role focussed functions, which I think in isolation are likely to be of more use on the client side. We have a new dev starting on Monday so hopefully in the near future I'll find time to do that.

mtrmac:
No, I don’t know about any security vulnerability in the `prepRepo` code; it is at most a technical debt, or more likely subjective coding style preference, matter.

server/handlers/validation.go
```diff
@@ -80,9 +80,21 @@ func validateUpdate(cs signed.CryptoService, gun string, updates []storage.MetaU
 			return nil, ErrBadTargets{msg: err.Error()}
 		}
 		repo.SetTargets(targetsRole, t)
+	} else {
+		targetsJSON, err := store.GetCurrent(gun, data.CanonicalTargetsRole)
+		if err != nil {
+			return nil, fmt.Errorf("could not load repo for snapshot generation: %v", err)
```
cyli: Actually it's great that this is moved here, since it'd be a failure if there were no targets metadata and the user did not send target metadata.

Should the error be different?  It's not loading just for snapshot generation, since presumably since if there's no targets file available that should be a problem with the repo.

cyli: Although, actually, now that I think about it, does the spec say that it's an error if there is no `targets.json`, so long as there is no entry for one in `snapshots.json`?

cyli: Answering my own question, it seems like it's implied [because when listing the top level metadata files, only mirror is listed as optional](https://github.com/theupdateframework/tuf/blob/develop/docs/tuf-spec.txt#L348).  Although it'd be nice if the spec explicitly said that those files must exist, or that the snapshot metadata must have hashes for all those roles, like [spec definition for the root.json metadata says it MUST have keys for all 4 roles](https://github.com/theupdateframework/tuf/blob/develop/docs/tuf-spec.txt#L519).  So it seems like maybe they're leaving the implementation open, but we can just interpret it as it always has to exist, even if it's empty?

endophage: The error isn't necessarily because there's no targets. It could be disrupted connectivity to the database or something like that. At this point in `validateUpdate` the sole purpose of loading the targets.json is to create/validate the snapshot.

So far I've gone with the understanding that the 4 main roles, root, targets, snapshot, and timestamp, must exist both in the root, and their associated json files.

cyli: Shouldn't it be a validation error if we can't load the targets, partially because then we can't validate the snapshot, but also because we need to also validate the old target file.  (We should probably load and validate all 4 pieces of metadata as a whole - for instance, I think currently, if we rotate the targets key and submit a new root.json, but don't submit a targets.json, which seems wrong because then the repo will be in a broken state?)

endophage: I believe the client always pushes a targets.json. We could try and be more intelligent on the client and only push it when necessary, but the (not correct but protective) assumption is that any change will require targets.json to be resigned.

I should assign a better error type, and I just realized they explicitly say "generation" which may not be correct (and could be troubling to some users) so I'll fix that too.

cyli: Anyway, sorry, this is just bikeshedding the error message :)  We can modify it later, too.

server/snapshot/snapshot.go
```diff
@@ -96,14 +102,11 @@ func contentExpired(gun string, sn *data.SignedSnapshot, store storage.MetaStore
...
 		if roleExp {
-			updatedMeta[role] = data.FileMeta{
-				Length: int64(len(curr)),
-				Hashes: data.Hashes{
-					"sha256": newHash,
-				},
-			}
+			updatedMeta[role] = newMeta
+		} else {
+			updatedMeta[role] = meta
```
cyli: Could we add a test for this case, where some of the content has expired and other content has not?

diogomonica:
It seems like tests got removed because the function went away, but functionally speaking it looks like they were covering some invalid cases that are worth still covering.

endophage:
@diogomonica the PrepRepo tests that got removed are no longer relevant. The cases they tested were cases that PrepRepo could encounter in isolation but can't actually be reached as part of `validateUpdate`. Now that PrepRepo is gone, there's no way to encounter those cases.

server/snapshot/snapshot.go
endophage: @cyli as we're progressing with the delegations work I'm seeing more and more that this attempt to update the snapshot is a poor idea. I'm just going to remove this code. Getting a "new" snapshot will just resign the previous one with an updated expiry if its expired.

endophage:
@cyli @diogomonica let's get this merged asap

PR #350, opened 2015-12-09
WIP + PoC: adding grpc interface for notary server

endophage:
This is totally just throwing something out there to start discussion on precisely what the interface might look like. There's copy pasted code that we'd want to refactor into common functions, and we need to make further changes to the current HTTP handler code to cleanly separate the HTTP stuff from repo management such that both the HTTP and RPC handlers can share the underlying code.

There's one big thing that needs to happen to make this a possibility: updating docker/distribution/registry/auth to not be reliant on the http.Request, at least providing an alternative method that allows me to directly pass the credentials (cc @stevvooe for thoughts on that).

cc @davedoesdev @HuKeping 


proto/server.proto
```diff
@@ -0,0 +1,51 @@
...
+}
+
+message Updates {
+	bytes jwt = 1;
+	string gun = 2;
+	Repository repo = 3;
+}
+
+message RepoRole {
+	bytes jwt = 1;
```
stevvooe: I am not sure if this can be changed, but it might be better to use a header/trailer to transmit the token, rather than including it on very request object. This way, we can use a credentials implementation to actually check it on each request. I am not yet sure how this should be implemented.

proto/server.proto
```diff
@@ -0,0 +1,51 @@
+syntax = "proto3";
```
stevvooe: Is there are `go generate` file somewhere for this?

endophage: Unfortunately not generated directly by go. We need to document building the protos, but in short you need to install protobuf3 beta 1, `go get -a github.com/golang/protobuf/protoc-gen-go`, `go get google.golang.org/grpc`, then you should be able to `make protos` (see the Makefile for the exact command that gets run)

proto/server.proto
endophage: Everything is open for suggestions. Hadn't looked deep enough into GRPC to find how it handles headers. Will look further.

proto/server.proto
stevvooe: That still all works with `go generate`.

endophage: Had to take another look at `go generate`. Sure it'll work, it's just
irrelevant. AFAIK the protoc command only needs $GOPATH, which `go
generate` doesn't set.

proto/server.proto
```diff
@@ -0,0 +1,51 @@
...
+	Repository repo = 3;
+}
+
+message RepoRole {
+	bytes jwt = 1;
+	string gun = 2;
+	string role = 3;
+}
+
+// Repository is a collection of TUF files
```
davedoesdev: Are the Notary TUF files compatible with https://github.com/theupdateframework/tuf ?

proto/server.proto
```diff
@@ -0,0 +1,51 @@
+syntax = "proto3";
+
+package proto;
+
+// NotaryServer defines the TUF operations that the server supports
+service Notary {
+	rpc Auth(Token) returns (AuthResponse) {}
+	rpc Update(Updates) returns (UpdateResponse) {}
+	rpc Get(RepoRole) returns (Repository) {}
+	rpc GetKey(RepoRole) returns (KeyResponse) {}
```
davedoesdev: Does Notary not store its keys in `root.json`?

endophage: Some keys (currently just timestamp, but we're working on snapshot as optional) are held by the server as it would be a pain for users to have to manage timestamping themselves. This function is a way for a client to request a timestamp key to sign into the root.json during initialization.

We're also talking about how to let users do server held key rotation and this function will probably also be used for that.

proto/server.proto
endophage: There is only one difference, and I have discussed it with the TUF designers prior to implementation. They said they would work it in to a future spec update.

In the official spec, the timestamp and snapshot files reference the other role files with a file extension, possibly including a compression extension, such as "snapshot.json.zip", with multiple entries for a role, each with different compression. The need to translate back and forth from the role name "snapshot" to the filename, and to do things like prefix matching and extension parsing to determine the right thing to request, creates a terrible chunk of code.

Instead, our timestamp and snapshot files simply contain the bare role name and compression is done at a lower level. We discussed if we could see any security vulnerabilities with that approach and couldn't come up with anything (there are discussed potential exploits in compression libraries but we didn't determine there was anything worrying to us for this use case, and major applications , like Chrome, are happy to assume decompression is a safe operation).

davedoesdev:
Thanks. I _think_ this is enough to write a backend for the Python TUF library. I'll add that as a work item for dtuf.

proto/server.proto
stevvooe: `go generate` honors `GOPATH`. You can set that for the Makefile environment it'll work out of the box. Having `go generate` supports a very succinct workflow when experiment with proto changes.

endophage: Yo'll have to show me. The Go docs for GRPC don't mention it unless it's buried somewhere... which it may well be, the docs are a little scattered.

stevvooe: Here is a solid setup:

https://github.com/stevvooe/continuity/blob/master/proto/gen.go

From the repo root, I can run `go generate ./proto` and it will "do the needful". It let's you get all the nasty details that might come with generation out of the Makefile.

diogomonica:
Closed for now. Let's reopen when ready.

PR #351, opened 2015-12-10
Propagate error validations from server to client

cyli:
This stuffs the server validation errors in the JSON returned back as the response, and the client can then parse said errors into agreed-upon validation errors.

This seems useful to return errors back to the client on update, but I don't really like how I'm serializing it here.  Any suggestions (or we can just throw this out) would be welcome.

Addendum:  `SetMeta` doesn't work because the server does not allow posting to individual metadata json files.  Not sure if we want to fix it or just disable setting individual metadata on the httpstore.


endophage:
This all looks good. As far as I can tell there's nothing that depends on an error being of type ErrInvalidOperation so it all appears to be non-breaking.

LGTM.

server/handlers/default_test.go
```diff
@@ -303,3 +311,77 @@ func TestGetHandlerNoStorage(t *testing.T) {
...
+
+type failStore struct {
+	storage.MemStorage
+}
+
+func (s failStore) GetCurrent(_, _ string) ([]byte, error) {
+	return nil, fmt.Errorf("oh no! storage has failed")
+}
+
+// a non-validation failure, such as the storage failing, will not be propogated
```
diogomonica: propagated

server/handlers/default_test.go
```diff
@@ -303,3 +311,77 @@ func TestGetHandlerNoStorage(t *testing.T) {
...
+// as a detail in the error (which gets serialized as the body of the response)
+func TestAtomicUpdateNonValidationFailureNotPropogated(t *testing.T) {
```
diogomonica: propagated

server/handlers/default_test.go
```diff
@@ -303,3 +311,77 @@ func TestGetHandlerNoStorage(t *testing.T) {
 	err := GetHandler(ctx, nil, req)
 	assert.Error(t, err)
 }
+
+// a validation failure, such as a snapshots file being missing, will be
+// propogated as a detail in the error (which gets serialized as the body of the
+// response)
+func TestAtomicUpdateValidationFailurePropogated(t *testing.T) {
```
diogomonica: propagated

server/handlers/default_test.go
```diff
@@ -303,3 +311,77 @@ func TestGetHandlerNoStorage(t *testing.T) {
 	err := GetHandler(ctx, nil, req)
 	assert.Error(t, err)
 }
+
+// a validation failure, such as a snapshots file being missing, will be
+// propogated as a detail in the error (which gets serialized as the body of the
```
diogomonica: propagated

cyli:
Looks like I can't spell propagated.  :D  I'll do a search and replace.

diogomonica:
LGTM. Was this the place where the conversation about GRPC originated? (strict error types?)

diogomonica:
Feel free to merge

cyli:
@diogomonica I think so, when we were discussing:  https://github.com/docker/notary/pull/335#discussion_r47043083

PR #352, opened 2015-12-10
Use canonical way to check if a map contains a key

HuKeping:
As the language spec:
https://golang.org/ref/spec#Index_expressions

And also fix a typo.

### DIFF
--- a/signer/api/api_test.go
+++ b/signer/api/api_test.go
@@ -225,3 +225,26 @@ func TestSignHandlerReturns404WithNonexistentKey(t *testing.T) {
 
 	assert.Equal(t, 404, res.StatusCode)
 }
+
+func TestCreateKeyHandlerWithInvalidAlgorithm(t *testing.T) {
+	keyStore := trustmanager.NewKeyMemoryStore(passphraseRetriever)
+	cryptoService := cryptoservice.NewCryptoService("", keyStore)
+	setup(signer.CryptoServiceIndex{data.ED25519Key: cryptoService, data.RSAKey: cryptoService, data.ECDSAKey: cryptoService})
+
+	// The `rbtree-algorithm` is expected as not supported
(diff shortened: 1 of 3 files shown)


mtrmac:
:+1:

endophage:
LGTM.

While you're in here it would be great if you add a short test to trigger the `http.Error`. Will let this sit for 24 hours and merge on Friday anyway if you don't have time.

diogomonica:
Agreed on the test, but otherwise LGTM

HuKeping:
Test ready :)

HuKeping:
BTW, I found that we have injected a specific `http.ResponseWriter` object for them to write back the http response. 

Wonder if we could do that (I mean, writing back the http response) outside the `getCryptoService`. Anyway it is not what this PR would like to deal with. If it worth to fix, I'll make another PR some days later.

signer/api/api_test.go
```diff
@@ -225,3 +225,26 @@ func TestSignHandlerReturns404WithNonexistentKey(t *testing.T) {
 
 	assert.Equal(t, 404, res.StatusCode)
 }
+
+func TestCreateKeyHandlerWithInvalidAlgorithm(t *testing.T) {
+	keyStore := trustmanager.NewKeyMemoryStore(passphraseRetriever)
+	cryptoService := cryptoservice.NewCryptoService("", keyStore)
+	setup(signer.CryptoServiceIndex{data.ED25519Key: cryptoService, data.RSAKey: cryptoService, data.ECDSAKey: cryptoService})
+
+	// The `rbtree-algorithm` is expected as not suppoted
```
cyli: Non-blocking nitpick - "supported" is mispelled.  :)  Minor though, I will merge in morning whether or not it's fixed.

cyli:
LGTM!  Thank you for your contribution!

HuKeping:
updated @cyli  :)

cyli:
@HuKeping thanks for fixing so quickly!!

PR #354, opened 2015-12-11
Delegation primitives up to changelist level

endophage:
I debated on whether `UpdateDelegations` should create the delegations file in the case it doesn't exist, but that's not necessarily possible in the general TUF case due to what keys the parent owner wants present on the delegation. Creation of this file will happen at a higher level (i.e. in NotaryRepository). When adding that functionality, we will likely want to update `repo.InitTargets` to take a role name.

### DIFF
--- a/client/changelist/change.go
+++ b/client/changelist/change.go
@@ -77,3 +77,26 @@ func (c TufChange) Path() string {
 func (c TufChange) Content() []byte {
 	return c.Data
 }
+
+// TufDelegation represents a modification to a target delegation
+// this includes creating a delegations. This format is used to avoid
+// unexpected race conditions between humans modifying the same delegation
+type TufDelegation struct {
+	NewName                string       `json:"new_name,omitempty"`
+	NewThreshold           int          `json:"threshold, omitempty"`
+	AddKeys                data.KeyList `json:"add_keys, omitempty"`
+	RemoveKeys             []string     `json:"remove_keys,omitempty"`
+	AddPaths               []string     `json:"add_paths,omitempty"`
+	RemovePaths            []string     `json:"remove_paths,omitempty"`
+	AddPathHashPrefixes    []string     `json:"add_prefixes,omitempty"`
+	RemovePathHashPrefixes []string     `json:"remove_prefixes,omitempty"`
+}
+
+// ToNewRole creates a fresh role object from the TufDelegation data
+func (td TufDelegation) ToNewRole(scope string) (*data.Role, error) {
+	name := scope
+	if td.NewName != "" {
+		name = td.NewName
+	}
+	return data.NewRole(name, td.NewThreshold, td.AddKeys.IDs(), td.AddPaths, td.AddPathHashPrefixes)
+}
--- a/client/helpers.go
+++ b/client/helpers.go
@@ -67,37 +67,51 @@ func applyTargetsChange(repo *tuf.Repo, c changelist.Change) error {
 
 func changeTargetsDelegation(repo *tuf.Repo, c changelist.Change) error {
 	switch c.Action() {
-	case changelist.ActionCreate, changelist.ActionUpdate:
+	case changelist.ActionCreate:
 		td := changelist.TufDelegation{}
 		err := json.Unmarshal(c.Content(), &td)
 		if err != nil {
 			return err
 		}
 		r, err := repo.GetDelegation(c.Scope())
-		if err == nil {
-			// role exists, merge
-			if err := r.AddPaths(td.AddPaths); err != nil {
-				return err
-			}
-			if err := r.AddPathHashPrefixes(td.AddPathHashPrefixes); err != nil {
-				return err
-			}
-			r.RemoveKeys(td.RemoveKeys)
-			r.RemovePaths(td.RemovePaths)
-			r.RemovePathHashPrefixes(td.RemovePathHashPrefixes)
-			return repo.UpdateDelegations(r, td.AddKeys, "")
-		} else if _, ok := err.(data.ErrNoSuchRole); ok {
-			// role doesn't exist, create brand new
-			r, err = td.ToNewRole(c.Scope())
-			if err != nil {
-				return err
-			}
-			return repo.UpdateDelegations(r, td.AddKeys, "")
-		} else {
-			// any error other than ErrNoSuchRole indicates
-			// bad data
-			return fmt.Errorf("could not apply delegations change because: \"%v\"", err)
+		if _, ok := err.(data.ErrNoSuchRole); err != nil && !ok {
+			// error that wasn't ErrNoSuchRole
+			return err
 		}
+		if err == nil {
+			// role existed
+			return data.ErrInvalidRole{
+				Role:   c.Scope(),
+				Reason: "cannot create a role that already exists",
+			}
+		}
+		// role doesn't exist, create brand new
+		r, err = td.ToNewRole(c.Scope())
+		if err != nil {
+			return err
+		}
+		return repo.UpdateDelegations(r, td.AddKeys)
+	case changelist.ActionUpdate:
+		td := changelist.TufDelegation{}
+		err := json.Unmarshal(c.Content(), &td)
+		if err != nil {
+			return err
+		}
+		r, err := repo.GetDelegation(c.Scope())
+		if err != nil {
+			return err
+		}
+		// role exists, merge
+		if err := r.AddPaths(td.AddPaths); err != nil {
+			return err
+		}
+		if err := r.AddPathHashPrefixes(td.AddPathHashPrefixes); err != nil {
+			return err
+		}
+		r.RemoveKeys(td.RemoveKeys)
+		r.RemovePaths(td.RemovePaths)
+		r.RemovePathHashPrefixes(td.RemovePathHashPrefixes)
+		return repo.UpdateDelegations(r, td.AddKeys)
 	case changelist.ActionDelete:
 		r := data.Role{Name: c.Scope()}
 		return repo.DeleteDelegation(r)
--- a/client/helpers.go
+++ b/client/helpers.go
@@ -55,6 +55,59 @@ func applyChangelist(repo *tuf.Repo, cl changelist.Changelist) error {
 }
 
 func applyTargetsChange(repo *tuf.Repo, c changelist.Change) error {
+	switch c.Type() {
+	case changelist.TypeTargetsTarget:
+		return changeTargetMeta(repo, c)
+	case changelist.TypeTargetsDelegation:
+		return changeTargetsDelegation(repo, c)
+	default:
+		return fmt.Errorf("only target meta and delegations changes supported")
+	}
+}
+
+func changeTargetsDelegation(repo *tuf.Repo, c changelist.Change) error {
+	switch c.Action() {
+	case changelist.ActionCreate, changelist.ActionUpdate:
+		td := changelist.TufDelegation{}
+		err := json.Unmarshal(c.Content(), &td)
+		if err != nil {
+			return err
+		}
+		r, err := repo.GetDelegation(c.Scope())
+		if err == nil {
+			// role exists, merge
+			if err := r.AddPaths(td.AddPaths); err != nil {
+				return err
+			}
+			if err := r.AddPathHashPrefixes(td.AddPathHashPrefixes); err != nil {
+				return err
+			}
+			r.RemoveKeys(td.RemoveKeys)
+			r.RemovePaths(td.RemovePaths)
+			r.RemovePathHashPrefixes(td.RemovePathHashPrefixes)
+			return repo.UpdateDelegations(r, td.AddKeys, "")
+		} else if _, ok := err.(data.ErrNoSuchRole); ok {
+			// role doesn't exist, create brand new
+			r, err = td.ToNewRole(c.Scope())
+			if err != nil {
+				return err
+			}
+			return repo.UpdateDelegations(r, td.AddKeys, "")
+		} else {
+			// any error other than ErrNoSuchRole indicates
+			// bad data
+			return fmt.Errorf("could not apply delegations change because: \"%v\"", err)
+		}
+	case changelist.ActionDelete:
+		r := data.Role{Name: c.Scope()}
+		return repo.DeleteDelegation(r)
+	default:
+		return fmt.Errorf("unsupported action against delegations: %s", c.Action())
+	}
+
+}
+
+func changeTargetMeta(repo *tuf.Repo, c changelist.Change) error {
 	var err error
 	switch c.Action() {
 	case changelist.ActionCreate:
@@ -72,10 +125,7 @@ func applyTargetsChange(repo *tuf.Repo, c changelist.Change) error {
 	default:
 		logrus.Debug("action not yet supported: ", c.Action())
 	}
-	if err != nil {
-		return err
-	}
-	return nil
+	return err
 }
 
 func applyRootChange(repo *tuf.Repo, c changelist.Change) error {
--- a/client/helpers_test.go
+++ b/client/helpers_test.go
@@ -9,6 +9,7 @@ import (
 	tuf "github.com/docker/notary/tuf"
 	"github.com/docker/notary/tuf/data"
 	"github.com/docker/notary/tuf/keys"
+	"github.com/docker/notary/tuf/testutils"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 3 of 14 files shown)


tuf/tuf.go
```diff
@@ -183,18 +183,23 @@ func (tr *Repo) UpdateDelegations(role *data.Role, keys []data.PublicKey, before
...
 
-	i := -1
-	var r *data.Role
-	for i, r = range p.Signed.Delegations.Roles {
-		if r.Name == role.Name {
-			break
-		}
+	// if the role has fewer keys than the threshold, it
+	// will never be able to create a valid targets file
+	// and should be considered invalid.
```
HuKeping: I always love these comments in code, thanks

tuf/tuf.go
```diff
@@ -183,18 +183,23 @@ func (tr *Repo) UpdateDelegations(role *data.Role, keys []data.PublicKey, before
...
+	if len(role.KeyIDs) < role.Threshold {
+		return data.ErrInvalidRole{Role: role.Name}
```
diogomonica: Is it worth to have a specific "Did not meet threshold" error?

tuf/utils/utils.go
```diff
@@ -105,3 +105,51 @@ func DoHash(alg string, d []byte) []byte {
...
+func RemoveUnusedKeys(t *data.SignedTargets, ids []string) {
+	unusedIDs := UnusedDelegationKeys(*t, ids)
+	for _, id := range unusedIDs {
+		delete(t.Signed.Delegations.Keys, id)
+	}
+}
+
+// FindRoleIndex returns the index of the role named <name> or -1 if no
+// matching role is found.
+func FindRoleIndex(rs []*data.Role, name string) int {
```
diogomonica: Shouldn't we use the more goesque way and return (int, error)?

tuf/tuf.go
endophage: I don't think so, it would get confused with the validation threshold error. Might be worth adding a "Reason" field to ErrInvalidRole though that can indicate why the role is invalid.

tuf/utils/utils.go
```diff
@@ -105,3 +105,51 @@ func DoHash(alg string, d []byte) []byte {
 	}
 	return nil
 }
+
+// UnusedDelegationKeys prunes a list of keys, returning those that are no
+// longer in use for a given targets file
+func UnusedDelegationKeys(t data.SignedTargets, ids []string) []string {
```
diogomonica: Why not receiving a `t *data.SignedTargets`?

tuf/tuf.go
diogomonica: That works.

tuf/utils/utils.go
endophage: c.f. https://golang.org/pkg/strings/#Index

tuf/utils/utils.go
endophage: Explicitly states that we won't modify the object.

tuf/utils/utils.go
```diff
@@ -105,3 +105,51 @@ func DoHash(alg string, d []byte) []byte {
...
+		}
+	}
+	return discard
+}
+
+// RemoveUnusedKeys determines which keys in the slice of IDs are no longer
+// used in the given targets file and removes them from the delegated keys
+// map
+func RemoveUnusedKeys(t *data.SignedTargets, ids []string) {
+	unusedIDs := UnusedDelegationKeys(*t, ids)
```
diogomonica: Why not pass the pointer?

tuf/tuf.go
endophage: Done!

client/changelist/change.go
```diff
@@ -42,6 +42,20 @@ type TufRootData struct {
 	RoleName string       `json:"role"`
 }
 
+// TufDelegationUpdate represents a modification to a target delegation
+// this includes creating a delegations. This format is used to avoid
+// unexpected race conditions between humans modifying the same delegation
+type TufDelegationUpdate struct {
+	NewName            string       `json:"new_name,omitempty"`
+	NewThreshold       int          `json:"threshold, omitempty"`
+	AddKeys            data.KeyList `json:"add_keys, omitempty"`
```
diogomonica: Why not a `[]string` like everything else?

client/helpers.go
```diff
@@ -55,6 +55,28 @@ func applyChangelist(repo *tuf.Repo, cl changelist.Changelist) error {
...
+		err = changeTargetDelegation(repo, c)
+	}
+	return err
+}
+
+func changeTargetDelegation(repo *tuf.Repo, c changelist.Change) error {
+	var err error
+	switch c.Action() {
+	case changelist.ActionCreate, changelist.ActionUpdate:
+
```
diogomonica: Intentionally empty?

tuf/utils/utils.go
diogomonica: Should we start being consistent then?

client/helpers.go
endophage: It's where the WIP is happening right now :-P

tuf/utils/utils.go
endophage: Yes, we should. I've been trying to be but I'm sure it's not perfect.

client/changelist/change.go
endophage: KeyList is a `[]PublicKey`, not a `[]string`, and it's explicitly defined so I can have a custom `UnmarshalJSON` that unmarshals keys into the correct types based on their algorithm.

endophage:
@diogomonica @cyli before it gets too big this should be a merge-able unit in its own right.

It's almost 900 lines of tests and only like 300 lines of functionality

tuf/data/roles.go
```diff
@@ -176,3 +189,75 @@ func (r Role) IsDelegation() bool {
...
+		kill[e] = true
+	}
+	write := 0
+	for _, e := range orig {
+		if !kill[e] {
+			orig[write] = e
+			write++
+		}
+	}
+	return orig[:write]
```
diogomonica: why keep a write counter instead of just creating a new `[]string` with items from `orig` that simply do not appear in `remove`?

diogomonica:
Great testing coverage. LGTM.

tuf/data/roles.go
endophage: >.< I was looking for set operations and came across this algorithm. Will switch out to just append to a new slice.

client/changelist/change.go
```diff
@@ -77,3 +77,26 @@ func (c TufChange) Path() string {
...
+	AddPathHashPrefixes    []string     `json:"add_prefixes,omitempty"`
+	RemovePathHashPrefixes []string     `json:"remove_prefixes,omitempty"`
+}
+
+// ToNewRole creates a fresh role object from the TufDelegation data
+func (td TufDelegation) ToNewRole(scope string) (*data.Role, error) {
+	name := scope
+	if td.NewName != "" {
+		name = td.NewName
+	}
```
cyli: Might be overly defensive, but should we raise an error if the scope is empty and the name is empty?

endophage: It'll fail on `GetDelegation` (https://github.com/docker/notary/pull/354/files#diff-56f9cc15696bff3d3adb8091efa7a8eeR76) before every reaching this point if the scope is empty.

endophage: (specifically it'll fail with `ErrInvalidRole` because the empty string doesn't represent a valid delegated role)

cyli: Ok, ould we add a comment to that effect (what the assumptions for `scope` and `NewName` are), and that applying changes/validation will blow up if the name is invalid?

client/helpers_test.go
```diff
@@ -147,3 +148,431 @@ func TestApplyChangelistMulti(t *testing.T) {
...
+
+	tgts := repo.Targets[data.CanonicalTargetsRole]
+	assert.Len(t, tgts.Signed.Delegations.Roles, 1)
+	assert.Len(t, tgts.Signed.Delegations.Keys, 1)
+
+	_, ok := tgts.Signed.Delegations.Keys[newKey.ID()]
+	assert.True(t, ok)
+
+	role := tgts.Signed.Delegations.Roles[0]
+	assert.Equal(t, newKey.ID(), role.KeyIDs[0])
```
cyli: Should this assert that there is just one key?

client/helpers.go
```diff
@@ -55,6 +55,59 @@ func applyChangelist(repo *tuf.Repo, cl changelist.Changelist) error {
...
+	case changelist.TypeTargetsDelegation:
+		return changeTargetsDelegation(repo, c)
+	default:
+		return fmt.Errorf("only target meta and delegations changes supported")
+	}
+}
+
+func changeTargetsDelegation(repo *tuf.Repo, c changelist.Change) error {
+	switch c.Action() {
+	case changelist.ActionCreate, changelist.ActionUpdate:
```
cyli: Should this be treated as just one action, since they are functionally the same?  Otherwise, would it make sense to error if there is no existing role on update?

client/helpers_test.go
```diff
@@ -147,3 +148,431 @@ func TestApplyChangelistMulti(t *testing.T) {
...
+	ch := changelist.NewTufChange(
+		changelist.ActionCreate,
+		"bad role",
+		changelist.TypeTargetsDelegation,
+		"",
+		tdJSON,
+	)
+
+	err = applyTargetsChange(repo, ch)
+	assert.Error(t, err)
```
cyli: Would it make sense to assert the `ErrInvalidRole` type here, so we don't conflate with some other error?

client/helpers_test.go
```diff
@@ -147,3 +148,431 @@ func TestApplyChangelistMulti(t *testing.T) {
...
+		changelist.TypeTargetsDelegation,
+		"",
+		tdJSON,
+	)
+
+	err = applyTargetsChange(repo, ch)
+	assert.NoError(t, err)
+
+	// add prefixes and update
+	td.AddPathHashPrefixes = []string{"abc"}
```
cyli: Could we add a docstring for this test stating that the conflict is that an error is returned if `PathHashPrefixes` is added to a delegation that already has a path prefix, because a role cannot have both?  Similarly for the one below?

It was not immediately obvious without digigng further in the `Role` code that adding a duplicate path is not an error.

tuf/data/roles.go
```diff
@@ -176,3 +189,74 @@ func (r Role) IsDelegation() bool {
...
+	return nil
+}
+
+// RemoveKeys removes the ids from the current list of key ids
+func (r *Role) RemoveKeys(ids []string) {
+	r.KeyIDs = subtractStrSlices(r.KeyIDs, ids)
+}
+
+// RemovePaths removes the paths from the current list of role paths
+func (r *Role) RemovePaths(paths []string) {
```
cyli: Should this and `RemovePathHashPrefixes` check to see if the opposite is 0?  [According to the spec](https://github.com/theupdateframework/tuf/blob/develop/docs/tuf-spec.txt#L729) one of these must be specified (although I guess `NewRole` allows the creation of a role with both as empty arrays).

endophage: The spec is written in a misleading way on that point. My understanding of
its meaning is that only one of paths, and path hash prefixes can be
specified, and having both specified is invalid. It is however perfectly
fine for neither to be specified, in which case the delegation is limited
by what its parent is capable of signing.

tuf/tuf.go
```diff
@@ -165,15 +184,14 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
 // will be added in front of (i.e. higher priority) in the delegation list.
 // An empty before string indicates to add the role to the end of the
 // delegation list.
-// A new, empty, targets file will be created for the new role.
 func (tr *Repo) UpdateDelegations(role *data.Role, keys []data.PublicKey, before string) error {
```
cyli: I think I may just be missing it, but is 'before' used anywhere?

endophage: No. It's mean to be a way to support the TUF specs idea about order of
roles indicating priority. I've yet to climb down off the fence on whether
I'm going to implement it or not (because I don't think we're going to use
it) so I keep leaving the argument there... :-/

client/helpers.go
endophage: Yeah, I wasn't sure. This was an effort to do the right thing even in the
face of incoming server changes that may have removed a role from under
you. If 2 admins are both making updates, one deletes a role while the
other is trying to modify it, I figure they're both making valid actions,
although it would be fair to assume the second action may be making some
assumptions about things like the threshold or name (I haven't actually
implemented renaming roles, I'm not planning to support it for friday...),
so possibly it does make sense to separate them.

tuf/data/roles.go
cyli: Hmm... can the path or path prefix be outside if the parent's path or path prefix?  If a parent specifies `/a` can the delegation specify `/b` as it's path?

tuf/tuf.go
cyli: If the role already existed, would it then ignore the `before` argument, or would the update also move the role?  Would a separate API to re-prioritize delegations all in one go make sense?

tuf/tuf.go
```diff
@@ -183,18 +201,21 @@ func (tr *Repo) UpdateDelegations(role *data.Role, keys []data.PublicKey, before
...
-		p.Signed.Delegations.Roles[i] = role
+
+	foundAt := utils.FindRoleIndex(p.Signed.Delegations.Roles, role.Name)
+
+	if foundAt >= 0 {
+		p.Signed.Delegations.Roles[foundAt] = role
 	} else {
 		p.Signed.Delegations.Roles = append(p.Signed.Delegations.Roles, role)
 	}
+	// We've made a change to parent. Set it to dirty
```
cyli: I think I'm just fuzzy on how this works - does the delegation role need to be marked as dirty as well?

tuf/tuf.go
```diff
@@ -202,6 +223,47 @@ func (tr *Repo) UpdateDelegations(role *data.Role, keys []data.PublicKey, before
...
 
+	utils.RemoveUnusedKeys(p)
+
+	return nil
+}
+
+// DeleteDelegation removes a delegated targets role from its parent
+// targets object. It also deletes the delegation from the snapshot.
+// DeleteDelegation will only make use of the role Name field.
+func (tr *Repo) DeleteDelegation(role data.Role) error {
```
cyli: Should this also remove any child roles?  Or error maybe if a child role is then orphaned?

tuf/data/roles.go
endophage: No. Technically one could specify that, but when searching for a path, one
would never reach the delegation because the parent would never be
traversed.

tuf/tuf.go
endophage: I'm not even convinced it's a good way to do delegations, hence why I
haven't implemented it...

tuf/data/roles.go
cyli: Ah, ok - should there be a check for it, since if specified, the users would shoot themselves in the foot?  (not necessarily in this PR)

tuf/tuf.go
endophage: Technically when the delegation is created, only the parent is modified.
The parent contains all the information about the delegation.

tuf/tuf.go
endophage: Removing a delegation is essentially orphaning it by removing it from the
parent, such that it won't be visible when traversing the repository via
the top level targets file. That naturally cuts off all the children of the
delegation too.

tuf/tuf.go
cyli: Ah, so any change to one delegation means all the delegations are updated?

tuf/data/roles.go
endophage: We should document it somewhere. Although we're unlikely to expose path
hash prefixes any time soon. They come with some significant caveats that
in the best case are confusing, and in the worst, create exploits.

tuf/tuf.go
endophage: The list of delegations in the parent is updated certainly. The individual
delegations within the list are only modified if the name matches the
delegation we're looking for.

tuf/tuf.go
cyli: Ah right, sorry, I was conflating updating the repo struct with deleting the metadata files.

tuf/tuf.go
```diff
@@ -202,6 +223,47 @@ func (tr *Repo) UpdateDelegations(role *data.Role, keys []data.PublicKey, before
...
+	if !role.IsDelegation() {
+		return data.ErrInvalidRole{Role: role.Name, Reason: "not a valid delegated role"}
+	}
+	// the role variable must not be used past this assignment for safety
+	name := role.Name
+
+	parent := filepath.Dir(name)
+	p, ok := tr.Targets[parent]
+	if !ok {
+		return data.ErrInvalidRole{Role: name, Reason: "parent role not found"}
```
cyli: Non-blocking question:  any difference between the parent not being found (which errors), and the role itself not being found (which does not)?  In both case, a role of that name will not exist.

tuf/tuf.go
cyli: Sorry :)  By "update" I meant everything is re-signed and sent to the server, in the client's case?

cyli:
Since I think this current code supports it - can we add tests for a 2-level deep delegation?  (Add delegation, add sub-delegation)?  Alternately, can we explicitly add a check that errors if trying to add a delegation whose parent is not target?  I'm just afraid that if we don't explicitly reject it or support it, it might happen by accident in an integration, and that use case would be untested.

tuf/data/roles.go
cyli: Should we explicitly de-support them then, if they are dangerous and unlikely to ever be supported?  (like maybe not serialize/deserialize them them in JSON roles?)

cyli: E.g. if they are serialized, and we store this information, then we every add support for path validation/pushing, a repo may already be in a bad/exploitable state if our API accepts and stores the path/path-prefix info.

endophage: Paths are fine and safe, mainly because they're easy to understand. Hash
prefixes have a strong use case, but require a significant level of
understanding to be used safely. We can add sufficient validation to make
them safe, it's just a lot of work and so far, we have no need for them.

client/changelist/change.go
endophage: Actually, this will fail on the NewRole call at the end. That checks the name is valid. Technically you could pass NewRole one of the non-delegation valid role names (like "root") and that would then be caught as invalid in UpdateDelegations

cyli: Ah, ok, thanks!

tuf/tuf.go
endophage: Yes, the targets file containing the list of delegations will get resigned in its entirety.

tuf/tuf.go
endophage: It's a very esoteric difference. For the purposes of delete I could probably just return nil in the case the parent doesn't exist. The error though will help us behave correctly in the case of failures where the parent role hasn't been loaded (via logic errors or otherwise).

endophage:
@cyli will add a test for 2 deep delegation!

client/helpers.go
```diff
@@ -55,6 +55,66 @@ func applyChangelist(repo *tuf.Repo, cl changelist.Changelist) error {
...
+		err := json.Unmarshal(c.Content(), &td)
+		if err != nil {
+			return err
+		}
+		r, err := repo.GetDelegation(c.Scope())
+		if _, ok := err.(data.ErrNoSuchRole); err != nil && !ok {
+			// error that wasn't ErrNoSuchRole
+			return err
+		}
+		// role doesn't exist, create brand new
```
cyli: Should there be an `if err == nil { return fmt.Errorf(...) }` before this line?  Otherwise, it seems like if the repo exists, this will just overwrite the existing one?

tuf/tuf.go
cyli: In that case should we just leave it out for now?  Alternately, document the parameter as unused and ignored for now?

client/helpers_test.go
```diff
@@ -147,3 +148,516 @@ func TestApplyChangelistMulti(t *testing.T) {
...
+	)
+
+	err = applyTargetsChange(repo, ch)
+	assert.NoError(t, err)
+
+	assert.Len(t, tgts.Signed.Delegations.Roles, 0)
+	assert.Len(t, tgts.Signed.Delegations.Keys, 0)
+}
+
+func TestApplyTargetsDelegationCreate2SharedKey(t *testing.T) {
```
cyli: Awesome test!

client/helpers.go
cyli: Could we also add a test for creating a delegation that already exists?  Whether the behavior is supposed to be overwriting the existing role or if the behavior is supposed to be to return an error.

client/helpers.go
```diff
@@ -55,6 +55,66 @@ func applyChangelist(repo *tuf.Repo, cl changelist.Changelist) error {
...
+			return err
+		}
+		return repo.UpdateDelegations(r, td.AddKeys, "")
+	case changelist.ActionUpdate:
+		td := changelist.TufDelegation{}
+		err := json.Unmarshal(c.Content(), &td)
+		if err != nil {
+			return err
+		}
+		r, err := repo.GetDelegation(c.Scope())
```
cyli: Could we add a test for updating a role that doesn't exist?

client/helpers_test.go
endophage: Done. Added a comment above the first of the 2 conflict functions describing why they exist.

tuf/tuf.go
endophage: Yeah, I'll take it out for now

client/helpers.go
endophage: ahhh good catch!

client/helpers.go
endophage: Done: `TestApplyTargetsDelegationEditNonExisting`

tuf/tuf.go
```diff
@@ -165,15 +184,14 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
 // will be added in front of (i.e. higher priority) in the delegation list.
 // An empty before string indicates to add the role to the end of the
```
cyli: Non-blocking nitpick:  This comment no longer applies.

cyli:
Pending CI +1, LGTM :)  Awesome job on test coverage!

tuf/tuf.go
endophage: No point in leaving it hanging around. Removed. Will merge after CI completes

PR #359, opened 2015-12-12
Adding a volume to MySQL container to ensure data persistence

diogomonica:
In case someone is deploying in production directly by doing `docker-compose up` from notary, we might as well make sure they don't lose data by accident.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,6 +20,8 @@ notarysigner:
    - notarymysql
   command: -config=fixtures/signer-config.json
 notarymysql:
+  volumes:
+    - notarymysql:/var/lib/mysql
   build: ./notarymysql/
   ports:
     - "3306:3306"


docker-compose.yml
```diff
@@ -20,6 +20,8 @@ notarysigner:
    - notarymysql
   command: -config=fixtures/signer-config.json
 notarymysql:
+  volumes:
+    - notarymysql:/var/lib/mysql
```
cyli: Non-blocking - should this be an absolute path?

cyli:
Other than the above LGTM!

docker-compose.yml
vikstrous: compose resolves the path for you

cyli: Ah ok, thanks!

thaJeztah: careful with that, because that may change in future (because of "named" volumes). If you want to create a bind-mounted volume, using `./notarymysql/:/var/lib/mysql/` may be safer. Compose will translate `./..` to relative to the `docker-compose.yml`

thaJeztah: Actually, this may already be the case; https://github.com/docker/compose/blob/master/docs/compose-file.md#volumes-volume_driver. Is the intent here to create a "named" volume, or a bind-mount?

cyli: A bind-mount.  Thanks @thaJeztah!  https://github.com/docker/notary/pull/362

PR #362, opened 2015-12-14
Use a .-relative bind-mounted volume in notarymysql compose section.

cyli:
See https://github.com/docker/notary/pull/359#discussion_r47570016

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,7 +21,7 @@ notarysigner:
   command: -config=fixtures/signer-config.json
 notarymysql:
   volumes:
-    - notarymysql:/var/lib/mysql
+    - ./notarymysql:/var/lib/mysql
   build: ./notarymysql/
   ports:
     - "3306:3306"


thaJeztah:
thanks! lgtm (not a maintainer here)

diogomonica:
The idea was to use a named volume originally, but local bind might be easier to do development on, and still makes sure data isn't lost between invocations. LGTM

thaJeztah:
If a named volume was the intent, then I think the original PR is good (and avoids possible "permissions" issues when using boot2docker). Up to you :smile:

diogomonica:
using `./notarymysql:/var/lib/mysql` doesn't work for me locally.

thaJeztah:
because of permissions probably? In that case, it's probably best to close this PR, and use a named volume as in the original PR. Sorry for the confusion, I thought the intent was using a bind-mounted volume

cyli:
@thaJeztah Same.  Thanks for testing @diogomonica - happy just to go with named volume then.

PR #368, opened 2015-12-15
download all delegated roles when doing downloadTargets

endophage:

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -440,8 +440,13 @@ func (r *NotaryRepository) ListTargets(roles ...string) ([]*Target, error) {
 	return targetList, nil
 }
 
-// GetTargetByName returns a target given a name
-func (r *NotaryRepository) GetTargetByName(name string) (*Target, error) {
+// GetTargetByName returns a target given a name. If no roles are passed
+// it uses the targets role and does a search of the entire delegation
+// graph, finding the first entry in a breadth first search of the delegations.
+// If roles are passed, they should be passed in ascending priority and
+// the target entry found in the subtree of the highest priority role
+// will be returned
+func (r *NotaryRepository) GetTargetByName(name string, roles ...string) (*Target, error) {
 	c, err := r.bootstrapClient()
 	if err != nil {
 		return nil, err
@@ -455,11 +460,25 @@ func (r *NotaryRepository) GetTargetByName(name string) (*Target, error) {
 		return nil, err
 	}
 
-	meta, err := c.TargetMeta(name)
+	if len(roles) == 0 {
+		roles = append(roles, data.CanonicalTargetsRole)
+	}
+	var (
+		meta *data.FileMeta
+	)
+	for i := len(roles) - 1; i >= 0; i-- {
+		meta, err = c.TargetMeta(roles[i], name)
+		if err != nil {
+			// important to error here otherwise there might be malicious
+			// behaviour that prevents a legitimate version of a target
+			// being found
+			return nil, err
+		} else if meta != nil {
+			break
+		}
+	}
 	if meta == nil {
 		return nil, fmt.Errorf("No trust data for %s", name)
-	} else if err != nil {
-		return nil, err
 	}
 
 	return &Target{Name: name, Hashes: meta.Hashes, Length: meta.Length}, nil
--- a/client/client.go
+++ b/client/client.go
@@ -403,8 +403,9 @@ func (r *NotaryRepository) RemoveTarget(targetName string, roles ...string) erro
 	return addChange(cl, template, roles...)
 }
 
-// ListTargets lists all targets for the current repository
-func (r *NotaryRepository) ListTargets() ([]*Target, error) {
+// ListTargets lists all targets for the current repository. The list of
+// roles should be passed in order from lowest to highest priority
+func (r *NotaryRepository) ListTargets(roles ...string) ([]*Target, error) {
 	c, err := r.bootstrapClient()
 	if err != nil {
 		return nil, err
@@ -419,9 +420,16 @@ func (r *NotaryRepository) ListTargets() ([]*Target, error) {
 	}
 
 	var targetList []*Target
-	for name, meta := range r.tufRepo.Targets["targets"].Signed.Targets {
-		target := &Target{Name: name, Hashes: meta.Hashes, Length: meta.Length}
-		targetList = append(targetList, target)
+	for _, r := range roles {
+		tgts, ok := r.tufRepo.Targets[r]
+		if !ok {
+			// not every role has to exist
+			continue
+		}
+		for name, meta := range r.tufRepo.Targets[r].Signed.Targets {
+			target := &Target{Name: name, Hashes: meta.Hashes, Length: meta.Length}
+			targetList = append(targetList, target)
+		}
 	}
 
 	return targetList, nil
--- a/client/client.go
+++ b/client/client.go
@@ -420,13 +420,13 @@ func (r *NotaryRepository) ListTargets(roles ...string) ([]*Target, error) {
 	}
 
 	var targetList []*Target
-	for _, r := range roles {
-		tgts, ok := r.tufRepo.Targets[r]
+	for _, role := range roles {
+		tgts, ok := r.tufRepo.Targets[role]
 		if !ok {
 			// not every role has to exist
 			continue
 		}
-		for name, meta := range r.tufRepo.Targets[r].Signed.Targets {
+		for name, meta := range tgts.Signed.Targets {
 			target := &Target{Name: name, Hashes: meta.Hashes, Length: meta.Length}
 			targetList = append(targetList, target)
 		}
--- a/client/client.go
+++ b/client/client.go
@@ -405,6 +405,11 @@ func (r *NotaryRepository) RemoveTarget(targetName string, roles ...string) erro
 
 // ListTargets lists all targets for the current repository. The list of
 // roles should be passed in order from highest to lowest priority.
+// IMPORTANT: if you pass a set of roles such as [ "targets/a", "targets/x"
+// "targets/a/b" ], even though "targets/a/b" is part of the "targets/a" subtree
+// its entries will be strictly shadowed by those in other parts of the "targets/a"
+// subtree and also the "targets/x" subtree, as we will defer parsing it until
+// we explicitly reach it in our iteration of the provided list of roles.
 func (r *NotaryRepository) ListTargets(roles ...string) ([]*Target, error) {
 	c, err := r.bootstrapClient()
 	if err != nil {
@@ -471,6 +476,7 @@ func (r *NotaryRepository) listSubtree(targets map[string]*Target, role string, 
 // If roles are passed, they should be passed in descending priority and
 // the target entry found in the subtree of the highest priority role
 // will be returned
+// See the IMPORTANT section on ListTargets above. Those roles also apply here.
 func (r *NotaryRepository) GetTargetByName(name string, roles ...string) (*Target, error) {
 	c, err := r.bootstrapClient()
 	if err != nil {
--- a/client/client.go
+++ b/client/client.go
@@ -404,7 +404,7 @@ func (r *NotaryRepository) RemoveTarget(targetName string, roles ...string) erro
 }
 
 // ListTargets lists all targets for the current repository. The list of
-// roles should be passed in order from lowest to highest priority.
+// roles should be passed in order from highest to lowest priority.
 func (r *NotaryRepository) ListTargets(roles ...string) ([]*Target, error) {
 	c, err := r.bootstrapClient()
 	if err != nil {
@@ -453,8 +453,10 @@ func (r *NotaryRepository) listSubtree(targets map[string]*Target, role string, 
 			continue
 		}
 		for name, meta := range tgts.Signed.Targets {
-			target := &Target{Name: name, Hashes: meta.Hashes, Length: meta.Length}
-			targets[name] = target
+			if _, ok := targets[name]; !ok {
+				target := &Target{Name: name, Hashes: meta.Hashes, Length: meta.Length}
+				targets[name] = target
+			}
 		}
 		for _, d := range tgts.Signed.Delegations.Roles {
 			if !excl[d.Name] {
@@ -467,7 +469,7 @@ func (r *NotaryRepository) listSubtree(targets map[string]*Target, role string, 
 // GetTargetByName returns a target given a name. If no roles are passed
 // it uses the targets role and does a search of the entire delegation
 // graph, finding the first entry in a breadth first search of the delegations.
-// If roles are passed, they should be passed in ascending priority and
+// If roles are passed, they should be passed in descending priority and
 // the target entry found in the subtree of the highest priority role
 // will be returned
 func (r *NotaryRepository) GetTargetByName(name string, roles ...string) (*Target, error) {
@@ -490,8 +492,8 @@ func (r *NotaryRepository) GetTargetByName(name string, roles ...string) (*Targe
 	var (
 		meta *data.FileMeta
 	)
-	for i := len(roles) - 1; i >= 0; i-- {
-		meta = c.TargetMeta(roles[i], name, roles...)
+	for _, role := range roles {
+		meta = c.TargetMeta(role, name, roles...)
 		if meta != nil {
 			return &Target{Name: name, Hashes: meta.Hashes, Length: meta.Length}, nil
 		}
--- a/client/client.go
+++ b/client/client.go
@@ -404,7 +404,7 @@ func (r *NotaryRepository) RemoveTarget(targetName string, roles ...string) erro
 }
 
 // ListTargets lists all targets for the current repository. The list of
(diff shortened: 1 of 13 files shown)


client/client.go
```diff
@@ -303,8 +303,9 @@ func (r *NotaryRepository) RemoveTarget(targetName string) error {
 	return nil
 }
 
-// ListTargets lists all targets for the current repository
-func (r *NotaryRepository) ListTargets() ([]*Target, error) {
+// ListTargets lists all targets for the current repository. The list of
+// roles should be passed in order from lowest to highest priority
+func (r *NotaryRepository) ListTargets(roles ...string) ([]*Target, error) {
```
diogomonica: Having an API that is order dependent is weird. What is the goal here?

client/client.go
```diff
@@ -303,8 +303,9 @@ func (r *NotaryRepository) RemoveTarget(targetName string) error {
 	return nil
 }
 
-// ListTargets lists all targets for the current repository
-func (r *NotaryRepository) ListTargets() ([]*Target, error) {
```
diogomonica: If a target gets signed with the same name under different roles (and different hashes), how does the caller do conflict resolution?

The name is something like 'latest', 'edge' or '2.6', so there is nothing that distinguishes which role this particular target came from after calling ListTargets. And since this is a list, the outcome will simply be an array with several 'latest' targets in it.

client/client.go
diogomonica: After reading the method I don't understand the comment around ordering.

endophage: Order dependent on arguments is weird, but plenty of things are order dependent on values. I could have written this as `ListTargets(roles []string)` but if we want to support not having to pass roles, which we well might, this format is easier for passing no roles (i.e. `ListTargets()` vs `ListTargets(nil)`).

TUF's spec talks about using the ordering of delegations to determine priority. We probably don't want to do exactly that as we want to manipulate priority across multiple levels (like putting "targets/releases" as a high priority than just "targets"). I've just realized I didn't update the function quite right, but the intent is to load the highest priority metadata last, having it overwrite the lower priority data.

Updating now

tuf/client/client.go
```diff
@@ -369,34 +369,47 @@ func (c *Client) downloadSnapshot() error {
...
-		return err
-	}
+	stack := utils.NewStack()
+	stack.Push(role)
+	for !stack.Empty() {
+		role, err := stack.PopString()
+		if err != nil {
+			return err
+		}
+		role = data.RoleName(role) // this will really only do something for base targets role
```
diogomonica: aren't the next 30 lines exactly what `downloadTargets` does?

client/client.go
diogomonica: @endophage not sure why this says it's on an outdated diff, but I see no changes.

endophage: Have another look. Both the prioritization and returned data should make more sense now.

tuf/client/client.go
endophage: This is `downloadTargets`... What are you asking?

client/client.go
endophage: I believe it's because you keep commenting on the deleted code rather than the added code.

tuf/client/client.go
diogomonica: I was working on a different version of the code, ignore that question.

Where are we calling `downloadTargets` now? We deleted from TargetMeta, but I don't see any new calls for `downloadTargets`.

endophage: It's called once as part of `update` and downloads the whole graph of delegations. This was actually the original algorithm and I improved on it because when I was testing against the PyPI TUF sample it was soooo huge it took ages to download. We'll work out how to optimize later, but for now this is less efficient but easier to work with in places like `NotaryRepository.Publish`

client/client.go
cyli: Non-blocking question - is there a case where we want to pass no roles?  Maybe just to trigger an update from the server?

cyli:
Would be lovely to have some tests for `ListTargets` to make sure that targets in the higher priority roles displayed instead of the ones in the lower priority roles with the same name.

endophage:
@cyli yep, working on those now.

client/client_test.go
```diff
@@ -918,6 +930,77 @@ func testListTarget(t *testing.T, rootType string) {
 	assert.Equal(t, currentTarget, newCurrentTarget, "current target does not match")
 }
 
+func testListTargetWithDelegates(t *testing.T, rootType string) {
```
cyli: I might just be missing it, but I don't see where this function is called.  Does it need to be added to `TestListTarget`? (or a new test case?)

cyli: Actually, apologies - I'm wondering if we need to test both the root being ECDSA and RSA for all these cases.  Maybe this would be fine just being its own test case with a capitalized T?  Again, only if I haven't just blindly missed the function call

endophage: It uses the type for the delegation key so it's probably worth doing it with both key types.

client/client.go
```diff
@@ -374,11 +392,25 @@ func (r *NotaryRepository) GetTargetByName(name string) (*Target, error) {
...
+	)
+	for i := len(roles) - 1; i >= 0; i-- {
+		meta, err = c.TargetMeta(roles[i], name)
+		if err != nil {
+			// important to error here otherwise there might be malicious
+			// behaviour that prevents a legitimate version of a target
+			// being found
+			return nil, err
+		} else if meta != nil {
+			break
```
cyli: Non-blocking: it might be slightly easier to read if this just returned the target instead of breaking.  Then we wouldn't have to test for metadata being nil after the loop.

client/client_root_validation_test.go
```diff
@@ -17,7 +17,7 @@ var passphraseRetriever = func(string, string, bool, int) (string, bool, error)
 // sure the repository looks correct on disk.
 // We test this with both an RSA and ECDSA root key
 func TestValidateRoot(t *testing.T) {
-	logrus.SetLevel(logrus.DebugLevel)
+	logrus.SetLevel(logrus.ErrorLevel)
```
cyli: heh, yes, thank you!

tuf/client/client.go
```diff
@@ -505,9 +518,9 @@ func (c Client) RoleTargetsPath(role string, hashSha256 string, consistent bool)
 	return role, nil
 }
 
-// TargetMeta ensures the repo is up to date, downloading the minimum
-// necessary metadata files
-func (c Client) TargetMeta(path string) (*data.FileMeta, error) {
+// TargetMeta ensures the repo is up to date. It assumes downloadTargets
+// has already downloaded all delegated roles
+func (c Client) TargetMeta(role, path string) (*data.FileMeta, error) {
 	c.Update()
```
cyli: I know you didn't change this line, but I think the only thing that calls this might be `GetTargetName`, which does a `c.Update` right before this.  Is it necessary to update again before getting the metadata for every role?  If so, should we check for an error at least and return it?

client/client_test.go
```diff
@@ -918,6 +936,81 @@ func testListTarget(t *testing.T, rootType string) {
...
+	assert.True(t, ok)
+	_, ok = repo.tufRepo.Targets["targets/level1"].Signed.Targets["other"]
+	assert.True(t, ok)
+
+	fakeServerData(t, repo, mux, keys)
+
+	targets, err := repo.ListTargets(data.CanonicalTargetsRole, "targets/level1")
+	assert.NoError(t, err)
+
+	// Should be two targets
```
cyli: "three targets"  :)  Although the assert is pretty clear, so this comment may just be superfluous.

tuf/utils/stack_test.go
```diff
@@ -0,0 +1,69 @@
...
+	assert.Len(t, s.s, 0)
+	assert.Equal(t, "foo", i)
+}
+
+func TestPopStringWrongType(t *testing.T) {
+	s := NewStack()
+	s.Push(123)
+	_, err := s.PopString()
+	assert.Error(t, err)
+	assert.IsType(t, ErrBadTypeCast{}, err)
```
cyli: Should we add an assert that it was not actually popped and that the length of the stack is still 1?

server/server.go
```diff
@@ -89,7 +89,7 @@ func RootHandler(ac auth.AccessController, ctx context.Context, trust signed.Cry
 		prometheus.InstrumentHandlerWithOpts(
 			prometheusOpts("UpdateTuf"),
 			hand(handlers.AtomicUpdateHandler, "push", "pull")))
-	r.Methods("GET").Path("/v2/{imageName:.*}/_trust/tuf/{tufRole:(root|targets|snapshot|timestamp)}.json").Handler(
+	r.Methods("GET").Path("/v2/{imageName:.*}/_trust/tuf/{tufRole:(root|targets(?:/.+)?|snapshot|timestamp)}.json").Handler(
```
cyli: As per slack conversation, should this be something like `(?:/[^/\s]+)?` instead?  (+ tests :D)

tuf/client/client.go
```diff
@@ -369,34 +369,47 @@ func (c *Client) downloadSnapshot() error {
...
-	}
+	stack := utils.NewStack()
+	stack.Push(role)
+	for !stack.Empty() {
+		role, err := stack.PopString()
+		if err != nil {
+			return err
+		}
+		role = data.RoleName(role) // this will really only do something for base targets role
+		if c.local.Snapshot == nil {
```
cyli: Any reason not to check root too?

tuf/client/client.go
```diff
@@ -369,34 +369,47 @@ func (c *Client) downloadSnapshot() error {
...
+			return ErrMissingMeta{role: role}
```
cyli: Should this missing role be `data.CanonicalSnapshotRole`?

tuf/client/client_test.go
```diff
@@ -330,6 +330,49 @@ func TestDownloadTargetsHappy(t *testing.T) {
 	assert.NoError(t, err)
 }
 
+func TestDownloadTargetsDeepHappy(t *testing.T) {
```
cyli: Wondering if we can make this a deeper and wider tree?  (maybe targets, targets/level1, targets/level2, targets/level1/level3, targets/level1/level4, to test (1) that it does all the children delegations for each role that it loops through, and (2) that it can do more than 1 level?)

client/client_test.go
cyli: Ah ok, thanks!

server/server.go
cyli: Also, we mentioned blacklisting `..` in `IsDelegation` (which could be used to check the role), but I'm not sure if `..` does something wonky with URLs and mux routing, so just wanted to mention it here too.

tuf/client/client_test.go
endophage: Done!

tuf/utils/stack_test.go
endophage: done

tuf/client/client.go
```diff
@@ -369,34 +369,52 @@ func (c *Client) downloadSnapshot() error {
...
+		snap := c.local.Snapshot.Signed
+		root := c.local.Root.Signed
+		r := c.keysDB.GetRole(role)
+		if r == nil {
+			return fmt.Errorf("Invalid role: %s", role)
+		}
+		keyIDs := r.KeyIDs
+		s, err := c.getTargetsFile(role, keyIDs, snap.Meta, root.ConsistentSnapshot, r.Threshold)
+		if err != nil {
+			if _, ok := err.(store.ErrMetaNotFound); ok {
```
cyli: `c.GetTargetsFile` seems to return `ErrMissingMeta` - I'm not sure if it also returns `ErrMetaNotFound`, but if we are not publishing delegations unnecessarily, then we also don't have any snapshot hashes for them, which results in a `ErrMissingMeta`

cyli: Also, we probably don't want to continue if it's the actual targets file

cyli: I made some minor changes here:  https://github.com/docker/notary/commit/74cbbdcc7f129175d11dbb5d8ca6a7a12d3a0f04, but without tests for it.

With that change my PR seems to pass tests.

client/client.go
```diff
@@ -418,17 +419,59 @@ func (r *NotaryRepository) ListTargets() ([]*Target, error) {
...
 
+	if len(roles) == 0 {
+		roles = []string{data.CanonicalTargetsRole}
+	}
+	targets := make(map[string]*Target)
+	for _, role := range roles {
+		// we don't need to do anything special with removing role from
+		// roles because listSubtree always processes role and only excludes
+		// descendent delegations that appear in roles.
+		r.listSubtree(targets, role, roles...)
```
cyli: Just want to verify this behavior -  if I have (an admittedly stupid) priority list like: `"targets/a", "targets/x", "targets/a/b", this would list *only* the parts of `targets/a` that are not in `targets/a/b` first.  So that targets in `targets/x` would have priority over in targets in `targets/a/b` even though `targets/a` is listed first.

Is that correct?  Or doe the whole `targets/a` subtree come first?  If so, we should probably exclude only the previous roles.

I'm fine with this either way, and the current way makes sense, but what happens if a subtree of a higher priority role is given lower priority than some other role is kind of confusing.  It might be good to add a test for it and clearer docstrings.

client/client.go
```diff
@@ -418,17 +419,59 @@ func (r *NotaryRepository) ListTargets() ([]*Target, error) {
 		return nil, err
 	}
 
+	if len(roles) == 0 {
+		roles = []string{data.CanonicalTargetsRole}
```
cyli: Would be awesome to have test coverage for the default case :D

tuf/client/client.go
endophage: Ah, I think you're right. ErrMissingMeta is probably the appropriate error to check here.

client/client.go
endophage: That is correct. I also think that behaviour is correct. By passing that list of roles, you're explicitly saying you want `targets/a/b` to be treated as lower priority than `targets/x`, even though the rest of the `targets/a` subtree is treated as higher priority than `targets/x`

server/server.go
```diff
@@ -89,7 +89,7 @@ func RootHandler(ac auth.AccessController, ctx context.Context, trust signed.Cry
 		prometheus.InstrumentHandlerWithOpts(
 			prometheusOpts("UpdateTuf"),
 			hand(handlers.AtomicUpdateHandler, "push", "pull")))
-	r.Methods("GET").Path("/v2/{imageName:.*}/_trust/tuf/{tufRole:(root|targets|snapshot|timestamp)}.json").Handler(
+	r.Methods("GET").Path("/v2/{imageName:.*}/_trust/tuf/{tufRole:(root|targets(?:/[^/\\s]+)*|snapshot|timestamp)}.json").Handler(
```
cyli: Non-blocking - we can always make this change later - do we want to include any of the regex stuff from @riyazdf's awesome `IsDelegation` fix?

client/client.go
endophage: Can do!

server/server.go
endophage: hmmm, probably not. Making it too explicit here will result in 404s. It would be better to validate later and return detailed error messages on why a role name is rejected (and do the same for upload).

endophage: And yes, we would still return a 404, but it would be good to have an informative message in the body

endophage: Or actually, maybe in the case of a truly invalid role name, we'd want to return a 400

tuf/client/client.go
cyli: I'm fine with merging this as is, and I can fix it in my PR (I'll add some tests for it), so this is non-blocking.

cyli:
LGTM pending CI!

tuf/client/client.go
cyli: (as per discussion, we were going to look at restructuring this code and fixing the checks in a different PR)

client/client_test.go
```diff
@@ -826,6 +845,20 @@ func fakeServerData(t *testing.T, repo *NotaryRepository, mux *http.ServeMux,
...
+	mux.HandleFunc("/v2/docker.com/notary/_trust/tuf/targets/level1.json",
+		func(w http.ResponseWriter, r *http.Request) {
+			level1JSON, err := json.Marshal(signedLevel1)
+			assert.NoError(t, err)
+			fmt.Fprint(w, string(level1JSON))
+		})
+
+	mux.HandleFunc("/v2/docker.com/notary/_trust/tuf/targets/level2.json",
+		func(w http.ResponseWriter, r *http.Request) {
+			level1JSON, err := json.Marshal(signedLevel2)
```
riyazdf: nit: `level2JSON`?  This is non-blocking since it shouldn't matter with the function scope though, but it might be nice for clarity.

tuf/client/client.go
endophage: Changed to ErrMissingMeta. In the general case it's fine for the targets file to be missing. I put it in the commit message but that would just indicate an empty repo with lazy targets file generation.

riyazdf:
LGTM after CI!

PR #370, opened 2015-12-16
server side delegations support in validation and snapshot generation

endophage:

### DIFF
--- a/server/handlers/default.go
+++ b/server/handlers/default.go
@@ -128,7 +128,7 @@ func getHandler(ctx context.Context, w http.ResponseWriter, r *http.Request, var
 
 	out, err := store.GetCurrent(gun, tufRole)
 	if err != nil {
-		if _, ok := err.(*storage.ErrNotFound); ok {
+		if _, ok := err.(storage.ErrNotFound); ok {
 			logrus.Error(gun + ":" + tufRole)
 			return errors.ErrMetadataNotFound.WithDetail(nil)
 		}
@@ -174,7 +174,7 @@ func getTimestamp(ctx context.Context, w http.ResponseWriter, logger ctxu.Logger
 	out, err := timestamp.GetOrCreateTimestamp(gun, store, cryptoService)
 	if err != nil {
 		switch err.(type) {
-		case *storage.ErrNoKey, *storage.ErrNotFound:
+		case *storage.ErrNoKey, storage.ErrNotFound:
 			logger.Error("404 GET timestamp")
 			return errors.ErrMetadataNotFound.WithDetail(nil)
 		default:
@@ -199,7 +199,7 @@ func getSnapshot(ctx context.Context, w http.ResponseWriter, logger ctxu.Logger,
 	out, err := snapshot.GetOrCreateSnapshot(gun, store, cryptoService)
 	if err != nil {
 		switch err.(type) {
-		case *storage.ErrNoKey, *storage.ErrNotFound:
+		case *storage.ErrNoKey, storage.ErrNotFound:
 			logger.Error("404 GET snapshot")
 			return errors.ErrMetadataNotFound.WithDetail(nil)
 		default:
--- a/server/handlers/validation.go
+++ b/server/handlers/validation.go
@@ -79,7 +79,6 @@ func validateUpdate(cs signed.CryptoService, gun string, updates []storage.MetaU
 		}
 	}
 
-	// TODO: validate delegated targets roles.
 	targetsToUpdate, err := loadAndValidateTargets(gun, repo, roles, kdb, store)
 	if err != nil {
 		return nil, err
--- a/server/handlers/validation.go
+++ b/server/handlers/validation.go
@@ -37,7 +37,6 @@ func validateUpdate(cs signed.CryptoService, gun string, updates []storage.MetaU
 	// updates with only the things we should actually update
 	updatesToApply := make([]storage.MetaUpdate, 0, len(updates))
 
-	// check that the necessary roles are present:
 	roles := make(map[string]storage.MetaUpdate)
 	for _, v := range updates {
 		roles[v.Role] = v
@@ -163,6 +162,12 @@ func loadAndValidateTargets(gun string, repo *tuf.Repo, roles map[string]storage
 			err error
 		)
 		if t, err = validateTargets(role, roles, kdb); err != nil {
+			if err == signed.ErrUnknownRole {
+				// role wasn't found in its parent. It has been removed
+				// or never existed. Drop this role from the update
+				// (by not adding it to updatesToApply)
+				continue
+			}
 			logrus.Error("ErrBadTargets: ", err.Error())
 			return nil, validation.ErrBadTargets{Msg: err.Error()}
 		}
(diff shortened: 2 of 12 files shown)


tuf/utils/role_sort.go
```diff
@@ -0,0 +1,28 @@
...
+
+// RoleList is a list of roles
+type RoleList []string
+
+// Len returns the length of the list
+func (r RoleList) Len() int {
+	return len(r)
+}
+
+// Less returns true if the item at i should be sorted
```
diogomonica: It doesn't hurt if we make this also alphabetically ordered within the same length bucket. The potential to have non-reproducible bugs because the final ordering is non-deterministic should overweight whatever CPU penalty we take for the alphabetic ordering operations.

endophage:
@diogomonica @cyli @riyazdf no longer WIP, PTAL.

server/handlers/default.go
```diff
@@ -199,7 +199,7 @@ func getSnapshot(ctx context.Context, w http.ResponseWriter, logger ctxu.Logger,
 	out, err := snapshot.GetOrCreateSnapshot(gun, store, cryptoService)
 	if err != nil {
 		switch err.(type) {
-		case *storage.ErrNoKey, *storage.ErrNotFound:
+		case *storage.ErrNoKey, storage.ErrNotFound:
```
diogomonica: Whats the story behind one of storage errors being a * and the other not?

server/handlers/validation.go
```diff
@@ -127,9 +126,67 @@ func validateUpdate(cs signed.CryptoService, gun string, updates []storage.MetaU
...
+
+func loadAndValidateTargets(gun string, repo *tuf.Repo, roles map[string]storage.MetaUpdate, kdb *keys.KeyDB, store storage.MetaStore) ([]storage.MetaUpdate, error) {
+	targetsRoles := make(utils.RoleList, 0)
+	for role := range roles {
+		if role == data.RoleName(data.CanonicalTargetsRole) || data.IsDelegation(role) {
+			targetsRoles = append(targetsRoles, role)
+		}
+	}
+
+	// by sorting, we'll always process shallower targets updates before deeper
```
diogomonica: should be: by sorting by segment, we'll...

Doesn't make sense when the first thought of the reader is "alphabetic sorting"

server/handlers/default.go
endophage: I'm slowly fixing all our errors as I encounter them. I just crossed paths with ErrNotFound writing some tests so I fixed it. Without very good reasons, none of our errors should be pointers, but we were all still learning Go when we wrote some of this.

server/handlers/validation.go
endophage: I'll move the last line of the comment to the top.

tuf/utils/role_sort.go
endophage: Go's sort itself is an unstable ordering because even with lexical sorting, technically 2 identical items won't always be sorted the same way around.

cyli: I thought `sort.Stable` got around that by preserving the original order if 2 equal items?

endophage: Yeah, looks like that's the intent with the penalty of some extra swap calls. There's just no point in us doing an alphabetic ordering. I can do it though if it'll make people happy.

cyli: I agree that I think the algorithm does not require alphabetic ordering, and that segment ordering should be fine.  I think @diogomonica's point was that if we got it wrong, or we otherwise changed things later to introduce a bug which might be affected by the ordering, the tests wouldn't fail deterministically.

I'm fine with no alphabetic ordering, since it adds some extra computational penalty, but if so it might be good to actually test that deterministic ordering within a bucket doesn't matter - (e.g. if we can inject the sorting function in in a test, we can have it return a bunch of different sort orderings per bucket).

tuf/tuf_test.go
```diff
@@ -16,21 +16,13 @@ import (
 
 func initRepo(t *testing.T, cryptoService signed.CryptoService, keyDB *keys.KeyDB) *Repo {
 	rootKey, err := cryptoService.Create("root", data.ED25519Key)
-	if err != nil {
```
cyli: Yay this file is so much cleaner!!

server/handlers/validation.go
```diff
@@ -28,9 +30,13 @@ func validateUpdate(cs signed.CryptoService, gun string, updates []storage.MetaU
...
 	rootRole := data.RoleName(data.CanonicalRootRole)
-	targetsRole := data.RoleName(data.CanonicalTargetsRole)
 	snapshotRole := data.RoleName(data.CanonicalSnapshotRole)
 
+	// some delegated targets role may be invalid based on other updates
+	// that have been made by other clients. We'll rebuild the slice of
+	// updates with only the things we should actually update
+	updatesToApply := make([]storage.MetaUpdate, 0, len(updates))
+
 	// check that the necessary roles are present:
```
cyli: Non-blocking - sorry, this was not part of your change at all, but this particular comment seems no longer applicable.

cyli:
Could we add another test that updates both "targets" and "targets/level1", but the update to targets deletes "targets/level1"?  In which case I think there should only be 1 update?

tuf/utils/role_sort.go
cyli: To be clear, I'm 99% certain you are right and this is pointless, I'm just subtracting a point because I tend to be wrong a lot :D  I would also be happy to add it later instead, since I don't want to block this PR necessarily.  (It'd just be nice to get in before the docker integration)

endophage: @cyli Added in the additional lexical sorting shortly after I commented :-P

tuf/tuf_test.go
endophage: Doing my best to fix things as I go :-)

server/handlers/validation.go
endophage: Ah yeah, I think I used to have the heirarchy check there but took that out in an earlier PR.

tuf/utils/role_sort.go
cyli: You are awesome!

tuf/utils/role_sort_test.go
```diff
@@ -0,0 +1,47 @@
...
+	rl := RoleList{"foo", "bar"}
+	rl.Swap(0, 1)
+	assert.Equal(t, "bar", rl[0])
+	assert.Equal(t, "foo", rl[1])
+}
+
+func TestRoleListSort(t *testing.T) {
+	rl := RoleList{"foo/bar", "foo", "bar", "bar/foo/baz", "bar/foo"}
+	sort.Sort(rl)
+	for i, s := range rl {
```
cyli: This can probably just assert an exact ordering now.

cyli:
Other than the above, LGTM!  :D

server/handlers/validation.go
```diff
@@ -74,32 +80,23 @@ func validateUpdate(cs signed.CryptoService, gun string, updates []storage.MetaU
 	}
 
 	// TODO: validate delegated targets roles.
```
cyli: Oop, one last thing:  yay this can be removed :D

tuf/tuf_test.go
diogomonica: :+1:

tuf/utils/role_sort_test.go
diogomonica: :+1:

diogomonica:
LGTM

PR #371, opened 2015-12-16
Add/Remove targets takes a list of roles to modify

cyli:
This allows us to add to more than 1 role.

From discussion with @endophage:  at publish time, we will try to sign the role - if there is no such role, and fall back the roles parent, and so forth.  If the role exists, and we just don't have a key, fail to sign.

This way if we integrate with docker, when adding targets it can specify the releases role to sign, so new clients will try to sign the releases role, and old clients will continue to sign the targets role.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/Sirupsen/logrus"
@@ -258,8 +259,10 @@ func (r *NotaryRepository) Initialize(rootKeyID string, serverManagedRoles ...st
 	return r.saveMetadata(serverManagesSnapshot)
 }
 
-// AddTarget adds a new target to the repository, forcing a timestamps check from TUF
-func (r *NotaryRepository) AddTarget(target *Target) error {
+// AddTarget adds a new target to the repository to the given roles, forcing a
+// timestamps check from TUF. If roles are unspecified, AddTarget the default
+// role is "target"
+func (r *NotaryRepository) AddTarget(target *Target, roles ...string) error {
 	cl, err := changelist.NewFileChangelist(filepath.Join(r.tufRepoPath, "changelist"))
 	if err != nil {
 		return err
@@ -273,16 +276,40 @@ func (r *NotaryRepository) AddTarget(target *Target) error {
 		return err
 	}
 
-	c := changelist.NewTufChange(
-		changelist.ActionCreate,
-		changelist.ScopeTargets,
-		changelist.TypeTargetsTarget,
-		target.Name,
-		metaJSON,
-	)
-	err = cl.Add(c)
-	if err != nil {
-		return err
+	if len(roles) == 0 {
+		roles = []string{data.CanonicalTargetsRole}
+	}
+
+	var changes []*changelist.TufChange
+	for _, role := range roles {
+		role = strings.ToLower(role)
+
+		if !data.ValidRole(role) {
+			return data.ErrInvalidRole{Role: role}
+		}
+
+		_, ok := data.ValidRoles[role]
+		if ok && role != data.CanonicalTargetsRole {
+			return data.ErrInvalidRole{
+				Role:   role,
+				Reason: "cannot add targets to this role",
+			}
+		}
+
+		changes = append(changes, changelist.NewTufChange(
+			changelist.ActionCreate,
+			role,
+			changelist.TypeTargetsTarget,
+			target.Name,
+			metaJSON,
+		))
+	}
+
+	for _, c := range changes {
+		err = cl.Add(c)
+		if err != nil {
+			return err
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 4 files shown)


tuf/tuf.go
```diff
@@ -179,7 +179,7 @@ func (tr *Repo) GetDelegation(role string) (*data.Role, error) {
 // UpdateDelegations updates the appropriate delegations, either adding
 // a new delegation or updating an existing one. If keys are
 // provided, the IDs will be added to the role (if they do not exist
-// there already), and the keys will be added to the targets file.
+// there already), and the keys will be added to the parent role file.
```
cyli: I originally thought I needed to add a delegations to test AddTarget and RemoveTarget, and I wanted to verify this assumption.  As it turns out, I didn't need it, but doesn't hurt to have the test for it.

client/client.go
```diff
@@ -258,8 +259,49 @@ func (r *NotaryRepository) Initialize(rootKeyID string, serverManagedRoles ...st
...
+			return err
+		}
+	}
+	return nil
+}
+
+// AddTarget creates new changelist entries to add a target to the given roles
+// in the repository when the changelist gets appied at publish time.
+// If roles are unspecified, the default role is "target".
+func (r *NotaryRepository) AddTarget(target *Target, roles ...string) error {
```
endophage: It would be allowable to pass no roles, what's the expected behaviour in the case. At the moment I believe `addChange` will create no change objects if no roles are passed.

endophage:
Would be good to add tests to see what happens when Add and Delete are called with no roles, then I suspect there will be some additional checks that need to be done.

client/client.go
```diff
@@ -258,8 +259,49 @@ func (r *NotaryRepository) Initialize(rootKeyID string, serverManagedRoles ...st
...
+
+	var changes []changelist.Change
+	for _, role := range roles {
+		role = strings.ToLower(role)
+
+		if !data.ValidRole(role) {
+			return data.ErrInvalidRole{Role: role}
+		}
+
+		if _, ok := data.ValidRoles[role]; ok && role != data.CanonicalTargetsRole {
```
diogomonica: If `data.ValidRoles` is 

```
var ValidRoles = map[string]string{
	CanonicalRootRole:      CanonicalRootRole,
	CanonicalTargetsRole:   CanonicalTargetsRole,
	CanonicalSnapshotRole:  CanonicalSnapshotRole,
	CanonicalTimestampRole: CanonicalTimestampRole,
}
```
Doesn't this mean that we only allow `addChange` for the 4 canonical roles?

diogomonica: Also, shouldn't we be verifying that the translation of `data.ValidRoles[role]` is `!= data.CanonicalTargetsRole` instead of doing `role !=` ?

client/client.go
cyli: [addChange sets the default role to be data.CanonicalTargetsRole if there are no roles passed](https://github.com/docker/notary/pull/371/files#diff-1cd41a2611da49ce63630d5a51bd82c5R265).

https://github.com/docker/notary/pull/371/files#diff-fce84029d346504e44323c30d21a3808R449 tests this for `AddTarget`, and https://github.com/docker/notary/pull/371/files#diff-fce84029d346504e44323c30d21a3808R449 tests it for `DeleteTarget`

client/client.go
cyli: `ValidRole` returns true for any of the canonical roles, and also if it's a delegation role.  I wanted to exclude canonical roles that were not the target role, (see https://github.com/docker/notary/pull/371/files#diff-fce84029d346504e44323c30d21a3808R584)

But this is confusing - I think if I steal @endophage's change in [this file](https://github.com/docker/notary/pull/370/files#diff-a84805db69378bc67da7d4ae04da989aR113) this will be easier to read.

diogomonica: Maybe its worth creating an issue to rip out the ValidRoles translation then?

client/client.go
endophage: Ah sorry, somehow totally missed the check at the start of `addChange`

client/client.go
cyli: Good point, thanks!   https://github.com/docker/notary/issues/372

client/client.go
cyli: Well, I tend to wrap function signatures, and I forgot to space between the actual function definition and the wrapping, so it's hard to visually distinguish.  :|  Sorry about that.

cyli:
@diogomonica @endophage Ready for another look.  :)

diogomonica:
Non-blocking comments. LGTM :+1:

endophage:
@diogomonica IsDelegation is also included in one of my PRs with tests. No need to add those tests here unless codecov complains about reduced coverage (which it isn't right now)

Really, ValidRole is more likely to call IsDelegation than the other way round. IsDelegation operates on a very narrow rule, all it checks if that the role name is prefixed with the base "targets" role name plus a slash.

cyli:
@diogomonica David was yes, I stole it to make the addChange code easier to read.  There are tests for it, and yeah there will be a conflict :|  I asked him already.  As far as calling `ValidRole`, a valid delegation role is a subset of all the roles that would be approved by `ValidRole`, so `ValidRole` calls it, not vice versa.

endophage:
LGTM

PR #378, opened 2015-12-18
Publishing delegation changes, and targets to delegations

cyli:
This makes the following logic changes:

- publish now iterates over all targets marked dirty and publishes only those that are dirty (this means canonical targets will not always get pushed)
- if changes are applied to a role that doesn't exist, publish will fallback on its parent role and so forth until it finds a role that exists.  the changes will be applied to that role.
- delegations metadata is not always created when a delegation is created
- role checks now check keyDB rather than the metadata, because the metadata may not exist
- if the role exists, but metadata doesn't, it is now created before being modified (as opposed to just failing before)

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -584,12 +584,17 @@ func (r *NotaryRepository) Publish() error {
 		updatedFiles[data.CanonicalRootRole] = rootJSON
 	}
 
-	// we will always re-sign targets
-	targetsJSON, err := serializeCanonicalRole(r.tufRepo, data.CanonicalTargetsRole)
-	if err != nil {
-		return err
+	// iterate through all the targets files - if they are dirty, or if they
+	// are the canonical target role, then sign and update
+	for roleName, roleObj := range r.tufRepo.Targets {
+		if roleName == data.CanonicalTargetsRole || roleObj.Dirty {
+			targetsJSON, err := serializeCanonicalRole(r.tufRepo, roleName)
+			if err != nil {
+				return err
+			}
+			updatedFiles[roleName] = targetsJSON
+		}
 	}
-	updatedFiles[data.CanonicalTargetsRole] = targetsJSON
 
 	// if we initialized the repo while designating the server as the snapshot
 	// signer, then there won't be a snapshots file.  However, we might now
--- a/client/client.go
+++ b/client/client.go
@@ -245,7 +245,7 @@ func (r *NotaryRepository) Initialize(rootKeyID string, serverManagedRoles ...st
 		logrus.Debug("Error on InitRoot: ", err.Error())
 		return err
 	}
-	err = r.tufRepo.InitTargets(data.CanonicalTargetsRole)
+	_, err = r.tufRepo.InitTargets(data.CanonicalTargetsRole)
 	if err != nil {
 		logrus.Debug("Error on InitTargets: ", err.Error())
 		return err
--- a/client/client.go
+++ b/client/client.go
@@ -584,10 +584,9 @@ func (r *NotaryRepository) Publish() error {
 		updatedFiles[data.CanonicalRootRole] = rootJSON
 	}
 
-	// iterate through all the targets files - if they are dirty, or if they
-	// are the canonical target role, then sign and update
+	// iterate through all the targets files - if they are dirty, sign and update
 	for roleName, roleObj := range r.tufRepo.Targets {
-		if roleName == data.CanonicalTargetsRole || roleObj.Dirty {
+		if roleObj.Dirty {
 			targetsJSON, err := serializeCanonicalRole(r.tufRepo, roleName)
 			if err != nil {
 				return err
--- a/client/client.go
+++ b/client/client.go
@@ -522,7 +522,7 @@ func (r *NotaryRepository) GetChangelist() (changelist.Changelist, error) {
 // Publish pushes the local changes in signed material to the remote notary-server
 // Conceptually it performs an operation similar to a `git rebase`
 func (r *NotaryRepository) Publish() error {
-	var updateRoot bool
+	var initialPublish bool
 	// attempt to initialize the repo from the remote store
 	c, err := r.bootstrapClient()
 	if err != nil {
@@ -538,10 +538,11 @@ func (r *NotaryRepository) Publish() error {
 				return err
 			}
 			// We had local data but the server doesn't know about the repo yet,
-			// ensure we will push the initial root file.  The root may not
-			// be marked as Dirty, since there may not be any changes that
-			// update it, so use a different boolean.
-			updateRoot = true
+			// ensure we will push the initial root and targets file.  Either or
+			// both of the root and targets may not be marked as Dirty, since
+			// there may not be any changes that update them, so use a
+			// different boolean.
+			initialPublish = true
 		} else {
 			// The remote store returned an error other than 404. We're
 			// unable to determine if the repo has been initialized or not.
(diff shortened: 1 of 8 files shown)


client/client.go
```diff
@@ -528,12 +578,16 @@ func (r *NotaryRepository) Publish() error {
 		updatedFiles[data.CanonicalRootRole] = rootJSON
 	}
 
-	// we will always re-sign targets
-	targetsJSON, err := serializeCanonicalRole(r.tufRepo, data.CanonicalTargetsRole)
-	if err != nil {
-		return err
+	// iterate through all the targets files - if they are dirty, sign and update
+	for roleName, roleObj := range r.tufRepo.Targets {
```
cyli: Note to reviewer:  I am slightly worried about no longer sending targets all the time - I *think* it should be fine, and all the tests seem to pass, but would appreciate it if folks took a harder look at this.

endophage: We should confirm targets is marked as dirty on a new repo so that it always gets pushed the first time.

client/client_test.go
```diff
@@ -1398,6 +1424,222 @@ func TestPublishSnapshotLocalKeysCreatedFirst(t *testing.T) {
...
+	defer ts.Close()
+
+	repo, _ := initializeRepo(t, data.ECDSAKey, tempBaseDir, gun, ts.URL, false)
+	assertPublishToRolesSucceeds(t, repo, []string{"targets/a/b", "targets/b/c"},
+		[]string{data.CanonicalTargetsRole})
+}
+
+// If a changelist specifies a particular role to push targets to, and there
+// is a role but no key, publish not fall back and just fail.
+func TestPublishTargetsDelgationScopeNoFallbackIfNoKeys(t *testing.T) {
```
endophage: I hadn't even thought of this behaviour and I think you picked exactly the right approach! Great job and great test!

client/client.go
cyli: Good point, thanks!

client/helpers.go
```diff
@@ -216,14 +245,14 @@ func addKeyForRole(kdb *keys.KeyDB, role string, key data.PublicKey) error {
 // signs and serializes the metadata for a canonical role in a tuf repo to JSON
 func serializeCanonicalRole(tufRepo *tuf.Repo, role string) (out []byte, err error) {
 	var s *data.Signed
-	switch role {
-	case data.CanonicalRootRole:
+	if role == data.CanonicalRootRole {
```
endophage: Rather than all the if ... else if ..., you can use an empty `switch {}` and put the same boolean checks in the cases.

cyli: I tried that first, but apparently `case` doesn't seem to work with `_, ok := tufRepo.Targets[role]; ok` - I kept getting syntax errors :|

tuf/tuf.go
```diff
@@ -185,10 +211,19 @@ func (tr *Repo) UpdateDelegations(role *data.Role, keys []data.PublicKey) error
...
+
+	// check the parent role
+	if parentRole := tr.keysDB.GetRole(parent); parentRole == nil {
 		return data.ErrInvalidRole{Role: role.Name, Reason: "parent role not found"}
 	}
+
+	// check the parent role's metadata
+	p, ok := tr.Targets[parent]
+	if !ok { // the parent targetfile may not exist yet - if not, then create it
+		p = data.NewTargets()
```
cyli: @riyazdf brought up a great point about the assumption of `keysDB` being correct (we check `keysDB` to be sure the parent role exists first).  If it doesn't have the parent role, and should, then the repo may just be in a bad state, and we correctly error.  If it shouldn't have the parent role, but does though, that would let UpdateDelegations proceed.  I'm wondering if it would be too defensive to check if the grandparent metadata exists before adding the parent metadata?

endophage: Do we want to check if we have signing keys before creating? Also:
```
tr.InitTargets(parent)
```

client/client.go
cyli: Good call!  this was definitely an error and would have made it impossible to push new bare repos.

client/helpers.go
endophage: Ah yeah, you can't do the assignment. You could switch that one to `tufRepo.Targets[role] != nil` as an explicitly set nil value would be just as invalid as an unset value for us.

tuf/tuf.go
```diff
@@ -99,8 +99,25 @@ func (tr *Repo) AddBaseKeys(role string, keys ...data.PublicKey) error {
...
 
+	// also, whichever role was switched out needs to be re-signed
+	// root has already been marked dirty
+	switch role {
+	case data.CanonicalSnapshotRole:
+		if tr.Snapshot != nil {
+			tr.Snapshot.Dirty = true
+		}
+	case data.CanonicalTargetsRole:
+		target, ok := tr.Targets[data.CanonicalTargetsRole]
```
endophage:
``` 
if target, ok := tr.Targets[data.CanonicalTargetsRole]; ok {
    target.Dirty = true
}
```

tuf/tuf.go
endophage: @cyli @riyazdf can't that logic be applied ad infinitum?

cyli: Hmm... I wouldn't mind doing so - we can check for signing keys before updating any delegations at all.  Would that be better than adding and letting the signing fail?

client/helpers.go
cyli: Ah right, thanks!

endophage:
We have the code:
```
r := tr.keysDB.GetRole(role)
if r == nil {
    return nil, data.ErrInvalidRole{Role: role, Reason: "does not exist"}
}

t, ok := tr.Targets[role]
if !ok { // the targetfile may not exist yet - if not, then create it
    t = data.NewTargets()
    tr.Targets[role] = t
}
```

This will operate correctly but it feels like the semantics should be:
```
t, ok := tr.Targets[role]
if !ok { 
    // the targetfile may not exist yet. Check if the role is valid and create targets file if it is
    r := tr.keysDB.GetRole(role)
    if r == nil {
        return nil, data.ErrInvalidRole{Role: role, Reason: "does not exist"}
    }
    t = data.NewTargets()
    tr.Targets[role] = t
}
```

Not a biggie, just my mental model of how the Repo object operates.

tuf/tuf.go
endophage: Yeah, that would be even better. Maybe a Repo method `CanSign(role string) bool` and as long as the CryptoService has at least one signing key for the role it returns true.

cyli:
@endophage sure, that'd work for me.  If I wanted to check for signing keys before operations though, the other way is easier:

```
	// check the role exists
	r := tr.keysDB.GetRole(role)
	if r == nil {
		return nil, data.ErrInvalidRole{Role: role, Reason: "does not exist"}
	}

	// check that we have a signing key for the parent
	allKeys := tr.cryptoService.ListAllKeys()
	haveKey := false
	for _, keyID := range r.KeyIDs {
		_, haveKey = allKeys[keyID]
		if haveKey {
			break
		}
	}
	if !haveKey {
		return signed.ErrNoKeys{KeyIDs: r.KeyIDs}
	}

	// check the role's metadata
	t, ok := tr.Targets[role]
	if !ok { // the targetfile may not exist yet - if not, then create it
		t, err = tr.InitTargets(role)
		if err != nil {
			return error
		}
	}
```

I think I'm indifferent as to whether we should check signing keys before creating or any other operation (although I think if we are going to check signing keys, we should fail the operation if there is no signing key whether or not a metadata creation was required), and I'm indifferent between checking the role first or the metadata file first.  :)  So I'll go with whichever one you pick.

endophage:
I'm thinking we would want to add the signing keys check to all methods that modify a role in the repo

client/client_test.go
cyli: Thanks!  I just didn't want us falling back just because someone forgot to plug in their key.  :)

endophage: Yeah, this is totally the right behaviour!

endophage:
LGTM

riyazdf:
LGTM!

endophage:
@cyli can you rebase just to make sure there aren't any conflicts with the delegated Paths update I just merged?

cyli:
@endophage minor changes, and added https://github.com/docker/notary/pull/378/files#diff-5f5c6e767e5dbb849cfc678ad1655d98R904

endophage:
@cyli that makes sense

diogomonica:
Amazing test coverage btw.

PR #383, opened 2015-12-22
Cleanup and removing filename aliases in favour of PEM headers

endophage:


trustmanager/filestore_test.go
```diff
@@ -416,3 +333,47 @@ func generateRandomFile(filePath string, perms os.FileMode) ([]byte, error) {
...
 	return rndBytes, nil
 }
+
+func TestFileStoreConsistency(t *testing.T) {
+	// Temporary directory where test files will be created
+	tempBaseDir, err := ioutil.TempDir("", "notary-test-")
+	assert.NoError(t, err)
+	defer os.RemoveAll(tempBaseDir)
+
+	s, err := NewPrivateSimpleFileStore(tempBaseDir, "txt")
```
diogomonica: Maybe also create a `NewPrivateSimpleFileStore(tempBaseDir, ".txt")` to also test that behavior?

trustmanager/keyfilestore.go
```diff
@@ -175,18 +177,27 @@ func addKey(s LimitedFileStore, passphraseRetriever passphrase.Retriever, cached
...
+			}
+			block, _ := pem.Decode(d)
+			if block != nil {
+				if role, ok := block.Headers["role"]; ok {
+					return role, nil
+				}
+			}
+
 			aliasPlusDotKey := strings.TrimPrefix(filename, name+"_")
 			retVal := strings.TrimSuffix(aliasPlusDotKey, "."+keyExtension)
```
diogomonica: I thought ListFiles was not supposed to return something that has `"."+keyExtension`

trustmanager/keyfilestore.go
```diff
@@ -228,28 +238,39 @@ func listKeys(s LimitedFileStore) map[string]string {
...
-			f = strings.TrimPrefix(f, nonRootKeysSubdir+"/")
+			keyIDFull = strings.TrimPrefix(f, nonRootKeysSubdir+"/")
 		}
 
-		// Remove the extension from the full filename
-		// abcde_root.key becomes abcde_root
-		keyIDFull := strings.TrimSpace(strings.TrimSuffix(f, filepath.Ext(f)))
+		keyIDFull = strings.TrimSpace(keyIDFull)
 
 		// If the key does not have a _, it is malformed
```
diogomonica: Change comment

trustmanager/x509utils.go
```diff
@@ -440,6 +451,11 @@ func EncryptPrivateKey(key data.PrivateKey, passphrase string) ([]byte, error) {
 		return nil, err
 	}
 
+	if encryptedPEMBlock.Headers == nil {
```
diogomonica: I don't think this condition should ever happen, and if it does, we should error out, because clearly the key has not been correctly encrypted.

diogomonica:
Can we add a test to ensure we keep backwards compatible behavior? (we can read the role from `abcd_targets.key` if it doesn't have the role header inside of the pem.

trustmanager/keyfilestore.go
endophage: It doesn't any more. Note no highlight on the line = legacy code. Will clean that up.

trustmanager/x509utils.go
endophage: I'm inclined to agree. This felt safer though as `EncryptPEMBlock` returns a specific error and if for some reason `Headers` isn't initialized attempting to assign will cause a panic. I can certainly error instead of initializing, that just felt presumptive give the explicit error return.

endophage:
@diogomonica will do on the backwards compat test.

server/server_test.go
```diff
@@ -42,12 +40,14 @@ func TestRunReservedPort(t *testing.T) {
...
 
-	if _, ok := err.(*net.OpError); !ok {
-		t.Fatalf("Received unexpected err: %s", err.Error())
-	}
-	if !strings.Contains(err.Error(), "bind: permission denied") {
-		t.Fatalf("Received unexpected err: %s", err.Error())
-	}
+	assert.Error(t, err)
+	assert.IsType(t, &net.OpError{}, err)
+	assert.True(
```
cyli: Non-blocking - assert also has an `assert.Contains` which calls `string.Contains` for you.  This is great though, comment was just in case you wanted to use it in the future.

trustmanager/keyfilestore.go
```diff
@@ -175,25 +177,37 @@ func addKey(s LimitedFileStore, passphraseRetriever passphrase.Retriever, cached
...
 
-func getKeyAlias(s LimitedFileStore, keyID string) (string, error) {
-	files := s.ListFiles()
-
+// getKeyRole finds the role for the given keyID. It attempts to look
+// both in the newer format PEM headers, and also in the legacy filename
+// format. It returns: the role, whether it was found in the legacy format
+// (true == legacy), and an error
+func getKeyRole(s LimitedFileStore, keyID string) (string, bool, error) {
 	name := strings.TrimSpace(strings.TrimSuffix(filepath.Base(keyID), filepath.Ext(keyID)))
```
cyli: Non-blocking question - I know this wasn't changed in your code, but I was wondering if we ever pass `getKeyRole` a `keyID` that has an extension?

trustmanager/keyfilestore.go
```diff
@@ -228,44 +241,59 @@ func listKeys(s LimitedFileStore) map[string]string {
...
-		keyIDFull := strings.TrimSpace(strings.TrimSuffix(f, filepath.Ext(f)))
+		keyIDFull = strings.TrimSpace(keyIDFull)
 
-		// If the key does not have a _, it is malformed
+		// If the key does not have a _, we'll attempt to
+		// read it as a PEM
 		underscoreIndex := strings.LastIndex(keyIDFull, "_")
 		if underscoreIndex == -1 {
-			continue
+			keyID := keyIDFull
```
cyli: Non-blocking nitpick - this assignment seems extraneous

cyli:
Two questions:

1.  This change would make it so that if you created repos with a new client, the old client cannot read those repos - I guess we don't have a changelist or anything, but maybe this should be noted in the merge commit for when we build the changelist for the next release.

2.  Should we add a backwards compatibility test for import-export?  E.g. Can we import a zip file containing keys in the old format?  Do we care about this?

Otherwise LGTM, excepting for the error-ing out if headers are missing!

trustmanager/keyfilestore.go
endophage: With the changes to ListFiles that will no longer happen.

endophage:
@cyli we should discuss point 1 in more depth outside the context of this PR. My opinion is that we should always try to make new code backwards compatible, but there is no expectation that old code should be future compatible. In this case, I believe it will only affect the signer and a single user shouldn't be running multiple different versions of the client (because of the backwards/forwards compatibility problems that exist in all software).

On 2, it would be good to add a test. I have a high degree of confidence it'll work but that's no substitute for a test :-)

cyli:
@endophage That makes sense - I'm not saying we should try to make it also work with older clients, just that we should note it somewhere people would see if they upgraded.  :)  People might be running different versions of the client due to running different versions of the docker client, but that's about it.

endophage:
Ooooh, yeah, we should create a wiki page here to document expected behaviours between clients.

cyli:
WRT to the tests for backwards-compatibility import-export, I've merged this into #387 locally and all those tests pass with the key change, so import-export looks good :)

This needs to be rebased because there is a semantic conflict with current master.

riyazdf:
I agree that we should error out on a nil header.

LGTM pending adding that error and rebase!

cyli:
Review comments addressed (and rebase fixes) in https://github.com/docker/notary/commit/0465365fb6fb9d1f5f995e0f19f772f5334fbe09.

trustmanager/x509utils.go
cyli: @diogomonica fixed in https://github.com/docker/notary/commit/0465365fb6fb9d1f5f995e0f19f772f5334fbe09

diogomonica:
LGTM

diogomonica:
Closed in favor of #388

Issue #363, opened 2015-12-15
Client side writing/publishing tests


cyli:
# Initializing a repo

## Success
- [ ] no prior data; initialize a new repo, server manages timestamp key
	- all metadata and keys created except timestamp
- [ ] no prior data; initialize a new repo, server manages timestamp and
	- root and targets metadata nad keys created

## Failure
- [ ] no prior data; remote server unreachable when getting timestamp key;
	initialize a repo
- [ ] no prior data; remote server unreachable when getting snapshot key;
	initialize a repo with remotely-managed snapshot key
- [ ] no prior data; initialize repo; failure to create root key; no metadata
	files saved
- [ ] no prior data; initialize repo; failure to create targets key; no metadata
	files saved
- [ ] no prior data; initialize repo with locally-managed snapshot signing;
	failure to create snapshot key; no metadata files saved

## Failure (fixme: these would not currently fail in notary - may want to check remote server first)
- [ ] prior repo with any metadata whatsoever - fails without need to reach server
	- nothing created
- [ ] prior repo with corrupt root.json - fails without need to reach server
	- nothing created

# Changelists + Publishing a repo

## Success - no changes
- [ ] no prior data; initialize a new repo with locally managed snapshot key;
	publish a repo - all root/target/snapshot/delegations metadata should be
	pushed, local snapshot key on disk
	(tests initializing with local snapshot key, all metadata gets pushed)
- [ ] no prior data; initialize a new repo; publish; publish;
	all root/target/(snapshot?)/delegations metadata re-signed and pushed again
	(tests initializing with local snapshot key, all metadata gets pushed)
- [ ] no prior data; initialize a new repo with remotely managed snapshot key;
	publish a repo - all root/target/delegations metadata should be pushed,
	no snapshot key on disk
	(tests initializing with remote snapshot key, all metadata except snapshot
	gets pushed)

## Success - content changes (with and without remote signing - with remote
	signing snapshot doesn't get pushed, without it does)
- [ ] initialized repo; add target twice; should be 2 changefiles, publish;
	0 changefiles, published targets file has 1 target
- [ ] initialized repo with 1 published target; add same target again;
	should be 1 changefile; publish; 0 change files,
	published targets file has 1 target
- [ ] initialized repo; add target, remove target; should be 2 changefiles;
	publish; published targets file has 0 targets
- [ ] initialized repo with 1 published target; remove same target;
	should be 1 changefile; publish; 0 change files, published targets file has
	0 targets
[...]

PR #384, opened 2015-12-22
More client writing/publishing negative tests

cyli:
A couple more negative test cases from https://github.com/docker/notary/issues/363

Semantic conflict with #385

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -1184,6 +1184,12 @@ func TestPublishUninitializedRepo(t *testing.T) {
 	assert.NoError(t, err, "error generating root key: %s", err)
 
 	assert.NoError(t, repo.Initialize(rootPubKey.ID()))
+
+	// now metadata is created
+	assertRepoHasExpectedMetadata(t, repo, data.CanonicalRootRole, true)
+	assertRepoHasExpectedMetadata(t, repo, data.CanonicalSnapshotRole, true)
+	assertRepoHasExpectedMetadata(t, repo, data.CanonicalTargetsRole, true)
+
 	assert.NoError(t, repo.Publish())
 }
 
@@ -1224,7 +1230,8 @@ func assertPublishSucceeds(t *testing.T, repo1 *NotaryRepository) {
 	assertPublishToRolesSucceeds(t, repo1, nil, []string{data.CanonicalTargetsRole})
 }
 
-// asserts that adding to the given roles results in the targets actually
+// asserts that adding to the given roles results in the targets actually being
+// added only to the expected roles and no others
 func assertPublishToRolesSucceeds(t *testing.T, repo1 *NotaryRepository,
 	publishToRoles []string, expectedPublishedRoles []string) {
 
@@ -1365,20 +1372,22 @@ func TestPublishSnapshotCorrupt(t *testing.T) {
 	ts := fullTestServer(t)
 	defer ts.Close()
 
-	// do not publish first - publish should fail
+	// do not publish first - publish should fail with corrupt snapshot data even with server signing snapshot
 	repo, _ := initializeRepo(t, data.ECDSAKey, "docker.com/notary1", ts.URL, true)
 	defer os.RemoveAll(repo.baseDir)
 	testPublishBadMetadata(t, data.CanonicalSnapshotRole, repo, false, false)
 
+	// do not publish first - publish should fail with corrupt snapshot data with local snapshot signing
 	repo, _ = initializeRepo(t, data.ECDSAKey, "docker.com/notary2", ts.URL, false)
 	defer os.RemoveAll(repo.baseDir)
 	testPublishBadMetadata(t, data.CanonicalSnapshotRole, repo, false, false)
 
-	// publish first - should succeed
+	// publish first - publish again should succeed despite corrupt snapshot data (server signing snapshot)
 	repo, _ = initializeRepo(t, data.ECDSAKey, "docker.com/notary3", ts.URL, true)
 	defer os.RemoveAll(repo.baseDir)
 	testPublishBadMetadata(t, data.CanonicalSnapshotRole, repo, true, true)
 
+	// publish first - publish again should succeed despite corrupt snapshot data (local snapshot signing)
 	repo, _ = initializeRepo(t, data.ECDSAKey, "docker.com/notary4", ts.URL, false)
 	defer os.RemoveAll(repo.baseDir)
 	testPublishBadMetadata(t, data.CanonicalSnapshotRole, repo, true, true)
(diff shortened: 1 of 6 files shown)


client/client_test.go
```diff
@@ -1370,64 +1320,117 @@ func testPublishNoOneHasSnapshotKey(t *testing.T, rootType string) {
...
+	defer os.RemoveAll(repo.baseDir)
+	testPublishBadMetadata(t, data.CanonicalTargetsRole, repo, true, true)
+}
+
+// If the root metadata is corrupt or the root metadata is unreadable,
+// we can't publish for the first time or the second time.  Root is the most
+// important and what we used to pin trust, so if it's corrupt, we can't
+// verify downloaded updates.
+func TestPublishRootCorrupt(t *testing.T) {
+	t.Skip("Test currently fails - not sure what the correct behavior is.")
```
cyli: Not sure if this is the expected behavior?

client/client_test.go
```diff
@@ -1686,17 +1663,38 @@ func TestPublishTargetsDelgationSuccessNeedsToDownloadRoles(t *testing.T) {
 		[]string{"targets/a/b"})
 }
 
+// If the delegation data is corrupt or unreadable, it doesn't matter because
+// all the delegation information is just re-downloaded.  When bootstrapping
+// the repository from disk, we just don't load the data from disk because
+// there should not be anything there yet.
+func TestPublishSucceedsDespiteDelegationCorrupt(t *testing.T) {
```
cyli: I don't think this behavior causes any problems, but I wanted to double-check.

client/client_test.go
```diff
@@ -1186,6 +1156,37 @@ func testPublishNoData(t *testing.T, rootType string, serverManagesSnapshot bool
...
+	// uninitialized repo should fail to publish
+	tempBaseDir, err := ioutil.TempDir("", "notary-tests")
+	assert.NoError(t, err)
+	defer os.RemoveAll(tempBaseDir)
+
+	repo, err := NewNotaryRepository(tempBaseDir, gun, ts.URL,
+		http.DefaultTransport, passphraseRetriever)
+	assert.NoError(t, err, "error creating repository: %s", err)
+	err = repo.Publish()
+	assert.Error(t, err)
```
cyli: Todo: check for uninitialized repo error once @riyazdf's checks land

client/client_test.go
```diff
@@ -123,6 +133,23 @@ func createRepoAndKey(t *testing.T, rootType, tempBaseDir, gun, url string) (
 	return repo, rootPubKey.ID()
 }
 
+// creates a new notary repository with the same gun and url as the previous
+// repo, so that it can be used to pull the trust data the original repo pushed
+func newRepoToTestRepo(t *testing.T, existingRepo *NotaryRepository) *NotaryRepository {
```
diogomonica: Is this to avoid deleting files in a tmpDir mid way to replicate a clean Repository state?

client/client_test.go
```diff
@@ -34,10 +34,12 @@ import (
 	"golang.org/x/net/context"
 )
 
-func simpleTestServer(t *testing.T) (
+func simpleTestServer(t *testing.T, roles ...string) (
```
diogomonica: Cool change.

client/client_test.go
cyli: Yes

client/client_test.go
```diff
@@ -1186,6 +1156,37 @@ func testPublishNoData(t *testing.T, rootType string, serverManagesSnapshot bool
...
+	// no metadata created
+	assertRepoHasExpectedMetadata(t, repo, data.CanonicalRootRole, false)
+	assertRepoHasExpectedMetadata(t, repo, data.CanonicalSnapshotRole, false)
+	assertRepoHasExpectedMetadata(t, repo, data.CanonicalTargetsRole, false)
+
+	// now, initialize and republish in the same directory
+	rootPubKey, err := repo.CryptoService.Create("root", data.ECDSAKey)
+	assert.NoError(t, err, "error generating root key: %s", err)
+
+	assert.NoError(t, repo.Initialize(rootPubKey.ID()))
```
diogomonica: Should we do assert that metadata got created after the `Initialize()` call?

client/client_test.go
```diff
@@ -1208,17 +1209,12 @@ func TestPublishAfterInitServerHasSnapshotKey(t *testing.T) {
...
-	assert.NoError(t, err, "failed to create a temporary directory: %s", err)
-
-	gun := "docker.com/notary"
 	ts := fullTestServer(t)
 	defer ts.Close()
 
-	repo, _ := initializeRepo(t, rootType, tempBaseDir, gun, ts.URL,
+	repo, _ := initializeRepo(t, rootType, "docker.com/notary", ts.URL,
 		serverManagesSnapshot)
+	defer os.RemoveAll(repo.baseDir)
```
diogomonica: There seems to be an unfinished comment on line 1227:

` // asserts that adding to the given roles results in the targets actually`

client/client_test.go
```diff
@@ -1253,15 +1249,9 @@ func assertPublishToRolesSucceeds(t *testing.T, repo1 *NotaryRepository,
...
-	tempBaseDir, err := ioutil.TempDir("", "notary-test-")
-	defer os.RemoveAll(tempBaseDir)
-
-	assert.NoError(t, err, "failed to create a temporary directory: %s", err)
-
-	repo2, err := NewNotaryRepository(tempBaseDir, repo1.gun, repo1.baseURL,
-		http.DefaultTransport, passphraseRetriever)
-	assert.NoError(t, err, "error creating repository: %s", err)
+	// use another repo to check metadata
+	repo2 := newRepoToTestRepo(t, repo1)
```
diogomonica: what exactly are we doing with repo2?

cyli: I check that both the original repo and the new clean repo have the expected roles (in the for loop)

client/client_test.go
```diff
@@ -1370,64 +1351,117 @@ func testPublishNoOneHasSnapshotKey(t *testing.T, rootType string) {
...
-// If the snapshot metadata is unreadable, whether the client or server has the
-// snapshot key, we can't publish.
-// We test this with both an RSA and ECDSA root key
-func TestPublishSnapshotUnreadable(t *testing.T) {
-	testPublishBadExistingSnapshot(t, data.ECDSAKey, true, false)
-	testPublishBadExistingSnapshot(t, data.ECDSAKey, false, false)
+	// do not publish first - publish should fail
+	repo, _ := initializeRepo(t, data.ECDSAKey, "docker.com/notary1", ts.URL, true)
+	defer os.RemoveAll(repo.baseDir)
+	testPublishBadMetadata(t, data.CanonicalSnapshotRole, repo, false, false)
```
diogomonica: aren't these tests only testing 2 behaviors? (false, false and true, true)

There doesn't seem to be any difference between 1 and 2; 3 and 4.

client/client_test.go
cyli: That is the assumption, yes :)  There is a test here https://github.com/docker/notary/pull/384/files#diff-fce84029d346504e44323c30d21a3808R402 that tests that initializing a repo creates the right metadata, keys, and certs.  But I can also make the same assertions here.

client/client_test.go
cyli: The difference between 1 and 2 is server snapshot signing vs not, similarly with 3 and 4.  I'll add comments.

client/client_test.go
diogomonica: So this situation we're trying to publish with a local corrupt root metadata, correct?

The trust anchor is actually the X509 certificate that we have in trusted_certificates. So technically we could remove a corrupt local `root.json` and still have trust work as intended (on download we will validate if the new root.json was signed with the private key corresponding to the X509 cert on disk).

/cc @endophage

client/client_test.go
diogomonica: Yes you do, I missed that (didn't collapse the code all the way).

client/client_test.go
```diff
@@ -1370,64 +1351,117 @@ func testPublishNoOneHasSnapshotKey(t *testing.T, rootType string) {
...
+	defer os.RemoveAll(repo.baseDir)
+	testPublishBadMetadata(t, data.CanonicalRootRole, repo, false, false)
+
+	// publish first - publish should still succeed if root corrupt
+	repo, _ = initializeRepo(t, data.ECDSAKey, "docker.com/notary2", ts.URL, false)
+	defer os.RemoveAll(repo.baseDir)
+	testPublishBadMetadata(t, data.CanonicalRootRole, repo, true, false)
+}
+
+// When publishing snapshot, root, or target, if the repo hasn't been published
```
diogomonica: Is this true for root?

client/client_test.go
cyli: But do we need the `root.json` to check that the version hasn't been rolled back, or do we just not care if local is corrupt?

diogomonica: You're right, we would lose the protection against rollbacks, but since it's a local corruption, I think that behavior might be ok.

diogomonica:
This is looking really good.

LGTM :+1:

client/client_test.go
cyli: Yes, if the repo hasn't been published yet, corrupt root data will cause a publish to fail.  If it *has* been published, it seems to be fine-ish because we re-download.  (there's a weird error when I can't read the file - it's `ErrLocalRootExpired` instead of a path error, which I expected, but I wanted to make sure that it *should* be fine, and not just blow up)

client/client_test.go
riyazdf: I made a [similar test](https://github.com/docker/notary/commit/c40ec37f50d3a3198e1bf8a5a78c8ab995530b20#diff-fce84029d346504e44323c30d21a3808R1434) that checks for the error in my PR as well -- your test is more thorough so feel free to remove mine and just add the error check if you prefer!

riyazdf:
LGTM, thanks for the thorough comments!

client/client_test.go
cyli: I'm happy to have it in both places!

PR #385, opened 2015-12-22
Add more delegation writing/publishing tests.

cyli:

Should address some of the tests in https://github.com/docker/notary/issues/363

Semantic conflict with #384

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -1604,7 +1604,7 @@ func TestPublishTargetsDelgationScopeNoFallbackIfNoKeys(t *testing.T) {
 	assert.Empty(t, targets)
 }
 
-// If a changelist specifies a particular role to push targets to, and is such
+// If a changelist specifies a particular role to push targets to, and such
 // a role and the keys are present, publish will write to that role only, and
 // not its parents.  This tests the case where the local machine knows about
 // all the roles (in fact, the role creations will be applied before the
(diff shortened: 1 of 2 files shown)


client/client_test.go
```diff
@@ -1632,6 +1632,40 @@ func TestPublishTargetsDelgationSuccessLocallyHasRoles(t *testing.T) {
...
+
+	for _, delgName := range []string{"targets/a", "targets/a/b"} {
+		assert.NoError(t,
+			repo.AddDelegation(delgName, 1, []data.PublicKey{delgKey}, []string{""}),
+			"error creating delegation")
+	}
+	assert.NoError(t, repo.Publish())
+
+	// remove targets key - it is not even needed
+	targetsKeys := repo.CryptoService.ListKeys(data.CanonicalTargetsRole)
```
diogomonica: I think we should assert that there is only one, and we should delete `[0]` instead of doing a loop.

client/client_test.go
```diff
@@ -1686,6 +1720,188 @@ func TestPublishTargetsDelgationSuccessNeedsToDownloadRoles(t *testing.T) {
...
+	}
+
+	gun := "docker.com/notary"
+	ts := fullTestServer(t)
+	defer ts.Close()
+
+	// this happens to be the client that creates the repo, but can also
+	// write a delegation
+	repo1, _ := initializeRepo(t, data.ECDSAKey, tempDirs[0], gun, ts.URL, true)
+	defer os.RemoveAll(repo1.baseDir)
```
diogomonica: do we need this if we already have a defer `os.RemoveAll()` on line 1730?

client/client_test.go
```diff
@@ -1686,6 +1720,188 @@ func TestPublishTargetsDelgationSuccessNeedsToDownloadRoles(t *testing.T) {
...
+		for _, targetName := range []string{"first", "second", "third"} {
+			_, ok := found[targetName]
+			assert.True(t, ok)
+		}
+	}
+}
+
+// A client who could publish before can no longer publish once the owner
+// revokes their delegation.
+func TestPublishRevokeDelgation(t *testing.T) {
```
diogomonica: We shouldn't use the word revoke. It implies signing of something that stands on it's own. In this case, remove is more appropriate.

diogomonica:
LGTM

client/client_test.go
cyli: Sounds good

client/client_test.go
cyli: No, not at all - good catch, thanks :) This change was originally part of #384, so it's an artifact of that refactor.

client/client_test.go
```diff
@@ -1632,6 +1632,38 @@ func TestPublishTargetsDelgationSuccessLocallyHasRoles(t *testing.T) {
...
+		assert.NoError(t,
+			repo.AddDelegation(delgName, 1, []data.PublicKey{delgKey}, []string{""}),
+			"error creating delegation")
+	}
+	assert.NoError(t, repo.Publish())
+
+	// remove targets key - it is not even needed
+	targetsKeys := repo.CryptoService.ListKeys(data.CanonicalTargetsRole)
+	assert.Len(t, targetsKeys, 1)
+	assert.NoError(t, repo.CryptoService.RemoveKey(targetKeys[0]))
```
riyazdf: small typo, I think this should be `targetsKeys[0]` (plural)

client/client_test.go
```diff
@@ -1686,6 +1718,187 @@ func TestPublishTargetsDelgationSuccessNeedsToDownloadRoles(t *testing.T) {
...
+	addTarget(t, repo1, "first", "../fixtures/root-ca.crt", "targets/a")
+	assert.NoError(t, repo1.Publish())
+	addTarget(t, repo2, "second", "../fixtures/root-ca.crt", "targets/a")
+	assert.NoError(t, repo2.Publish())
+
+	// first repo can publish again
+	addTarget(t, repo1, "third", "../fixtures/root-ca.crt", "targets/a")
+	assert.NoError(t, repo1.Publish())
+
+	// both should see both targets
```
riyazdf: nit: Should we switch "both" to "all" since there three targets?

riyazdf:
LGTM pending CI!

client/client_test.go
cyli: Yep, thanks for the catch!  :)  that was the error - I think I deleted it while switching windows :(

client/client_test.go
cyli: Definitely, thanks!

PR #387, opened 2015-12-23
Tests for backwards-compatibility reading/writing/exporting/importing old repo format

cyli:
As a test for #383


riyazdf:
awesome to have fixtures and docs around this, LGTM!

fixtures/compatibility/README.md
```diff
@@ -0,0 +1,17 @@
...
+	- GUN: `docker.io/notary0.1/samplerepo`
+	- key passwords: "randompass"
+	- targets:
+
+		```
+		  NAME                                DIGEST                                SIZE (BYTES)
+		------------------------------------------------------------------------------------------
+		  v1     454fd6f9e5a71c34902a673f66cc73be5890785b55014afb0eee6cbbe784cca9   243
+		```
+	- timestamp metadata has been removed - we have to rotating the timestamp key in order for the server to be able to sign the timestamp, and the `timestamp.json` generated by the server will be version 1, since as far as it knows this is a new repo.  If we kept the `timestamp.json` with the higher version number, we would not be able to publish.
```
diogomonica: "we have to rotating the timestamp"

diogomonica:
This is great. Thank you for taking the time to make sure we test the backwards behavior.

LGTM

fixtures/compatibility/README.md
cyli: Oops!  hehe, thanks!  :D

PR #389, opened 2015-12-24
Do not panic with invaild log level

HuKeping:
Since we have already specify a default log level for the project,
we can use it when coming across a invaild log level from configuration
file.

Besides, it is not a good idea to panic the whole project just because
of a wrong log level. Instead we can log it out and use the default one.

### DIFF
--- a/cmd/notary-server/main.go
+++ b/cmd/notary-server/main.go
@@ -190,10 +190,7 @@ func main() {
 	}
 
 	// default is error level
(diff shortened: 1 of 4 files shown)


endophage:
For server and signer this is intentional. When not setting a log level it is acceptable to use a default. However, when an explicit value is set, it must be a valid value. Your statement about panic is generally correct, but really applies when a system has reached a running state. It is acceptable to panic when problems are found during startup such that the running state should not be reached.

Does that make sense?

HuKeping:
Sure, closing this.

PR #390, opened 2015-12-25
Stop injecting to the helper function

HuKeping:
GetCryptoService is a helper function and we have injected a specific
http.ResponseWriter object for it to write back error message. Meanwhile
the caller for that function checks whether the cryptoService is nil
or not and return immediately if it is nil.

I think it's not a good idea to write back HTTP response in the helper function,
it's the caller's work and thus there is no need to inject the specific
ResponseWriter object into it.

### DIFF
--- a/signer/api/api.go
+++ b/signer/api/api.go
@@ -3,6 +3,7 @@ package api
 import (
 	"crypto/rand"
 	"encoding/json"
+	"fmt"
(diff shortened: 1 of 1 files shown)


diogomonica:
LGTM

cyli:
LGTM!

PR #391, opened 2015-12-26
Do not use strings.Compare

hqhq:
According to https://golang.org/src/strings/compare.go?s=491:520#L3
It's suggested to use the built-in string comparison operators.

And also, seems like this function does not exist in some arch
such as arm64. So we better not use it.

Signed-off-by: Qiang Huang <h.huangqiang@huawei.com>


endophage:
LGTM

PR #393, opened 2015-12-30
use path instead of filepath to express TUF roles

riyazdf:
Replaces `path/filepath` operations with just `path` operations when dealing with TUF roles because `path/filepath` is OS-specific (for Windows it will produce a different path, for example).  `path` functions will adhere to our TUF role format of `X/Y/Z`

Inspired by discussion in this PR https://github.com/docker/docker/pull/18887#discussion-diff-48414110L40

### DIFF
--- a/tuf/data/roles_test.go
+++ b/tuf/data/roles_test.go
@@ -1,7 +1,7 @@
 package data
 
 import (
-	"path/filepath"
+	"path"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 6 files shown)


tuf/data/roles_test.go
```diff
@@ -194,46 +194,46 @@ func TestErrInvalidRole(t *testing.T) {
...
+	assert.False(t, IsDelegation(path.Join(CanonicalTargetsRole, "level1")+"/"))
+	assert.False(t, IsDelegation(path.Join(CanonicalTargetsRole, "UpperCase")))
 
 	assert.False(t, IsDelegation(
-		filepath.Join(CanonicalTargetsRole, "directory")+"/../../traversal"))
+		path.Join(CanonicalTargetsRole, "directory")+"/../../traversal"))
 
 	assert.False(t, IsDelegation(
-		filepath.Join(CanonicalTargetsRole)+"///test/middle/slashes"))
+		path.Join(CanonicalTargetsRole)+"///test/middle/slashes"))
```
thaJeztah: is there actually something "joined" here?

tuf/data/roles_test.go
```diff
@@ -194,46 +194,46 @@ func TestErrInvalidRole(t *testing.T) {
...
 
 	assert.False(t, IsDelegation(
-		filepath.Join(CanonicalTargetsRole)+"/./././"))
+		path.Join(CanonicalTargetsRole)+"/./././"))
```
thaJeztah: same here; looks like there's nothing "joined"?

riyazdf: thanks for catching these, the `join` is definitely unnecessary so I've removed them

cyli:
Wondering if https://github.com/docker/notary/blob/28128c28af78ab176c96017b1f6864dd3ca2cc98/tuf/data/roles.go#L129 should also be replaced with `path.Clean` instead of `filepath.Clean`.

Also, https://github.com/docker/notary/blob/db9705dd97509e71adb1c74d466ccf5272b9990a/tuf/client/client.go#L513 calls `filepath.Dir` - should this also be changed to `path.Dir`?

cyli:
This otherwise LGTM!  :)

riyazdf:
I think the first [`filepath.Clean`](https://github.com/docker/notary/blob/28128c28af78ab176c96017b1f6864dd3ca2cc98/tuf/data/roles.go#L129) should indeed be `path.Clean` since we're only referring to the role itself here.  I'll update the PR 

My intuition on the [`filepath.Dir`](https://github.com/docker/notary/blob/db9705dd97509e71adb1c74d466ccf5272b9990a/tuf/client/client.go#L513 ) was that it's eventually being used for a json filepath, but this actually seems like we could have some incompatibilities with Windows paths.
I think you're right that we need to change this to `path.Dir` but I think we would then need to update the dir to have the correct [OS separators](https://golang.org/pkg/os/#pkg-constants) and the next [`path.Join`](https://github.com/docker/notary/blob/db9705dd97509e71adb1c74d466ccf5272b9990a/tuf/client/client.go#L518) to a `filepath.Join`, right?

endophage:
Outside of storage structs that specifically operate on the FS, we should probably be using `path` functions rather than `filepath` everywhere.

We should update any storage methods that need it to convert to the system specific path.

cyli:
If we want gun/targets/delegation to be translated to something like `C:\notary\..path\to\tuf\gun\targets\delegation.json`, we probably need to split by `/` and re-join with `filepath.Join` everywhere we do file stuff  :|

endophage:
@cyli I think we can just use this within the file storage methods: https://golang.org/pkg/path/filepath/#FromSlash

It converts from a `/` separated path to whatever the appropriate path separator is on the platform.

riyazdf:
@endophage I used `filepath.FromSlash` in the `RoleTargetsPath` method directly since the function description seems to imply that we should return a valid filepath.

We can handle this in the file storage code instead, if you'd prefer this function to return a `/` separated path regardless of OS?

endophage:
@riyazdf it looks like `RoleTargetsPath` is being used to generate URLs for HTTP requests (at least that's one of its functions) so it might be better to update the comment. Either way, it probably shouldn't be using `filepath.FromSlash` as that will generate incorrect URLs on Windows.

riyazdf:
@endophage that makes a lot of sense, I was confused with the function's description but after more digging I definitely agree with you that it'd be good to update the comment -- this is in my latest change

cyli:
LGTM!

diogomonica:
LGTM.

aardvark

PR #394, opened 2015-12-30
fix up style according to comments

riyazdf:
Style cleanup from @aaronlehmann's helpful comments!

### DIFF
--- a/tuf/data/roles.go
+++ b/tuf/data/roles.go
@@ -115,12 +115,10 @@ func ValidRole(name string) bool {
 
 // IsDelegation checks if the role is a delegation or a root role
 func IsDelegation(role string) bool {
-	targetsBase := fmt.Sprintf("%s/", ValidRoles[CanonicalTargetsRole])
+	targetsBase := ValidRoles[CanonicalTargetsRole] + "/"
 
-	whitelistedChars, err := regexp.MatchString("^[-a-z0-9_/]+$", role)
-	if err != nil {
-		return false
-	}
+	delegationRegexp := regexp.MustCompile("^[-a-z0-9_/]+$")
+	whitelistedChars := delegationRegexp.MatchString(role)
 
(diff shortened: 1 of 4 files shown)


tuf/data/roles.go
```diff
@@ -274,7 +272,9 @@ func mergeStrSlices(orig, new []string) []string {
 	}
 	for _, e := range new {
 		if !have[e] {
-			orig = append(orig, e)
+			origCopy := make([]string, len(orig))
+			copy(origCopy, orig)
+			orig = append(origCopy, e)
```
riyazdf: I'm not sure if this is the most go-esque way to do what you had in mind @aaronlehmann (using a copy to avoid slice corruption), please let me know if you think there's a better way to do this!

endophage: Move the first two lines above the loop, assign the append result into origCopy, then return origCopy (might also want to call it something else, like `merged`):

``` 
origCopy := make([]string, len(orig), len(orig) + len(new))
copy(origCopy, orig)
for _, e := range new {
    if !have[e] {
        origCopy = append(origCopy, e)
    }
}
return origCopy
```

This should be more efficient for large `new` slices

riyazdf: Thanks for the suggestion!  I'll change this

cyli:
LGTM

endophage:
LGTM pending codecov :-)

diogomonica:
LGTM

aaronlehmann:
Sorry I missed the opportunity to review this, as I was on vacation. I'll add a belated LGTM.

My only comment is that `delegationRegexp := regexp.MustCompile("^[-a-z0-9_/]+$")` could be moved to a global variable declaration, to avoid recompiling the regexp every time the function is called. This only matters if `IsDelegation` is called frequently. The error handling issue that I pointed out before has been fixed by using `MustCompile`, so whether it's called inside the function or not is just a matter of optimization.

PR #401, opened 2016-01-04
When listing targets, the role the target belongs to is also listed

cyli:
The output looks like:

```
  NAME                                DIGEST                                SIZE (BYTES)    ROLE    
----------------------------------------------------------------------------------------------------
  v1     454fd6f9e5a71c34902a673f66cc73be5890785b55014afb0eee6cbbe784cca9   243            targets  
```

I'm not sure I like the role in `Target` being ignored when adding - should there just be two objects?  An adding `Target` and a listing `TargetWithRole`?

(It's being ignored because otherwise it would be a much larger change to how `AddTargets` works, and also adding a target to multiple roles would be more tedious - I don't feel super strongly though, so if anyone thinks it should go the other way I'll be happy to make the change)

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -518,6 +518,35 @@ func TestAddTargetToTargetRoleByDefault(t *testing.T) {
 		[]string{data.CanonicalTargetsRole})
 }
 
+// Adding a target ignores the role of the target object, and goes by the
+// roles passed to AddTarget (or the default targets role, if no roles are passed)
+func TestAddTargetIgnoresTargetRole(t *testing.T) {
+	ts, _, _ := simpleTestServer(t)
+	defer ts.Close()
+
+	repo, _ := initializeRepo(t, data.ECDSAKey, "docker.com/notary", ts.URL, false)
+	defer os.RemoveAll(repo.baseDir)
+
+	targetObj, err := NewTarget("latest", "../fixtures/root-ca.crt")
+	assert.NoError(t, err, "error creating target")
+	targetObj.Role = "targets/whosits"
+
+	// if no roles are passed, the scope is targets, not whatever targetObj has
+	// specified
+	assert.NoError(t, repo.AddTarget(targetObj))
+	changes := getChanges(t, repo)
+	assert.Len(t, changes, 1)
+	assert.Equal(t, data.CanonicalTargetsRole, changes[0].Scope())
+
+	// if roles are passed, the scope is one of those roles, not whatever
+	// targetObj has specified
+	assert.NoError(t, repo.AddTarget(targetObj, "targets/one", "targets/two"))
+	changes = getChanges(t, repo)
+	assert.Len(t, changes, 3)
+	assert.Equal(t, "targets/one", changes[1].Scope())
+	assert.Equal(t, "targets/two", changes[2].Scope())
+}
+
 // Tests that adding a target to a repo or deleting a target from a repo,
 // with the given roles, makes a change to the expected scopes
 func testAddOrDeleteTarget(t *testing.T, repo *NotaryRepository, action string,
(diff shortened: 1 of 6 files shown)


client/client_test.go
```diff
@@ -910,6 +940,10 @@ func testListTarget(t *testing.T, rootType string) {
 
 	sort.Stable(targetSorter(targets))
 
+	// the targets should both be find in the targets role
```
riyazdf: nit: comment should read `found`

riyazdf:
Awesome feature and tests!  I'm ok with not adding roles immediately upon adding targets but I'm open to discussion -- I guess I'm not sure if there are other cases where we'd want to have this information ready right after adding the target?

I have two super small non-blocking nits, where the comment should be changed from "two targets" to "four targets" (outside the diff but would be nice to fix)
https://github.com/docker/notary/pull/401/files#diff-fce84029d346504e44323c30d21a3808L985
https://github.com/docker/notary/pull/401/files#diff-fce84029d346504e44323c30d21a3808R1040

LGTM pending codecov!

client/client_test.go
```diff
@@ -518,6 +518,35 @@ func TestAddTargetToTargetRoleByDefault(t *testing.T) {
...
+func TestAddTargetIgnoresTargetRole(t *testing.T) {
+	ts, _, _ := simpleTestServer(t)
+	defer ts.Close()
+
+	repo, _ := initializeRepo(t, data.ECDSAKey, "docker.com/notary", ts.URL, false)
+	defer os.RemoveAll(repo.baseDir)
+
+	targetObj, err := NewTarget("latest", "../fixtures/root-ca.crt")
+	assert.NoError(t, err, "error creating target")
+	targetObj.Role = "targets/whosits"
```
diogomonica: should it be targets/whosisit?

cyli: Heh, sorry, this is a misremembered [character from A Wrinkle In Time](https://en.wikipedia.org/wiki/A_Wrinkle_in_Time#Mrs_Whatsit) (I got the name wrong).  Will change it to `targets/whoisit` :)

cyli:
@riyazdf Done!

cmd/notary/prettyprint_test.go
```diff
@@ -167,9 +167,9 @@ func TestPrettyPrintSortedTargets(t *testing.T) {
 		assert.NoError(t, err)
 	}
 	unsorted := []*client.Target{
-		{Name: "zebra", Hashes: data.Hashes{"sha256": hashes[0]}, Length: 8},
-		{Name: "abracadabra", Hashes: data.Hashes{"sha256": hashes[1]}, Length: 1},
-		{Name: "bee", Hashes: data.Hashes{"sha256": hashes[2]}, Length: 5},
+		{Name: "zebra", Role: "targets/b", Hashes: data.Hashes{"sha256": hashes[0]}, Length: 8},
+		{Name: "abracadabra", Role: "targets", Hashes: data.Hashes{"sha256": hashes[1]}, Length: 1},
```
diogomonica: I don't know what kind of animal abracadabra is. Can you provide a photo evidence of its existence? Not-LGTM otherwise.

diogomonica:
The ignoring the role in target isn't great :(

@endophage ideas on the best path forward? I'm inclined to fixing the API of AddTarget.

cmd/notary/prettyprint_test.go
cyli: It's a unicorn:  http://www.amazon.com/Douglas-Cuddle-Toys-343-Abracadbra/dp/B000C24LPG

cyli:
@diogomonica Also, `RemoveTarget` takes a list of roles, so changing `AddTarget` to not take roles and only add a target to one role at a time would make them slightly inconsistent.  It doesn't bother me that much though.

Alternately, we have `ListTargets` and `GetTargetByName` return a `FoundTarget` object which has a `Target` embedded and an extra `Role` field.

diogomonica:
The alternative seems better actually. Having a role a specific artifact of a FoundTarget object.

endophage:
I vote for `TargetWithRole` (or `FoundTarget` if we prefer that name, the desired functionality seems the same) as a struct that embeds `Target` and adds a role field.

riyazdf:
I also like `TargetWithRole` and prefer that name over `FoundTarget` for clarity, especially for integrations with docker

riyazdf:
I like the embedding, LGTM!

diogomonica:
LGTM. This Aardvark says: ship it! aardvark

PR #413, opened 2016-01-05
fixing bootstrapClient to prefer cached root

endophage:
Necessary precursor to making root rotations workable. Currently we always download the root.json, regardless of whether we have it in cache. This changes that logic to only download if we don't have one in cache. A newer root may then be downloaded later, as part of a repo update, if a problem is detected per the TUF update rules.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -774,6 +774,12 @@ func (r *NotaryRepository) bootstrapClient(checkInitialized bool) (*tufclient.Cl
 			if err != nil {
 				return nil, err
 			}
+
+			err = r.fileStore.SetMeta("root", tmpJSON)
+			if err != nil {
+				// if we can't write cache we should still continue, just log error
+				logrus.Errorf("could not save root to cache: %s", err.Error())
+			}
 		}
 	}
 
--- a/client/client.go
+++ b/client/client.go
@@ -737,6 +737,11 @@ func (r *NotaryRepository) Update(forWrite bool) (*tufclient.Client, error) {
 	return c, nil
 }
 
+// bootstrapClient attempts to bootstrap a root.json to be used as the trust
+// anchor for a repository. The checkInitialized argument indicates whether
+// we should always attempt to contact the server to determine if the repository
+// is initialized or not. If set to true, we will always attempt to download
+// and return an error if the remote repository errors.
 func (r *NotaryRepository) bootstrapClient(checkInitialized bool) (*tufclient.Client, error) {
 	var (
 		rootJSON   []byte
--- a/client/client.go
+++ b/client/client.go
@@ -420,7 +420,7 @@ func (r *NotaryRepository) RemoveTarget(targetName string, roles ...string) erro
 // subtree and also the "targets/x" subtree, as we will defer parsing it until
 // we explicitly reach it in our iteration of the provided list of roles.
 func (r *NotaryRepository) ListTargets(roles ...string) ([]*TargetWithRole, error) {
-	_, err := r.Update()
+	_, err := r.Update(false)
 	if err != nil {
 		return nil, err
 	}
@@ -480,7 +480,7 @@ func (r *NotaryRepository) listSubtree(targets map[string]*TargetWithRole, role 
 // will be returned
 // See the IMPORTANT section on ListTargets above. Those roles also apply here.
 func (r *NotaryRepository) GetTargetByName(name string, roles ...string) (*TargetWithRole, error) {
-	c, err := r.Update()
+	c, err := r.Update(false)
 	if err != nil {
 		return nil, err
 	}
@@ -515,7 +515,7 @@ func (r *NotaryRepository) GetChangelist() (changelist.Changelist, error) {
 func (r *NotaryRepository) Publish() error {
 	var initialPublish bool
 	// update first before publishing
-	_, err := r.Update()
+	_, err := r.Update(true)
 	if err != nil {
 		// If the remote is not aware of the repo, then this is being published
 		// for the first time.  Try to load from disk instead for publishing.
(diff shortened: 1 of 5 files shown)


client/client.go
```diff
@@ -744,32 +744,36 @@ func (r *NotaryRepository) saveMetadata(ignoreSnapshot bool) error {
...
-			return nil, err
+func (r *NotaryRepository) bootstrapClient(checkInitialized bool) (*tufclient.Client, error) {
+	var (
+		rootJSON []byte
+		err      error
+	)
+	// try to read root from cache first. We will trust this root
+	// until we detect a problem during update which will cause
+	// us to download a new root and perform a rotation.
+	rootJSON, cacheErr := r.fileStore.GetMeta("root", maxSize)
```
diogomonica: We need to check the validity of this root.

diogomonica:
@cyli would love your input around tests we can add here. Specifically the logic around when roots are downloaded, and if roots are downloaded when they expire, or the behavior when a invalid root is taken from cache, or when an invalid root comes from remote.

My major concerns are situations where the client never recovers.

client/client.go
```diff
@@ -744,53 +744,49 @@ func (r *NotaryRepository) saveMetadata(ignoreSnapshot bool) error {
 	return r.fileStore.SetMeta(data.CanonicalSnapshotRole, snapshotJSON)
 }
 
-func (r *NotaryRepository) bootstrapClient() (*tufclient.Client, error) {
-	var rootJSON []byte
-	remote, err := getRemoteStore(r.baseURL, r.gun, r.roundTrip)
-	if err == nil {
-		// if remote store successfully set up, try and get root from remote
-		rootJSON, err = remote.GetMeta("root", maxSize)
+func (r *NotaryRepository) bootstrapClient(checkInitialized bool) (*tufclient.Client, error) {
```
riyazdf: It'd be great if you could add a comment about `checkInitialized`, if possible?

endophage: Done!

client/client.go
```diff
@@ -804,6 +811,28 @@ func (r *NotaryRepository) bootstrapClient() (*tufclient.Client, error) {
 	), nil
 }
 
+// validateRoot MUST only be used during bootstrapping. It will only validate
+// signatures of the root based on known keys, not expiry or other metadata.
+// This is so that an out of date root can be loaded to be used in a rotation
+// should the TUF update process detect a problem.
+func (r *NotaryRepository) validateRoot(rootJSON []byte) (*data.SignedRoot, error) {
```
cyli: Non-blocking nitpick:  maybe it can be called `validateRootSignatures`, given the comment?

client/client.go
```diff
@@ -744,53 +744,60 @@ func (r *NotaryRepository) saveMetadata(ignoreSnapshot bool) error {
...
-			// the store and it doesn't know about the repo.
-			return nil, err
-		}
-		result, cacheErr := r.fileStore.GetMeta("root", maxSize)
-		if cacheErr != nil {
-			// if cache didn't return a root, we cannot proceed - just return
-			// the original error.
+	remote, remoteErr := getRemoteStore(r.baseURL, r.gun, r.roundTrip)
+	if remoteErr != nil {
+		logrus.Error(remoteErr)
```
cyli: Should there be a return here?  But also, wondering if this should be done inside the `if cachedRootErr != nil || checkInitialized` because the URL does not need to be checked unless we are definitely going to download?

cyli: Oh wait, nm about being inside the `if`, sorry.  Forgot about the new client needing the remote.  :)

cyli: Could we also add a test where we create a new notary repository with an invalid URL and show that it fails to `ListTargets`/`GetTargetByName`/`Publish`?

cyli:
@diogomonica Those would be great unit tests for bootstrapClient.  Although a lot of them are sort of indirectly tested through the tests for `Publish`, `ListTargets`, and `GetTargetsByName`.

- replacing a corrupt/non-existent root cache with a downloaded remote root is tested by `TestPublishUninitializedRepo`, `TestPublishRootCorrupt`
- downloading roots if there is no cache is tested by most of the tests that create a new repo to list (most of the publish tests, which include tests for list and get target by name)
- not being able to download a root from remote if there is no cache is tested by `TestRemoteServerUnavailableNoLocalCache`.

It looks like we are missing explicit tests the cases where (although there seems to be coverage for it - I'm not sure where it's coming from):

- local root may be corrupt, so we download a remote root, but it is corrupt/has signatures that do not match our existing trust certificates
- local root is expired, but it can still be used as the trust anchor

@endophage Do you know if we have those tests anywhere else?

client/client.go
cyli: Yay, thanks for adding the test where the URL is valid but not a server.  And apologies, I wasn't clear:  I actually meant if the url is actually invalid.  E.g. something like "hello" - if it's invalid, `getRemoteStore` should fail with an error, and I think we might get a panic because `remote` is `nil`, since we only log the `remoteErr` rather than return.

endophage: Good call. Test added and required some changes.

client/client.go
```diff
@@ -745,53 +745,64 @@ func (r *NotaryRepository) saveMetadata(ignoreSnapshot bool) error {
...
-	// can't just unmarshal into SignedRoot because validate root
-	// needs the root.Signed field to still be []byte for signature
-	// validation
-	root := &data.Signed{}
-	err = json.Unmarshal(rootJSON, root)
-	if err != nil {
-		return nil, err
-	}
+		if cachedRootErr != nil {
+			// if we entered the remote download code because there was a cache
```
cyli: Non-blocking nit:  this comment is a little confusing, because we'd enter this block if there was a cache error whether or not we wanted to check the initialization setting.  Maybe something like "we only want to use the downloaded data as our root if there was a cache error"?

cyli:
LGTM!

endophage:
Need to rebase against @cyli's PR that modified Publish/Update which is going to make it into master first. That may have an effect on this.

riyazdf:
LGTM pending rebase

endophage:
@cyli @riyazdf @diogomonica rebase done

cyli:
:+1:

riyazdf:
:+1:

client/client.go
```diff
@@ -797,6 +808,28 @@ func (r *NotaryRepository) bootstrapClient() (*tufclient.Client, error) {
...
+	// can't just unmarshal into SignedRoot because validate root
+	// needs the root.Signed field to still be []byte for signature
+	// validation
+	root := &data.Signed{}
+	err := json.Unmarshal(rootJSON, root)
+	if err != nil {
+		return nil, err
+	}
+
+	err = r.CertManager.ValidateRoot(root, r.gun)
```
diogomonica: We should probably rename this `ValidateRoot` too.

PR #416, opened 2016-01-06
removing the ability to configure role names. 

### LINKED ISSUE #372: Remove role name translation
https://github.com/docker/notary/blob/master/tuf/data/roles.go#L20 has a mapping of canonical role names to custom role names, the intent originally being users could set their own role name mappings.

However, this is not currently used consistently across the codebase, and it would add a lot of extra complexity to fully support (a ton of extra tests, etc.)

However, the map itself serves a useful function as a set of all the valid canonical roles that we can easily match another role against and iterate over.  Possibly instead of removing the map, we can:

1. convert it to a `map[string]interface{}` or something, so we can key into it and iterate over it
1. convert it to a list that can be iterated over, and add a helper function `IsPrimaryRole` or `IsCanonicalRole` or something (e.g. not a delegation)

We should probably just remove `SetValidRoles` either way.

It's a small fix but a big change to update all the usages of `ValidRoles` all over the codebase.

endophage:
It adds a lot of complexity without adding much value. If somebody wants custom role names they can implement it at the display level.

Closes #372 

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -393,7 +393,7 @@ func requireRepoHasExpectedMetadata(t *testing.T, repo *NotaryRepository,
 		require.Len(t, decodedRoot.Roles, len(data.ValidRoles),
 			"wrong number of roles in root.json")
 
-		for role := range data.ValidRoles {
+		for _, role := range data.ValidRoles {
 			_, ok := decodedRoot.Roles[role]
 			require.True(t, ok, "Missing role %s in root.json", role)
 		}
@@ -1172,7 +1172,7 @@ func testPublishNoData(t *testing.T, rootType string, serverManagesSnapshot bool
 	require.NoError(t, err)
 	require.Empty(t, targets)
 
-	for role := range data.ValidRoles {
+	for _, role := range data.ValidRoles {
 		// we don't cache timstamp metadata
 		if role != data.CanonicalTimestampRole {
 			requireRepoHasExpectedMetadata(t, repo2, role, true)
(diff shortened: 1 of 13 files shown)


cyli:
If we don't have a function to normalize roles, should we then also remove this check https://github.com/docker/notary/pull/416/files#diff-a84805db69378bc67da7d4ae04da989aR57 that lowercases a role before determining if it's valid?

cyli:
Also, does this mean that `data.ValidRoles` is more of an easy way to enumerate the top level roles?  If so, should it be renamed?  Also, should then all invocations of `data.ValidRoles["root"]` or `data.ValidRoles[data.CanonicalRootRole]` be changed to just `data.CanonicalRootRole` (or other top level roles)?

endophage:
@cyli re. `ValidRole` check - yeah, that's a good idea, in the spirit of being strict we should definitely remove the lowercase.

re. `data.ValidRoles` - yes, I was trying to take out the references to `data.ValidRoles[role]` but I'm sure there are still some hanging around. It does essentially just become easy enumeration to be used in the `ValidRole` check. I should probably just switch it to be a slice.

cyli:
LGTM!

riyazdf:
I'm open to the idea of renaming `data.ValidRoles` to indicate top level roles only since the `ValidRole` function also checks for delegation roles.  What do we think?

endophage:
@riyazdf @cyli how about `BaseRoles` in place of `ValidRoles`? Now that roles aren't configurable it seems like a suitable name.

cyli:
@endophage +1

riyazdf:
@endophage +1

PR #418, opened 2016-01-06
Fix here document

HuKeping:
It seems the prompt doesn't work here.


cyli:
Awesome, thanks for noticing and fixing!

PR #419, opened 2016-01-06
improve message for ErrMetaNotFound

riyazdf:
Improve the error message for `ErrMetaNotFound` by always specifying the missing resource in the error message string.  This is the first part of https://github.com/docker/docker/issues/15463 (the other errors are from the docker repo)

### DIFF
--- a/tuf/client/client.go
+++ b/tuf/client/client.go
@@ -272,7 +272,7 @@ func (c *Client) downloadTimestamp() error {
 			if err == nil {
 				// couldn't retrieve data from server and don't have valid
 				// data in cache.
-				return store.ErrMetaNotFound{}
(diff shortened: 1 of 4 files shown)


cyli:
LGTM - thanks for noticing that issue in docker and fixing!  I'm wondering if maybe `ErrMetaNotFound` should also be added to https://github.com/docker/docker/blob/master/api/client/trust.go#L267 and translated the same as the JSON syntax error?

Is the codecov issue just a mismatch between the base of this branch and master, maybe?

riyazdf:
@cyli Sounds good, I'm also thinking it would be more helpful to add there.  I was planning to also improve those error messages in a separate docker commit so I'll include it then

As for codecov, I'll try a rebase.

endophage:
LGTM

PR #421, opened 2016-01-06
Update jose2go to latest to fix #239

### LINKED ISSUE #239: update jose2go dependency to prevent go build errors
When building the Dockerfile from the https://github.com/docker/docker/blob/release/v1.9/Dockerfile#L143 on an ARM server, notary fails to build as described here: https://github.com/dvsekhvalnov/jose2go/issues/8

That's due to an issue in jose2go which has been fixed in the meantime.

Could you please update jose2go to the latest version?

Thanks.

PS: Do you have any clue why the notary version that is included in the Dockerfile https://github.com/docker/docker/blob/release/v1.9/Dockerfile#L143 is fixed to such an old commit?

justincormack:
This seems to have never been applied; fixes Arm build

Signed-off-by: Justin Cormack <justin.cormack@unikernel.com>


cyli:
LGTM!

Issue #366, opened 2015-12-15
Reserialize metadata JSON bytes before storing in DB


cyli:
This way if the user sends us extra fields (either just garbage or their own tuf extensions), only the data we support will be stored in the DB.

endophage:
I've found 2 problems with doing this:

1. It turns out clients aren't using canonical json for the `data.Signed` -> `[]byte` final serialization. This makes it problematic to consistently reserialize on the server. I've addressed this in #426

2. If a client is managing the snapshot key itself, we can't (or at least there's no point) reserialize targets/root JSON because if it changes we won't be able to update the hash/size in the snapshot file.

mtrmac:
If the server drops JSON fields received from the author, and therefore cannot reproduce the same JSON when sending it to other clients, AFAICS the author’s signature cannot validate. What am I missing?

endophage:
@mtrmac that is also true. I should close this, we're not looking to do it any more, but I also need to open another issue to track fixing the canonical serialization problem (we can't reserialize but we have determined a path to get to canonical serialization for all TUF data.

PR #426, opened 2016-01-06
use canonical json for all marshalling of TUF data

endophage:
Related to #366.

It's safe to change serialization because all deserialization is done AFTER checking hashes and sizes.


cyli:
LGTM!  Thanks for catching this!  And the `require` as `assert` should make the test output much easier to read.  :)  

The only other non-test place I see `json.Marshal` being used is in `tuf/encrypted/encrypted.go` which I don't think is used anywhere.

riyazdf:
LGTM!

PR #427, opened 2016-01-07
Move regex compilation to outside var for optimization

riyazdf:
Moves regex compilation to outer variable as an optimization, since `IsDelegation` is called in quite a few places.  From @aaronlehmann's comment in https://github.com/docker/notary/pull/394#issuecomment-169396464


aaronlehmann:
LGTM

cyli:
LGTM!

PR #429, opened 2016-01-07
using JSON logging format

endophage:
This defaults notary server and signer to use JSON as its output format for logging so that we can integrate it with Heka/Kibana.

It's possible to revert to the logfmt output by passing the CLI flag `-logf=logfmt` (default is `-logf=json`)

cc @diogomonica 


diogomonica:
LGTM

riyazdf:
LGTM!

PR #430, opened 2016-01-07
Fix the exposed port in Dockerfile

HuKeping:
4443 is default used by Server and 4444 is for Signer.


cyli:
Thanks for noticing this and fixing!

PR #431, opened 2016-01-07
added /usr/lib64 to search paths

useidel:
On some x86_64 Linux distributions the libaries are installed in /usr/lib64, e.g. Fedora. Without that notary does not work with the Yubikey on these Linux distributions. Adding that path to the search list fixes that issue and makes workrounds like copying/installing the libraries in /usr/lib or /usr/local/lib unnecessary.

### DIFF
--- a/trustmanager/yubikey/pkcs11_linux.go
+++ b/trustmanager/yubikey/pkcs11_linux.go
@@ -4,6 +4,7 @@ package yubikey
 
 var possiblePkcs11Libs = []string{
 	"/usr/lib/libykcs11.so",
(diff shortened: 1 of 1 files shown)


useidel:
@GordonTheTurtle: Ah .. ok .. will do so ..

thaJeztah:
@useidel in future, please don't open a new PR, but update the existing PR using the instructions that @GordonTheTurtle showed.

useidel:
Yes ... sure. Misread the instructions at the first and then I could not re-opened the already closed PR. Sorry for that.

PR #432, opened 2016-01-07
added /usr/lib64 to search paths

useidel:
On some x86_64 Linux distributions the libaries are installed in /usr/lib64, e.g. Fedora. Without that notary does not work with the Yubikey on these Linux distributions. Adding that path to the search list fixes that issue and makes workrounds like copying/installing the libraries in /usr/lib or /usr/local/lib unnecessary.

Signed-off-by: Udo Seidel <udoseidel@gmx.de>


cyli:
LGTM!  Thank you for your contribution!

PR #433, opened 2016-01-07
Verbose health check information

HuKeping:
When I deploy Notary, the health check of Notary-Server began to warn me:
```
level=error msg="Trust not fully operational: Trust is not healthy"
```  
I think the error message is not enough to find the problem, I spent much time to figured out that was because I forget to start the MySQL for Notary-Signer.

With this PR, if the same thing happened, the error message will be quite useful:
```
level=error msg="Trust not fully operational: Trust is not healthy: map[DB operational:Cannot access table: private_keys]"
```

### DIFF
--- a/signer/client/signer_trust.go
+++ b/signer/client/signer_trust.go
@@ -185,11 +185,16 @@ func (trust *NotarySigner) CheckHealth(timeout time.Duration) error {
 	status, err := trust.kmClient.CheckHealth(ctx, &pb.Void{})
 	defer cancel()
 	if err == nil && len(status.Status) > 0 {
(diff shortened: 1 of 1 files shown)


cyli:
Thank you for improving our error logging!  Since the status is just a `map[string]string` would it be possible to format them so it looks nicer than just a map?  Something like:

"Trust not fully operational: Cannot access table: private_keys; <next error message>; <next error message"?

HuKeping:
Sure, done!

cyli:
Thanks!  LGTM!

riyazdf:
LGTM!

PR #438, opened 2016-01-09
adding sha256 column to tuf_files database table

### LINKED ISSUE #437: Add sha256 to tuf_files table
Given a timestamp we do not know the version of the snapshot. To support "download all" for old versions, add a sha256 column to the database such that given a timestamp file, we can easily query for the snapshot it references (and similar for snapshot -> targets/root).

endophage:
Closes #437 

### DIFF
--- a/migrations/server/mysql/0003_add_sha256_tuf_files.up.sql
+++ b/migrations/server/mysql/0003_add_sha256_tuf_files.up.sql
@@ -1,0 +1,5 @@
+ALTER TABLE `tuf_files` ADD COLUMN `sha256` CHAR(64) DEFAULT NULL, ADD INDEX `sha256` (`sha256`);
+
+-- SHA2 function takes the column name or a string as the first parameter, and the 
+-- hash size as the second argument. It returns a hex string.
+UPDATE `tuf_files` SET `sha256` = SHA2(`data`, 256);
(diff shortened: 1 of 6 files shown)


riyazdf:
LGTM, thanks for updating the tests to `require` too :+1:

migrations/server/mysql/0003_add_sha256_tuf_files.up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE `tuf_files` ADD COLUMN `sha256` CHAR(64) DEFAULT NULL, ADD INDEX `sha256` (`sha256`);
+
+UPDATE `tuf_files` SET `sha256` = SHA2(`data`, 256);
```
riyazdf: nit: might be helpful to link to the [docs for SHA2](https://dev.mysql.com/doc/refman/5.5/en/encryption-functions.html#function_sha2), so it's clear that it's compatible with the golang sha256 -- I have less experience with mysql so I had to look it up, but it might be clearer for others.

endophage: Will do

diogomonica:
LGTM

PR #439, opened 2016-01-11
Log: improve error logging

HuKeping:
When pushing an image to Distribution server with Notary enabled at the
first time, client will try to get the root.json and it will fail since
there is nothing in Notary yet.

This patch enhances the error message by adding the error detail.
- "404 GET notary-distribution:5678/busybox:root"

And also update some comments

### DIFF
--- a/server/handlers/default.go
+++ b/server/handlers/default.go
@@ -129,7 +129,7 @@ func getHandler(ctx context.Context, w http.ResponseWriter, r *http.Request, var
 	out, err := store.GetCurrent(gun, tufRole)
 	if err != nil {
 		if _, ok := err.(storage.ErrNotFound); ok {
-			logrus.Error("404 GET " + gun + ":" + tufRole)
+			logrus.Errorf("404 GET %s:%s, error: %v", gun, tufRole, err)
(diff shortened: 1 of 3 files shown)


server/handlers/default.go
```diff
@@ -129,7 +129,7 @@ func getHandler(ctx context.Context, w http.ResponseWriter, r *http.Request, var
 	out, err := store.GetCurrent(gun, tufRole)
 	if err != nil {
 		if _, ok := err.(storage.ErrNotFound); ok {
-			logrus.Error("404 GET " + gun + ":" + tufRole)
+			logrus.Error("404 GET " + gun + ":" + tufRole + ", error:" + err.Error())
```
endophage: You did the right thing preserving the existing code format, but while you're here, could you switch this over to `logrus.Errorf("404 GET %s:%s, error: %v", gun, tufRole, err)`

HuKeping: sure, done.

endophage:
LGTM. Thanks!

riyazdf:
LGTM!

PR #440, opened 2016-01-11
delegation command for notary-cli

### LINKED ISSUE #434: Add delegate command to notary-cli
Add the ability to use the `notary delegation` to delegate a particular Public key in X509 or DER formats to a particular role.

```
➜  notary git:(diogo-cli-adding-delegations) ✗ bin/notary delegation
Operations on TUF delegations.

Usage:
  notary delegation [command]

Available Commands:
  list        Lists delegations for the Global Unique Name.
  remove      Remove a Role delegation for the KeyID.
  add         Add a Role delegation for the provided public key (X509 or PEM encoded public keys).
```

riyazdf:
Adds a `notary delegation` command to the cli.  There are three commands:
- `add`: adds a delegation using a given gun, X509 cert(s) with public key, role (ex: "targets/delegation"), and optional comma-separated `--paths`. Currently we only validate the expiry date on the certificate since there are no restrictions on the signature, but it could be backed by the TUF root key or existing PKI in the future. This command creates a TUF change offline, which will need to be published to take effect.
- `remove`: removes delegation info using a gun and role with either key IDs (standard arguments) or comma-separated `--paths`.  The command can also just remove all information for the specified role using a `--force` flag (this requires confirmation).  Creates a TUF change offline (update or delete, depending on `--force`) that will require a publish for effect.
- `list`: updates the notary repository online before listing all published delegations in a table format.

Closes #434.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -353,7 +353,7 @@ func (r *NotaryRepository) AddDelegation(name string, threshold int,
 // the repository when the changelist gets applied at publish time.
 // This does not validate that the delegation exists, since one might exist
 // after applying all changes.
-func (r *NotaryRepository) RemoveDelegation(name string) error {
+func (r *NotaryRepository) RemoveDelegation(name string, keyIDs, paths []string, removeAll bool) error {
 
 	if !data.IsDelegation(name) {
 		return data.ErrInvalidRole{Role: name, Reason: "invalid delegation role name"}
@@ -366,14 +366,35 @@ func (r *NotaryRepository) RemoveDelegation(name string) error {
 	defer cl.Close()
 
 	logrus.Debugf(`Removing delegation "%s"\n`, name)
+	var template *changelist.TufChange
 
-	template := changelist.NewTufChange(
-		changelist.ActionDelete,
-		name,
-		changelist.TypeTargetsDelegation,
-		"", // no path
-		nil,
-	)
+	// We use the Delete action only for force removal, Update is used for removing individual keys and paths
+	if removeAll {
+		template = changelist.NewTufChange(
+			changelist.ActionDelete,
+			name,
+			changelist.TypeTargetsDelegation,
+			"",  // no path
+			nil, // deleting role, no data needed
+		)
+
+	} else {
+		tdJSON, err := json.Marshal(&changelist.TufDelegation{
+			RemoveKeys:  keyIDs,
+			RemovePaths: paths,
+		})
+		if err != nil {
+			return err
+		}
+
+		template = changelist.NewTufChange(
+			changelist.ActionUpdate,
+			name,
+			changelist.TypeTargetsDelegation,
+			"", // no path
+			tdJSON,
+		)
+	}
 
 	return addChange(cl, template, name)
 }
@@ -513,6 +534,45 @@ func (r *NotaryRepository) GetChangelist() (changelist.Changelist, error) {
 		return nil, err
 	}
 	return cl, nil
+}
+
+// GetDelegationRoles returns the keys and roles of the repository's delegations
+func (r *NotaryRepository) GetDelegationRoles() ([]*data.Role, error) {
+	// Update state of the repo to latest
+	if _, err := r.Update(false); err != nil {
+		return nil, err
+	}
+
+	// All top level delegations (ex: targets/level1) are stored exclusively in targets.json
+	targets, ok := r.tufRepo.Targets[data.CanonicalTargetsRole]
+	if !ok {
+		return nil, store.ErrMetaNotFound{data.CanonicalTargetsRole}
+	}
+
+	allDelegations := targets.Signed.Delegations.Roles
+
+	// make a copy for traversing nested delegations
+	delegationsList := make([]*data.Role, len(allDelegations))
+	copy(delegationsList, allDelegations)
+
+	// Now traverse to lower level delegations (ex: targets/level1/level2)
+	for len(delegationsList) > 0 {
+		// Pop off first delegation to traverse
+		delegation := delegationsList[0]
+		delegationsList = delegationsList[1:]
+
+		// Get metadata
+		delegationMeta, ok := r.tufRepo.Targets[delegation.Name]
+		// If we get an error, don't try to traverse further into this subtree because it doesn't exist or is malformed
+		if !ok {
+			continue
+		}
+
+		// Add nested delegations to return list and exploration list
+		allDelegations = append(allDelegations, delegationMeta.Signed.Delegations.Roles...)
+		delegationsList = append(delegationsList, delegationMeta.Signed.Delegations.Roles...)
+	}
+	return allDelegations, nil
 }
 
 // Publish pushes the local changes in signed material to the remote notary-server
--- a/cmd/notary/delegations.go
+++ b/cmd/notary/delegations.go
@@ -1,0 +1,212 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+
+	"github.com/docker/notary"
+	notaryclient "github.com/docker/notary/client"
+	"github.com/docker/notary/passphrase"
+	"github.com/docker/notary/trustmanager"
+	"github.com/docker/notary/tuf/data"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+var cmdDelegationTemplate = usageTemplate{
+	Use:   "delegation",
+	Short: "Operates on delegations.",
+	Long:  `Operations on TUF delegations.`,
+}
+
+var cmdDelegationListTemplate = usageTemplate{
+	Use:   "list [ GUN ]",
+	Short: "Lists delegations for the Global Unique Name.",
+	Long:  "Lists all delegations known to notary for a specific Global Unique Name.",
+}
+
+var cmdDelegationRemoveTemplate = usageTemplate{
+	Use:   "remove [ GUN ] [ Role ] <KeyID 1> ...",
+	Short: "Remove KeyID(s) from the specified Role delegation.",
+	Long:  "Remove KeyID(s) from the specified Role delegation in a specific Global Unique Name.",
+}
+
+var cmdDelegationAddTemplate = usageTemplate{
+	Use:   "add [ GUN ] [ Role ] <PEM file path 1> ...",
+	Short: "Add a keys to delegation using the provided public key certificate PEMs.",
+	Long:  "Add a keys to delegation using the provided public key certificate PEMs in a specific Global Unique Name.",
+}
+
+var paths []string
+var removeAll, removeYes bool
+
+type delegationCommander struct {
+	// these need to be set
+	configGetter func() *viper.Viper
+	retriever    passphrase.Retriever
+}
+
+func (d *delegationCommander) GetCommand() *cobra.Command {
+	cmd := cmdDelegationTemplate.ToCommand(nil)
+	cmd.AddCommand(cmdDelegationListTemplate.ToCommand(d.delegationsList))
+
+	cmdRemDelg := cmdDelegationRemoveTemplate.ToCommand(d.delegationRemove)
+	cmdRemDelg.Flags().StringSliceVar(&paths, "paths", nil, "List of paths to remove")
+	cmdRemDelg.Flags().BoolVarP(&removeYes, "yes", "y", false, "Answer yes to the removal question (no confirmation)")
+	cmd.AddCommand(cmdRemDelg)
+
+	cmdAddDelg := cmdDelegationAddTemplate.ToCommand(d.delegationAdd)
+	cmdAddDelg.Flags().StringSliceVar(&paths, "paths", nil, "List of paths to add")
+	cmd.AddCommand(cmdAddDelg)
+	return cmd
+}
+
+// delegationsList lists all the delegations for a particular GUN
+func (d *delegationCommander) delegationsList(cmd *cobra.Command, args []string) error {
+	if len(args) != 1 {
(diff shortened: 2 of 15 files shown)


cmd/notary/delegations.go
```diff
@@ -0,0 +1,172 @@
...
+
+	// no online operations are performed by add so the transport argument
+	// should be nil
+	nRepo, err := notaryclient.NewNotaryRepository(config.GetString("trust_dir"), gun, getRemoteTrustServer(config), nil, retriever)
+	if err != nil {
+		fatalf(err.Error())
+	}
+
+	// Add the delegation to the repository
+	err = nRepo.AddDelegation(role, 1, []data.PublicKey{pubKey}, paths)
```
NathanMcCauley: What is `1` in this context? Can we make a const with a descriptive name?

trustmanager/x509utils.go
```diff
@@ -299,6 +299,39 @@ func ParsePEMPrivateKey(pemBytes []byte, passphrase string) (data.PrivateKey, er
...
+	default:
+		return nil, fmt.Errorf("unsupported PEM block type %q, expected certificate", pemBlock.Type)
+	}
+}
+
+// ValidateCertificate returns an error if the certificate is not valid for notary
+// Currently, this is only a time expiry check
+func ValidateCertificate(c *x509.Certificate) error {
+	now := time.Now()
+	if now.Before(c.NotBefore) || now.After(c.NotAfter) {
```
NathanMcCauley: Sometimes makes sense to have a bit of a grace period on newly created certs, in case it was just created. Maybe give it a day of leeway on creation time?

cmd/notary/delegations.go
riyazdf: This is the number of thresholds, I can definitely make a const to make things clearer.  We could also try to make this configurable, though I don't think we've surfaced a way to add keys to non-root roles or configure role thresholds at the moment

trustmanager/x509utils.go
riyazdf: Good idea, will do

client/client.go
```diff
@@ -509,6 +509,58 @@ func (r *NotaryRepository) GetChangelist() (changelist.Changelist, error) {
...
+// GetDelegationRoles returns the keys and roles of the repository's delegations
+func (r *NotaryRepository) GetDelegationRoles() ([]*data.Role, error) {
+	// Update state of the repo to latest
+	_, err := r.Update()
+	if err != nil {
+		return nil, err
+	}
+
+	// All top level delegations (ex: targets/level1) are stored exclusively in targets.json
+	targetsJSON, err := r.fileStore.GetMeta("targets", 0)
```
endophage: After calling `r.Update()`, the validated repository is already de-serialized into `r.tufRepo`. No need to load the files from the filestore.

cmd/notary/delegations.go
```diff
@@ -0,0 +1,174 @@
...
+	// Parse PEM bytes into type PublicKey
+	pubKey, err := trustmanager.ParsePEMPublicKey(pubKeyBytes)
+	if err != nil {
+		return fmt.Errorf("unable to parse valid public key certificate from PEM file %s: %v", pubKeyPath, err)
+	}
+
+	keyID := pubKey.ID()
+
+	// Check to see if it is an invalid ID
+	if len(keyID) != idSize {
```
endophage: Because we're generating this `keyID` via `pubKey.ID()` it's redundant and inflexible to have this check, should just be deleted.

cmd/notary/delegations.go
```diff
@@ -0,0 +1,174 @@
...
+
+	config := d.configGetter()
+
+	gun := args[0]
+	pubKeyPath := args[1]
+	role := args[2]
+	paths := args[3:]
+
+	// Check to see if the public key file exists
+	if _, err := os.Stat(pubKeyPath); os.IsNotExist(err) {
```
endophage: Can probably just do the `ioutil.ReadFile` below and not bother with the `Stat`.

cmd/notary/delegations.go
```diff
@@ -0,0 +1,174 @@
...
+		"Removal of delegation of key \"%s\" to role %s, to repository \"%s\" staged for next publish.\n",
+		keyID, role, gun)
+	cmd.Println("")
+
+	return nil
+}
+
+// delegationAdd adds a public key from a certificate to a specific role in a GUN
+func (d *delegationCommander) delegationAdd(cmd *cobra.Command, args []string) error {
+	if len(args) < 4 {
```
endophage: Add a short test with 3 args just to his this case?

cmd/notary/delegations.go
```diff
@@ -0,0 +1,174 @@
...
+	cmd.AddCommand(cmdDelegationListTemplate.ToCommand(d.delegationsList))
+	cmd.AddCommand(cmdDelegationRemoveTemplate.ToCommand(d.delegationRemove))
+	cmd.AddCommand(cmdDelegationAddTemplate.ToCommand(d.delegationAdd))
+
+	return cmd
+}
+
+// delegationsList lists all the delegations for a particular GUN
+func (d *delegationCommander) delegationsList(cmd *cobra.Command, args []string) error {
+	if len(args) != 1 {
```
endophage: Add a short test with an incorrect number of args.

client/client.go
```diff
@@ -509,6 +509,58 @@ func (r *NotaryRepository) GetChangelist() (changelist.Changelist, error) {
...
+		// If we get an error, don't try to traverse further into this subtree because it doesn't exist or is malformed
+		if err != nil {
+			continue
+		}
+		delegationMeta := &data.SignedTargets{}
+		err = json.Unmarshal(delegationMetaJSON, delegationMeta)
+		if err != nil {
+			return nil, err
+		}
+		for _, d := range delegationMeta.Signed.Delegations.Roles {
```
endophage: replace this whole loop with 

```
allDelegations = append(allDelegations, delegationMeta.Signed.Delegations.Roles...)
delegationsList = append(delegationsList, delegationMeta.Signed.Delegations.Roles...)
```

riyazdf: go is awesome, thanks for the tip!

endophage: Yeah, it's nice :-) 

I actually saw this and knew I'd done something similar elsewhere so I had to go double check my code. I was doing map->slice conversion so unfortunately couldn't make use of this.

cmd/notary/delegations.go
riyazdf: good idea, done!

cmd/notary/delegations.go
riyazdf: done!  Also for remove too

client/client.go
```diff
@@ -509,6 +509,51 @@ func (r *NotaryRepository) GetChangelist() (changelist.Changelist, error) {
 	return cl, nil
 }
 
+// GetDelegationRoles returns the keys and roles of the repository's delegations
+func (r *NotaryRepository) GetDelegationRoles() ([]*data.Role, error) {
+	// Update state of the repo to latest
+	_, err := r.Update(false)
```
endophage: As we're not using the non-err return value:
```
if _, err := r.Update(false); err != nil {
    return nil, err
}
```

endophage: Also, for test purposes, if I remember right you can make `Update` fail by setting `r.tufRepo = nil`

client/client.go
```diff
@@ -509,6 +509,51 @@ func (r *NotaryRepository) GetChangelist() (changelist.Changelist, error) {
...
+	copy(delegationsList, allDelegations)
+
+	// Now traverse to lower level delegations (ex: targets/level1/level2)
+	for len(delegationsList) > 0 {
+		// Pop off first delegation to traverse
+		delegation := delegationsList[0]
+		delegationsList = delegationsList[1:]
+
+		// Get metadata
+		delegationMetaJSON, err := r.fileStore.GetMeta(delegation.Name, 0)
```
endophage: Can replace this like you did above:
```
delegationMeta, ok := r.tufRepo.Targets[delegation.Name]
```

cmd/notary/delegations.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	config := d.configGetter()
+
+	gun := args[0]
+
+	// initialize repo with transport to get latest state of the world before listing delegations
+	nRepo, err := notaryclient.NewNotaryRepository(config.GetString("trust_dir"), gun, getRemoteTrustServer(config), getTransport(config, gun, true), retriever)
+	if err != nil {
+		fatalf(err.Error())
```
endophage: There's a mix of `fatalf` and `return err`, it seems like it should always be `return err` based on function signature?

riyazdf:
notes on my end after some testing and discussion this evening, marking as WIP and open to discussion:

- I'm going to change how `notary delegation add` behaves from this PR's implementation, so that we add new keys if possible (this PR overwrites).  If the `paths` conflict between the existing delegation and the one we want to add with the same name, we reject the new delegation -- this check must be done online to avoid race conditions.

- I'll also change the `notary delegation remove` command to remove key IDs one by one, and remove the delegation once no keys remain.  This will probably change `RemoveDelegation` as well, and also has to be checked online for race conditions (if another client concurrently adds, etc.)

endophage:
Been thinking for consistency with the other sub-commands, which are plural nouns (`keys` and `certs`), we might also want to make it `notary delegates ...`

riyazdf:
@endophage unless I'm missing something, I think all of notary's sub-commands are singular, so maybe leave as `notary delegation ...` or `notary delegate ...`?

endophage:
Wow... yeah, I had a complete mental inversion. I prefer `notary delegate`. Happy to bow to the consensus though if people disagree, it's only a 2 character difference.

riyazdf:
removed WIP -- I had to change the logic in `helpers.go` to handle adding and removing one key at a time from a delegation:
- If the delegation role already exists, `delegation add` will add to the pre-existing delegation with the new `keyID` if and only if the paths match.  If paths do not match, this change is rejected on publish.
- `delegation remove` will remove single keys by `keyID` at a time until the last key, when the delegation is deleted in entirety

endophage:
When adding a key to an existing delegation, having to duplicate the paths seems like terrible UX. Would it be possible to add a `--update` flag to the add command that indicates that the key and any new paths should be added to the role, regardless of keys or paths it currently contains?

riyazdf:
updating this PR for documentation, after some offline discussion with @endophage, it seems most intuitive to eventually have finer grained flags such as `--key` and `--path` to operate on these single items within a `delegation add` or `delegation remove` command.

riyazdf:
Removing WIP.

I added the ability for multiple keys/paths to add/remove.  Keys are the default argument, paths are specified by a flag `--paths` which takes a comma-separated list of paths to change: `notary delegation add GUN ROLE KEY1 KEY2 --paths=PATH1,PATH2`

I also added a force flag to remove, so an admin can easily `notary delegation remove GUN ROLE --force` to remove all data (there's an accompanying `--yes` flag to avoid a confirmation)

endophage:
Do we need `--force` and `--yes`? What's the difference in behaviour between `notary delegation remove GUN ROLE` and `notary delegation remove GUN ROLE --force`?

riyazdf:
I could go either way with having the `--force` flag but I think we should at least prompt for confirmation ( and have a `--yes` flag to bypass this like when we remove certs) if you're about to delete all information for a delegation role.

Right now `notary delegation remove GUN ROLE` is an invalid command unless `--force` is included.

diogomonica:
Haven't tried this since last week, but code LGTM.

diogomonica:
@riyazdf rebase please

riyazdf:
@diogomonica done!

diogomonica:
@riyazdf me and @endophage have a long standing disagreement regarding whether notary should be 100% invocable from the cmd-line without expect scripts (the --yes was added by me a while back).

I think we need @cyli to weigh in and decide.

cyli:
@diogomonica what is the agreement?  Whether `--yes` should be included?

Maybe it would be less confusing if `--force` is changed to `--all` or `--delegation` for removing all delegation information?  `--force` to me already implies that it will just proceed without asking.  Aside from that, I'm mostly indifferent between `--force` and `--yes`.

diogomonica:
I like --yes for stuff that would ask and won't. I agree with --all.

riyazdf:
I also agree with `--all` (just updated)

endophage:
@diogomonica I'm fine with having one of `--yes` and `--force`, I was just questioning that they appeared to be redundant.

I don't think `--all` semantically makes sense. If the command is `notary delegation remove GUN DELEGATION`, what does `--all` mean? The `DELEGATION` is singular. I would say we should just have `notary delegation remove GUN DELEGATION` work with a prompt, and add the `--yes` flag to preemptively confirm (i.e. no prompt).

diogomonica:
Agreed with `notary delegation remove GUN DELEGATION` + `--yes`

riyazdf:
I think this is good, I was originally thinking that `--all` made sense to specify multiple parts of a delegation but I think keeping abstracting it to a singular delegation makes sense.  Just updated.

cmd/notary/integration_test.go
```diff
@@ -144,6 +145,223 @@ func TestClientTufInteraction(t *testing.T) {
 	assert.False(t, strings.Contains(string(output), target))
 }
 
+// Initialize repo and test delegations commands by adding, listing, and removing delegations
+func TestClientDelegationsInteraction(t *testing.T) {
```
cyli: This is an awesome integration test!

cyli:
LGTM!

Issue #423, opened 2016-01-06
server.storage.ErrOldVersion not exposed to clients


mtrmac:
`server.storage.ErrOldVersion` is converted by `atomicUpdateHandler` to generic `ErrUpdating`; it seems useful to expose this specific case to clients (in particular for #408 ).

PR #441, opened 2016-01-12
Pass original invalid version storage error details into standardized err

riyazdf:
Introduces `ErrOldVersion` to distinguish between errors for invalid versions of metadata and errors for during a DB rollback (`ErrUpdating` was propagated both for errors previously from the `atomicUpdateHandler`).  Closes https://github.com/docker/notary/issues/423

### DIFF
--- a/server/errors/errors.go
+++ b/server/errors/errors.go
@@ -44,6 +44,12 @@ var (
 		Description:    "An error occurred when attempting to apply an update at the storage layer.",
 		HTTPStatusCode: http.StatusInternalServerError,
 	})
+	ErrOldVersion = errcode.Register(errGroup, errcode.ErrorDescriptor{
+		Value:          "VERSION",
+		Message:        "A newer version of metadata is already available.",
+		Description:    "A newer version of the repository's metadata is already available in storage.",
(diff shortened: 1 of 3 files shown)


server/errors/errors.go
```diff
@@ -44,6 +44,12 @@ var (
 		Description:    "An error occurred when attempting to apply an update at the storage layer.",
 		HTTPStatusCode: http.StatusInternalServerError,
 	})
+	ErrOldVersion = errcode.Register(errGroup, errcode.ErrorDescriptor{
+		Value:          "VERSION",
+		Message:        "A newer version of metadata is already available.",
+		Description:    "A newer version of the repository's metadata is already available in storage.",
+		HTTPStatusCode: http.StatusInternalServerError,
```
endophage: It's more likely this will be raised because a user sends us an old version. Snapshot generation means it could technically be an internal server error, but the majority case is expected to be bad user input so `http.StatusBadRequest` is probably more appropriate.

riyazdf: Agreed, will change

endophage:
LGTM

PR #442, opened 2016-01-12
when doing getTransport readOnly needs to be false 

endophage:
Write permissions are required to access the *.key endpoints as they may transparently result in key creation.


riyazdf:
might be helpful to add in this information as a comment, otherwise LGTM.

cyli:
LGTM!

PR #450, opened 2016-01-13
Drop mailing list link

mtrmac:
Per today’s phone call, the mailing list is not currently used (and is closed to external contributors). So, drop the link.


endophage:
LGTM

PR #453, opened 2016-01-14
Ensure that we do not unnecessarily re-sign/serialize a root.json file on publish

cyli:
Adds additional tests to ensure that keys aren't unnecessarily created on error,
and that only the required keys to sign are used.

Probably the better way to do this would be to refactor the code some to inject a remoteStore into NotaryRepository, and compare the metadata (which we should probably do anyway).

Then again, this does ensure that we aren't re-signing or pulling out private keys unnecessarily.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -555,9 +555,21 @@ func (r *NotaryRepository) Publish() error {
 	// we send anything to remote
 	updatedFiles := make(map[string][]byte)
 
-	// check if our root file is nearing expiry. Resign if it is.
-	if nearExpiry(r.tufRepo.Root) || r.tufRepo.Root.Dirty || initialPublish {
+	// check if our root file is nearing expiry or dirty. Resign if it is.  If
+	// root is not dirty but we are publishing for the first time, then just
+	// publish the existing root we have.
+	if nearExpiry(r.tufRepo.Root) || r.tufRepo.Root.Dirty {
 		rootJSON, err := serializeCanonicalRole(r.tufRepo, data.CanonicalRootRole)
+		if err != nil {
+			return err
+		}
+		updatedFiles[data.CanonicalRootRole] = rootJSON
+	} else if initialPublish {
+		root, err := r.tufRepo.Root.ToSigned()
+		if err != nil {
+			return err
+		}
+		rootJSON, err := json.Marshal(root)
 		if err != nil {
 			return err
 		}
--- a/client/client.go
+++ b/client/client.go
@@ -565,11 +565,7 @@ func (r *NotaryRepository) Publish() error {
 		}
 		updatedFiles[data.CanonicalRootRole] = rootJSON
 	} else if initialPublish {
-		root, err := r.tufRepo.Root.ToSigned()
-		if err != nil {
-			return err
-		}
-		rootJSON, err := json.Marshal(root)
+		rootJSON, err := r.tufRepo.Root.MarshalJSON()
 		if err != nil {
 			return err
 		}
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -35,6 +35,60 @@ import (
 	"golang.org/x/net/context"
 )
 
+const password = "passphrase"
+
+type passRoleRecorder struct {
+	rolesCreated []string
+	rolesAsked   []string
+}
+
+func newRoleRecorder() *passRoleRecorder {
+	return &passRoleRecorder{}
+}
+
+func (p *passRoleRecorder) clear() {
+	p.rolesCreated = nil
+	p.rolesAsked = nil
+}
+
+func (p *passRoleRecorder) retriever(_, alias string, createNew bool, _ int) (string, bool, error) {
+	if createNew {
+		p.rolesCreated = append(p.rolesCreated, alias)
+	} else {
+		p.rolesAsked = append(p.rolesAsked, alias)
+	}
+	return password, false, nil
+}
+
+func (p *passRoleRecorder) compareRolesRecorded(t *testing.T, expected []string, created bool,
+	args ...interface{}) {
+
+	var actual, useExpected sort.StringSlice
+	copy(expected, useExpected) // don't sort expected, since we don't want to mutate it
+	sort.Stable(useExpected)
+
+	if created {
+		copy(p.rolesCreated, actual)
+	} else {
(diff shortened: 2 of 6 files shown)


client/client.go
```diff
@@ -555,13 +555,25 @@ func (r *NotaryRepository) Publish() error {
 	// we send anything to remote
 	updatedFiles := make(map[string][]byte)
 
-	// check if our root file is nearing expiry. Resign if it is.
-	if nearExpiry(r.tufRepo.Root) || r.tufRepo.Root.Dirty || initialPublish {
+	// check if our root file is nearing expiry or dirty. Resign if it is.  If
+	// root is not dirty but we are publishing for the first time, then just
+	// publish the existing root we have.
+	if nearExpiry(r.tufRepo.Root) || r.tufRepo.Root.Dirty {
 		rootJSON, err := serializeCanonicalRole(r.tufRepo, data.CanonicalRootRole)
```
diogomonica: Would you mind adding a comment on:

```
func (r SignedRoot) ToSigned() (*Signed, error) {
	s, err := json.MarshalCanonical(r.Signed)
	if err != nil {
		return nil, err
	}
	signed := json.RawMessage{}
	err = signed.UnmarshalJSON(s)
	if err != nil {
		return nil, err
	}
```
To make it explicit that this Marshling and Unmarsheling is a type cast to json.RawMessage.

client/client.go
```diff
@@ -555,13 +555,25 @@ func (r *NotaryRepository) Publish() error {
...
 		if err != nil {
 			return err
 		}
 		updatedFiles[data.CanonicalRootRole] = rootJSON
+	} else if initialPublish {
+		root, err := r.tufRepo.Root.ToSigned()
```
diogomonica: shouldn't these 8 lines of code be replaced by something like:

```
rootJSON, err := root, err := r.tufRepo.Root.ToSignedJSON()
if err != nil {
    return err
}
```

client/client_test.go
```diff
@@ -1559,20 +1822,46 @@ func TestPublishSnapshotLocalKeysCreatedFirst(t *testing.T) {
 	assert.False(t, requestMade)
 }
 
+func createKey(t *testing.T, repo *NotaryRepository, role string, x509 bool) data.PublicKey {
+	key, err := repo.CryptoService.Create(role, data.ECDSAKey)
+	assert.NoError(t, err, "error creating key")
+
+	if x509 {
+		start := time.Now()
```
diogomonica: When generating TLS certs, we should add a start time some time in the past. It makes it slightly safer (ensures that clock drifts don't make the cert not be valid).

client/client_test.go
```diff
@@ -2143,25 +2443,6 @@ func assertRotationSuccessful(t *testing.T, repo1 *NotaryRepository,
 	}
 }
 
-// Initialize repo to have the server sign snapshots (remote snapshot key)
```
cyli: This was in error - adding it back now...

tuf/data/root.go
```diff
@@ -11,6 +11,7 @@ type SignedRoot struct {
 	Signatures []Signature
 	Signed     Root
 	Dirty      bool
+	Serializer Serializer
```
cyli: This was basically my attempt at getting more coverage by doing dependency injection.  This may be just overkill, since we have to propagate JSON errors everywhere.  I'm happy to take it out, if we are ok with merging with slightly lower code coverage (because I'm adding lines that propagate JSON marshalling errors).

The two arguments I have in favor are:
- if we switch canonical JSON libraries, or if go adds one, that has a different API (specifying a canonical boolean for instance) then we can gloss over it with the serializer
- we can possibly at a later point, if we wanted to adopt this model, randomly changing the serializer in different pieces of metadata and seeing what will break and what won't

But again, happy to remove it.  It adds extra complexity for not a ton of gain.

client/client.go
cyli: Done

tuf/data/root.go
```diff
@@ -43,10 +43,11 @@ func NewRoot(keys map[string]PublicKey, roles map[string]*RootRole, consistent b
 
 // ToSigned partially serializes a SignedRoot for further signing
 func (r SignedRoot) ToSigned() (*Signed, error) {
-	s, err := json.MarshalCanonical(r.Signed)
+	s, err := defaultSerializer.MarshalCanonical(r.Signed)
```
cyli: This was basically my attempt at getting more coverage by doing dependency injection, since we have to propagate JSON errors everywhere.  I'm not sure we can isolate all or most of the JSON marshalling/unmarshalling to this package alone, but if we can, we can get more coverage that way.  Also that way, the data structures can handle the marshalling with canonical JSON themselves and I think everywhere else can just call regular `json.Marshsal` (since it will call their own `MarshalJSON` function?)

Thoughts?  This may be just overkill, since we have to propagate JSON errors everywhere, and this introduces complexity just for a test. I'm happy to take it out, if we are ok with merging with slightly lower code coverage (because I'm adding lines that propagate JSON marshaling errors).

riyazdf: In general I like the idea of isolating the JSON marshaling/unmarshaling into its own package, it's nice to keep that logic separate from the TUF objects themselves

client/client_test.go
```diff
@@ -35,6 +35,60 @@ import (
 	"golang.org/x/net/context"
 )
 
+const password = "passphrase"
+
+type passRoleRecorder struct {
```
riyazdf: this is awesome for testing, thanks for adding this in

riyazdf:
LGTM after rebase

tuf/data/serializer.go
```diff
@@ -0,0 +1,36 @@
...
+	return json.Marshal(from)
+}
+
+// Unmarshal unmarshals some JSON bytes
+func (c canonicalJSON) Unmarshal(from []byte, to interface{}) error {
+	return json.Unmarshal(from, to)
+}
+
+// defaultSerializer is a canonical JSON serializer
+var defaultSerializer = serializer(canonicalJSON{})
```
endophage: `var defaultSerializer serializer = canonicalJSON{}`

cyli: Thanks!  I forgot I could do that.

endophage:
LGTM

PR #454, opened 2016-01-14
client library and cli deletion functionality

### LINKED ISSUE #449: add the ability to clear trust data for a given repo
This is needed primarily in the library. A CLI feature would be useful, but outside the scope of this issue.

### LINKED ISSUE #462: Add notary-cli untrust call to cert remove
This is essentially exposing #454 on the cli:

My proposal is that `notary cert remove GUN`, also removes the TUF metadata for the GUN.

riyazdf:
Delete local client repo TUF cache, certs, and keys via the client library.  Closes #449

Also updates `notary cert remove gun` to remove TUF data for that gun.  Closes #462 

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -13,7 +13,6 @@ import (
 	"time"
 
 	"github.com/Sirupsen/logrus"
-
 	"github.com/docker/notary"
 	"github.com/docker/notary/certs"
 	"github.com/docker/notary/client/changelist"
@@ -907,6 +906,30 @@ func (r *NotaryRepository) rootFileKeyChange(role, action string, key data.Publi
 	err = cl.Add(c)
 	if err != nil {
 		return err
+	}
+	return nil
+}
+
+// DeleteTrustData removes the trust data stored for this repo in the TUF cache and certificate store on the client side
+func (r *NotaryRepository) DeleteTrustData() error {
+	// Clear TUF files and cache
+	if err := r.fileStore.RemoveAll(); err != nil {
+		return fmt.Errorf("error clearing TUF repo data: %v", err)
+	}
+	r.tufRepo = tuf.NewRepo(nil, nil)
+	// Clear certificates
+	certificates, err := r.CertStore.GetCertificatesByCN(r.gun)
+	if err != nil {
+		// If there were no certificates to delete, we're done
+		if _, ok := err.(*trustmanager.ErrNoCertificatesFound); ok {
+			return nil
+		}
+		return fmt.Errorf("error retrieving certificates for %s: %v", r.gun, err)
+	}
+	for _, cert := range certificates {
+		if err := r.CertStore.RemoveCert(cert); err != nil {
+			return fmt.Errorf("error removing certificate: %v: %v", cert, err)
+		}
 	}
 	return nil
 }
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -2836,3 +2836,117 @@ func testPublishTargetsDelgationCanUseUserKeyWithArbitraryRole(t *testing.T, x50
 
 	delgRec.assertAsked(t, []string{"targets/a/b"})
 }
+
+// TestDeleteRepo tests that local repo data, certificate, and keys are deleted from the client library call
+func TestDeleteRepo(t *testing.T) {
+	gun := "docker.com/notary"
+
+	ts, _, _ := simpleTestServer(t)
+	defer ts.Close()
+
+	repo, rootKeyID := initializeRepo(t, data.ECDSAKey, gun, ts.URL, false)
+	defer os.RemoveAll(repo.baseDir)
+
+	// Assert initialization was successful before we delete
+	assertRepoHasExpectedKeys(t, repo, rootKeyID, true)
+	assertRepoHasExpectedCerts(t, repo)
+	assertRepoHasExpectedMetadata(t, repo, data.CanonicalRootRole, true)
+	assertRepoHasExpectedMetadata(t, repo, data.CanonicalTargetsRole, true)
+	assertRepoHasExpectedMetadata(t, repo, data.CanonicalSnapshotRole, true)
+
+	// Delete all client trust data for repo
+	err := repo.DeleteTrustData()
+	assert.NoError(t, err)
+
+	// Assert no metadata for this repo exists locally
+	assertRepoHasExpectedMetadata(t, repo, data.CanonicalRootRole, false)
+	assertRepoHasExpectedMetadata(t, repo, data.CanonicalTargetsRole, false)
+	assertRepoHasExpectedMetadata(t, repo, data.CanonicalSnapshotRole, false)
+	assertRepoHasExpectedMetadata(t, repo, data.CanonicalTimestampRole, false)
+
+	// Assert no certs for this repo exist locally
+	_, err = repo.CertStore.GetCertificatesByCN(gun)
+	assert.Error(t, err)
+	assert.IsType(t, &trustmanager.ErrNoCertificatesFound{}, err)
+	assert.NotNil(t, err)
+
+	// Assert keys for this repo exist locally
+	assertRepoHasExpectedKeys(t, repo, rootKeyID, true)
+}
+
+type brokenRemoveFilestore struct {
+	store.MetadataStore
+}
+
+func (s *brokenRemoveFilestore) RemoveAll() error {
+	return fmt.Errorf("can't remove from this broken filestore")
+}
+
+// TestDeleteRepoBadFilestore tests that we properly error when trying to remove against a faulty filestore
+func TestDeleteRepoBadFilestore(t *testing.T) {
+	gun := "docker.com/notary"
+
+	ts, _, _ := simpleTestServer(t)
+	defer ts.Close()
+
+	repo, rootKeyID := initializeRepo(t, data.ECDSAKey, gun, ts.URL, false)
+	defer os.RemoveAll(repo.baseDir)
+
+	// Assert initialization was successful before we delete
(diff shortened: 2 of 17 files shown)


client/client.go
```diff
@@ -904,3 +904,32 @@ func (r *NotaryRepository) rootFileKeyChange(role, action string, key data.Publi
...
+	if err != nil {
+		return fmt.Errorf("error retrieving certificates for %s: %v", r.gun, err)
+	}
+	for _, cert := range certificates {
+		if err := certStore.RemoveCert(cert); err != nil {
+			return fmt.Errorf("error removing certificate: %v: %v", cert, err)
+		}
+	}
+	// Clear keys
+	repoKeys := r.CryptoService.ListAllKeys()
```
cyli: I think this also lists all the keys without a gun, such as the root, which maybe we don't want to remove because it can be shared with other repositories in the same configuration directory.  Can we check the role?  (e.g. only remove the repository keys, and delegation specific keys but not user keys?)

riyazdf: Ah yes, thanks for the catch.  I'll change this to check the key id and role to only remove non-root keys for the repo

riyazdf: Done!

client/client.go
```diff
@@ -910,3 +909,34 @@ func (r *NotaryRepository) rootFileKeyChange(role, action string, key data.Publi
 	}
 	return nil
 }
+
+// DeleteTrustData removes the trust data stored for this repo in the TUF cache on the client side,
+// if clearCertificate is set, it will also delete the certificate data for this GUN
```
diogomonica: `clearCertificate` no longer exists

diogomonica:
LGTM after comments

client/client.go
endophage: If we're moving towards a flat key store, I wouldn't bother deleting the old keys. It doesn't affect what we trust, only what we can sign. I also worry that we may delete keys we shouldn't be deleting. I don't think that's happening here but deleting a key you need leaves you in a much worse state than having an obsolete key taking up disk space.

client/client_test.go
```diff
@@ -424,7 +424,7 @@ func assertRepoHasExpectedKeys(t *testing.T, repo *NotaryRepository,
 		"there should be no timestamp key because the server manages it")
 }
 
-// This creates a new certificate store in the repo's base directory and
+// This creates a new certificate manager in the repo's base directory and
```
endophage: I think "store" was correct?

riyazdf: oops that's a rebase issue, will fix!

client/client_test.go
```diff
@@ -2836,3 +2836,77 @@ func testPublishTargetsDelgationCanUseUserKeyWithArbitraryRole(t *testing.T, x50
...
+
+	// Assert no metadata for this repo exists locally
+	assertRepoHasExpectedMetadata(t, repo, data.CanonicalRootRole, false)
+	assertRepoHasExpectedMetadata(t, repo, data.CanonicalTargetsRole, false)
+	assertRepoHasExpectedMetadata(t, repo, data.CanonicalSnapshotRole, false)
+	assertRepoHasExpectedMetadata(t, repo, data.CanonicalTimestampRole, false)
+
+	// Assert no certs for this repo exist locally
+	certs, err := repo.CertStore.GetCertificatesByCN(gun)
+	assert.IsType(t, &trustmanager.ErrNoCertificatesFound{}, err)
```
endophage: Check for `assert.Nil(t, err)` as well. There's a weirdness with Go interfaces that a `nil` value can have an underlying type. (c.f. https://endophage.com/articles/nil-interface-type/)

endophage: Sorry, got that backwards. You want to check not nil. This should be fine as it.

cmd/notary/cert.go
```diff
@@ -63,29 +64,30 @@ func certRemove(cmd *cobra.Command, args []string) {
...
 
 	var certsToRemove []*x509.Certificate
+	var certFoundByID *x509.Certificate
 
 	// If there is no GUN, we expect a cert ID
 	if certRemoveGUN == "" {
 		certID := args[0]
 		// This is an invalid ID
-		if len(certID) != idSize {
+		if len(certID) != notary.IDSize {
```
endophage: I'm actually convinced this check isn't at all necessary... Regardless of ID size, if a file exists with the ID, we should delete it, if it doesn't we won't. Similar logic applies to other places we have these checks for the ID size.

endophage: cc @diogomonica

diogomonica: My reasoning was that a message saying that the ID is invalid is better than people receiving a "ID does not exist" because they forgot 1 char, or something of the sort.

endophage: Seems inflexible.

Without these size checks, an old client would happily accept newer TUF files that used a longer hash if we wanted to increase the size. With the check, they won't.

client/client_test.go
riyazdf: ah this is interesting, thanks for the catch!  I'll add in the not nil check

cmd/notary/cert.go
endophage: Maybe we can do this check _if_ we fail to find the key and print it as a warning message?

diogomonica: I'm fine either way. Up to @riyazdf .

const.go
```diff
@@ -2,6 +2,7 @@ package notary
 
 // application wide constants
 const (
+	IDSize          = 64
```
endophage: Rather than `IDSize` something more descriptive like `Sha256HexSize`. Makes it obvious how we expect IDs to be generated.

tuf/store/filestore.go
```diff
@@ -84,3 +84,11 @@ func (f *FilesystemStore) SetMeta(name string, meta []byte) error {
 	}
 	return nil
 }
+
+// RemoveAll clears the existing filestore by removing its base directory
+func (f *FilesystemStore) RemoveAll() error {
```
endophage: Should be:
```
func (f *FilesystemStore) RemoveAll() error {
    return os.RemoveAll(f.baseDir)
}
```

client/client.go
riyazdf: Fine with me, since it doesn't affect the trust.  cc @diogomonica and @cyli if they have other opinions

cmd/notary/cert.go
riyazdf: I'll move this check to only if we fail to find the cert (@endophage's suggestion) and print a different message if the ID is invalid

endophage:
Would like to see a test with certs for multiple GUNs and ensure deleting only deletes the expected certs.

client/client_test.go
endophage: I think `assert.Error` does the nil check appropriately

const.go
riyazdf: will do!

client/client_test.go
endophage: Yeah, it does. Just testing it against my little go playground example linked from that post. Using `isNil` from the assert lib (which `assert.Error` eventually calls), correctly prints `nil`.

riyazdf:
@endophage `TestClientCertInteraction` in `integration_test.go` already tests with two different guns, but I can add more if you'd like?

endophage:
Ah cool. That looks good. The relevant part of the test (the second init and one of the removes) was folded.

riyazdf:
@endophage sounds good!  Just updated with your comments and rebased :)

client/client.go
```diff
@@ -910,3 +909,23 @@ func (r *NotaryRepository) rootFileKeyChange(role, action string, key data.Publi
...
+func (r *NotaryRepository) DeleteTrustData() error {
+	// Clear TUF files and cache
+	if err := r.fileStore.RemoveAll(); err != nil {
+		return fmt.Errorf("error clearing TUF repo data: %v", err)
+	}
+	r.tufRepo = tuf.NewRepo(nil, nil)
+	// Clear certificates
+	certificates, err := r.CertStore.GetCertificatesByCN(r.gun)
+	if err != nil {
+		return fmt.Errorf("error retrieving certificates for %s: %v", r.gun, err)
```
cyli: Non-blocking - do we want to try to catch `err.(*trustmanager.ErrNoCertificatesFound)` and count that as a job well done?  If there are no certificates, then that seems like it shouldn't be an error :)

cmd/notary/cert.go
```diff
@@ -63,29 +64,30 @@ func certRemove(cmd *cobra.Command, args []string) {
...
-			fatalf("Invalid certificate ID provided: %s", certID)
-		}
-		// Attempt to find this certificates
-		cert, err := certStore.GetCertificateByCertID(certID)
+		// Attempt to find this certificate
+		certFoundByID, err = certStore.GetCertificateByCertID(certID)
 		if err != nil {
+			// This is an invalid ID, the user might have forgotten a character
+			if len(certID) != notary.Sha256HexSize {
+				fatalf("Unable to retrieve certificate with invalid certificate ID provided: %s", certID)
```
cyli: :D  awesome error messaging!

cyli:
Totally nonblocking:  Given that we seem to be removing all certs with the same GUN as the cert the user specified by ID, would it make sense to deprecate deleting by ID?

Apologies for the late nitpicks.  LGTM aside from nonblocking!

diogomonica:
@cyli @riyazdf we should allow delete by ID to delete a single certificate.

client/client.go
riyazdf: Good idea, not erroring seems like expected behavior if there are no certs to delete.  I'll update this

cyli:
:+1:

endophage:
LGTM

PR #455, opened 2016-01-14
Fix bug where the yubikey store was not prioritized over the filestore in a client repo.

cyli:
Also, fix a test with exporting/importing all keys - because a key
that is imported into the yubikey is also backed up on disk, when exporting
all keys, it also gets exported.

@diogomonica was correct and setting up the yubikey and such should have just happened in the init.  because we added other tests (e.g. keys_test.go), the tests for the package cmd/notary break now because those tests don't call setup.

### DIFF
--- a/cmd/notary/integration_nonpkcs11_test.go
+++ b/cmd/notary/integration_nonpkcs11_test.go
@@ -9,9 +9,8 @@ import (
 )
 
 func init() {
-	fake := passphrase.ConstantRetriever("pass")
-	retriever = fake
-	getRetriever = func() passphrase.Retriever { return fake }
+	retriever = passphrase.ConstantRetriever("pass")
+	getRetriever = func() passphrase.Retriever { return retriever }
 }
 
 func rootOnHardware() bool {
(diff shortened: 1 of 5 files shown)


diogomonica:
LGTM

cmd/notary/integration_nonpkcs11_test.go
```diff
@@ -8,27 +8,18 @@ import (
...
-func setUp(t *testing.T) func() {
-	oldRetriever := retriever
-
-	var fake = func(k, a string, c bool, n int) (string, bool, error) {
-		return testPassphrase, false, nil
-	}
-
+func init() {
+	fake := passphrase.ConstantRetriever("pass")
 	retriever = fake
```
riyazdf: I might be missing something, but why do we need both `retriever` and `fake` here?

riyazdf:
LGTM pending non-blocking comment

cmd/notary/integration_nonpkcs11_test.go
cyli: You are not missing anything, and are absolutely correct.  :)

PR #461, opened 2016-01-15
Add a user_keys subdirectory to the KeyFileStore ...

cyli:
where users can keep user keys.

This makes it so that we can sign with any key in the user_keys subdir.  Currently,
created keys are created in the user_keys subdirectory under the gun subdirectory,
but this can be a future change.

Note that user keys can be ANY role that is not a valid role, so we don't write
ourselves into a corner yet.

We've also tested this manually by building a binary, by creating a key and moving to `../private/user_keys/SHA.key`, and were able to add a delegation and add a target to a delegation using a hacked up version of #440.

### DIFF
--- a/trustmanager/keyfilestore.go
+++ b/trustmanager/keyfilestore.go
@@ -14,6 +14,7 @@ import (
 
 const (
 	rootKeysSubdir    = "root_keys"
+	userKeysSubdir    = "user_keys"
 	nonRootKeysSubdir = "tuf_keys"
 	privDir           = "private"
 )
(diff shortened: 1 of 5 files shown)


diogomonica:
LGTM

trustmanager/keyfilestore.go
```diff
@@ -301,10 +302,14 @@ func removeKey(s LimitedFileStore, cachedKeys map[string]*cachedKey, name string
 
 // Assumes 2 subdirectories, 1 containing root keys and 1 containing tuf keys
 func getSubdir(alias string) string {
-	if alias == "root" {
+	switch {
+	case alias == data.CanonicalRootRole:
 		return rootKeysSubdir
+	case data.ValidRole(alias):
```
riyazdf: I think we should change this to `data.BaseRoles` so that delegations and canonical roles don't have keys in the same place?

endophage: `data.ValidRole` returns true for delegations too so this function will still always return the `nonRootKeysSubdir`

riyazdf: After discussion, this is fine as is.

endophage: Also, if I remember right, for anything but the root role, the `alias` includes the `gun` so nothing but root will ever be valid.

cyli: :|  I've changed this to BaseRole as suggested, so we put all delegations in there.  The tests seem to pass @endophage do you remember what was failing

cyli:
Closing this in favor of just adding a test to see if dropping keys in `tuf_keys` works.

PR #463, opened 2016-01-15
Enhance expiry function

HuKeping:

### DIFF
--- a/tuf/signed/verify.go
+++ b/tuf/signed/verify.go
@@ -11,6 +11,13 @@ import (
 	"github.com/jfrazelle/go/canonical/json"
 )
 
+const (
+	// expiryDelta determines how earlier a time value should be considered
+	// expired than its actual expiration time. It is used to avoid late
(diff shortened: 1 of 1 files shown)


endophage:
Just to confirm I'm understanding the change correctly. This will cause certs to be identified as expired 10 seconds early?

HuKeping:
Yes @endophage , actually I'm not very certain about whether the check for TUF files needs this buffer-time or not. It's just what we do on almost our Token verification things.

Feel free to close this if it was not a good idea.

endophage:
If you want to take it in another direction: I don't think this is necessary when checking expiry but for server signed timestamps and snapshots, it would be good to build in a buffer time at which a new one gets produced even though the current one has not expired.

diogomonica:
Agreed with @endophage. As is, this PR isn't very useful. Closing for now.

HuKeping:
Agree with @endophage

PR #467, opened 2016-01-15
Add FileMeta for previous timestamp to current timestamp

endophage:
Marked as PoC to prevent preemptive merging. This is complete working and tested code.

This creates a verifiable history chain. Still waiting for feedback on this idea from various advisors.

### DIFF
--- a/server/timestamp/timestamp.go
+++ b/server/timestamp/timestamp.go
@@ -72,7 +72,7 @@ func GetOrCreateTimestamp(gun string, store storage.MetaStore, cryptoService sig
 			return d, nil
 		}
 	}
-	sgnd, version, err := CreateTimestamp(gun, ts, snapshot, store, cryptoService)
+	sgnd, version, err := CreateTimestamp(gun, d, ts, snapshot, store, cryptoService)
 	if err != nil {
 		logrus.Error("Failed to create a new timestamp")
 		return nil, err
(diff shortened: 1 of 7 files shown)


server/timestamp/timestamp.go
```diff
@@ -111,37 +114,42 @@ func snapshotExpired(ts *data.SignedTimestamp, snapshot []byte) bool {
 // version number one higher than prev. The store is used to lookup the current
 // snapshot, this function does not save the newly generated timestamp.
 func CreateTimestamp(gun string, prev *data.SignedTimestamp, snapshot []byte, store storage.MetaStore, cryptoService signed.CryptoService) (*data.Signed, int, error) {
-	algorithm, public, err := store.GetKey(gun, data.CanonicalTimestampRole)
+	kdb := keys.NewDB()
+	repo := tuf.NewRepo(kdb, cryptoService)
+
+	// load the current root to ensure we use the correct timestamp key.
+	root, err := store.GetCurrent(gun, "root")
```
cyli: Non-blocking:  Should we still be using `data.CanonicalRootRole` everywhere?

cyli:
I think there is a rebase problem with the number of arguments for `EmptyRepo`, but otherwise this LGTM.

tuf/data/timestamp.go
```diff
@@ -23,26 +26,16 @@ type Timestamp struct {
 }
 
 // NewTimestamp initializes a timestamp with an existing snapshot
```
riyazdf: We should update this comment if we aren't using the snapshot here.

tuf/tuf.go
```diff
@@ -394,18 +395,8 @@ func (tr *Repo) InitSnapshot() error {
 }
 
 // InitTimestamp initializes a timestamp based on the current snapshot
```
riyazdf: We should also update this comment to not include the snapshot since we haven't updated the timestamp with the current state yet

tuf/tuf.go
```diff
@@ -625,7 +616,7 @@ func (tr *Repo) UpdateSnapshot(role string, s *data.Signed) error {
 }
 
 // UpdateTimestamp updates the snapshot meta in the timestamp based on the Signed object
```
riyazdf: We should also update this comment to indicate that this method can be used to set previous timestamp data for this timestamp.

server/timestamp/timestamp.go
riyazdf: :+1:

riyazdf:
LGTM after clarifying comments

diogomonica:
closing for now, while we decide what to do for append-only history.

Issue #468, opened 2016-01-16
Notary not falling back on 404 and DNS errors


diogomonica:
If notary is pointed to a random remote server that returns a 404, it should fallback, but it doesn't:

```
➜  notary list docker.io/library/ubuntu -s https://example.com    
ERRO[0000] Client Update (Timestamp): timestamp trust data unavailable.  Has a notary repository been initialized? 
ERRO[0000] client Update (Root):root trust data unavailable.  Has a notary repository been initialized? 
* fatal: example.com does not have trust data for docker.io/library/ubuntu
```

If notary can't resolve the DNS of the remote host, it also should fallback, but it doesn't:

```
➜  notary list docker.io/library/ubuntu -s https://notary.docker.io
* fatal: Get https://notary.docker.io/v2/: dial tcp: lookup notary.docker.io on [::1]:53: read udp [::1]:39327->[::1]:53: read: connection refused
notary list docker.io/library/ubuntu -s https://notary.docker.io  0.00s user 0.00s system 102% cpu 0.006 total
```

endophage:
In the first instance I'm wondering if this should be considered valid behaviour. A 404 means you successfully connected to the server and it didn't recognize what you were asking for. Fixing this feels like the new behaviour could only be considered correct in the case that a user mis-typed the URL (to some other valid URL), in which case falling back to local data seems like the wrong thing to do.

endophage:
With the bugfix to #469 merged, I believe this is now fixed. There is more than can be done (i.e. return a specific JSON body or header to get a positive ID), but the update should be sufficient for the majority of cases.

PR #469, opened 2016-01-16
Offline operation when server not reachable

endophage:
This only solves for the second error in #468.

~~TODO: need to dig up what specific errors I should be matching when checking the http response error. It probably shouldn't be generic.~~

According to net/http docs, non-2XX responses don't result in an err, which is exclusively reserved for protocol problems or client configuration directives. Added a check for non-2XX StatusCodes and also setting to offline on those.

### DIFF
--- a/cmd/notary/tuf.go
+++ b/cmd/notary/tuf.go
@@ -426,7 +426,9 @@ func tokenAuth(config *viper.Viper, baseTransport *http.Transport, gun string,
 	}
 	resp, err := pingClient.Do(req)
 	if err != nil {
-		fatalf(err.Error())
+		logrus.Errorf("could not reach %s: %s", trustServerURL, err.Error())
+		logrus.Info("continuing in offline mode")
+		return nil
 	}
 	defer resp.Body.Close()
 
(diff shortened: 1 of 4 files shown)


riyazdf:
awesome tests, LGTM

cmd/notary/tuf.go
```diff
@@ -426,9 +426,20 @@ func tokenAuth(config *viper.Viper, baseTransport *http.Transport, gun string,
...
 	resp, err := pingClient.Do(req)
 	if err != nil {
-		fatalf(err.Error())
+		logrus.Errorf("could not reach %s: %s", trustServerURL, err.Error())
+		logrus.Info("continuing in offline mode")
+		return nil
 	}
+	// non-nil err means we must close body
 	defer resp.Body.Close()
+	if resp.StatusCode < http.StatusOK || resp.StatusCode >= http.StatusMultipleChoices {
```
diogomonica: I was going to say that there has to be a better way of checking for a 2XX response, but then I saw this: https://github.com/kubernetes/deployment-manager/blob/master/manager/manager/deployer.go#L160

diogomonica:
LGTM

cmd/notary/tuf.go
endophage: Yeah... there's no good way. I've written this exact line in half a dozen languages :-P

PR #471, opened 2016-01-16
Change testutils.EmptyRepo() to use a cert as the root.json root key.

cyli:
This involves making it use ECDSA keys since we can't generate ED25519 certs.

This is useful when using `EmptyRepo()` to generate metadata to be served when testing.


diogomonica:
LGTM

endophage:
If we passed the testing object into `EmptyRepo` we wouldn't have to add all the asserts every time it's called. You've already done it so it would be creating more work to remove them so I'm kind of indifferent either way.

LGTM

PR #472, opened 2016-01-18
Fix a wrong function call

HuKeping:

### DIFF
--- a/tuf/client/client.go
+++ b/tuf/client/client.go
@@ -327,7 +327,7 @@ func (c *Client) downloadSnapshot() error {
 		}
 		err := json.Unmarshal(raw, old)
 		if err == nil {
-			snap, err := data.TimestampFromSigned(old)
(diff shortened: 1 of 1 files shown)


endophage:
Oh man, that was probably me. 

Why did it work? It worked because All TUF files share a certain amount of high level structure so even though we were unmarshalling into the wrong struct, it still parsed `snapshot.Signed.Version` into the expected field, and we were retrieving the correct raw JSON from cache.

Thank you!

LGTM!

riyazdf:
good find, LGTM!

cyli:
Nice catch!  LGTM!

HuKeping:
>  It worked because All TUF files share a certain amount of high level structure 

Yes it is, and I also noticed that the `[Targets/Timestamp/Snapshort]FromSigned` are almost the same, does it worth to create a common function?

PR #474, opened 2016-01-19
Consistent TUF Files

endophage:
Allows downloading TUF files by hash. Forward compatible for when we add Sha384/Sha512.

### DIFF
--- a/server/handlers/default.go
+++ b/server/handlers/default.go
@@ -11,7 +11,6 @@ import (
 	"github.com/gorilla/mux"
 	"golang.org/x/net/context"
 
-	"github.com/Sirupsen/logrus"
 	"github.com/docker/notary/server/errors"
 	"github.com/docker/notary/server/snapshot"
 	"github.com/docker/notary/server/storage"
@@ -115,6 +114,7 @@ func GetHandler(ctx context.Context, w http.ResponseWriter, r *http.Request) err
 
 func getHandler(ctx context.Context, w http.ResponseWriter, r *http.Request, vars map[string]string) error {
 	gun := vars["imageName"]
+	checksum := vars["checksum"]
 	tufRole := vars["tufRole"]
 	s := ctx.Value("metaStore")
 	store, ok := s.(storage.MetaStore)
@@ -124,30 +124,7 @@ func getHandler(ctx context.Context, w http.ResponseWriter, r *http.Request, var
 
 	logger := ctxu.GetLoggerWithFields(ctx, map[string]interface{}{"gun": gun, "tufRole": tufRole})
 
-	switch tufRole {
-	case data.CanonicalTimestampRole:
-		return getTimestamp(ctx, w, logger, store, gun)
-	case data.CanonicalSnapshotRole:
-		return getSnapshot(ctx, w, logger, store, gun)
-	}
-
-	out, err := store.GetCurrent(gun, tufRole)
-	if err != nil {
-		if _, ok := err.(storage.ErrNotFound); ok {
-			logrus.Errorf("404 GET %s:%s, error: %v", gun, tufRole, err)
-			return errors.ErrMetadataNotFound.WithDetail(nil)
-		}
-		logger.Error("500 GET")
-		return errors.ErrUnknown.WithDetail(err)
-	}
-	if out == nil {
-		logger.Error("404 GET")
-		return errors.ErrMetadataNotFound.WithDetail(nil)
-	}
-	w.Write(out)
-	logger.Debug("200 GET")
-
-	return nil
+	return getRole(ctx, logger, w, store, gun, tufRole, checksum)
 }
 
 // DeleteHandler deletes all data for a GUN. A 200 responses indicates success.
@@ -165,56 +142,6 @@ func DeleteHandler(ctx context.Context, w http.ResponseWriter, r *http.Request) 
 		logger.Error("500 DELETE repository")
 		return errors.ErrUnknown.WithDetail(err)
 	}
-	return nil
-}
-
-// getTimestampHandler returns a timestamp.json given a GUN
-func getTimestamp(ctx context.Context, w http.ResponseWriter, logger ctxu.Logger, store storage.MetaStore, gun string) error {
-	cryptoServiceVal := ctx.Value("cryptoService")
-	cryptoService, ok := cryptoServiceVal.(signed.CryptoService)
-	if !ok {
-		return errors.ErrNoCryptoService.WithDetail(nil)
-	}
-
-	out, err := timestamp.GetOrCreateTimestamp(gun, store, cryptoService)
-	if err != nil {
-		switch err.(type) {
-		case *storage.ErrNoKey, storage.ErrNotFound:
-			logger.Error("404 GET timestamp")
-			return errors.ErrMetadataNotFound.WithDetail(nil)
-		default:
-			logger.Error("500 GET timestamp")
-			return errors.ErrUnknown.WithDetail(err)
-		}
-	}
-
-	logger.Debug("200 GET timestamp")
-	w.Write(out)
-	return nil
-}
-
-// getTimestampHandler returns a timestamp.json given a GUN
-func getSnapshot(ctx context.Context, w http.ResponseWriter, logger ctxu.Logger, store storage.MetaStore, gun string) error {
-	cryptoServiceVal := ctx.Value("cryptoService")
-	cryptoService, ok := cryptoServiceVal.(signed.CryptoService)
-	if !ok {
-		return errors.ErrNoCryptoService.WithDetail(nil)
-	}
-
-	out, err := snapshot.GetOrCreateSnapshot(gun, store, cryptoService)
-	if err != nil {
-		switch err.(type) {
-		case *storage.ErrNoKey, storage.ErrNotFound:
-			logger.Error("404 GET snapshot")
-			return errors.ErrMetadataNotFound.WithDetail(nil)
-		default:
-			logger.Error("500 GET snapshot")
-			return errors.ErrUnknown.WithDetail(err)
-		}
-	}
-
-	logger.Debug("200 GET snapshot")
-	w.Write(out)
 	return nil
 }
 
--- a/server/handlers/default.go
+++ b/server/handlers/default.go
@@ -122,9 +122,7 @@ func getHandler(ctx context.Context, w http.ResponseWriter, r *http.Request, var
 		return errors.ErrNoStorage.WithDetail(nil)
 	}
 
-	logger := ctxu.GetLoggerWithFields(ctx, map[string]interface{}{"gun": gun, "tufRole": tufRole})
-
-	return getRole(ctx, logger, w, store, gun, tufRole, checksum)
+	return getRole(ctx, w, store, gun, tufRole, checksum)
 }
 
 // DeleteHandler deletes all data for a GUN. A 200 responses indicates success.
--- a/server/handlers/logic.go
+++ b/server/handlers/logic.go
@@ -53,7 +53,7 @@ func getRole(ctx context.Context, logger ctxu.Logger, w io.Writer, store storage
 // getMaybeServerSigned writes the current snapshot or timestamp (based on the
 // role passed) to the provided writer or returns an error. In retrieving
 // the timestamp and snapshot, based on the keys held by the server, a new one
-// might be generated and signed due to expiry of the previoud one or updates
+// might be generated and signed due to expiry of the previous one or updates
 // to other roles.
 func getMaybeServerSigned(ctx context.Context, w io.Writer, logger ctxu.Logger, store storage.MetaStore, gun, role string) error {
 	cryptoServiceVal := ctx.Value("cryptoService")
--- a/server/handlers/logic.go
+++ b/server/handlers/logic.go
@@ -1,0 +1,89 @@
+package handlers
+
+import (
+	"io"
+
+	"github.com/Sirupsen/logrus"
+	ctxu "github.com/docker/distribution/context"
+	"golang.org/x/net/context"
+
+	"github.com/docker/notary/server/errors"
+	"github.com/docker/notary/server/snapshot"
+	"github.com/docker/notary/server/storage"
+	"github.com/docker/notary/server/timestamp"
+	"github.com/docker/notary/tuf/data"
+	"github.com/docker/notary/tuf/signed"
+)
+
+func getRole(ctx context.Context, logger ctxu.Logger, w io.Writer, store storage.MetaStore, gun, role, checksum string) error {
+	var (
+		out []byte
+		err error
+	)
+	if checksum == "" {
+		// the timestamp and snapshot might be server signed so are
+		// handled specially
+		switch role {
+		case data.CanonicalTimestampRole, data.CanonicalSnapshotRole:
+			return getMaybeServerSigned(ctx, w, logger, store, gun, role)
+		}
+		out, err = store.GetCurrent(gun, role)
+	} else {
+		out, err = store.GetChecksum(gun, role, checksum)
+	}
+
+	if err != nil {
+		if _, ok := err.(storage.ErrNotFound); ok {
+			logrus.Errorf("404 GET %s:%s@%s, error: %v", gun, role, checksum, err)
+			return errors.ErrMetadataNotFound.WithDetail(nil)
+		}
+		logger.Error("500 GET")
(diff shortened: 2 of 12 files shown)


server/handlers/default.go
```diff
@@ -168,56 +145,6 @@ func DeleteHandler(ctx context.Context, w http.ResponseWriter, r *http.Request)
...
-		default:
-			logger.Error("500 GET snapshot")
-			return errors.ErrUnknown.WithDetail(err)
-		}
-	}
-
-	logger.Debug("200 GET snapshot")
-	w.Write(out)
-	return nil
-}
```
riyazdf: it's great to see this red and the new simplification :+1:

server/handlers/logic.go
```diff
@@ -0,0 +1,89 @@
...
+	w.Write(out)
+	logger.Debug("200 GET")
+
+	return nil
+}
+
+// getMaybeServerSigned writes the current snapshot or timestamp (based on the
+// role passed) to the provided writer or returns an error. In retrieving
+// the timestamp and snapshot, based on the keys held by the server, a new one
+// might be generated and signed due to expiry of the previoud one or updates
```
riyazdf: small typo `previoud`

server/storage/database.go
```diff
@@ -127,7 +127,23 @@ func (db *SQLStorage) UpdateMany(gun string, updates []MetaUpdate) error {
...
+func (db *SQLStorage) GetChecksum(gun, tufRole, checksum string) ([]byte, error) {
+	var row TUFFile
+	q := db.Select("data").Where(
+		&TUFFile{
+			Gun:    gun,
+			Role:   tufRole,
+			Sha256: checksum,
+		},
+	).First(&row)
+	return returnRead(q, row)
```
riyazdf: it might be nice to have a simple unit test for this in `database_test.go`

endophage: was writing the test as you were writing this comment :-) It's up now

server/handlers/logic.go
```diff
@@ -0,0 +1,89 @@
+package handlers
```
diogomonica: Not sure if `logic.go` is a good name for this file.

server/handlers/logic.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		if _, ok := err.(storage.ErrNotFound); ok {
+			logrus.Errorf("404 GET %s:%s@%s, error: %v", gun, role, checksum, err)
+			return errors.ErrMetadataNotFound.WithDetail(nil)
+		}
+		logger.Error("500 GET")
+		return errors.ErrUnknown.WithDetail(err)
+	}
+	if out == nil {
+		logger.Error("404 GET")
```
diogomonica: Do we want to add more detail to these errors? Like the gun, role and checksum being requested in the first place?

I would also love to get some unique request ID that we use for all error/info logs to see the path of a request.

server/handlers/logic.go
```diff
@@ -0,0 +1,89 @@
...
+		return errors.ErrMetadataNotFound.WithDetail(nil)
+	}
+	w.Write(out)
+	logger.Debug("200 GET")
```
diogomonica: More detail?

server/handlers/logic.go
```diff
@@ -0,0 +1,89 @@
...
+	var (
+		out []byte
+		err error
+	)
+	if checksum == "" {
+		// the timestamp and snapshot might be server signed so are
+		// handled specially
+		switch role {
+		case data.CanonicalTimestampRole, data.CanonicalSnapshotRole:
+			return getMaybeServerSigned(ctx, w, logger, store, gun, role)
```
diogomonica: why not follow the same pattern and have this be `out, err = getMaybeServerSigned(...)`?

server/handlers/logic.go
```diff
@@ -0,0 +1,89 @@
...
+	return nil
+}
+
+// getMaybeServerSigned writes the current snapshot or timestamp (based on the
+// role passed) to the provided writer or returns an error. In retrieving
+// the timestamp and snapshot, based on the keys held by the server, a new one
+// might be generated and signed due to expiry of the previous one or updates
+// to other roles.
+func getMaybeServerSigned(ctx context.Context, w io.Writer, logger ctxu.Logger, store storage.MetaStore, gun, role string) error {
+	cryptoServiceVal := ctx.Value("cryptoService")
```
diogomonica: What about defining a context key that provide type-safe accessors to access the `CryptoService`?

https://godoc.org/golang.org/x/net/context

server/handlers/logic.go
```diff
@@ -0,0 +1,89 @@
...
+	logger.Debug("200 GET")
+
+	return nil
+}
+
+// getMaybeServerSigned writes the current snapshot or timestamp (based on the
+// role passed) to the provided writer or returns an error. In retrieving
+// the timestamp and snapshot, based on the keys held by the server, a new one
+// might be generated and signed due to expiry of the previous one or updates
+// to other roles.
```
diogomonica: @cyli this is one of the things that keeps me up at night. Did we end up creating any tests that show that a server can't convince a client to sign TUF metadata with new keys for targets or even root?

server/handlers/logic.go
```diff
@@ -0,0 +1,89 @@
...
+	cryptoService, ok := cryptoServiceVal.(signed.CryptoService)
+	if !ok {
+		return errors.ErrNoCryptoService.WithDetail(nil)
+	}
+
+	var (
+		out []byte
+		err error
+	)
+	switch role {
```
diogomonica: Where is the check for "is this key server signed"? if `GetOrCreateSnapshot` is faithful to it's name, if it doesn't find one it will create it, regardless of whether the server owns it or not.

server/server.go
```diff
@@ -89,12 +89,16 @@ func RootHandler(ac auth.AccessController, ctx context.Context, trust signed.Cry
 		prometheus.InstrumentHandlerWithOpts(
 			prometheusOpts("UpdateTuf"),
 			hand(handlers.AtomicUpdateHandler, "push", "pull")))
-	r.Methods("GET").Path("/v2/{imageName:.*}/_trust/tuf/{tufRole:(root|targets(?:/[^/\\s]+)*|snapshot|timestamp)}.json").Handler(
```
diogomonica: We no longer need regexp groups?

server/storage/database.go
```diff
@@ -127,7 +127,23 @@ func (db *SQLStorage) UpdateMany(gun string, updates []MetaUpdate) error {
 func (db *SQLStorage) GetCurrent(gun, tufRole string) ([]byte, error) {
 	var row TUFFile
 	q := db.Select("data").Where(&TUFFile{Gun: gun, Role: tufRole}).Order("version desc").Limit(1).First(&row)
+	return returnRead(q, row)
+}
+
+// GetChecksum gets a specific TUF record by its hex checksum
+func (db *SQLStorage) GetChecksum(gun, tufRole, checksum string) ([]byte, error) {
```
diogomonica: Do we want to enforce validity of the length here?

server/server.go
```diff
@@ -89,12 +89,16 @@ func RootHandler(ac auth.AccessController, ctx context.Context, trust signed.Cry
...
+	r.Methods("GET").Path("/v2/{imageName:.*}/_trust/tuf/{tufRole:root|targets(?:/[^/\\s]+)*|snapshot|timestamp}.json").Handler(
 		prometheus.InstrumentHandlerWithOpts(
 			prometheusOpts("GetRole"),
 			hand(handlers.GetHandler, "pull")))
+	r.Methods("GET").Path("/v2/{imageName:.*}/_trust/tuf/{checksum:[a-fA-F0-9]{64}|[a-fA-F0-9]{96}|[a-fA-F0-9]{128}}.{tufRole:root|targets(?:/[^/\\s]+)*|snapshot|timestamp}.json").Handler(
```
riyazdf: To clarify, are the 96 and 128 hex matches here meant for future-proofing, or is there another purpose?  I know we do have SHA512 helper functions in notary but I don't think we have any SHA384 functions, and we don't seem to use either for checksums currently anyway?

riyazdf: This also ties in with @diogomonica's comment https://github.com/docker/notary/pull/474/files#r50304600

server/handlers/logic.go
endophage: Agreed. Just didn't want to get hung up on a filename which is irrelevant in the context of Go. Happy to change it to anything else but I haven't come up with a good name for it :-/

server/handlers/logic.go
endophage: The logger will already add the URL (which contains all those variables) _and_ a request ID :-)

(in kibana you should see there's a `http.request.id` field.

server/handlers/logic.go
endophage: The possible errors are different so it felt clearer to have it separated.

server/handlers/logic.go
endophage: I don't know what you're suggesting. The argument and return types are all defined as `interface{}`, there is no type safety around `ctx.WithValue` and `ctx.Value`, which we would always be using.

server/handlers/logic.go
endophage: I could give you a longer function name but I think `GetSnapshotOrMaybeCreateIfServerHasKey` is kind of a silly name :-P

Joking aside, it's perfectly valid for `GetOrCreateSnapshot` to return `ErrNoKey`. It loads the current root.json to determine which key it should use. If the server doesn't then have that key, it errors and that error gets translated below.

server/server.go
endophage: They were actually incorrect. It's not explicitly stated in the mux docs anywhere but after debugging my URL regex problems yesterday and getting help in IRC, we determined that not only are they not needed, they shouldn't be used. It causes unexpected behaviour.

endophage: Filed an issue about it with gorilla/mux: https://github.com/gorilla/mux/issues/143

server/storage/database.go
endophage: I don't believe so. It shouldn't be the job of a storage engine to know how long a checksum should be or which hashing methods are valid.

server/server.go
endophage: Future proofing because I can't see any good reason not to and I want to get sha384 and/or sha512 into whatever version we release next after 0.2

server/handlers/logic.go
cyli: @diogomonica Apologies, could you elaborate on this more?

server/handlers/logic.go
cyli: If the logger is already going to log the information, does having an additional `logger.Error` here helpful for debugging?  (I guess on the terminal it's red, so easier to see).

endophage: @cyli probably not useful. I'm not sure who (it may even have been me) added all this status logging but I'm happy to remove it.

cyli: I definitely added some, but can't claim all of them :)

server/server_test.go
```diff
@@ -85,3 +89,54 @@ func TestGetKeysEndpoint(t *testing.T) {
...
+	assert.NoError(t, err)
+	update := storage.MetaUpdate{
+		Role:    data.CanonicalTimestampRole,
+		Version: 1,
+		Data:    j,
+	}
+	checksumBytes := sha256.Sum256(j)
+	checksum := hex.EncodeToString(checksumBytes[:])
+
+	store.UpdateCurrent("gun", update)
```
cyli: Would it be possible to add a second timestamp in here?  Since there is only one timestamp in the store right now, and it's the current one, this test would still pass if checksum returned only the current timestamp.

server/storage/memory.go
```diff
@@ -68,6 +75,17 @@ func (st *MemStorage) GetCurrent(gun, role string) (data []byte, err error) {
...
+	defer st.lock.Unlock()
+	data, ok := st.checksums[checksum]
+	if !ok || len(data) == 0 {
+		return nil, ErrNotFound{}
+	}
+	return data, nil
+}
+
 // Delete delets all the metadata for a given GUN
 func (st *MemStorage) Delete(gun string) error {
```
cyli: Should delete also update the checksum map?

server/storage/database_test.go
```diff
@@ -440,3 +441,43 @@ func TestDBCheckHealthSucceeds(t *testing.T) {
...
+	require.NoError(t, err)
+	update := MetaUpdate{
+		Role:    data.CanonicalTimestampRole,
+		Version: 1,
+		Data:    j,
+	}
+	checksumBytes := sha256.Sum256(j)
+	checksum := hex.EncodeToString(checksumBytes[:])
+
+	store.UpdateCurrent("gun", update)
```
cyli: Similarly here - could we add a second timestamp?  That way, if someone accidentally deletes the checksum line in the function, this test will fail?

server/storage/database_test.go
```diff
@@ -440,3 +441,43 @@ func TestDBCheckHealthSucceeds(t *testing.T) {
...
+	require.NoError(t, err)
+	require.EqualValues(t, j, data)
+}
+
+func TestDBGetChecksumNotFound(t *testing.T) {
+	tempBaseDir, err := ioutil.TempDir("", "notary-test-")
+	_, store := SetUpSQLite(t, tempBaseDir)
+	defer os.RemoveAll(tempBaseDir)
+
+	_, err = store.GetChecksum("gun", data.CanonicalTimestampRole, "12345")
```
cyli: Could we also add a timestamp to the DB, so that we can be sure that it's the checksum check that's failing?

server/handlers/default.go
cyli: :fireworks:

server/handlers/logic.go
diogomonica: On repository initialization or key rotation from local to remote, the server provides back information about the keyID that should be added to the TUF metadata and signed by the client with the root key.

The question is: can the server provide more than the data for the timestamp and snapshot in that case, potentially having the client sign arbitrary TUF metadata?

cyli: The data provided has to be JSON-unmarshallable as a public key, and we add the public key to the role's list of keys.  If they provide a list of keys, it won't be unmarshallable as a public key. 

They can definitely provide extra data inside the JSON, because only the fields we care about is unmarshalled into the data structure, and that structure specifically is used to add a key to a role, the extra data provided will be ignored.

There are no specific tests for fuzzing the key data coming from the server yet, although there are tests for the server failing to provide a key.

server/handlers/logic.go
riyazdf: maybe `roles.go`?  I'm not sold on that name either but seems a little more descriptive

server/storage/memory.go
```diff
@@ -68,6 +75,17 @@ func (st *MemStorage) GetCurrent(gun, role string) (data []byte, err error) {
 	return space[len(space)-1].data, nil
 }
 
+// GetChecksum returns the metadada for a given role, under a GUN
```
riyazdf: typo: metadata

server/storage/memory.go
```diff
@@ -68,6 +75,17 @@ func (st *MemStorage) GetCurrent(gun, role string) (data []byte, err error) {
...
+	st.lock.Lock()
+	defer st.lock.Unlock()
+	data, ok := st.checksums[checksum]
+	if !ok || len(data) == 0 {
+		return nil, ErrNotFound{}
+	}
+	return data, nil
+}
+
 // Delete delets all the metadata for a given GUN
```
riyazdf: while you're in here, it'd be great to fix the typo on `delets` :)

server/storage/memory.go
riyazdf: Agreed, I think it's necessary for our memory store implementation since the tufMeta and checksum maps are separate.

endophage: Had to modify the memory store but it should work consistently now.

endophage:
@cyli I think I've addressed your requests to add a second timestamp in the tests. I now create and add 2 timestamps and try to get the older one by checksum.

server/storage/database_test.go
cyli: There is no checksum in this table, but the other two tests should cover the case I think of filtering by checksum

cyli:
@endophage Aside from the file renaming, if you want to do it, this LGTM!

endophage:
@cyli @riyazdf renamed logic.go and test file to role.go

riyazdf:
LGTM!

Issue #446, opened 2016-01-12
Provide a way for notary client to get more details about a repo


vikstrous:
Notary client needs to tell us at least:
* What keys are currently allowed to sign the main targets file in a repository (what targets keys exist?)
* What key was used to sign a given tag
* What roles exist
* What key(s) can sign for which role
* What usernames are in each delegated key
* Which role a given tag is coming from
* Timestamps for everything
* What's the root key for this repository

We need to query all of this data to display information about a repository, so it would be ideal if we can make O(1) requests for this data.

endophage:
What key was used to sign a tag requires determining what version of the repo added the tag and who signed that version. I didn't think analyzing history was in scope for dtr/1.5. 

The rest seems reasonable as it's inspection of current state.

vikstrous:
Writing down what we talked about: "What key was used to sign a given tag" means "what role is the tag in and what key signed that role last".

riyazdf:
These functions exist in the library but we're moving toward a static notary repository object that can serve this info.  I'll close this and keep #548 and #550 as open issues

PR #475, opened 2016-01-19
get all current role information for a repo

riyazdf:
client library function `ListRoles` to get all current state for a repository's roles including keys and signatures.  Adds a new `RoleWithSignatures` object to express role state that wraps each `Role` with its signatures in a `RoleWithSignatures` object.

Also fixes some typos.

Partially addresses #446.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -169,7 +169,7 @@ func (r *NotaryRepository) Initialize(rootKeyID string, serverManagedRoles ...st
 	// currently we only support server managing timestamps and snapshots, and
 	// nothing else - timestamps are always managed by the server, and implicit
 	// (do not have to be passed in as part of `serverManagedRoles`, so that
-	// the API of Initialize doens't change).
+	// the API of Initialize doesn't change).
 	var serverManagesSnapshot bool
 	locallyManagedKeys := []string{
 		data.CanonicalTargetsRole,
@@ -396,7 +396,7 @@ func (r *NotaryRepository) RemoveDelegation(name string, keyIDs, paths []string,
 }
 
 // AddTarget creates new changelist entries to add a target to the given roles
-// in the repository when the changelist gets appied at publish time.
+// in the repository when the changelist gets applied at publish time.
 // If roles are unspecified, the default role is "targets".
 func (r *NotaryRepository) AddTarget(target *Target, roles ...string) error {
 
@@ -454,7 +454,7 @@ func (r *NotaryRepository) ListTargets(roles ...string) ([]*TargetWithRole, erro
 	for _, role := range roles {
 		// we don't need to do anything special with removing role from
 		// roles because listSubtree always processes role and only excludes
-		// descendent delegations that appear in roles.
+		// descendant delegations that appear in roles.
 		r.listSubtree(targets, role, roles...)
 	}
 
@@ -569,6 +569,56 @@ func (r *NotaryRepository) GetDelegationRoles() ([]*data.Role, error) {
 		delegationsList = append(delegationsList, delegationMeta.Signed.Delegations.Roles...)
 	}
 	return allDelegations, nil
+}
+
+// RoleWithSignatures is a Role with its associated signatures
+type RoleWithSignatures struct {
+	Signatures []data.Signature
+	data.Role
+}
+
+// GetRepoRoleMetaInfo returns a list of RoleWithSignatures objects for this repo
+// This represents the latest metadata for each role in this repo
+func (r *NotaryRepository) GetRepoRoleMetaInfo() ([]RoleWithSignatures, error) {
+	// Update to latest repo state
+	_, err := r.Update(false)
+	if err != nil {
+		return nil, err
+	}
+
+	// Get all role info from our updated keysDB
+	roles, err := r.tufRepo.GetAllRoles()
+	if err != nil {
+		return nil, err
+	}
+
+	var roleWithSigs []RoleWithSignatures
+
+	// Populate RoleWithSignatures with Role from keysDB and signatures from TUF metadata
+	for _, role := range roles {
+		roleWithSig := RoleWithSignatures{Role: *role, Signatures: nil}
+		switch role.Name {
+		case data.CanonicalRootRole:
+			roleWithSig.Signatures = r.tufRepo.Root.Signatures
+		case data.CanonicalTargetsRole:
+			roleWithSig.Signatures = r.tufRepo.Targets[data.CanonicalTargetsRole].Signatures
+		case data.CanonicalSnapshotRole:
+			roleWithSig.Signatures = r.tufRepo.Snapshot.Signatures
+		case data.CanonicalTimestampRole:
+			roleWithSig.Signatures = r.tufRepo.Timestamp.Signatures
+		default:
+			// If the role isn't a delegation, we should error -- this is only possible if we have invalid keyDB state
+			if !data.IsDelegation(role.Name) {
+				return nil, data.ErrInvalidRole{Role: role.Name, Reason: "invalid role name"}
+			}
+			if _, ok := r.tufRepo.Targets[role.Name]; ok {
+				// We'll only find a signature if we've published any targets with this delegation
+				roleWithSig.Signatures = r.tufRepo.Targets[role.Name].Signatures
+			}
+		}
+		roleWithSigs = append(roleWithSigs, roleWithSig)
+	}
+	return roleWithSigs, nil
 }
 
 // Publish pushes the local changes in signed material to the remote notary-server
--- a/client/client.go
+++ b/client/client.go
@@ -577,20 +577,17 @@ type RoleWithSignatures struct {
 	data.Role
 }
 
-// GetRepoRoleMetaInfo returns a list of RoleWithSignatures objects for this repo
+// ListRoles returns a list of RoleWithSignatures objects for this repo
 // This represents the latest metadata for each role in this repo
-func (r *NotaryRepository) GetRepoRoleMetaInfo() ([]RoleWithSignatures, error) {
+func (r *NotaryRepository) ListRoles() ([]RoleWithSignatures, error) {
 	// Update to latest repo state
 	_, err := r.Update(false)
 	if err != nil {
 		return nil, err
 	}
 
-	// Get all role info from our updated keysDB
-	roles, err := r.tufRepo.GetAllRoles()
-	if err != nil {
-		return nil, err
-	}
+	// Get all role info from our updated keysDB, can be empty
+	roles := r.tufRepo.GetAllLoadedRoles()
 
 	var roleWithSigs []RoleWithSignatures
 
--- a/tuf/keys/db.go
+++ b/tuf/keys/db.go
@@ -58,6 +58,15 @@ func (db *KeyDB) AddRole(r *data.Role) error {
 	return nil
 }
 
+// GetAllRoles gets all roles from the database
+func (db *KeyDB) GetAllRoles() []*data.Role {
+	roles := []*data.Role{}
+	for _, role := range db.roles {
+		roles = append(roles, role)
+	}
+	return roles
+}
+
 // GetKey pulls a key out of the database by its ID
 func (db *KeyDB) GetKey(id string) data.PublicKey {
 	return db.keys[id]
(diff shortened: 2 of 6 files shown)


client/client.go
```diff
@@ -515,6 +515,68 @@ func (r *NotaryRepository) GetChangelist() (changelist.Changelist, error) {
...
+	}
+
+	keys := make(map[string][]string)
+
+	rootSignatures := r.tufRepo.Root.Signatures
+	for _, sig := range rootSignatures {
+		keys[data.CanonicalRootRole] = append(keys[data.CanonicalRootRole], sig.KeyID)
+	}
+	// Traverse into the targets map, this includes delegations
+	targets := r.tufRepo.Targets
```
cyli: I think this may only get the first level delegations?  Delegations can be nested (so "targets/a" will have the keys for "targets/a/b" for instance.  So I think you may have to walk down the tree of targets to get all the key IDs.

client/client.go
```diff
@@ -515,6 +515,68 @@ func (r *NotaryRepository) GetChangelist() (changelist.Changelist, error) {
 	return cl, nil
 }
 
+// GetRepoKeyIDs returns a map of associations between this repo's roles and its key IDs
+func (r *NotaryRepository) GetRepoKeyIDs() (map[string][]string, error) {
+	// Update to latest repo state
+	_, err := r.Update(false)
```
cyli: We can probably get away with skipping the update here, since `Update` only pulls down the tuf files and this is just looking at keys on disk.

client/client.go
cyli: Ignore me, I misunderstood this :D

cyli:
Apologies, could you remind me again what `GetRepoKeyIDs` will be used for?  (e.g. is it going to take the place of `notary key list`?)  Also, it'd be awesome if we could add a test for it that includes keys from other repos, so that we can make sure the other keys are filtered out.  :)

riyazdf:
@cyli I've refactored to include all key and signature information for each role in a `RoleWithSignature` object, and a list of these is returned from `GetRepoRoleMetaInfo`

Also, great idea on the test!  I've made a setup for that in `TestGetRepoRoleMetaInfo` in `client_test.go`

tuf/keys/db.go
```diff
@@ -58,6 +58,15 @@ func (db *KeyDB) AddRole(r *data.Role) error {
 	return nil
 }
 
+// GetAllRoles gets all roles from the database
+func (db *KeyDB) GetAllRoles() []*data.Role {
+	roles := []*data.Role{}
```
cyli: Non-blocking, as this is totally fine - I think you can also use `copy()` here - it probably does the same thing :D

cyli: Oops, nm!  I was confused and forgot `db.roles` was a map, since we were iterating over it and ignoring the key.  :)

cyli:
LGTM!

tuf/tuf.go
```diff
@@ -173,6 +173,15 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
 	return nil
 }
 
+// GetAllRoles returns a list of all role entries for this TUF repo
+func (tr *Repo) GetAllRoles() ([]*data.Role, error) {
+	roles := tr.keysDB.GetAllRoles()
+	if len(roles) == 0 {
+		return nil, data.ErrNoRoles{}
```
diogomonica: Not sure if I would return a No roles error, instead of an empty array.

client/client.go
```diff
@@ -574,6 +574,56 @@ func (r *NotaryRepository) GetDelegationRoles() ([]*data.Role, error) {
...
 
+// RoleWithSignatures is a Role with its associated signatures
+type RoleWithSignatures struct {
+	Signatures []data.Signature
+	data.Role
+}
+
+// GetRepoRoleMetaInfo returns a list of RoleWithSignatures objects for this repo
+// This represents the latest metadata for each role in this repo
+func (r *NotaryRepository) GetRepoRoleMetaInfo() ([]RoleWithSignatures, error) {
```
diogomonica: Do we want to expose this as an API? Not sure if we will see it in a future PR, but I was expecting a GetRepoInfo, or something of the sort, that would return the full status of the repository (all targets, all roles, all signatures).

tuf/tuf.go
riyazdf: I'm fine either way, though it seemed odd to not have any roles in a TUF repo since we have the base roles.  If you have a preference for empty array, I can definitely change this :)

diogomonica: @endophage @cyli what should our API do?

client/client.go
riyazdf: I was thinking that we would leave this as public functions on the `NotaryRepository`, but by "API" do you mean a server endpoint or notary CLI command?

I was originally thinking that we would have this function for retrieving the status of roles for a repo, and make use of `ListTargets` just for considering the status of targets (the two could be called together in another function as well), but I could also modify this to be a full `GetRepoInfo` as you're suggesting such that we would also only require to call `Update()` once.

tuf/tuf.go
cyli: There might be no roles if nothing has been loaded yet - like if it hasn't been initialized, for instance.  I guess that isn't really an error - maybe it'd be clearer if this was named `GetAllLoadedRoles` and that could return an empty array?  As opposed to `GetAllRoles` which may imply all base roles which must be in a valid TUF repo?

client/client.go
riyazdf: ah, maybe I misinterpreted your comment -- instead were you saying that we should just have a single public facing API for getting the full status of the repository and keep this function private?  I think that's fine too, @cyli @endophage any opinions?

tuf/tuf.go
riyazdf: I agree that `GetAllLoadedRoles` clarifies the distinction between what's actually loaded and what must be a valid TUF repo -- unless there are any objections I'll update the function name and return an empty array here instead of `data.ErrNoRoles{}`

client/client.go
endophage: I might shorten the function name to simply `ListRoles` (for consistency with `ListTargets` and maybe it can take a single argument `withDetail bool`. If `withDetail` is false, we only populate the basic role info (i.e. the content of a `data.Role` object translated into a `RoleWithSignatures` object). If `withDetail` is `true`, we return the public keys and signatures data too (which I believe is the current behaviour).

I don't think there's a need for a single function that that returns combined targets and roles info. Yes we could make it marginally easier for others to get one big report with a single function, but I don't think it's worthwhile over having 2 functions with clearly defined responsibilities.

cyli: @endophage If we want to display the combined results though, we should probably provide a singular function that calls `Update` once, since if we call the 2 separate functions, the results might be inconsistent (the state of the repo may have changed in between ListTargets and ListRoles on the server).  Or otherwise call one, take the repo offline, and then call the other.

tuf/tuf.go
```diff
@@ -173,6 +173,11 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
 	return nil
 }
 
+// GetAllLoadedRoles returns a list of all role entries loaded in this TUF repo, could be empty
+func (tr *Repo) GetAllLoadedRoles() []*data.Role {
```
endophage: I actually wouldn't expect this function to exist. As the `keysDB` has to be injected when creating a new Repo object, the caller should hold onto its reference if it needs to call `GetAllRoles`. These kinds of passthrough were what caused confusion and got us into trouble in other places.

endophage: Looking at the current code, this just makes me realize even more we need to merge /client/client.go and /tuf/client/client.go. The one in /tuf keeps a reference to the keyDB, the other doesn't :-(

client/client.go
diogomonica: I think a method that returns the whole state of the repository in a well delineated object is preferable, exactly because of what @cyli said.

I would prefer people calling ListTargets when they merely want to get all the targets (probably the most common occurence), and have another method to get a snapshot of the current repository with all the info that it provides.

endophage: @cyli I'd go the other way. Add a flag that we can check to determine we have performed an update. On the basis this won't be the only time we want to perform multiple ops against a loaded repo, having a consistent way for any method that _might_ need to load the repo to check if it's already been loaded seems like it would be useful.

Maybe even have the flag as a `time.Time` and allow `Update` to decide if a sufficient amount of time has passed that it should actually run an update again.

tuf/tuf.go
diogomonica: If it would make our code/API easier to understand and develop, I think we should take the time to do it.

client/client.go
riyazdf: @cyli good point on the `Update` calls, I agree that the `GetRepoInfo` can't be two separate calls to this function and `ListTargets` (at least with how they're currently written).  I think what we can do here is call `ListTargets` to update and retrieve targets, and then follow the rest of `GetRepoRoleMetaInfo`'s logic without updating again: this can be our `GetRepoInfo` API and we can wrap `TargetsWithRoles` and `RoleWithSignatures` in a `RepoInfo` object.  How does that sound?

@endophage I don't think our API calls should rely on when other APIs were called, I think it's most expected to have each call update independently and always return the latest state at the time it was called.

riyazdf: I added a new commit that follows the logic I outlined above, exposing a `GetRepoInfo` API that only performs a single update and wraps the targets and roles (with signatures) in a `RepoInfo` object

tuf/tuf.go
endophage: Not in the 0.2 timeframe, it's a big change with a lot of caution needed, but yeah, I'd like to do it sooner rather than later.

endophage:
I really don't think a monolithic method that returns everything is the right way to go... Smaller scope methods can always be efficiently composed by downstream consumers into larger contexts. Big monolithic methods don't give the choice. By the given reasoning we should ditch `ListTargets` and have `GetRepoInfo` be the way to get targets out too.

If we insist on have a big utility method, I'd still like to see a `ListRoles` and the changes I mentioned about `Update` (the `Update` fixes don't need to be in this PR). If anyone wants to use the library in a non-oneshot environment, having every function update is both inefficient, and possibly provides an inconsistent view of data over time. In that use case, it should be possible to load a repo once and perform multiple operations on that particular version of the repo.

cyli:
I think I generally agree on having separate smaller functions, but the multiple update is problematic.

@endophage What do you think about none of these functions `Update` in any function, and leave that to user of the library to call first?  For instance, in the CLI.

endophage:
@cyli that would also be acceptable, but I think in the current configuration it will mean we need to do a bunch of nil checks to ensure various things have been populated, which means we would probably still want a `isLoaded` flag or some such, on which we can error if we try and say `ListTargets` before `Update`

riyazdf:
Looping back here after discussion with @cyli and @endophage : It seems best for the API to simply have a `ListRoles` API, and change the mechanics of `Update` (in a separate PR) to have a wrapper that decides whether or not to actually update.  We agree that it makes more sense to compose calls of `ListRoles` and `ListTargets` as needed rather than getting flooded with a giant repo info call and having to parse out information.  I'll update this PR.

diogomonica:
Not sure I agree with this path.

I think I agree with an explicit call to Update to update a repository. This has the advantage of making bootstrapRepo be completely offline as long as there is a cache. It does feel like a major change in the way things are done/structured, and would love some more thought around it.

As is, if we merge this, we have an API where a call to ListRoles, followed by a call to ListTargets might not return information about the same thing. That seems wrong.

cyli:
@diogomonica Would this + the api change to `Update` be ok?  The plan for `Update` was, as @riyazdf mentioned, to have a wrapper to check if it's been updated already.  So a call to `ListTargets` would actually update, and if it's followed by a call to `ListRoles`, it would not update because the update flag would have been set on the client.  This would prevent the consistency mismatch.

diogomonica:
@cyli the update flag seems like a hack, but if the rest of the team is ok with this, I'm ok with it too. We're all going to have to maintain this code for the foreseeable future. Choose wisely :

cyli:
@diogomonica Is it ok to merge this right now, and do the update refactor as per IRL discussions in a later PR?

diogomonica:
Sounds good.

PR #479, opened 2016-01-19
Do not lowercase role names when adding a change

### LINKED ISSUE #444: ToLower on path prefix?
https://github.com/docker/notary/pull/371/files#diff-1cd41a2611da49ce63630d5a51bd82c5R271 calls `ToLower` on a role name.

AFAICS the delegation role names are path prefixes, which are typically case-sensitive (UNIX, URLs), so the `ToLower` call seems to be losing information.

What is the intent here? AFAICS the canonical role names are expected to be case-insensitive, but the delegated roles are case sensitive (looking at `data.CanonicalRole`, `utils.FindRoleIndex`, uses of `filepath.Dir`.

cyli:

Fixes #444

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -9,7 +9,6 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
-	"strings"
 	"time"
 
 	"github.com/Sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


diogomonica:
LGTM

mtrmac:
:+1:, thanks!

diogomonica:
Is it worth it to add some testing around this?

cyli:
@diogomonica Good point, thanks, will add

cyli:
@diogomonica https://github.com/docker/notary/pull/479/files#diff-fce84029d346504e44323c30d21a3808R809

diogomonica:
LGTM

riyazdf:
LGTM after ci!

PR #482, opened 2016-01-19
Add some comments about the notary constants, and other lint fixes.

cyli:
It seems that `make vet` sometimes disagrees locally vs on CircleCI.  This
just fixes my local `make vet` complaints.

Sometimes my local `make vet` seems to let me get away without commenting, too, so I'm not sure where the disparity is.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -545,7 +545,7 @@ func (r *NotaryRepository) GetDelegationRoles() ([]*data.Role, error) {
 	// All top level delegations (ex: targets/level1) are stored exclusively in targets.json
 	targets, ok := r.tufRepo.Targets[data.CanonicalTargetsRole]
 	if !ok {
-		return nil, store.ErrMetaNotFound{data.CanonicalTargetsRole}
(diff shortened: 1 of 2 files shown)


riyazdf:
It seems I get a different `make vet` error than you have: 
```
This file should end with '_test':
+ vet
make: *** [vet] Error 1
```

I'll look into this further and see if I have any suggestions, it seems like I'm hitting an empty filename or something...

riyazdf:
That said, it's probably something just for my local setup, this PR LGTM :+1:

diogomonica:
LGTM

cyli:
@riyazdf And you said that `find . -iname *test*.go | grep -v _test.go | grep -v Godeps` returns nothing, right?

Could you let me know what `which xargs` returns?  And what `echo | xargs echo "hey":` returns?

riyazdf:
I did some digging and it seems it was an issue with my `xargs` and homebrew findutils version, all fixed now :).  `make vet` passes for me now so LGTM!

cyli:
@riyazdf Thanks!

Issue #361, opened 2015-12-14
Client side validation tests (on update/pull, with delegations)


cyli:
# Success - all metadata added/updated
- [x] new repo - download valid timestamp, root, targets, snapshot, and a
	2-level tree of delegations, all with the right hashes and signatures and
	thresholds (tests completely new repo, and downloading delegations if there
	are delegations within the targets and delegations file) 

     (any test that calls `assertPublishToRolesSucceeds` does this because `assertPublishToRolesSucceeds` creates a new bare repo with no new information and downloads.  Same with `TestPublishDelegations`)
- [x] prev repo has invalid everything; download new timestamp, snapshot, root,
	targets, and delegations (tests updating all data over previous repo)

    (`TestUpdateReplacesCorruptOrMissingMetadata`)
- [x] prev repo has valid root only; download valid timestamp, snapshot,
	targets, and delegations (tests content changed, but keys have not)

    (`TestUpdateReplacesCorruptOrMissingMetadata`)
- [x] prev repo has valid root and targets; download valid timestamp, snapshot,
	and delegations (tests downloading new delegations data)

    (`TestUpdateReplacesCorruptOrMissingMetadata`)
- [x] prev repo has valid root, targets, and partial delegations (1 level);
	download valid timestamp, snapshot, and other delegations (2nd level) (
	tests downloading new delegations data if data is not changed in targets)

    (`TestUpdateReplacesCorruptOrMissingMetadata`)
- [x] prev repo has valid root, targets, and delegations; download valid
	timestamp and snapshot (tests download a new snapshot if the previous
	snapshot was re-signed, possibly due to expiry)

    (`TestUpdateReplacesCorruptOrMissingMetadata`)
- [x] prev repo has valid root, targets, delegations, and snapshot; download
	valid timestamp (tests the case that the local repo is the latest repo)

    (`TestUpdateReplacesCorruptOrMissingMetadata`)
- [x] prev repo has invalid root (new root has 1 key added, threshold 1, so
	signatures are still good);

    (`TestUpdateReplacesCorruptOrMissingMetadata`)

## Key rotation
- [x] prev repo, new repo has timestamp key rotated; download timestamp and
	root, timestamp (tests timestamp key rotation)

     (WIP on server-rotate-key branch: `TestRotateBeforePublishFromRemoteKeyToLocalKey`, `TestRotateKeyAfterPublishNoServerManagementChange`, `TestRotateKeyAfterPublishServerManagementChange`)
[...]

PR #484, opened 2016-01-20
Add tests for when the server 404's or 50X's when updating

cyli:
I re-worked the first couple of tests in `client_update_tests` to cover more cases and moved them to the bottom of the file.

I also changed client so that:

1.  we don't get an unexpected end of JSON error when the server errors while downloading a timestamp
2.  we just use the cached timestamp now when the server 404's on getting a timestamp (previously, we used the cached timestamp when we got any other error than 404 - not sure we should default to cached if we do get a 404, but the behavior should probably be consistent for non-200 status codes)
3.  warn users that we are using a stale timestamp file and hence may have a stale view of the repo, if we use the cached timestamp

Partially addresses #361.

### DIFF
--- a/client/client_update_test.go
+++ b/client/client_update_test.go
@@ -461,7 +461,6 @@ func TestUpdateRemoteRoot50XCannotUseLocalCache(t *testing.T) {
 // but is missing other data, then we propagate the ErrMetaNotFound.  Skipping
 // force check, because that only matters for root.
 func TestUpdateNonRootRemoteMissingMetadataNoLocalCache(t *testing.T) {
-	// TODO: fix json syntax error
 	for _, role := range append(data.BaseRoles, "targets/a", "targets/a/b") {
 		if role == data.CanonicalRootRole {
 			continue
@@ -531,7 +530,6 @@ func TestUpdateNonRootRemoteMissingMetadataCannotUseLocalCache(t *testing.T) {
 // If there is no local cache, we just update. If the server 50X's when getting
 // metadata, we propagate ErrServerUnavailable.
 func TestUpdateNonRootRemote50XNoLocalCache(t *testing.T) {
-	// TODO: fix json syntax error
 	for _, role := range append(data.BaseRoles, "targets/a", "targets/a/b") {
 		if role == data.CanonicalRootRole {
 			continue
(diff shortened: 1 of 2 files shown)


tuf/client/client.go
```diff
@@ -277,17 +276,18 @@ func (c *Client) downloadTimestamp() error {
 			}
 			return err
 		}
-		logrus.Debug("using cached timestamp")
+		logrus.Debug(err.Error())
+		logrus.Warn("Using cached timestamp - this may mean we cannot get the latest version of the repository")
```
diogomonica: can we rephrase this?

"Error while downloading remote metadata, using cached timestamp - this might not be the latest version available remotely"

I would avoid the word "Repository" anywhere in notary.

diogomonica:
Skimmed through the tests, they all seem to make sense. LGTM

tuf/client/client.go
```diff
@@ -54,7 +54,7 @@ func (c *Client) Update() error {
 	if err != nil {
 		logrus.Debug("Error occurred. Root will be downloaded and another update attempted")
 		if err := c.downloadRoot(); err != nil {
-			logrus.Error("client Update (Root):", err)
+			logrus.Error("Client Update (Root):", err)
```
riyazdf: thanks for making this consistent :+1:

riyazdf:
I really like how you structured the test options and look forward to using them :)  LGTM!

tuf/client/client.go
cyli: Sounds good!

tuf/client/client.go
cyli: heh, thanks.  :D  it was annoying me to see the upper vs lower cases appear in the tests :)

client/client_update_test.go
```diff
@@ -28,109 +28,46 @@ func newBlankRepo(t *testing.T, url string) *NotaryRepository {
...
-// still returns an ErrRepositoryNotExist.
-func TestUpdateNotExistWithLocalCache(t *testing.T) {
-	testUpdateNotExistWithLocalCache(t, false)
-	testUpdateNotExistWithLocalCache(t, true)
-}
-
-func testUpdateNotExistWithLocalCache(t *testing.T, forWrite bool) {
-	ts, _, _ := simpleTestServer(t)
-	defer ts.Close()
+var metadataDelegations = []string{"targets/a", "targets/a/b", "targets/a/b/c"}
```
endophage: Would be good to have a `target/b` too just so we have more than a single linear chain.

cyli: Good point, thanks!

endophage:
LGTM

Issue #424, opened 2016-01-06
Consider supporting passphrase change for private keys


mtrmac:
It is probably easier for a passphrase to get compromised (by someone just watching a keyboard, or a window manager unexpectedly switching focus) than for the private key to get compromised; so it might be useful to support changing passphrases on existing keys.

riyazdf:
This has been added for root keys, we're tracking adding this functionality (and import/export, since they are also root-only) in https://github.com/docker/notary/issues/483

PR #485, opened 2016-01-21
passwd command and tests

riyazdf:
Introduces a `notary key passwd` command that takes a root key ID and allows the user to change its encryption passphrase.

Partially addresses https://github.com/docker/notary/issues/424 since this only considers root keys, but allowing passphrase changes on individual targets or snapshot keys would require changes to the passphrase retriever since we bundle the keys together for usability (we don't expose how there are two separate keys with the same targets passphrase to the user on repo creation)

```
$ notary key passwd b96f10739347d960ef964ec946b9c3b26d443d57b09cfa574f4ab93a8614f954
Enter passphrase for root key with ID b96f107:
You are about to create a new root signing key passphrase. This passphrase
will be used to protect the most sensitive key in your signing system. Please
choose a long, complex passphrase and be careful to keep the password and the
key file itself secure and backed up. It is highly recommended that you use a
password manager to generate the passphrase and keep it safe. There will be no
way to recover this key. You can find the key in your config directory.
Enter passphrase for new root key with ID b96f107:
Repeat passphrase for new root key with ID b96f107:

Successfully updated passphrase for key ID: b96f10739347d960ef964ec946b9c3b26d443d57b09cfa574f4ab93a8614f954
```

### DIFF
--- a/cmd/notary/integration_test.go
+++ b/cmd/notary/integration_test.go
@@ -844,6 +844,36 @@ func TestLogLevelFlags(t *testing.T) {
 	assert.Equal(t, "debug", logrus.GetLevel().String())
 }
 
+func TestClientKeyPassphraseChange(t *testing.T) {
+	// -- setup --
+	setUp(t)
+
+	tempDir := tempDirWithConfig(t, "{}")
+	defer os.RemoveAll(tempDir)
+
+	server := setupServer()
+	defer server.Close()
+
+	// -- tests --
+	_, err := runCommand(t, tempDir, "-s", server.URL, "init", "gun1")
+	assert.NoError(t, err)
+
+	// we should have three keys stored locally in total: root, targets, snapshot
+	rootIDs, _ := assertNumKeys(t, tempDir, 1, 2, true)
+
+	// only one rootID, try changing the private key passphrase
+	rootID := rootIDs[0]
+	_, err = runCommand(t, tempDir, "-s", server.URL, "key", "passwd", rootID)
+	assert.NoError(t, err)
+	// make sure we can init a new repo with this key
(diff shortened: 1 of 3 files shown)


NathanMcCauley:
> individual targets or snapshot keys would require changes to the passphrase retriever since we bundle the keys together for usability

Doesn't it make this simpler now that snapshots are server managed?

riyazdf:
@NathanMcCauley snapshots aren't server managed by default in notary (or when docker uses notary), the user has to manually `notary key rotate -r`.

If the snapshot key is local, if we change the passphrase of only one of the targets or snapshot keys we'll fail the passphrase check against the cached entry in the passphrase retriever. This will cause us to ask for the snapshot key password, which the user has never been explicitly asked for before.

I can definitely refactor the passphrase retriever to separate out targets and snapshots, but I'm worried about the UX around the snapshot key -- we never mention anything about it on `notary init`, but I'm not sure the best alternative is asking for an additional snapshot password?  Maybe some clever caching?

cyli:
This LGTM as a root key password rotator.

Maybe, for the UX of the passphrase retriever, on `init` when it asks for a passphrase for the targets password, after the user enters the password it can say something like "using the same passphrase for new snapshot key with ID...." after?

Similarly during signing, if it uses the same passphrase for targets and signing, and it succeeds, maybe the passphrase retriever can just mention that?  (if it doesn't succeed, it'll fall back on asking for the snapshot passphrase)

diogomonica:
I'm ok with the functionality as is for root. 

Also, I think it's ok for us to: turn remote snapshots by default on notary; add the distinct key types as different passphrase requests.

PR #486, opened 2016-01-21
tokenAuth should also 'succeed' if we get a 401

endophage:
I had an incorrect understanding of the response codes that would be returned from `/v2/`. A 401 is considered valid and will result in authentication being done later.

Tests have been added to ensure tokenAuth succeeds when the server returns a 401


cyli:
LGTM!   Listing trust data from hub now works!

riyazdf:
LGTM

PR #491, opened 2016-01-22
allow for configurable log level, change default for CLI

riyazdf:
Currently the default logging level is error - it'd be nice to have configurable logging levels for notary, and only default to a higher level such as `fatal` for a less noisy CLI.

This is motivated by when I was testing remote snapshot rotation via the CLI, and any online command after the `key rotate` command brought about an Info log for a signature mismatch for Snapshot, and an associated Error log.  This is particularly noisy in this instance because these logs occur before we download a new root and retry the TUF update, ultimately running the command successfully to completion:

```
$ notary key rotate snap3 -t snapshot -r true
$ notary publish snap3
Pushing changes to snap3
Enter passphrase for root key with ID b96f107:
$ notary list snap3
INFO[0000] failed ECDSA signature validation
ERRO[0000] Client Update (Snapshot): valid signatures did not meet threshold
  NAME                                DIGEST                                SIZE (BYTES)    ROLE
----------------------------------------------------------------------------------------------------
  v1     e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855   0              targets
```

### DIFF
--- a/cmd/notary/main.go
+++ b/cmd/notary/main.go
@@ -20,6 +20,7 @@ const (
 )
 
 var (
+	debug             bool
 	verbose           bool
 	trustDir          string
 	configFile        string
@@ -37,10 +38,7 @@ func init() {
 }
 
 func parseConfig() *viper.Viper {
-	if verbose {
-		logrus.SetLevel(logrus.DebugLevel)
-		logrus.SetOutput(os.Stderr)
-	}
+	setVerbosityLevel()
 
 	// Get home directory for current user
 	homeDir, err := homedir.Dir()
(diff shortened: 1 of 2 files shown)


cmd/notary/main.go
```diff
@@ -180,3 +179,20 @@ func getPassphraseRetriever() passphrase.Retriever {
 		return baseRetriever(keyName, alias, createNew, numAttempts)
 	}
 }
+
+// If the user specified verbose, have it override any log-level
+func setVerbosityLevel() {
+	if verbose {
```
endophage: `logLevel` should probably override verbose as it's a finer grain control.

endophage:
If we're going to have both a verbose and a loglevel flag, then verbose should turn on just enough logging for the typical user (probably Error level).

Alternatively, I really only see 3 useful levels: default (maybe Fatal as you suggest), Error, and Debug. I'd love to see this implemented as no flag = default, `-v` = Error and `-vv` = Debug. I can't remember which tools do it but I've always really liked tools that just add more `v`'s to the verbosity flag to crank up the level. There's something innately satisfying about it, it feels like turning up a volume knob.

riyazdf:
@endophage I really like that idea.  My only concern is that the behavior of `-v` will now be different, but I think this is fine as long as its properly documented in release notes and on the notary CLI

riyazdf:
Actually it seems that we can't have a shorthand flag with more than one letter, so `-vv` is not possible with `pflags` unless we change to `--v` and `--vv` which doesn't feel right :disappointed: .  We could try different flag names like `-D` for debug level and `-v` just for error level, but I'm not sure if I like this too much either.

I've updated to @endophage's first suggestion of making `-v` Error level and having `--logLevel` override it to any specified level.  I'll continue to think of other ways to make this better -- if anyone has other suggestions or ideas, they are more than welcome!

diogomonica:
Debug seems to override verbose level. LGTM.

cyli:
LGTM!

PR #492, opened 2016-01-22
List timestamps

endophage:
Server code for listing timestamps.

It doesn't implement a rigid pagination but instead relies on clients to keep track of pages themselves, letting them send a non-inclusive starting hash and a number of items to return.

### DIFF
--- a/server/errors/errors.go
+++ b/server/errors/errors.go
@@ -29,7 +29,7 @@ var (
 	ErrBadPagination = errcode.Register(errGroup, errcode.ErrorDescriptor{
 		Value:          "BAD_PAGINATION",
 		Message:        "The pagination parameters are invalid.",
-		Description:    "Either the start or number parameters could not be parsed as integers greater than or equal to zero.",
+		Description:    "The cursor or limit parameters could not be parsed. The cursor parameter must be a hex encoded sha256 checksum, and the limit must be an integer >= 0",
 		HTTPStatusCode: http.StatusBadRequest,
 	})
 	ErrInvalidRole = errcode.Register(errGroup, errcode.ErrorDescriptor{
--- a/server/errors/errors.go
+++ b/server/errors/errors.go
@@ -26,6 +26,12 @@ var (
 		Description:    "No file/role name is provided to associate an update with.",
 		HTTPStatusCode: http.StatusBadRequest,
 	})
+	ErrBadPagination = errcode.Register(errGroup, errcode.ErrorDescriptor{
+		Value:          "BAD_PAGINATION",
+		Message:        "The pagination parameters are invalid.",
+		Description:    "Either the start or number parameters could not be parsed as integers greater than or equal to zero.",
+		HTTPStatusCode: http.StatusBadRequest,
+	})
 	ErrInvalidRole = errcode.Register(errGroup, errcode.ErrorDescriptor{
 		Value:          "INVALID_ROLE",
 		Message:        "The role you are attempting to operate on is invalid.",
--- a/server/handlers/default.go
+++ b/server/handlers/default.go
@@ -3,8 +3,10 @@ package handlers
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"strconv"
 	"strings"
 
@@ -212,6 +214,7 @@ func getKeyHandler(ctx context.Context, w http.ResponseWriter, r *http.Request, 
 	return nil
 }
 
+// ListVersionsHandler serves a paginated list of versions of a given TUF role
 func ListVersionsHandler(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
 	vars := mux.Vars(r)
 	gun, ok := vars["imageName"]
@@ -228,21 +231,9 @@ func ListVersionsHandler(ctx context.Context, w http.ResponseWriter, r *http.Req
 		return errors.ErrUnknown.WithDetail(err)
 	}
 
-	var (
-		start  = 0
-		number = 0
-	)
-	startStr := r.Form.Get("start")
-	if startStr != "" {
-		if start, err = strconv.Atoi(startStr); err != nil || start < 0 {
-			return errors.ErrBadPagination.WithDetail(nil)
-		}
-	}
-	numberStr := r.Form.Get("number")
-	if numberStr != "" {
-		if number, err = strconv.Atoi(numberStr); err != nil || start < 0 {
-			return errors.ErrBadPagination.WithDetail(nil)
-		}
+	start, number, err := parsePageParams(r.Form)
+	if err != nil {
+		return errors.ErrBadPagination.WithDetail(err)
 	}
 
 	s := ctx.Value("metaStore")
@@ -250,11 +241,26 @@ func ListVersionsHandler(ctx context.Context, w http.ResponseWriter, r *http.Req
 	if !ok || store == nil {
 		return errors.ErrNoStorage.WithDetail("No storage configured")
 	}
+	return listVersionsHandler(ctx, w, store, gun, role, start, number)
+}
 
+func parsePageParams(f url.Values) (start string, number int, err error) {
+	start = f.Get("start")
+	numberStr := f.Get("number")
+	if numberStr != "" {
+		if number, err = strconv.Atoi(numberStr); err != nil || number < 0 {
+			return "", 0, fmt.Errorf("could not parse number parameter")
+		}
+	}
+	return
+}
+
+func listVersionsHandler(ctx context.Context, w http.ResponseWriter, store storage.MetaStore, gun, role, start string, number int) error {
 	versions, err := store.GetVersions(gun, role, start, number)
-	if _, ok := err.(storage.ErrNotFound); ok {
-		return errors.ErrMetadataNotFound.WithDetail(nil)
-	} else {
+	if err != nil {
+		if _, ok := err.(storage.ErrNotFound); ok {
+			return errors.ErrMetadataNotFound.WithDetail(nil)
+		}
 		return errors.ErrUnknown.WithDetail(err)
 	}
 
--- a/server/handlers/default.go
+++ b/server/handlers/default.go
@@ -7,10 +7,12 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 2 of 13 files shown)


endophage:
Moving back to WIP. After discussions we've decided the API should be <checksum> and <number of entries> as the inputs, not a start index (the server will loop up the start index in the DB using the checksum).

server/handlers/default.go
```diff
@@ -211,6 +214,65 @@ func getKeyHandler(ctx context.Context, w http.ResponseWriter, r *http.Request,
...
+	numberStr := f.Get("number")
+	if numberStr != "" {
+		if number, err = strconv.Atoi(numberStr); err != nil || number < 0 {
+			return "", 0, fmt.Errorf("could not parse number parameter")
+		}
+	}
+	return
+}
+
+func listVersionsHandler(ctx context.Context, w http.ResponseWriter, store storage.MetaStore, gun, role, start string, number int) error {
```
cyli: Non-blocking - I think in a previous PR, you were going to change these handler functions to just take a writer of any kind, which seemed like a pretty cool idea.  :D  Is that possible here?  This function doesn't seem to use the context, for instance.

server/handlers/default.go
```diff
@@ -211,6 +214,65 @@ func getKeyHandler(ctx context.Context, w http.ResponseWriter, r *http.Request,
...
+	s := ctx.Value("metaStore")
+	store, ok := s.(storage.MetaStore)
+	if !ok || store == nil {
+		return errors.ErrNoStorage.WithDetail("No storage configured")
+	}
+	return listVersionsHandler(ctx, w, store, gun, role, start, number)
+}
+
+func parsePageParams(f url.Values) (start string, number int, err error) {
+	start = f.Get("start")
```
cyli: Non-blocking - for an HTTP API, unless we are trying to make this pagination more general (so it applies to more things), maybe more descriptive query parameters would make it easier to understand how it works?  Maybe something like `from_checksum` and `num_versions` or some such?  (I'm bad at naming, so probably something else :D)

cyli: Update: or maybe `before_checksum` or the like, since the results seem to not include the starting checksum?

server/handlers/default.go
```diff
@@ -211,6 +214,65 @@ func getKeyHandler(ctx context.Context, w http.ResponseWriter, r *http.Request,
...
+	if numberStr != "" {
+		if number, err = strconv.Atoi(numberStr); err != nil || number < 0 {
+			return "", 0, fmt.Errorf("could not parse number parameter")
+		}
+	}
+	return
+}
+
+func listVersionsHandler(ctx context.Context, w http.ResponseWriter, store storage.MetaStore, gun, role, start string, number int) error {
+	versions, err := store.GetVersions(gun, role, start, number)
```
cyli: If `0` is a valid number of results to return, should we add an explicit check for it here and bail early, rather than make a DB query?

server/errors/errors.go
```diff
@@ -26,6 +26,12 @@ var (
 		Description:    "No file/role name is provided to associate an update with.",
 		HTTPStatusCode: http.StatusBadRequest,
 	})
+	ErrBadPagination = errcode.Register(errGroup, errcode.ErrorDescriptor{
+		Value:          "BAD_PAGINATION",
+		Message:        "The pagination parameters are invalid.",
+		Description:    "Either the start or number parameters could not be parsed as integers greater than or equal to zero.",
```
cyli: Maybe "The number parameter could not be parsed as an integer greater...", since we don't check the start except to see if it exists?  (If it doesn't, we return `ErrNotFound`, and not `ErrBadPagination`)

server/storage/database_test.go
```diff
@@ -501,3 +482,133 @@ func TestDBGetChecksumNotFound(t *testing.T) {
...
+
+	ts, err := repo.SignTimestamp(data.DefaultExpires("timestamp"))
+	tsJSON1, err := json.Marshal(ts)
+	require.NoError(t, err)
+	store.UpdateCurrent(
+		"gun",
+		MetaUpdate{Role: "timestamp", Version: 1, Data: tsJSON1},
+	)
+
+	// table is empty
```
cyli: Nitpick:  I think this comment might have been a copy-paste artifact?

server/storage/database_test.go
```diff
@@ -501,3 +482,133 @@ func TestDBGetChecksumNotFound(t *testing.T) {
...
+	require.EqualValues(t, tsJSON1, versions[2])
+
+	versions, err = store.GetVersions("gun", "timestamp", checksum, 1)
+	require.NoError(t, err)
+
+	require.Len(t, versions, 1)
+	require.EqualValues(t, tsJSON1, versions[0])
+}
+
+func TestGetVersionsNotFound(t *testing.T) {
```
cyli: Awesome test with all the cases of DB failure!  :D

server/storage/database_test.go
```diff
@@ -501,3 +482,133 @@ func TestDBGetChecksumNotFound(t *testing.T) {
...
+		"gun",
+		"timestamp",
+		"173d71993529e47adf58301e58d916085f30ca08797581d7bf200da94595db89",
+		0,
+	)
+	require.Error(t, err)
+	require.IsType(t, ErrNotFound{}, err)
+}
+
+func TestTranslateOldVersionError(t *testing.T) {
```
cyli: Yay!  Thanks for adding this!

server/storage/interface.go
```diff
@@ -34,6 +34,18 @@ type MetaStore interface {
 	// found, it returns storage.ErrNotFound
 	GetChecksum(gun, tufRole, checksum string) (data []byte, err error)
 
+	// GetVersions returns an ordered list of versions, newest to oldest
+	// for the given gun and role. Only the binary data is returned as
+	// all other values stored in the DB for convenience should be
+	// generated/validated from this raw json by the consumer.
+	// start is a checksum which will be used as a non-inclusive offset
+	// An empty start value means start from (and include) the most recent version
+	// numToResturn can limit how many results are returned.
```
cyli: Nitpick: "numToReturn"

server/storage/database_test.go
```diff
@@ -501,3 +482,133 @@ func TestDBGetChecksumNotFound(t *testing.T) {
...
+	require.Len(t, versions, 3)
+	require.EqualValues(t, tsJSON3, versions[0])
+	require.EqualValues(t, tsJSON2, versions[1])
+	require.EqualValues(t, tsJSON1, versions[2])
+
+	versions, err = store.GetVersions("gun", "timestamp", checksum, 1)
+	require.NoError(t, err)
+
+	require.Len(t, versions, 1)
+	require.EqualValues(t, tsJSON1, versions[0])
```
cyli: Would it be possible to add a couple more checks here?  Some of these may be redundant (so please feel free to reject any or all):

1. `store.GetVersions("gun", "timestamp", "", 1)` should return just `versions[2]` (I think?)
1. `store.GetVersions("gun", "timestamp", checksum, 0)` should return `versions[0]`
1. `store.GetVersions("gun", "timestamp", checksum, 80)` should also return `versions[0]`
1. `store.GetVersions("gun", "timestamp", "", 80)` should also return all?

The number being > than what's there case is the particular one I wanted to test.

server/storage/memory.go
```diff
@@ -84,11 +90,50 @@ func (st *MemStorage) GetCurrent(gun, role string) (data []byte, err error) {
...
+// GetVersions for MemStorage only returns the latest version
+// as it does not store multiple versions
+func (st *MemStorage) GetVersions(gun, role, start string, numToReturn int) ([][]byte, error) {
+	var (
+		idx, stop int
+		id        = entryKey(gun, role)
+	)
+	if start != "" {
+		for st.tufMeta[id][idx].checksum != start {
+			if idx == len(st.tufMeta[id]) {
```
cyli: Should this be `len(st.tufMeta[id]) - 1`?  By the time it reaches `len(st.tufMeta[id])`, won't this be an index out of range error?  Perhaps we could add a test for searching an in-memory store for a checksum that doesn't exist?

cyli:
Could we also add a test in `server/handlers/default_test.go`'s `TestListVersionsHandler` that passing an invalid number (not parseable as string or negative number) in returns the predicted error? 

Also, can we add a test in `server/server_test.go`'s `TestListVersions` that a URL route that includes the start but no number returns an error (is it a 404?)

server/handlers/default.go
cyli: Oh oops, sorry, I missed the part where `0` meant "get all".  nm!  :)

endophage:
@cyli renamed `start` to `cursor` and `number` to `limit` (I genreally prefer generic terms so that they can be used consistently throughout an HTTP API rather than having custom parameter names for every endpoint).

A URL with a start but no number wouldn't (and shouldn't) have been an error, it just wouldn't have had the regex checks applied to ensure the checksum was valid. It would have been caught by the URL route for versions that had no matching query params. That made me realize any request to versions for which the params didn't match the regexes would get caught by that version of the route so I've just removed the other ones and added a check to parse params to make sure the cursor is a valid sha256 hex checksum.

server/storage/database_test.go
endophage: It helped me get the test coverage to acceptable levels :-P

server/storage/database_test.go
endophage: I totally missed testing what happens if the limit is larger than the total number of possible entries to return. Great catch.

server/handlers/default.go
```diff
@@ -211,6 +216,76 @@ func getKeyHandler(ctx context.Context, w http.ResponseWriter, r *http.Request,
...
+	}
+	return listVersionsHandler(ctx, w, store, gun, role, start, number)
+}
+
+func parsePageParams(f url.Values) (start string, number int, err error) {
+	start = f.Get("cursor")
+	numberStr := f.Get("limit")
+	if numberStr != "" {
+		if number, err = strconv.Atoi(numberStr); err != nil || number < 0 {
+			return "", 0, fmt.Errorf("could not parse number parameter")
```
riyazdf: should this read "limit" instead of "start"?

endophage: yep, fixed to "limit"

diogomonica:
Closing for now.

endophage:
The request for this feature has been tabled so this would have been unused code for some period. We will revisit this and either rewrite, or re-open as and when appropriate.

Issue #490, opened 2016-01-22
Notary usage says PEM file, when it only receives X509 certs


diogomonica:
```
➜  notary git:(master) notary delegation add docker.io/diogomonica/testingnot1 targets/diogo   
Usage:
  notary delegation add [ GUN ] [ Role ] <PEM file path 1> ... [flags]
```

Issue #488, opened 2016-01-22
Notary delegation yes/no has the wrong ordering


diogomonica:
It asks me to confirm the deletion and then fails with usage.

```
➜  notary git:(master) notary delegation remove docker.io/diogomonica/testingnot1 targets

Are you sure you want to remove all data for this delegation? (yes/no)
yes
Usage:
  notary delegation remove [ GUN ] [ Role ] <KeyID 1> ... [flags]

Flags:
      --paths=[]: List of paths to remove
  -y, --yes[=false]: Answer yes to the removal question (no confirmation)

Global Flags:
  -c, --configFile="": Path to the configuration file to use
  -s, --server="": Remote trust server location
  -d, --trustDir="": Directory where the trust data is persisted to
  -v, --verbose[=false]: Verbose output
* fatal: failed to remove delegation: tuf: invalid role targets. invalid delegation role name
```

PR #493, opened 2016-01-22
delegations CLI UX improvements

### LINKED ISSUE #487: Weird phrasing in notary
```
➜  notary git:(master) notary delegation list docker.io/diogomonica/testingnot1
No such roles published in this repository.
```

We should change the phrasing of this error.

riyazdf:
Would love any suggestions for better messages on the CLI.  Closes #487, #488, #490.

### DIFF
--- a/cmd/notary/delegations.go
+++ b/cmd/notary/delegations.go
@@ -32,9 +32,9 @@ var cmdDelegationRemoveTemplate = usageTemplate{
 }
 
 var cmdDelegationAddTemplate = usageTemplate{
-	Use:   "add [ GUN ] [ Role ] <PEM file path 1> ...",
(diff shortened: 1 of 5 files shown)


cyli:
LGTM!  Thanks for clarifying the X509 vs PEM :)

diogomonica:
Shouldn't we also have a `isDelegation` check on `addDelegation`?

riyazdf:
@diogomonica we have a `IsDelegation` check in `AddDelegation` (in `client/client.go`), so we still make a check there after the user has input the command.

For removing a delegation it was more tricky because the additional user yes/no confirmation after the command itself had to occur before `RemoveDelegation` also checked with an `IsDelegation`, so it wasn't as intuitive on the CLI without adding an additional check.

PR #494, opened 2016-01-22
Cleanup: remove unused code

HuKeping:
It is not possible that the hash keeps consistent when the length
has been changed.

### DIFF
--- a/tuf/client/client.go
+++ b/tuf/client/client.go
@@ -92,10 +92,9 @@ func (c *Client) update() error {
 	return nil
 }
 
-// checkRoot determines if the hash, and size are still those reported
-// in the snapshot file. It will also check the expiry, however, if the
-// hash and size in snapshot are unchanged but the root file has expired,
-// there is little expectation that the situation can be remedied.
+// checkRoot determines if the hash is still those reported in the snapshot file.
+// It will also check the expiry, however, if the hash in snapshot is unchanged but the
+// root file has expired, there is little expectation that the situation can be remedied.
 func (c Client) checkRoot() error {
 	role := data.CanonicalRootRole
 	size := c.local.Snapshot.Signed.Meta[role].Length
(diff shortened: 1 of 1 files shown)


tuf/client/client.go
```diff
@@ -111,10 +110,6 @@ func (c Client) checkRoot() error {
 		return fmt.Errorf("Cached root sha256 did not match snapshot root sha256")
```
riyazdf: Isn't it possible that another byte array could still collide with the same hash?  Especially if we are not restricting the length?  It is improbable, but the size check below make it even more so.

Also, while `c.cache.GetMeta` takes a byte `size` to read, it doesn't actually restrict the size in the `filestore.go` implementation.  I think this gives us more reason to leave in the sanity check for the length below.

endophage: `bytes.Equal` will only return true in the case that the 2 slices are identical, including length (the documentation explicitly confirms the length is checked).

endophage: Oh, sorry, I misunderstood when I read this the first time.

We should fix filestore.go so we can test all the stores at a uniform interface level. @HuKeping I think without the fix in filestore.go (probably memorystore.go too), this change is unsafe.

riyazdf: Agreed that we should fix this -- I think `memorystore.go` is fine so it'll only be a change to `filestore.go`

riyazdf: Just opened a PR for this: https://github.com/docker/notary/pull/495

diogomonica: Just FYI, as far as we're concerned hash collisions are impossible. The only reason to have the size checks is to protect against infinite data attacks.

HuKeping: Yes I see and I'm agree with that.

HuKeping: Closing for now.

PR #495, opened 2016-01-22
ensure filestore GetMeta only returns up to size bytes.  Add max size constant

riyazdf:
store/filestore.go's `GetMeta` now considers the `size` argument and will only return up to `size` bytes.  If `size` is -1, we get "all" data for the resource but stop at 100MB under the hood -- this is consistent across the memory, file, and remote stores.  I also refactored `MaxTimestampSize` to const.go for convenience.

This follows up on discussion from https://github.com/docker/notary/pull/494/files#r50593036

### DIFF
--- a/const.go
+++ b/const.go
@@ -2,9 +2,10 @@ package notary
 
 // application wide constants
 const (
-	// MaxMetaSize is the maximum size of metadata - 5MiB.
-	// Should be used only for testing, downloading new timestamps, or downloading roles when we have no local metadata
-	MaxMetaSize int64 = 5 << 20
+	// MaxDownloadSize is the maximum size we'll download for metadata if no limit is given
+	MaxDownloadSize int64 = 100 << 20
+	// MaxTimestampSize is the maximum size of timestamp metadata - 1MiB.
+	MaxTimestampSize int64 = 1 << 20
 	// MinRSABitSize is the minimum bit size for RSA keys allowed in notary
 	MinRSABitSize = 2048
 	// MinThreshold requires a minimum of one threshold for roles; currently we do not support a higher threshold
--- a/const.go
+++ b/const.go
@@ -2,7 +2,8 @@ package notary
 
 // application wide constants
 const (
-	// MaxMetaSize is the maximum size of metadata - 5MiB
+	// MaxMetaSize is the maximum size of metadata - 5MiB.
+	// Should be used only for testing, downloading new timestamps, or downloading roles when we have no local metadata
 	MaxMetaSize int64 = 5 << 20
 	// MinRSABitSize is the minimum bit size for RSA keys allowed in notary
 	MinRSABitSize = 2048
--- a/const.go
+++ b/const.go
@@ -2,6 +2,8 @@ package notary
 
 // application wide constants
 const (
+	// MaxMetaSize is the maximum size of metadata - 5MiB
+	MaxMetaSize int64 = 5 << 20
 	// MinRSABitSize is the minimum bit size for RSA keys allowed in notary
 	MinRSABitSize = 2048
 	// MinThreshold requires a minimum of one threshold for roles; currently we do not support a higher threshold
--- a/tuf/store/filestore.go
+++ b/tuf/store/filestore.go
@@ -2,6 +2,7 @@ package store
 
 import (
 	"fmt"
+	"github.com/docker/notary"
 	"io/ioutil"
 	"os"
 	"path"
@@ -45,6 +46,7 @@ func (f *FilesystemStore) getPath(name string) string {
 }
 
 // GetMeta returns the meta for the given name (a role) up to size bytes
+// If size is -1, this corresponds to "infinite," but we cut off at 100MB
 func (f *FilesystemStore) GetMeta(name string, size int64) ([]byte, error) {
 	meta, err := ioutil.ReadFile(f.getPath(name))
 	if err != nil {
@@ -52,6 +54,9 @@ func (f *FilesystemStore) GetMeta(name string, size int64) ([]byte, error) {
 			err = ErrMetaNotFound{Resource: name}
 		}
 		return nil, err
+	}
+	if size == -1 {
+		size = notary.MaxDownloadSize
 	}
 	// Only return up to size bytes
 	if int64(len(meta)) < size {
--- a/tuf/store/filestore.go
+++ b/tuf/store/filestore.go
@@ -44,7 +44,7 @@ func (f *FilesystemStore) getPath(name string) string {
 	return filepath.Join(f.metaDir, fileName)
 }
 
-// GetMeta returns the meta for the given name (a role)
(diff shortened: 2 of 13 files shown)


cyli:
Apologies - needs a rebase (and I think I may have used maxSize in some more places in that last PR that got merged).  But otherwise LGTM!

riyazdf:
@cyli no problem, rebased :)

cyli:
LGTM!

diogomonica:
I agree with moving `maxSize` into `const.go`, but this default should only be used when downloading `timestamp.json`.

In TUF, only `timestamp.json` has an unknown size (since its the first file downloaded on `.Update()`), so it is the only file that should be bounded by the default `maxSize` variable. For every other file, we should know exactly how many bytes we are supposed to download.

`timestamp.json` includes how many bytes to download for `snapshot.json`, which in turn includes how many bytes to download for both `root.json` and for `targets.json`.

In notary, even though `GetMeta()` on `tuf/filestore.go` should probably implement the size checks, we aren't vulnerable to infinite data attacks beause we use a `b := io.LimitReader(resp.Body, size)`  in `func (s HTTPStore) GetMeta(name string, size int64)`

/cc @endophage

diogomonica:
Auditing the code for maxSize, we only have three non-test usages.

`client.go:bootstrapClient() `:
```
rootJSON, cachedRootErr := r.fileStore.GetMeta("root", maxSize)
...
tmpJSON, err := remote.GetMeta("root", maxSize)
```
`/tuf/client/client.go:downloadRoot()`
```
size := maxSize
```
`/tuf/client/client.downloadTimestamp()`
```
cachedTS, err := c.cache.GetMeta(role, maxSize)
...
raw, s, err := c.downloadSigned(role, maxSize, nil)
```

- Since `bootstrapClient` is only used when we don't already have local metadata for the repository, it makes sense to download root with a maxSize (even though I don't necessarily think 5MB is the best maxSize for a root file).
- The use of `maxSize` on `downloadTimestamp()` make sense
- And finally, I think we have a bug on `downloadRoot`, because it's being called by `Update()` directly without attempting to see how much data should be downloaded.

diogomonica:
It would be great to add a slowtest that verifies if notary is capable of updating to a 5MB+ targets and snapshot files (since those are the most likely to get to that size).

riyazdf:
Good catch.  In `downloadRoot()` we do check later in the code for a size if we have a snapshot, and in the code preceding the `downloadRoot()` call we attempt to download a snapshot (but may fail) in `c.update()`.  It may not be guaranteed that we have a snapshot locally, though:

```
if c.local.Snapshot != nil {
		size = c.local.Snapshot.Signed.Meta[role].Length
		expectedSha256 = c.local.Snapshot.Signed.Meta[role].Hashes["sha256"]
	}
```

I can add a slow test for updating on 5MB+ targets/snapshot files

endophage:
@diogomonica the main reason I really want to see the size implemented across all the stores is so we can write consistent tests for them at the interface level. It'll help avoid any weird behaviour between tests and reality.

const.go
```diff
@@ -2,6 +2,9 @@ package notary
 
 // application wide constants
 const (
+	// MaxMetaSize is the maximum size of metadata - 5MiB.
```
endophage: Should be clear it's only the max size for a timestamp, which we download without any knowledge about exactly how large it will be.

cyli:
@riyazdf Sorry about yubikey failure.  :(  That is unfortunately the #498 issue.

riyazdf:
@diogomonica @endophage I restricted the use of the size constant only to timestamps, and introduced a `-1` option for `GetMeta` that will read "infinite" metadata but stop at 100MB under the hood.

For `downloadRoot()`, I talked with @cyli and have come to the conclusion that we can't place any guaranteed size restrictions on root with the current TUF metadata update cycle because it's possible that trying to download the timestamp or snapshot metadata to get the root size may fail due to mismatching signatures.

We could get around this by changing the update flow to download timestamp, snapshot, and then root in that order while checking hashes and sizes along the way, and then verifying the content on the three metadata files.  This is a pretty big change, what do you think?

diogomonica:
I think this is actually a problem with TUF as described.

If someone rotates a timestamp key, when the client goes to validate the timestamp.json and fails, it will immediately go to download root.json and is served with potentially infinite data (in our case, up to 100mb). The attacker only has to corrupt the timestamp.json to have this happen.

I'm fine with this as is.

tuf/store/filestore.go
```diff
@@ -53,7 +54,14 @@ func (f *FilesystemStore) GetMeta(name string, size int64) ([]byte, error) {
 		}
 		return nil, err
 	}
-	return meta, nil
+	if size == -1 {
+		size = 100 << 20
```
diogomonica: This should also be a constant, not in code.

riyazdf: done!

tuf/store/filestore.go
```diff
@@ -53,7 +55,14 @@ func (f *FilesystemStore) GetMeta(name string, size int64) ([]byte, error) {
 		}
 		return nil, err
 	}
-	return meta, nil
+	if size == -1 {
```
endophage: We should decide what we want our paradigm to be. I used the 0 value in listing timestamps to indicate "unlimited" because when requesting something that _must_ have a size, 0 is obviously non-sensical. I felt like it also fits in with Go's "zero" value idea. So if somebody created an int via `var x int`, then never assigned a more restricted value, the behaviour would be "get everything" rather than "get nothing".

riyazdf: I'm open to changing this to 0 if there's a consensus, though for me 0 feels more like a `nil` than "unlimited".  I agree that we should be consistent across notary with whatever we choose.

I chose a negative value here to be consistent with how Go handles `strings.Replace` -- a negative value replaces every occurrence in your input string.

cyli: I have no strong opinions.  0 it's basically a no-op, and seems silly to call a function with, but still seems to be a valid value ("please do this 0 times"), where as you can't do things `-1` times.

I'm also of the opinion that consistency is more important than the exact value, though :)

endophage: @riyazdf good point on `strings.Replace`. We should follow what Go's core libs are already doing. I'll update listing timestamps to use `-1`

riyazdf: Technically I think `strings.Replace` actually accepts any negative number for "unlimited" (https://golang.org/src/strings/strings.go?s=18112:18158#L681)  -- so we could change to any negative number or keep it as `-1`

endophage: I feel like I've seen -1 in other languages. I prefer the precision of saying it's a specific value, even if we code it to just be `foo < 0`

riyazdf: @endophage agreed, I like keeping it a specific value (and also have seen -1 in other languages as well, just can't recall where)

PR #496, opened 2016-01-22
Update when server metadata is corrupt

cyli:
Adds tests for if the server has invalid checksums, or the metadata is otherwise invalid.

This fixes an issue where, if the server has an invalid timestamp file but one that is unmarshallable to a `data.Signed` object, the update just fails.  It should instead try to fall back on the cached timestamp.  This is not true for any other role (which behave correctly).

There are also a bugs with downloading a new root that has the wrong version when the old root is expired, or downloading a new root during the bootstrap phase that doesn't have enough signatures.  I have not fixed these because I think they require a bigger refactor of `certs.ValidateRoot`, and we should maybe unify the root validations for both bootstrap and regular download.

I've skipped these cases for now.

### DIFF
--- a/client/client_update_test.go
+++ b/client/client_update_test.go
@@ -178,26 +178,30 @@ func TestUpdateSucceedsEvenIfCannotWriteExistingRepo(t *testing.T) {
 	}
 }
 
-type messUpMetadata func(role string) error
+type swizzleFunc func(*testutils.MetadataSwizzler, string) error
+type swizzleExpectations struct {
+	desc       string
+	swizzle    swizzleFunc
+	expectErrs []interface{}
+}
 
-func waysToMessUpLocalMetadata(repoSwizzler *testutils.MetadataSwizzler) map[string]messUpMetadata {
-	return map[string]messUpMetadata{
-		// for instance if the metadata got truncated or otherwise block corrupted
-		"invalid JSON": repoSwizzler.SetInvalidJSON,
-		// if the metadata was accidentally deleted
-		"missing metadata": repoSwizzler.RemoveMetadata,
-		// if the signature was invalid - maybe the user tried to modify something manually
-		// that they forgot (add a key, or something)
-		"signed with right key but wrong hash": repoSwizzler.InvalidateMetadataSignatures,
-		// if the user copied the wrong root.json over it by accident or something
-		"signed with wrong key": repoSwizzler.SignMetadataWithInvalidKey,
-		// self explanatory
-		"expired": repoSwizzler.ExpireMetadata,
+var waysToMessUpLocalMetadata = []swizzleExpectations{
+	// for instance if the metadata got truncated or otherwise block corrupted
+	{desc: "invalid JSON", swizzle: (*testutils.MetadataSwizzler).SetInvalidJSON},
+	// if the metadata was accidentally deleted
+	{desc: "missing metadata", swizzle: (*testutils.MetadataSwizzler).RemoveMetadata},
+	// if the signature was invalid - maybe the user tried to modify something manually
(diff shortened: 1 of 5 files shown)


client/client_update_test.go
```diff
@@ -606,3 +636,436 @@ func testUpdateRemoteNon200Error(t *testing.T, opts updateOpts, errExpected inte
...
+// If the local root is corrupt, and the remote root is corrupt, we should fail
+// to update.  Note - this one is really slow.
+func TestUpdateLocalAndRemoteRootCorrupt(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping test in short mode")
+	}
+	for _, localExpt := range waysToMessUpLocalMetadata {
+		for _, serverExpt := range waysToMessUpServer {
+			if localExpt.desc == "expired metadata" && serverExpt.desc == "lower metadata version" {
+				// TODO: bug right now where if the local metadata is invalid, we just download a
```
diogomonica: This one seems tricky. What we're saying is that: we know there is corrupt local metadata, but we're going to trust parts of that local metadata to make decisions about the validity of the remote state.

Feels like a fail open vs fail closed decision, more than a bug. Do we want the fail closed?

client/client_update_test.go
```diff
@@ -606,3 +636,436 @@ func testUpdateRemoteNon200Error(t *testing.T, opts updateOpts, errExpected inte
...
+				// new version - we verify the signatures and everything, but don't check the version
+				// against the previous if we can
+				continue
+			}
+			if serverExpt.desc == "insufficient signatures" {
+				// TODO: bug right now if we download the root during the bootstrap phase,
```
diogomonica: We should meet as a team and decide if we want to continue supporting thresholds at all. It might have to wait until we have a PoC of the signature "pipeline" feature.

client/client_update_test.go
```diff
@@ -606,3 +636,436 @@ func testUpdateRemoteNon200Error(t *testing.T, opts updateOpts, errExpected inte
...
+		}
+	}
+}
+
+// If there's is a local cache, but the remote server has new data (some corrupted),
+// we go immediately to check the remote server for root.  Having
+// extra spaces doesn't prevent it from validating during bootstrap,
+// but it will fail validation during update. So it will fail with or without
+// a force check (update for write).  If any of the other roles (except
+// timestamp, because there is no checksum for that) have an extra space, they
```
diogomonica: Small detail, but I think throughout all the comments, we should replace "have an extra space" for "have been modified in transit", or something that makes it more obvious that has been "malicious" intervention here.

This isn't immediately obvious what you mean by "have an extra space" until we reach `TestSwizzlerAddExtraSpace()`

diogomonica:
This seems pretty good coverage of the core behavior of notary around modified/corrupt metadata. Great job.

Small comments, but overall LGTM.

QQ: is the CI running only short tests? Have we configured CircleCI to run the longer ones every 12 hours or so, or are we, by merging this, no longer having that test coverage on circleCI?

client/client_update_test.go
cyli: True, I think that makes sense for all other cases of corruption.  Being expired feels like a special kind of invalid local metadata though.

riyazdf:
I agree with @diogomonica that it might be better to explain the motivation behind adding an extra space, specifically on the behavior of serialization vs. checksum in comments.  LGTM!

cyli:
@diogomonica CI is always running long tests.  I added the short skip so that if developers wanted to run short tests for quick sanity checks, these will be skipped

diogomonica:
@cyli rebased needed

cyli:
@diogomonica done!

PR #498, opened 2016-01-25
Root is always on disk unless restored from backup, in which case it may only be on a yubikey

cyli:


riyazdf:
thanks for the catch, LGTM!

PR #500, opened 2016-01-26
Don't run all the tests with yubikey

cyli:
Since we have a single builder, and they're slow, we don't want this to be a bottleneck.  If we can figure out how to have multiple VMs, one for each yubikey in the builder, then we can probably start running full tests again.


diogomonica:
LGTM

riyazdf:
LGTM!

PR #508, opened 2016-01-27
move verifier and update logs down to debug level

### LINKED ISSUE #507: Dial down info logging
For when notary is called as a library, we probably don't need signature verification logs at `Info` level -- this could be `Debug` since a delegated user won't be able to verify signatures he does not have keys for.

riyazdf:
Reduces log levels for signature verification and update() failures to debug, to avoid confusing situations when calling notary library.  The two main situations this helps are when rotating the snapshot key to server (snapshot validation fails before updating the root meta) and when publishing with delegations (signatures won't validate for keys you do not possess).  Closes #507 


diogomonica:
LGTM

PR #509, opened 2016-01-27
Add check for RSA key len before adding delegation

### LINKED ISSUE #504: Reject short keys when adding delegations
`notary-server` will reject short keys (< 2048 bits), but `notary delegation add` will accept them -- add a check during the CLI to reject the delegation.

riyazdf:
If the cert we're adding to the delegation has a RSA key and it's too short (< 2048), we'll reject before adding the delegation.  Without this check, the server will reject when we try to publish with the short key.  Closes #504.


HuKeping:
Seems great!

diogomonica:
LGTM

PR #510, opened 2016-01-27
return a better error message if the HTTP status code was 401

### LINKED ISSUE #288: Fix error output when notary returns a 401
It's really annoying that the 401 response is the only one that isn't JSON formatted. Additionally the behaviour is poor as there is little indication of what/where the problem is (c.f. #287).

It will need to be configurable for legacy support. The PR will need to be made against docker/distribution. If it's already configurable and I've missed it, then we need to update notary-server to configure it to use JSON.

endophage:
Closes #288 

We've made other changes that mean we can better message a 401 on the client. This just provides a clearer error message.

### DIFF
--- a/tuf/store/httpstore.go
+++ b/tuf/store/httpstore.go
@@ -33,6 +33,9 @@ type ErrServerUnavailable struct {
 }
 
 func (err ErrServerUnavailable) Error() string {
+	if err.code == 401 {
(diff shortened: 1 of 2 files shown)


cyli:
LGTM

riyazdf:
LGTM!

PR #513, opened 2016-01-28
Notary for 1.10 in docs.docker.com

moxiegirl:
For the 1.10 release we are moving to product-name taxonomy to help users find documentation about a product. Registry and Notary are moving under a menu Marketing calls Component Projects. This PR makes this menu:

image

Note: The rest of the menu may look a bit foobared --- don't freak. I have multiple PRs in play across several repos and when they all close the currently exposed codes will go away.

Signed-off-by: Mary Anthony <mary@docker.com>

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -1,5 +1,6 @@
 <!--[metadata]>
 +++
+draft=true
 title = "Notary CLI"
 description = "Description of the Notary CLI"
 keywords = ["docker, notary, trust, image, signing, repository, cli"]
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,16 +1,18 @@
-<!--[metadata]>
+<!--* [metadata]>
 +++
-title = "Docker Notary"
+title = "Notary"
 description = "List of Notary Documentation"
 keywords = ["docker, notary, trust, image, signing, repository, tuf"]
 [menu.main]
 identifier="mn_notary"
-parent="mn_docker_hub"
+parent="mn_components"
 weight=4
 +++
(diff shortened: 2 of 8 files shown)


carolfh:
@mary @NathanMcCauley 
cc: @rajat-docker   Hi Mary--I thought we were not using the name "notary", rather trusted content. Chatting with Rajat and he recommended you pinging Nathan. Let me know, as it affects my docs as well.

moxiegirl:
@carolfh This is is for the component not for the embedded trust functionality in DTR.

carolfh:
@moxiegirl Thanks!

diogomonica:
@moxiegirl tests for docs are failing now.

cyli:
@moxiegirl It looks like the problem is that our docs link to other notary documents using relative links - for example, `index.md` links to `cli.md`, which is in the same folder.  But the notary docs all get moved into the subdir `notary` in the global docker docs, so the links end up wrong.

If we changed the links to `notary/cli.md` viewing just our repo docs would be broken.

moxiegirl:
I'll rebase and resend.

moxiegirl:
@cyli Thanks for the feedback. The problem wasn't related to the move, Hugo actually handles that fine.  The Dockerfile in the project was pulling in old stuff. I've fixed that and couple of other issues.

moxiegirl:
@diogomonica @cyli @endophage  Hey gang, the test have past and I've updated the menu.  Can I get a  final review and merge?

moxiegirl:
Ty Dr. D :syringe:

PR #514, opened 2016-01-28
Ensure empty string path is properly handled, add --all-paths flag

### LINKED ISSUE #505: Ensure "" path is being used for adding delegations
It seems that currently the empty string path prefix (which corresponds to all paths) is not behaving as expected as a command line arg, or being set properly when running `notary delegation add`

riyazdf:
This improves notary CLI to properly display `""` in the console (before it was just blank) and also describes it as `<all paths>`.  This also adds a `--all-paths` flag that will add `""` when used with`delegation add`, and clear all paths on `delegation remove`.  Removing the individual `""` path will only remove that path.

Closes #505.

### DIFF
--- a/cmd/notary/delegations.go
+++ b/cmd/notary/delegations.go
@@ -42,8 +42,8 @@ type delegationCommander struct {
 	configGetter func() (*viper.Viper, error)
 	retriever    passphrase.Retriever
 
-	paths               []string
-	removeAll, forceYes bool
+	paths                         []string
+	allPaths, removeAll, forceYes bool
 }
 
 func (d *delegationCommander) GetCommand() *cobra.Command {
@@ -52,12 +52,13 @@ func (d *delegationCommander) GetCommand() *cobra.Command {
 
 	cmdRemDelg := cmdDelegationRemoveTemplate.ToCommand(d.delegationRemove)
 	cmdRemDelg.Flags().StringSliceVar(&d.paths, "paths", nil, "List of paths to remove")
-	cmdRemDelg.Flags().BoolVarP(
-		&d.forceYes, "yes", "y", false, "Answer yes to the removal question (no confirmation)")
+	cmdRemDelg.Flags().BoolVarP(&d.forceYes, "yes", "y", false, "Answer yes to the removal question (no confirmation)")
+	cmdRemDelg.Flags().BoolVar(&d.allPaths, "all-paths", false, "Remove all paths from this delegation")
 	cmd.AddCommand(cmdRemDelg)
 
 	cmdAddDelg := cmdDelegationAddTemplate.ToCommand(d.delegationAdd)
 	cmdAddDelg.Flags().StringSliceVar(&d.paths, "paths", nil, "List of paths to add")
+	cmdAddDelg.Flags().BoolVar(&d.allPaths, "all-paths", false, "Add all paths to this delegation")
 	cmd.AddCommand(cmdAddDelg)
 	return cmd
 }
@@ -121,18 +122,19 @@ func (d *delegationCommander) delegationRemove(cmd *cobra.Command, args []string
 	}
 
 	// If we're only given the gun and the role, attempt to remove all data for this delegation
-	if len(args) == 2 && d.paths == nil {
+	if len(args) == 2 && d.paths == nil && !d.allPaths {
 		d.removeAll = true
 	}
 
 	keyIDs := []string{}
-	// Change nil paths to empty slice for TUF
-	if d.paths == nil {
-		d.paths = []string{}
-	}
 
 	if len(args) > 2 {
 		keyIDs = args[2:]
+	}
+
+	// If the user passes --all-paths, don't use any of the passed in --paths
+	if d.allPaths {
+		d.paths = nil
 	}
 
 	// no online operations are performed by add so the transport argument
@@ -160,6 +162,12 @@ func (d *delegationCommander) delegationRemove(cmd *cobra.Command, args []string
 			return fmt.Errorf("failed to remove delegation: %v", err)
 		}
 	} else {
+		if d.allPaths {
+			err = nRepo.ClearDelegationPaths(role)
+			if err != nil {
+				return fmt.Errorf("failed to remove delegation: %v", err)
+			}
+		}
 		// Remove any keys or paths that we passed in
 		err = nRepo.RemoveDelegationKeysAndPaths(role, keyIDs, d.paths)
 		if err != nil {
@@ -175,6 +183,9 @@ func (d *delegationCommander) delegationRemove(cmd *cobra.Command, args []string
 		if len(keyIDs) > 0 {
 			removingItems = removingItems + fmt.Sprintf("with keys %s, ", keyIDs)
 		}
+		if d.allPaths {
+			removingItems = removingItems + "with all paths,"
+		}
 		if d.paths != nil {
 			removingItems = removingItems + fmt.Sprintf("with paths [%s], ", prettyPrintPaths(d.paths))
 		}
@@ -187,7 +198,7 @@ func (d *delegationCommander) delegationRemove(cmd *cobra.Command, args []string
 
 // delegationAdd creates a new delegation by adding a public key from a certificate to a specific role in a GUN
 func (d *delegationCommander) delegationAdd(cmd *cobra.Command, args []string) error {
-	if len(args) < 2 || len(args) < 3 && d.paths == nil {
+	if len(args) < 2 || len(args) < 3 && d.paths == nil && !d.allPaths {
 		cmd.Usage()
 		return fmt.Errorf("must specify the Global Unique Name and the role of the delegation along with the public key certificate paths and/or a list of paths to add")
 	}
@@ -219,6 +230,18 @@ func (d *delegationCommander) delegationAdd(cmd *cobra.Command, args []string) e
 		}
 	}
 
+	for _, path := range d.paths {
+		if path == "" {
+			d.allPaths = true
+			break
+		}
+	}
+
+	// If the user passes --all-paths (or gave the "" path in --paths), give the "" path
+	if d.allPaths {
+		d.paths = []string{""}
+	}
+
 	// no online operations are performed by add so the transport argument
 	// should be nil
 	nRepo, err := notaryclient.NewNotaryRepository(
@@ -244,9 +267,16 @@ func (d *delegationCommander) delegationAdd(cmd *cobra.Command, args []string) e
 	}
 
 	cmd.Println("")
+	addingItems := ""
+	if len(pubKeyIDs) > 0 {
+		addingItems = addingItems + fmt.Sprintf("with keys %s, ", pubKeys)
+	}
+	if d.paths != nil || d.allPaths {
+		addingItems = addingItems + fmt.Sprintf("with paths [%s], ", prettyPrintPaths(d.paths))
+	}
 	cmd.Printf(
-		"Addition of delegation role %s with keys %s to paths [%s], to repository \"%s\" staged for next publish.\n",
-		role, pubKeyIDs, prettyPrintPaths(d.paths), gun)
+		"Addition of delegation role %s %s to repository \"%s\" staged for next publish.\n",
+		role, addingItems, gun)
 	cmd.Println("")
 	return nil
 }
--- a/cmd/notary/delegations.go
+++ b/cmd/notary/delegations.go
@@ -184,7 +184,7 @@ func (d *delegationCommander) delegationRemove(cmd *cobra.Command, args []string
 			removingItems = removingItems + fmt.Sprintf("with keys %s, ", keyIDs)
 		}
 		if d.allPaths {
-			removingItems = removingItems + "with all paths,"
+			removingItems = removingItems + "with all paths, "
 		}
 		if d.paths != nil {
 			removingItems = removingItems + fmt.Sprintf("with paths [%s], ", prettyPrintPaths(d.paths))
@@ -198,7 +198,7 @@ func (d *delegationCommander) delegationRemove(cmd *cobra.Command, args []string
 
 // delegationAdd creates a new delegation by adding a public key from a certificate to a specific role in a GUN
 func (d *delegationCommander) delegationAdd(cmd *cobra.Command, args []string) error {
-    // We must have at least the gun and role name, and at least one key or path (or the --all-paths flag) to add
+	// We must have at least the gun and role name, and at least one key or path (or the --all-paths flag) to add
 	if len(args) < 2 || len(args) < 3 && d.paths == nil && !d.allPaths {
 		cmd.Usage()
 		return fmt.Errorf("must specify the Global Unique Name and the role of the delegation along with the public key certificate paths and/or a list of paths to add")
@@ -276,7 +276,7 @@ func (d *delegationCommander) delegationAdd(cmd *cobra.Command, args []string) e
 		addingItems = addingItems + fmt.Sprintf("with paths [%s], ", prettyPrintPaths(d.paths))
 	}
 	cmd.Printf(
-		"Addition of delegation role %s %s to repository \"%s\" staged for next publish.\n",
+		"Addition of delegation role %s %sto repository \"%s\" staged for next publish.\n",
 		role, addingItems, gun)
 	cmd.Println("")
 	return nil
--- a/cmd/notary/delegations.go
+++ b/cmd/notary/delegations.go
@@ -198,6 +198,7 @@ func (d *delegationCommander) delegationRemove(cmd *cobra.Command, args []string
 
 // delegationAdd creates a new delegation by adding a public key from a certificate to a specific role in a GUN
 func (d *delegationCommander) delegationAdd(cmd *cobra.Command, args []string) error {
+    // We must have at least the gun and role name, and at least one key or path (or the --all-paths flag) to add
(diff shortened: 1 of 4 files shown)


cyli:
This definitely makes everything a lot easier!  :D  Unfortunately I think it may need to take into account whether we are modifying a role or not?  I tried the following:

1. `notary init repo1`
2. `notary delegation add repo1 targets/releases <path-to-pem>`
3. `notary publish repo1`

Awesome, after updating the repo, the delegation has the empty path.

But if I do:

1. `notary init repo2`
2. `notary delegation add repo2 targets/releases <path-to-pem> --paths="/"`
3. `notary publish repo2`
4. `notary delegation add repo2 targets/releases <path-to-pem2>`
5. `notary publish repo2`

I get a delegation with both keys, but paths "/" and "", and "" is more permissive so it might take precedence.  I'm not sure that's desirable?  Maybe you had it right in the first place (nil by default) and we just have to better document how to add the empty string?  Not sure.  :|  @endophage?

riyazdf:
@cyli You're definitely right that it makes modifying a pre-existing delegation less intuitive.  I can't think of a good way to get around this unless we go with your suggestion of making `nil` the default, and documenting how to input empty string (which has to be escaped as `--paths=\"\"`).  I'm definitely open to ideas here, I'm leaning toward the `nil` default and stronger documentation -- maybe this can be made clearer in the output after running `notary delegation add`?

A second observation from your example: if a delegation already has a path of `/a`, and a user tries to add path `/a/b`, should we even bother showing `/a/b` if it is a successful publish?  We could probably try to do some smarter merging when adding to the delegation (right now we just merge string slices as is).

cyli:
@riyazdf I remember @endophage mentioning something like that before, where we were debating whether we wanted to check the prefixes of the paths when merging, and maybe the outcome was that it should probably be resolved by the admin?.  It would probably be a nice UI feature for the CLI to log a warning or something if it detects that, though.

endophage:
Yeah, inefficient/overlapping paths should be resolved by admin. Safer than us trying to be clever.

On the empty path problem. Are we able to move the addition of the empty path to the time at which we apply the changelist? At that point I believe we know whether the role already exists or not, so we can detect if we're creating a role, and if no specific paths have been supplied.

riyazdf:
@endophage We can detect if the role already exists when we apply the changelist, but we won't know if the user passed in `--paths=\"\"` or if the flag defaulted to `""`.

We could get around this by defaulting the `--paths` flag to `nil` and we default to a `""` path when applying changelists if we get a `nil` path on a brand new delegation?  What do you (and @cyli) think?  We'll need to make it very clear that we add `""` by default to new delegations without specified paths.

diogomonica:
`--all-paths`

riyazdf:
I added in the `--all-paths` flag, which required an additional key to the embedded delegation object in the changelist.

I also changed `""` to be printed as `"" <all paths>` on the console to make things clearer when adding all paths, but I can change this back if we'd prefer not to.

endophage:
Great idea on the console printing. Users accommodate with a lot of non-intuitive stuff when it's well messaged.

client/changelist/change.go
```diff
@@ -88,6 +88,7 @@ type TufDelegation struct {
 	RemoveKeys             []string     `json:"remove_keys,omitempty"`
 	AddPaths               []string     `json:"add_paths,omitempty"`
 	RemovePaths            []string     `json:"remove_paths,omitempty"`
+	ClearPaths             bool         `json:"clear_paths,omitempty"`
```
diogomonica: `ClearAllPaths`

client/client.go
```diff
@@ -349,7 +349,7 @@ func (r *NotaryRepository) AddDelegation(name string, threshold int,
 // the repository when the changelist gets applied at publish time.
 // This does not validate that the delegation exists, since one might exist
 // after applying all changes.
-func (r *NotaryRepository) RemoveDelegation(name string, keyIDs, paths []string, removeAll bool) error {
+func (r *NotaryRepository) RemoveDelegation(name string, keyIDs, paths []string, clearPaths, removeAll bool) error {
```
diogomonica: `clearAllPaths`

client/client.go
```diff
@@ -378,6 +378,7 @@ func (r *NotaryRepository) RemoveDelegation(name string, keyIDs, paths []string,
 		tdJSON, err := json.Marshal(&changelist.TufDelegation{
 			RemoveKeys:  keyIDs,
 			RemovePaths: paths,
+			ClearPaths:  clearPaths,
```
diogomonica: `ClearAllPaths`

client/helpers.go
```diff
@@ -121,7 +121,12 @@ func changeTargetsDelegation(repo *tuf.Repo, c changelist.Change) error {
 			return err
 		}
 		r.RemoveKeys(td.RemoveKeys)
-		r.RemovePaths(td.RemovePaths)
+
+		removePaths := td.RemovePaths
+		if td.ClearPaths {
```
diogomonica: Not sure if I'm convinced of the need of adding a new `bool`. What if we iterate through `removePaths`, and if we find a `""`, we simply remove all of them?

cmd/notary/delegations.go
```diff
@@ -160,7 +165,7 @@ func (d *delegationCommander) delegationRemove(cmd *cobra.Command, args []string
 
 // delegationAdd creates a new delegation by adding a public key from a certificate to a specific role in a GUN
 func (d *delegationCommander) delegationAdd(cmd *cobra.Command, args []string) error {
-	if len(args) < 2 || len(args) < 3 && paths == nil {
+	if len(args) < 2 || len(args) < 3 && addPaths == nil && !allPaths {
```
diogomonica: I think it needs a comment on intended behavior

cmd/notary/delegations.go
```diff
@@ -188,6 +193,11 @@ func (d *delegationCommander) delegationAdd(cmd *cobra.Command, args []string) e
 		}
 	}
 
+	// If the user passes --all-paths, give the "" path
+	if allPaths {
```
diogomonica: What if we special case this to remove all the current paths when `""` gets added?

client/helpers.go
riyazdf: We could smartly remove paths (and it would be simpler code), but we would have to make it clear that there is smarter path logic around removing `""` to users, since we don't do any smart path logic anywhere else.  Users can currently add `""` and `path` as paths to the same delegation and both will display on the CLI, even though it doesn't really make sense in notary.

cc @endophage since we were talking earlier about how to manage paths via CLI

cmd/notary/delegations.go
riyazdf: good idea, will include

client/changelist/change.go
riyazdf: :+1:

cmd/notary/delegations.go
riyazdf: I think this is similar to this: https://github.com/docker/notary/pull/514/files#r51305926

I think we should decide whether or not we should have any smart logic around adding/removing paths.  It would require changes to `AddPaths` and `RemovePaths` in `roles.go` -- right now they just merge/subtract string slices without any logic around the values of the included paths.

diogomonica: It might be ok to only do "smart logic" around the "all paths" exception. If you add `""` and other stuff, ignore the other stuff (or remove it). If you delete `""` also make sure everything else is empty.

We don't have to do more smart logic than that. Not sure what @endophage things, but I would rather do that VS having another bool on the api.

client/helpers.go
endophage: @diogomonica 's suggesting is too dependent on order of operations. If a user wants to take an existing role that has the `""` path permissions, and restrict it to `"foo/"`, this implicit behaviour would require them to remove `""` before adding `"foo/"`.

cmd/notary/delegations.go
riyazdf: I agree that this is a nice compromise, `""` is a special prefix in this case and I agree it's better than having to add more object keys and booleans :)  @endophage and @cyli what do you think?

The only downside I can see here is if someone mistakenly adds `""` when they didn't mean to, and loses track of the previous paths, but this is a small headache and easily recoverable.

riyazdf:
I just implemented the "smart logic" that @diogomonica outlined around the `""` path (`""` clears other paths that are passed in when adding, removing `""` removes all paths) because I think it's the best way to use notary here, and doesn't further complicate the API.

I also think that we can think about tweaking the logic around removing the `""` path in particular once we improve the API into finer-grained calls https://github.com/docker/notary/issues/525.  Once we have a simpler way of clearing paths vs. removing just the `""` path, the only way to accomplish this is by adding more flags to `RemoveDelegation`.

diogomonica:
Rebase needed.

riyazdf:
@diogomonica rebased!

diogomonica:
@endophage final review?

cmd/notary/delegations.go
```diff
@@ -38,8 +38,8 @@ var cmdDelegationAddTemplate = usageTemplate{
 	Long:  "Add a keys to delegation using the provided public key PEM encoded X509 certificates in a specific Global Unique Name.",
 }
 
-var paths []string
-var removeAll, removeYes bool
+var addPaths, removePaths []string
```
endophage: What's the necessity for these 2 separate vars? Also:

```
var (
    paths []string
    allPaths, removeAll, removeYes bool
)
```

cmd/notary/delegations.go
```diff
@@ -38,8 +38,8 @@ var cmdDelegationAddTemplate = usageTemplate{
...
+var allPaths, removeAll, removeYes bool
```
endophage: We probably want to use the `-y` flag in other places (or will do), so I'd just call it `forceYes` or something more generic than `removeYes`

cmd/notary/delegations.go
riyazdf: I'll make them one var -- earlier they had different default values and so we needed two separate variables.  I'll also reformat this as you suggested!

cmd/notary/delegations.go
riyazdf: Makes sense, will do

endophage:
Sorry to be the naysayer but I still completely disagree with the remove command using `""` as an alternative  to infer "remove all".

Also wondering whether the flags should be `allowAll` and `denyAll` rather than `allPaths`. It's clearer about the behaviour whereas `allPaths` is intuitive to us only because we understand the data structures involved.

diogomonica:
@endophage your alternative being?

endophage:
The same thing I was suggesting last week. We already have an explicit flag, there's no need for this implicit behaviour, just treat `""` like it was any other path. If it exists in the role's `Path` list it gets removed, if it doesn't, it's a no-op

riyazdf:
@endophage I think that's a fine idea but I'm not sure if I see the benefit of allowing for this behavior against adding a flag to the `RemoveDelegation` API, since `notary delegation remove` is an offline operation.  I think making the API simpler and more flexible (see https://github.com/docker/notary/issues/525) will improve this and maybe we could allow for this behavior then.

In any case, I don't think this special way of handling `""` on remove takes away any functionality from the user -- instead of being able to add paths and then remove `""` individually, we're just making the user remove `""` first before adding paths he wants to restrict the delegation to.

diogomonica:
Agree with @riyazdf I think we should refactor.

diogomonica:
@riyazdf this needs rebase after #528 got merged.

riyazdf:
@diogomonica @endophage refactored and rebased.

Now `notary delegation remove --all-paths` will clear all paths, and `notary delegation remove --paths=\"\"` will just remove the empty path.  I've added these to the integration test.

cmd/notary/integration_test.go
```diff
@@ -365,6 +385,86 @@ func TestClientDelegationsInteraction(t *testing.T) {
 	assert.Contains(t, output, keyID)
 	assert.Contains(t, output, keyID2)
 
+	// Add a bunch of individual paths so we can test a delegation remove --all-paths
+	output, err = runCommand(t, tempDir, "delegation", "add", "gun", "targets/delegation", "--paths", "abcdef,123456")
+	assert.NoError(t, err)
+
+	// Add more individual paths so we can test a delegation remove --all-paths
+	output, err = runCommand(t, tempDir, "delegation", "add", "gun", "targets/delegation", "--paths", "banana,apple,orange,kiwi")
```
diogomonica: maybe some more complex paths? `banana/split` comes to mind ;)

cmd/notary/prettyprint.go
```diff
@@ -198,6 +198,19 @@ func prettyPrintRoles(rs []*data.Role, writer io.Writer, roleType string) {
 	table.Render()
 }
 
+// Pretty-prints a list of delegation paths, and ensures the empty string is printed as "" in the console
+func prettyPrintPaths(paths []string) string {
+	prettyPaths := []string{}
+	for _, path := range paths {
+		// manually escape "" and designate that it is all paths with an extra print <all paths>
+		if path == "" {
+			path = "\"\" <all paths>"
```
diogomonica: Why include `""`?

diogomonica:
LGTM

diogomonica:
@riyazdf oops, sorry about this. I realized that this was going to need a rebase after I merged @cyli PR.

cmd/notary/prettyprint.go
riyazdf: I think we should still display the actual `""` path so that the `notary delegation remove --paths=\"\"` has meaning, what do you think?

diogomonica: Sounds good.

riyazdf:
@diogomonica no problem, just rebased.

cmd/notary/integration_test.go
```diff
@@ -362,6 +382,86 @@ func TestClientDelegationsInteraction(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	// list delegations - we should see all of our paths
+	output, err = runCommand(t, tempDir, "-s", server.URL, "delegation", "list", "gun")
+	assert.NoError(t, err)
+	assert.NotContains(t, output, "abcdef")
+	assert.NotContains(t, output, "123456")
+	assert.NotContains(t, output, "banana/split")
+	assert.NotContains(t, output, "apple/crumble/pie")
+	assert.NotContains(t, output, "orange.peel")
```
diogomonica: love it

cmd/notary/delegations.go
```diff
@@ -171,9 +179,17 @@ func (d *delegationCommander) delegationRemove(cmd *cobra.Command, args []string
...
 	} else {
-		cmd.Printf(
-			"Removal of delegation role %s with keys %s and paths %s, to repository \"%s\" staged for next publish.\n",
-			role, keyIDs, d.paths, gun)
+		removingItems := ""
+		if len(keyIDs) > 0 {
+			removingItems = removingItems + fmt.Sprintf("with keys %s, ", keyIDs)
+		}
+		if d.allPaths {
+			removingItems = removingItems + "with all paths,"
```
cyli: Non-blocking nitpick - should this string terminate with a space like the other ones?  Alternately should none of them have a trailing space and the format string should just include the space?

riyazdf: Good point!  I'll fix this.

cmd/notary/delegations.go
```diff
@@ -239,9 +268,16 @@ func (d *delegationCommander) delegationAdd(cmd *cobra.Command, args []string) e
...
+	if len(pubKeyIDs) > 0 {
+		addingItems = addingItems + fmt.Sprintf("with keys %s, ", pubKeys)
+	}
+	if d.paths != nil || d.allPaths {
+		addingItems = addingItems + fmt.Sprintf("with paths [%s], ", prettyPrintPaths(d.paths))
+	}
 	cmd.Printf(
-		"Addition of delegation role %s with keys %s and paths %s, to repository \"%s\" staged for next publish.\n",
-		role, pubKeyIDs, d.paths, gun)
+		"Addition of delegation role %s %s to repository \"%s\" staged for next publish.\n",
```
cyli: Non-blocking nitpick Should this be `%sto repository` like above in the `delegationRemove` command, since all the additions have trailing spaces?

riyazdf: yes!  Will update

cmd/notary/prettyprint.go
cyli: Totally non-blocking, and this may already be done elsewhere, but I probably missed it - would sorting the paths make it easier for a user to see if they have overlapping path permissions?  So the all paths would always be in front, and then paths that overlap each other would be right next to each other?

endophage: I like @cyli's suggestion. Sort in lexical order, maybe we can even add some output syntax to indicate shadowing?

cyli:
Could we also two test cases where both `--path` and `--all-paths` are given simultaneously, and that if so, the `--paths` are ignored?

cyli:
LGTM otherwise!  :D

cmd/notary/prettyprint.go
riyazdf: Great suggestion!

riyazdf:
@cyli added your suggested test cases to the integration test and also added sorting to path pretty printing!

cyli:
@riyazdf :+1: Thanks!! :D

diogomonica:
Great team-work. This is looking good.

PR #515, opened 2016-01-28
Roles for targets via notary CLI

### LINKED ISSUE #511: enable using delegations on targets with notary CLI
docker/docker will automatically attempt to add to and list from a `targets/releases` role, but in notary there's currently no way to specify which delegation roles to add targets to.  The same functionality also doesn't exist for listing targets from specified roles or removing targets.

I propose we add a `--roles` flag to `notary add`/`notary remove`/`notary list` to make this configurable in notary CLI.  I have a PoC from my own notary delegation testing that I can put up once I work out some tests

riyazdf:
Adds `--roles` flags to `notary add`/`notary remove`/`notary list` to allow for adding to specific delegations.  The `--roles` flag takes a list of roles.

I used this for testing end-to-end delegation publishing with notary, and there's an integration test that does a publish to `targets/releases` after removing the targets key and copying over the delegation private key to `private/tuf_keys`.  This private key needs either the name format of `canonicalID_releases` or a role in the PEM headers.

Closes #511

### DIFF
--- a/cmd/notary/integration_test.go
+++ b/cmd/notary/integration_test.go
@@ -384,7 +384,9 @@ func TestClientDelegationsPublishing(t *testing.T) {
 
 	privKey, err := trustmanager.GenerateRSAKey(rand.Reader, 2048)
 	assert.NoError(t, err)
-	privKeyBytes, err := trustmanager.KeyToPEM(privKey, "")
+	privKeyBytesNoRole, err := trustmanager.KeyToPEM(privKey, "")
+	assert.NoError(t, err)
+	privKeyBytesWithRole, err := trustmanager.KeyToPEM(privKey, "user")
 	assert.NoError(t, err)
 	startTime := time.Now()
 	endTime := startTime.AddDate(10, 0, 0)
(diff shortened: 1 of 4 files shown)


cmd/notary/integration_test.go
```diff
@@ -362,6 +368,164 @@ func TestClientDelegationsInteraction(t *testing.T) {
 	assert.Contains(t, output, "No delegations present in this repository.")
 }
 
+// Initialize repo and test publishing targets with delegation roles
+func TestClientDelegationsPublishing(t *testing.T) {
```
cyli: Awesome and very comprehensive test!!!

cyli:
LGTM!  Thank you for fixing this!  This makes the complete delegation add/publishing targets/etc. workflow much easier!

cmd/notary/tuf.go
```diff
@@ -163,8 +164,9 @@ func tufList(cmd *cobra.Command, args []string) {
 		fatalf(err.Error())
 	}
 
-	// Retreive the remote list of signed targets
-	targetList, err := nRepo.ListTargets(data.CanonicalTargetsRole, "targets/releases")
+	// Retrieve the remote list of signed targets, prioritizing the passed-in list over targets
+	roles = append(roles, data.CanonicalTargetsRole)
+	targetList, err := nRepo.ListTargets(roles...)
```
diogomonica: Isn't this a change behavior? before "targets/releases" was mandated, and now it's not.

riyazdf: I think notary itself should be completely customizable for adding/removing/listing targets from any created role, and shouldn't hardcode "targets/releases".  And even so, the order we have in notary right now will prioritize the targets role over "targets/releases," which is not desirable.

docker calls `ListTargets` directly with prioritizing "targets/releases" (falling back to targets role if releases doesn't exist), and in docker it makes more sense to hardcode "targets/releases" https://github.com/docker/docker/blob/master/api/client/trust.go#L313

diogomonica: Makes sense.

PR #516, opened 2016-01-28
use only canonical IDs for display on delegation CLI commands,

### LINKED ISSUE #506: Use canonical IDs for all delegation commands
Since we currently need canonical IDs for the right filename to drop in delegation private keys, we should always display their IDs as canonical IDs instead of calling `key.ID()`.  The TUF metadata doesn't use the canonical ID but this is never exposed to the end user directly

riyazdf:
Since we display the canonical key ID when asking a delegation user to passphrases and also need to store the key under its canonical key ID when saving to `tuf_files`, it makes sense to only show the canonical ID to the user on notary CLI.

Before this change, we used the TUF key IDs on notary CLI commands, but still required the canonical key ID in the filename and would show the user the canoncial key ID if they needed to enter a passphrase when signing with that key.

This does require us to do more work to translate between key ID types, but this is only done for `delegation list` (to translate TUF key IDs to canonical IDs to show to the user) or a `publish` after a `delegation remove` (to translate user-supplied canonical IDs to TUF key IDs to apply the changelist logic).

This is unfortunately not the most elegant solution, but it will only affect delegation keys -- behavior for root/targets/snapshot keys remains unchanged.

Closes #506 

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -23,6 +23,7 @@ import (
 	"github.com/docker/notary/tuf/keys"
 	"github.com/docker/notary/tuf/signed"
 	"github.com/docker/notary/tuf/store"
+	"github.com/docker/notary/tuf/utils"
 )
 
 func init() {
@@ -529,6 +530,7 @@ func (r *NotaryRepository) GetChangelist() (changelist.Changelist, error) {
 }
 
 // GetDelegationRoles returns the keys and roles of the repository's delegations
+// Also converts key IDs to canonical key IDs to keep consistent with signing prompts
 func (r *NotaryRepository) GetDelegationRoles() ([]*data.Role, error) {
 	// Update state of the repo to latest
 	if _, err := r.Update(false); err != nil {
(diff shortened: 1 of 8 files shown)


client/client.go
```diff
@@ -564,13 +570,42 @@ func (r *NotaryRepository) GetDelegationRoles() ([]*data.Role, error) {
...
+		allDelegations = append(allDelegations, canonicalDelegations...)
+		// Add nested delegations to the exploration list
 		delegationsList = append(delegationsList, delegationMeta.Signed.Delegations.Roles...)
 	}
+
+	// Convert all key IDs to canonical IDs:
 	return allDelegations, nil
 }
 
+func replaceDelegationsWithCanonicalIDs(delegationInfo data.Delegations) ([]*data.Role, error) {
```
diogomonica: this is not a replace as much as it is a translate.

diogomonica:
LGTM.

client/client.go
riyazdf: good point, I'll update the name

PR #517, opened 2016-01-28
Add a doc about how to upgrade to v0.2.

cyli:
So that we have something to link against from docker.

### DIFF
--- a/docs/upgrading-client.md
+++ b/docs/upgrading-client.md
@@ -1,0 +1,52 @@
+<!--[metadata]>
++++
+title = "Upgrading Notary client"
+description = "How to upgrade the Notary client"
+keywords = ["docker, notary, notary-client"]
+[menu.main]
+parent="mn_notary"
++++
+<![end-metadata]-->
+
+# Upgrading the client
+
+### From `< v0.2` to `v0.2`
+
+Please download the v0.2 binary from [the Notary releases page](https://github.com/docker/notary/releases).
+
+##### New features:
+
+- Support for adding, editing, removing, and viewing information about [delegation roles](https://github.com/theupdateframework/tuf/blob/develop/docs/tuf-spec.txt#L252), and adding targets to delegation roles (as opposed to the targets role, which is currently the default).
+
+	This will allow repository owners to add one or more delegated users as collaborators, allowing them to use their own keys to publish.  In Notary's previous workflow, a repository owner had to share the targets key with each collaborator in order to enable collaborators to publish.
+
+- Support for rotating the snapshot key to one managed by the Notary server.  This is necessary in order to facilitate delegations without forcing repository owners to share the snapshot key with collaborators.
+
+	A collaborator with their own user key may sign a delegation, but without server snapshot-signing, they would be unable to publish unless they also had the snapshot key.
+
+	If repository owners want to reclaim control of their snapshot key, they can rotate the key again to a locally-generated key.
+
+- Support for remote key rotation (forcing server-managed keys to rotate).  Forcing the server to rotate a server-managed key should be a very rare occurrence, necessarily only in the case of server compromise or remote key algorithm obsolescence.
+
+- Drop-in support for user keys - users can just place their private key to be used for delegations in `~/.notary/private/tuf_keys` as `<keyID>.key`.  The PEM header "role" should be "user", although any invalid role is currently allowed.
+
+##### Incompatible changes:
+
+- Due to a bug in previous versions of Notary, if delegation keys are added to a repository, old clients (including, for instance, Docker versions < 1.10, which bundle older versions of Notary client) will not be able to view the repository.  Any attempt to download updated repository information with old client will result in an error.
+
+	However, repositories created or edited by Notary Client v0.2 without delegations will be perfectly readable, downloadable, and publishable by old clients (with the caveat of server snapshot-signing, see below).
+
+- If server snapshot-signing is enabled for a repository, old clients will be able to read from and download updates for the repository, but they may not publish.
+
+- Due to a bug in previous versions of Notary, if new keys are created with Notary Client v0.2 and to disk, an old client (including, for instance, Docker versions < 1.10, which bundle older versions of Notary client) will not be able to use the same trust directory as the new client.  However, the new client will be able to and use read old-style keys.
+
+	If you need to switch between old style and new style clients, please use different trust directories (for example, `~/.notaryv1/` and `~/.notaryv2`).
+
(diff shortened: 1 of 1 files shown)


docs/upgrading-client.md
```diff
@@ -0,0 +1,52 @@
...
+
+- Due to a bug previous versions of notary, if delegation keys are added to a repository, old clients (including, for instance, Docker versions < 1.10, which bundle older versions of Notary client) will not be able to view the repository.  Any attempt to update with old client will result in an error.
+
+	However, repositories created/updated by Notary Client v0.2 without delegations will be perfectly readable, updatable, and publishable by old clients (with the caveat of server snapshot-signing, see below).
+
+- If server-snapshot-signing is enabled for a repository, old clients will be able to read from and update the repository, but they may not publish.
+
+- Due to a bug in previous versions of notary, if new keys are created with Notary Client v0.2 and to disk, an old client (including, for instance, Docker versions < 1.10, which bundle older versions of notary) will not be able to use the same trust directory as the new client.  However, the new client will be able to and use read old-style keys.
+
+	If you need to switch between old style and new stile clients, please use different trust directories (for example, `~/.notaryv1/` and `~/.notaryv2`).
```
riyazdf: typo nit: "style"

docs/upgrading-client.md
```diff
@@ -0,0 +1,52 @@
...
+
+### From `< v0.2` to `v0.2`
+
+Please download the v0.2 binary from [the Notary releases page](https://github.com/docker/notary/releases).
+
+##### New features:
+
+- Support for adding, editing, removing, and viewing information about [delegation roles](https://github.com/theupdateframework/tuf/blob/develop/docs/tuf-spec.txt#L252), and adding targets to delegation roles (as opposed to the targets role, which is currently the default).
+
+	This will allow repository owners to add one or more delegated users as collaborators, allowing them to use their own keys to publish.  Previous workflow, where a repository owner had share the targets key with a collaborator in order to enable them to publish.
```
riyazdf: Rewording the second sentence:

`In notary's previous workflow, a repository owner had to share the targets key with each collaborator in order to enable collaborators to publish.`

docs/upgrading-client.md
```diff
@@ -0,0 +1,52 @@
...
+
+- Support for rotating the snapshot key to one managed by the Notary server.  This is necessary in order to facility delegations without forcing users to share the snapshot key with collaborators.
```
riyazdf: Two small nits: `to facilitate` (instead of facility)

and `without forcing repository owners` (instead of users)

docs/upgrading-client.md
```diff
@@ -0,0 +1,52 @@
...
+
+	A collaborator with their own user key may sign a delegation, but without server-snapshot signing, they would be unable to publish unless they also had the snapshot key.
+
+	If users want to reclaim control of their snapshot key, they can rotate the key again to a locally-generated key.
```
riyazdf: I'd also replace users with `repository owners` here,

docs/upgrading-client.md
```diff
@@ -0,0 +1,52 @@
...
+
+- Support for remote key rotation (forcing server-managed keys to rotate).  Forcing the server to rotate a server-managed key should be a very rare occurrence, necessarily only in the case of server compromise or remote key algorithm obsolescence.
+
+- Drop-in support for user keys - users can just place their private key to be used for delegations in `~/.notary/private/tuf_keys` as `<keyID>.key`.  The PEM header "role" should be "user", although any invalid role is currently allowed.
```
riyazdf: Do we also want to mention `<keyID>_role.key` without a PEM header?  It's not as clean, however.  Also, this may change if we introduce a `delegation key import` command or something similar.

docs/upgrading-client.md
```diff
@@ -0,0 +1,52 @@
...
+
+##### Incompatible changes:
+
+- Due to a bug previous versions of notary, if delegation keys are added to a repository, old clients (including, for instance, Docker versions < 1.10, which bundle older versions of Notary client) will not be able to view the repository.  Any attempt to update with old client will result in an error.
```
riyazdf: typo: "Due to a bug in" (add "in")

docs/upgrading-client.md
```diff
@@ -0,0 +1,52 @@
...
+
+	However, repositories created/updated by Notary Client v0.2 without delegations will be perfectly readable, updatable, and publishable by old clients (with the caveat of server snapshot-signing, see below).
```
riyazdf: We should keep consistent with the hyphenation of `server snapshot-signing` (vs. `server-snapshot-signing`)

docs/upgrading-client.md
```diff
@@ -0,0 +1,52 @@
...
+
+- If server-snapshot-signing is enabled for a repository, old clients will be able to read from and update the repository, but they may not publish.
```
riyazdf: the way `update` is used here makes it seem like clients can somehow write to the repo.  Maybe we could remove the `update` bit, or change the old client part to: `old clients will be able to read from updated versions of repository`

cyli: Good point, thanks!

docs/upgrading-client.md
```diff
@@ -0,0 +1,52 @@
...
+
+	If you need to switch between old style and new style clients, please use different trust directories (for example, `~/.notaryv1/` and `~/.notaryv2`).
+
+##### Bugfixes and improvements:
+
+- The client no longer makes an extra request to the server when listing and getting targets.
+- The client can use the local cache now if there are problems updating the new timestamp from the server.
+- Improved error messages when there is a client failure
+- Reduced the verbosity of logging during normal client operation
+- The `-v`/`--verbose` command line option now display `INFO` level logging (normal logging level is `FATAL`)
```
riyazdf: I think this is `ERROR` level

docs/upgrading-client.md
```diff
@@ -0,0 +1,52 @@
+<!--[metadata]>
++++
+title = "Upgrading notary client"
```
riyazdf: nit: It seems that other titles use capitalized `Notary`, so we might want to update this usage here too for consistency

docs/upgrading-client.md
```diff
@@ -0,0 +1,52 @@
...
+description = "How to upgrade the notary client"
```
riyazdf: also capitalized `Notary` in descriptions too

riyazdf:
LGTM after comments, thanks so much for starting this! :smile:

diogomonica:
LGTM

PR #519, opened 2016-01-29
Consistent Download

endophage:
The client will use the checksum in downloading whenever a checksum is available. This means it will never use one for timestamp and will only use one for downloading root if a snapshot was successfully downloaded first, however it will always use one for downloading the snapshot and all targets (incl. delegations) files.

Also deleted dead code related to loading and downloading targets while I was in the store code. It has never seen use in Notary.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/Sirupsen/logrus"
@@ -885,7 +886,10 @@ func (r *NotaryRepository) Update(forWrite bool) (*tufclient.Client, error) {
 	}
 	err = c.Update()
 	if err != nil {
-		if notFound, ok := err.(store.ErrMetaNotFound); ok && notFound.Resource == data.CanonicalRootRole {
+		// notFound.Resource may include a checksum so when the role is root,
+		// it will be root.json or root.<checksum>.json. Therefore best we can
+		// do it match a "root." prefix
+		if notFound, ok := err.(store.ErrMetaNotFound); ok && strings.HasPrefix(notFound.Resource, data.CanonicalRootRole+".") {
 			return nil, r.errRepositoryNotExist()
 		}
 		return nil, err
--- a/server/server.go
+++ b/server/server.go
@@ -89,13 +89,13 @@ func RootHandler(ac auth.AccessController, ctx context.Context, trust signed.Cry
 		prometheus.InstrumentHandlerWithOpts(
 			prometheusOpts("UpdateTuf"),
 			hand(handlers.AtomicUpdateHandler, "push", "pull")))
-	r.Methods("GET").Path("/v2/{imageName:.*}/_trust/tuf/{tufRole:root|targets(?:/[^/\\s]+)*|snapshot|timestamp}.json").Handler(
-		prometheus.InstrumentHandlerWithOpts(
-			prometheusOpts("GetRole"),
-			hand(handlers.GetHandler, "pull")))
 	r.Methods("GET").Path("/v2/{imageName:.*}/_trust/tuf/{tufRole:root|targets(?:/[^/\\s]+)*|snapshot|timestamp}.{checksum:[a-fA-F0-9]{64}|[a-fA-F0-9]{96}|[a-fA-F0-9]{128}}.json").Handler(
 		prometheus.InstrumentHandlerWithOpts(
 			prometheusOpts("GetRoleByHash"),
+			hand(handlers.GetHandler, "pull")))
+	r.Methods("GET").Path("/v2/{imageName:.*}/_trust/tuf/{tufRole:root|targets(?:/[^/\\s]+)*|snapshot|timestamp}.json").Handler(
+		prometheus.InstrumentHandlerWithOpts(
+			prometheusOpts("GetRole"),
 			hand(handlers.GetHandler, "pull")))
 	r.Methods("GET").Path(
 		"/v2/{imageName:.*}/_trust/tuf/{tufRole:snapshot|timestamp}.key").Handler(
--- a/tuf/client/client.go
+++ b/tuf/client/client.go
@@ -177,7 +177,7 @@ func (c *Client) downloadRoot() error {
 	var raw []byte
 	if download {
 		// use consistent download if we have the checksum.
-		raw, s, err = c.downloadSigned(role, size, expectedSha256, len(expectedSha256) > 0)
+		raw, s, err = c.downloadSigned(role, size, expectedSha256)
 		if err != nil {
 			return err
 		}
@@ -265,7 +265,7 @@ func (c *Client) downloadTimestamp() error {
 	}
 	// unlike root, targets and snapshot, always try and download timestamps
 	// from remote, only using the cache one if we couldn't reach remote.
-	raw, s, err := c.downloadSigned(role, notary.MaxTimestampSize, nil, false)
+	raw, s, err := c.downloadSigned(role, notary.MaxTimestampSize, nil)
 	if err == nil {
 		ts, err = c.verifyTimestamp(s, version, c.keysDB)
 		if err == nil {
@@ -342,7 +342,7 @@ func (c *Client) downloadSnapshot() error {
 	}
 	var s *data.Signed
 	if download {
-		raw, s, err = c.downloadSigned(role, size, expectedSha256, true)
+		raw, s, err = c.downloadSigned(role, size, expectedSha256)
 		if err != nil {
 			return err
 		}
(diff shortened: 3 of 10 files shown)


server/server.go
```diff
@@ -89,14 +89,14 @@ func RootHandler(ac auth.AccessController, ctx context.Context, trust signed.Cry
...
-			prometheusOpts("GetRole"),
-			hand(handlers.GetHandler, "pull")))
 	r.Methods("GET").Path("/v2/{imageName:.*}/_trust/tuf/{tufRole:root|targets(?:/[^/\\s]+)*|snapshot|timestamp}.{checksum:[a-fA-F0-9]{64}|[a-fA-F0-9]{96}|[a-fA-F0-9]{128}}.json").Handler(
 		prometheus.InstrumentHandlerWithOpts(
 			prometheusOpts("GetRoleByHash"),
 			hand(handlers.GetHandler, "pull")))
+	r.Methods("GET").Path("/v2/{imageName:.*}/_trust/tuf/{tufRole:root|targets(?:/[^/\\s]+)*|snapshot|timestamp}.json").Handler(
+		prometheus.InstrumentHandlerWithOpts(
+			prometheusOpts("GetRole"),
+			hand(handlers.GetHandler, "pull")))
```
riyazdf: I diff'd the added handler against the removed one and they're exactly the same (unless I'm missing something?).  Any reason for moving the handler?

client/client.go
```diff
@@ -854,7 +855,7 @@ func (r *NotaryRepository) Update(forWrite bool) (*tufclient.Client, error) {
 	}
 	err = c.Update()
 	if err != nil {
-		if notFound, ok := err.(store.ErrMetaNotFound); ok && notFound.Resource == data.CanonicalRootRole {
+		if notFound, ok := err.(store.ErrMetaNotFound); ok && strings.HasPrefix(notFound.Resource, data.CanonicalRootRole+".") {
```
riyazdf: It might be worth leaving a comment about the new resource URL format from `URLFilePath` and why we can only check prefixes

tuf/utils/utils.go
```diff
@@ -146,3 +147,15 @@ func FindRoleIndex(rs []*data.Role, name string) int {
 	}
 	return -1
 }
+
+// URLFilePath generates the appropriate HTTP URL path for the targets file,
```
riyazdf: I think this is used for more than just the targets file -- it's any metadata we have a checksum for right?  We should update the comment if so

tuf/utils/utils.go
```diff
@@ -146,3 +147,15 @@ func FindRoleIndex(rs []*data.Role, name string) int {
...
+// based on whether the repo is marked as consistent. The RemoteStore
+// is responsible for adding file extensions.
+func URLFilePath(role string, hashSha256 []byte, consistent bool) string {
+	if consistent && len(hashSha256) > 0 {
+		// Use path instead of filepath since we refer to the TUF role directly instead of its target files
```
riyazdf: I think we can remove this comment (no path or filepath magic here)

server/server.go
endophage: Priority of matching. The checksum gets caught in the extended targets delegations match of this Path comes so you end up with `role = "targets/foo.<checksum>"` and `checksum = ""`.

endophage: p.s. this is a common problem with every URL router I've seen in ever language. Order of declaration matters.

riyazdf: Ah, that makes sense now :)  Thanks for clarifying, this is definitely what we want :+1:

tuf/utils/utils.go
endophage: Yep, sorry, updated filename because it specifically referenced targets but missed the comment :-/

tuf/store/memorystore.go
```diff
@@ -56,16 +80,16 @@ func (m *memoryStore) SetMultiMeta(metas map[string][]byte) error {
 
 // RemoveMeta removes the metadata for a single role - if the metadata doesn't
 // exist, no error is returned
-func (m *memoryStore) RemoveMeta(name string) error {
+func (m *MemoryStore) RemoveMeta(name string) error {
 	delete(m.meta, name)
```
riyazdf: Should we also try to remove the `m.consistent` entry?

tuf/store/memorystore.go
```diff
@@ -95,16 +119,16 @@ func (m *memoryStore) WalkStagedTargets(paths []string, targetsFn targetsWalkFun
...
+func (m *MemoryStore) GetKey(role string) ([]byte, error) {
+	return nil, fmt.Errorf("GetKey is not implemented for the MemoryStore")
 }
 
 // Clear this existing memory store by setting this store as new empty one
-func (m *memoryStore) RemoveAll() error {
+func (m *MemoryStore) RemoveAll() error {
 	m.meta = make(map[string][]byte)
 	m.files = make(map[string][]byte)
 	m.keys = make(map[string][]data.PrivateKey)
```
riyazdf: Same here, I think we should clear `m.consistent`

riyazdf: Also, it could be nice to have a small unit test for the `MemoryStore` focusing on `GetMeta`/`SetMeta` when we include hashes and potentially the `RemoveAll` and `RemoveMeta` functionality if those change.

tuf/testutils/corrupt_memorystore.go
```diff
@@ -0,0 +1,24 @@
+package testutils
+
+import (
+	"github.com/docker/notary/tuf/store"
+)
+
+// CorruptingMemoryStore corrupts all data returned by GetMeta
+type CorruptingMemoryStore struct {
```
riyazdf: This is awesome! :+1:

riyazdf:
Thanks for cleaning up the dead memory store code too, LGTM!

diogomonica:
@endophage rebase needed

tuf/client/client.go
```diff
@@ -421,8 +419,9 @@ func (c *Client) downloadTargets(role string) error {
 	return nil
 }
 
-func (c *Client) downloadSigned(role string, size int64, expectedSha256 []byte) ([]byte, *data.Signed, error) {
-	raw, err := c.remote.GetMeta(role, size)
+func (c *Client) downloadSigned(role string, size int64, expectedSha256 []byte, consistent bool) ([]byte, *data.Signed, error) {
```
diogomonica: I would rather not change the API and just infer whether we have a sha256 or not (if you don't want consistent download, pass `nil`)

endophage: I was going to disagree (unexported method, explicit behaviour, blah, blah, blah) with you but in the context of notary, if we have a sha256, we should _always_ do a consistent download so I'll happily make this change :-)

diogomonica: Did you end up making the change? I don't see any other commits.

tuf/client/client.go
```diff
@@ -422,7 +420,8 @@ func (c *Client) downloadTargets(role string) error {
 }
 
 func (c *Client) downloadSigned(role string, size int64, expectedSha256 []byte) ([]byte, *data.Signed, error) {
-	raw, err := c.remote.GetMeta(role, size)
+	rolePath := utils.URLFilePath(role, expectedSha256)
```
diogomonica: Is `URLFilePath` the name we want for this method? We're essentially assigning a "URL" to a path, and then sticking it into a getMeta.

Feels to me that the remote store should abstract that this is a remote call, so we shouldn't call this `URLFilePath`

endophage: It's not a perfect name and it doesn't even construct a full path, it literally just puts the checksum and role name together. The RemoteStore deals with adding any appropriate file extensions and path prefix (i.e. /v2/<gun>/_trust/)

diogomonica: Yeah, I read what it does, I just think that something like `CanonicalPath` is less confusing than it including URL in the name.

tuf/testutils/corrupt_memorystore.go
```diff
@@ -0,0 +1,71 @@
...
+	d = append(d, ' ')
+	return d, err
+}
+
+// ShortMemoryStore corrupts all data returned by GetMeta
+type ShortMemoryStore struct {
+	store.MemoryStore
+}
+
+// NewShortMemoryStore returns a new instance of memory store that
```
diogomonica: one byte too little data

tuf/testutils/corrupt_memorystore.go
diogomonica: Yeah, this enables some pretty cool tests!

diogomonica:
LGTM after small fixes.

tuf/client/client.go
endophage: how about `ConsistentName`?

PR #526, opened 2016-01-30
remove signatures that are no longer valid after signing TUF file

### LINKED ISSUE #466: Cleanup signatures that are no longer valid
The way I intended this to work was that Sign is called with all the KeyIDs that are valid for the role being signed, we sign with all the keys we actually have, then delete any prior signatures for which the KeyID does not appear in the list passed to Sign.

endophage:

Closes #466

### DIFF
--- a/certs/certs_test.go
+++ b/certs/certs_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/docker/notary/cryptoservice"
 	"github.com/docker/notary/trustmanager"
 	"github.com/docker/notary/tuf/data"
+	"github.com/docker/notary/tuf/keys"
 	"github.com/docker/notary/tuf/signed"
 	"github.com/stretchr/testify/assert"
 )
@@ -293,10 +294,14 @@ func testValidateSuccessfulRootRotation(t *testing.T, keyAlg, rootKeyType string
 	signedTestRoot, err := testRoot.ToSigned()
 	assert.NoError(t, err)
 
-	err = signed.Sign(cs, signedTestRoot, replRootKey)
+	kdb := keys.NewDB()
+	kdb.AddKey(origRootKey)
+	kdb.AddKey(replRootKey)
+
+	err = signed.Sign(cs, kdb, signedTestRoot, replRootKey)
 	assert.NoError(t, err)
 
-	err = signed.Sign(cs, signedTestRoot, origRootKey)
+	err = signed.Sign(cs, kdb, signedTestRoot, origRootKey)
 	assert.NoError(t, err)
 
 	// This call to ValidateRoot will succeed since we are using a valid PEM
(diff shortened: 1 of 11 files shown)


diogomonica:
It seems that this code is still not validating if the signature belongs to the right role. We should also add a test for that.

mtrmac:
Right now, this is rather pointless: _every single_ non-test user is breaking old signatures (typically by incrementing `Version`) just before calling `signed.Sign` (or `tr.sign`).

(More in #267.)

endophage:
@diogo correct, that does not happen during signing, but clients validate it while updating. I initially started going down that route but it required much more significant changes for only a minimal (non-security related) benefit.

@mtrmac I understand what you're saying and I think a difference in development paradigm is at play here. We should be able to move small pieces of code forward where possible. We may not have thresholds yet, that doesn't mean we can't move  parts of the code towards that eventual goal.

mtrmac:
I like moving towards the goal in smallish pieces, it's just I’m afraid I don’t see that this is moving us toward a good threshold support.

It is very counterintuitive for a “`Sign`” function to be validating, and selectively removing, signatures. And this not just an aesthetics argument:

- It is losing information: Due to the logistics of creating and transferring partial multi-signatures, the top-level caller _necessarily knows_ for certain whether this call is a supposed to be a complete single-instance signature, an initial multi-signature step (which should blow away old signatures), a next multi-signature step (appending a signature; all existing signatures should verify), or appending a final multi-signature (after which the threshold must be met). The automagic _do whatever seems to be the next step in signing_ `signed.Sign` function would benefit from knowing what its top-level caller _knows_ is the next step instead of having to guess.
- Because it is losing information, it is redundant: In a non-initial multi-signature step, the old signatures should be validated by the client downloading the partially-signed content before proceeding to append a new signature. This is inherent in the process, and `signed.Sign` would be validating again only because the context about the operation is being lost.
- Because it is losing information, it is unnecessarily silently failing: When appending a signature, if a previous signature were corrupted for whatever reason, we want to _fail_, not to silently remove an invalid signature and pretend that everything is fine and multi-signature signing can continue.

The above are all arguments targeting the potential multi-signing workflow, not just what suites the current code.

It seems to me it would be much cleaner to have a signing mechanism explicit about the threshold signing, say, `SignRoleCompletely`, `MultiThresholdFirstSignature`, `MultiThresholdAppendNextSignature`, `MultiThresholdAppendLastSignature`. And in such a design, none of these *Sign* functions would need to be validating any signatures nor selectively removing some of them.

endophage:
I agree with your premise about the caller having specific knowledge, but until somebody shows there is a meaningful performance penalty, this one function, in the format in this PR, satisfies the requirements of the 4 functions you describe. Where possible I greatly prefer a simple API, that by definition will have as few functions as possible, that fully implements the requirements.

diogomonica:
I think we should also bring into this discussion the fact that threshold signing doesn't really work in a dynamic environment. The necessity of having a "staging area" for partial signatures, and the fact that the repository can easily change between each signature, makes using thresholds for targets and snapshots unfeasible.

We're going down the path of using something like role pipelining, where the client can check wether a hash is present in multiple different roles (CI, staging, etc), without having the issues of the staging area and of the diverging collection.

mtrmac:
> Where possible I greatly prefer a simple API, that by definition will have as few functions as possible, that fully implements the requirements.

As simple as possible, *but no simpler*; the single-function API is so simple that it doesn’t allow users of the API to detect failures.

endophage:
This is so out of date I'm going to close and open a redone version.

PR #528, opened 2016-02-01
Break down client API for delegations

### LINKED ISSUE #525: Break down client API for delegations
The only exposed `client` libraries for manipulating delegations are `AddDelegation` and `RemoveDelegation`.  They are both offline and create `TufDelegation` structs to be consumed on a publish.  It is difficult to express more complicated functionality like clearing all paths from a delegation or removing all delegation data in these structs without having to introduce more flags to the library methods (ex: `removeAll` flag in `RemoveDelegation`)

I propose we break down these methods into simpler APIs for manipulating one piece of data at a time: `AddDelegationRole`, `AddDelegationKeys`, `ClearDelegationPaths`, `RemoveDelegationRole`, etc.  Creating and removing delegations and their associated data can be a composition of API calls.

riyazdf:
Changes the client API for delegations to compose smaller methods that each create their own TUF delegation changefiles, with only a subset of the keys set.  Also introduces `ClearAllPaths` in the TUF delegation object to remove all paths from a delegation (can be used for #514).

Includes tests that old-style changelists from the original `AddDelegation()` and `RemoveDelegation()` APIs are still valid.

Closes #525


client/client.go
```diff
@@ -302,12 +302,28 @@ func addChange(cl *changelist.FileChangelist, c changelist.Change, roles ...stri
...
 
-// AddDelegation creates a new changelist entry to add a delegation to the repository
-// when the changelist gets applied at publish time.  This does not do any validation
-// other than checking the name of the delegation to add - all that will happen
-// at publish time.
-func (r *NotaryRepository) AddDelegation(name string, threshold int,
-	delegationKeys []data.PublicKey, paths []string) error {
+// AddDelegation creates changelist entries to add provided delegation public keys and paths.
+// This method composes AddDelegationRoleAndKeys and AddDelegationPaths (each creates one changelist if called).
+func (r *NotaryRepository) AddDelegation(name string, delegationKeys []data.PublicKey, paths []string) error {
```
diogomonica: This is an API change. Are we intentionally starting the process of having all thresholds be one?

riyazdf: I'm fine adding back in the `threshold` argument both here and `AddDelegationRoleAndKeys`.  I originally didn't include `threshold` in these APIs because we never set them to any other value but 1 in our code, nor do we currently have all of the thresholding logic completely implemented.

We can future-proof this method by keeping the threshold argument, or we can leave as is since the change files that these APIs produce are still processed in the same fashion.  Either way is fine with me, @endophage @cyli WDYT?

cyli:
Non-blocking question - were we also going to update the CLI to have more fine-grained operations, or just the client library?

client/client.go
cyli: Normally I'd say it'd be good to future proof this, since thresholds are in the spec, but we don't have support for changing any other thresholds, nor do we know (nor does the spec seem to suggest) a good way for managing multiple signers.  So I don't feel particularly strongly.

If and when we support thresholds, we could add a "ChangeThreshold/UpdateKeys" function that takes any role maybe?  To that end, if we did want to futureproof, we can consider allowing adding keys for any role.  (I'm not saying we should do this now)

riyazdf:
@cyli I wasn't planning on changing any CLI behavior in this PR -- the only operation the CLI can't use yet is clearing paths, which I can handle in #514.

As is, the CLI can add/remove select keys and paths from a delegation, and don't need to specify both keys and paths in a single CLI command (we can add just keys or just paths, etc.).  The API will ensure that we only make one change file for key operations, and a separate change file for path operations even if we want to add/remove both.  Did you have anything else in mind?

cyli:
@riyazdf Ah, no, I think just misunderstood the discussion WRT to eliminating booleans.  I thought that we also wanted to eliminate single-use flags in the add-delegation and remove-delegation CLI commands.

client/client.go
diogomonica: I'm always for simplifying security critical code, and the majority of notary is pretty security critical.

If we don't see a legitimate use for thresholds, I say actually completely remove them. It's not that much extra code, but we should not ever justify carrying tech debt over some cloudy future promise.

I think we should make this as a unanimous team decision. If everyone agrees, we take it out.

/cc @endophage @NathanMcCauley

diogomonica:
@cyli I think those are ok. I just think API-level `(..., var1, var2 bool)` are terrible.

client/client.go
endophage: The main use I still see for thresholds is to require multiple signers on a root.json. Other than that, I think it's OK for us to enforce a threshold of 1 on delegations. As long as it's still handled correctly when applying the changelist, this side of the flow can be updated if we ever decide we do indeed want configurable thresholds.

endophage:
client/client.go is getting really big. Can we put all the delegations code into client/delegations.go (literally just a copy paste)? 

Otherwise LGTM.

diogomonica:
Agreed with `client/delegations.go`

riyazdf:
@endophage great idea!  Just updated with a new `client/delegations.go` file that has the delegations code.

PR #529, opened 2016-02-01
WIP: Add a no-op server endpoint for rotating keys, and make the client use it

cyli:
The PR for the full server rotation of keys was getting a bit big, so here are some cherry-picked parts.

Currently the endpoint always returns a 429 (rate limit).

Part of #347


server/handlers/default.go
```diff
@@ -143,70 +143,99 @@ func DeleteHandler(ctx context.Context, w http.ResponseWriter, r *http.Request)
...
 	}
-	keyAlgorithm := keyAlgo
+	var key data.PublicKey
 
-	var (
-		key data.PublicKey
-		err error
-	)
-	switch role {
+	switch s.role { // parseKeyParams ensures it's only timestamp or snapshot
```
cyli: Note - this logic goes away and `GetOrCreateXKey` gets unified in the next PR.

client/backwards_compatibility_test.go
```diff
@@ -77,7 +77,7 @@ func Test0Dot1RepoFormat(t *testing.T) {
 	assert.NoError(t, err, "error creating repo: %s", err)
 
 	// rotate the timestamp key, since the server doesn't have that one
-	timestampPubKey, err := getRemoteKey(ts.URL, gun, data.CanonicalTimestampRole, http.DefaultTransport)
+	timestampPubKey, err := getRemoteKey(ts.URL, gun, data.CanonicalTimestampRole, http.DefaultTransport, false)
```
NathanMcCauley: Can we have a const or enum for the bool value so its self documenting?

client/client_test.go
```diff
@@ -2530,6 +2531,49 @@ func testRotateKeySuccess(t *testing.T, serverManagesSnapshotInit bool,
...
+	role := data.CanonicalSnapshotRole
+
+	// original key is the key on the repo
+	origKey, err := getRemoteKey(ts.URL, repo.gun, role, repo.roundTrip, false)
+	assert.NoError(t, err)
+	assert.Len(t, repo.tufRepo.Root.Signed.Roles[role].KeyIDs, 1)
+	assert.Equal(t, origKey.ID(), repo.tufRepo.Root.Signed.Roles[role].KeyIDs[0])
+
+	// rotate keys, and assert that the first rotation doesn't fail but doesn't rotate,
+	// because the original key has not actually been published yet.  But the
```
riyazdf: incomplete comment at the end?

server/errors/errors.go
```diff
@@ -86,5 +86,11 @@ var (
 		Description:    "No key algorihtm has been configured for the server and it has been asked to perform an operation that requires generation.",
```
riyazdf: While you're in here, could you fix this typo on `algorithm`? :)

server/server_test.go
```diff
@@ -90,6 +91,34 @@ func TestGetKeysEndpoint(t *testing.T) {
...
+		context.Background(), "metaStore", storage.NewMemStorage())
+	ctx = context.WithValue(ctx, "keyAlgorithm", data.ED25519Key)
+
+	handler := RootHandler(nil, ctx, signed.NewEd25519())
+	ts := httptest.NewServer(handler)
+	defer ts.Close()
+
+	rolesToStatus := map[string]int{
+		data.CanonicalTimestampRole: 429, // not implemented yet
+		data.CanonicalSnapshotRole:  429, // not implemented yet
```
riyazdf: Maybe it could be worth making this our own constant in `const.go` until Go1.6?

riyazdf: (Even though it may not be used here after the snapshot rotation is implemented, but it could be nice)

tuf/store/httpstore.go
```diff
@@ -133,6 +134,8 @@ func translateStatusToError(resp *http.Response, resource string) error {
 		return ErrMetaNotFound{Resource: resource}
 	case http.StatusBadRequest:
 		return tryUnmarshalError(resp, ErrInvalidOperation{})
+	case 429: // HTTP over limit error - will be added as a constnat in Go 1.6
```
riyazdf: typo `constant`

tuf/store/httpstore_test.go
```diff
@@ -303,27 +303,116 @@ func TestHTTPStoreRemoveAll(t *testing.T) {
...
-			assert.Contains(t, err.Error(), "not authorized")
+			require.Contains(t, err.Error(), "not authorized")
 		} else {
-			assert.Contains(t, err.Error(), "unable to reach trust server")
+			require.Contains(t, err.Error(), "unable to reach trust server")
 		}
 	}
 }
+
+// If successful, GetKey and RotateKey both return public keys
```
riyazdf: This comment was slightly confusing when I first read it, maybe could we reword it to:
`GetKey and RotateKey both succeed if valid public keys are returned from the server`?

server/handlers/default.go
```diff
@@ -143,70 +143,99 @@ func DeleteHandler(ctx context.Context, w http.ResponseWriter, r *http.Request)
...
+		return nil, errors.ErrNoCryptoService.WithDetail(nil)
 	}
+
 	algo := ctx.Value("keyAlgorithm")
 	keyAlgo, ok := algo.(string)
-	if !ok || keyAlgo == "" {
-		logger.Error("500 GET key algorithm not configured")
-		return errors.ErrNoKeyAlgorithm.WithDetail(nil)
+	if !ok || keyAlgo != data.ECDSAKey && keyAlgo != data.RSAKey && keyAlgo != data.ED25519Key {
+		return nil, errors.ErrNoKeyAlgorithm.WithDetail(nil)
```
riyazdf: thanks for improving these checks :+1:

client/helpers.go
```diff
@@ -238,22 +238,15 @@ func nearExpiry(r *data.SignedRoot) bool {
 }
 
 // Fetches a public key from a remote store, given a gun and role
-func getRemoteKey(url, gun, role string, rt http.RoundTripper) (data.PublicKey, error) {
+func getRemoteKey(url, gun, role string, rt http.RoundTripper, rotate bool) (data.PublicKey, error) {
```
riyazdf: It seems that it might make it clearer to rename and update the description for this method, since it can also rotate the remote key if specified?

riyazdf:
awesome tests, LGTM pending comments!

cyli:
@riyazdf @NathanMcCauley thanks for the quick review!  I've just made a new function `rotateRemoteKey` instead of having the boolean.  Ideally if we could cache the remote store on the client at some point, we can just remove those two functions.  :)  But for now this seems easier to read - thanks!  :D

cyli:
Marking as WIP due to some discussion on #534 that may change how this works

cyli:
Since so much has changed, opening new request #533 instead.

PR #532, opened 2016-02-02
Stop logging out critical info of database

HuKeping:
The signer will print out the user name and password of the database
which could cause security problem.

The server side is OK.


cyli:
Good catch @HuKeping!  Thanks so much!

PR #534, opened 2016-02-02
WIP: Server rotate key (for reference)

cyli:
Non-database changes for allowing a the rotation of a server-managed key.

Some of this is already in #529.

In order to keep the changes small, I plan on cherry-picking the server/storage changes out and working on the database changes in another PR that should be merged before this PR.


server/storage/interface.go
```diff
@@ -1,14 +1,31 @@
...
-	SetKey(gun, role, algorithm string, public []byte) error
+// KeyStore provides a minimal interface for managing key persistence.  Any
+// or all of these functions can optionally clean up expired pending keys, or
+// no-longer-active non-pending keys.
+type KeyStore interface {
+	// GetLatestKey the most recently created, non-expired key for the given
+	// GUN and role. If no keys exist for the GUN+role, returns an ErrNoKeys.
+	GetLatestKey(gun, role string) (*ManagedPublicKey, error)
+
+	// HasAnyKeys returns true if any non-expired keys exist for the given GUN,
```
cyli: Only necessary because we validate that the root.json has timestamp keys that the server has. If we want to eliminate this requirement and allow users to manage their own timestamp keys, we can probably eliminate this interface.

It is also used as a shortcut to see if we can generate a snapshot, but we can probably just try to generate one and then fail if we don't have the keys.

server/storage/interface.go
```diff
@@ -1,14 +1,31 @@
...
-	// SetKey sets the algorithm and public key for the given GUN and role if
-	// it doesn't already exist.  Otherwise an error is returned.
-	SetKey(gun, role, algorithm string, public []byte) error
+// KeyStore provides a minimal interface for managing key persistence.  Any
+// or all of these functions can optionally clean up expired pending keys, or
+// no-longer-active non-pending keys.
+type KeyStore interface {
+	// GetLatestKey the most recently created, non-expired key for the given
+	// GUN and role. If no keys exist for the GUN+role, returns an ErrNoKeys.
+	GetLatestKey(gun, role string) (*ManagedPublicKey, error)
```
endophage: This won't necessarily be the correct signing key right? It'll literally be the most recently created one?

server/storage/interface.go
```diff
@@ -1,14 +1,31 @@
...
+	HasAnyKeys(gun, role string, keyIDs []string) (bool, error)
+
+	// AddKey adds the given public key for the given GUN and role, with an
+	// expiration time.  The key is added as a pending key - MarkActiveKeys
+	// must be called to make it active.
+	AddKey(gun, role string, key data.PublicKey, expires time.Time) error
+
+	// MarkActiveKeys marks the following key IDs as active.
+	// This does not fail if any of the key IDs doesn't exist.
+	MarkActiveKeys(gun, role string, keyIDs []string) error
```
endophage: Does it also mark keys for that gun/role that do not appear in the keyIDs list as inactive? Should there be a way to do that?

server/storage/interface.go
```diff
@@ -1,14 +1,31 @@
...
+	GetLatestKey(gun, role string) (*ManagedPublicKey, error)
+
+	// HasAnyKeys returns true if any non-expired keys exist for the given GUN,
+	// role, and key IDs.
+	HasAnyKeys(gun, role string, keyIDs []string) (bool, error)
+
+	// AddKey adds the given public key for the given GUN and role, with an
+	// expiration time.  The key is added as a pending key - MarkActiveKeys
+	// must be called to make it active.
+	AddKey(gun, role string, key data.PublicKey, expires time.Time) error
```
endophage: What are we intending to set `expires` to? Is it for timing out pending keys? Once a key is signed into a root.json does `expires` basically become irrelevant?

server/storage/interface.go
cyli: Yep

server/storage/interface.go
cyli: I was thinking about having it delete/clean out any previously active keys, and also any pending keys that were created before the most recently created active key.  I think I can do this without a select-for-update DB query (update all these keys, delete any pending=false keys that are not these keys).

I'm not sure if deleting any pending=true keys that were created after the max of the pending=false created dates - that might be too complex of a query, but we can probably just wait for the pending=true keys to expire and delete those when we add more keys.

server/storage/interface.go
cyli: Yep, it's just for timing out pending keys, and will be be set to nil (or otherwise ignored) when it gets marked active.

endophage:
I need to think about this more. I really like the naive implementation of just returning a new key whenever the user asks for it and letting them sign it in as and when they feel like it. That would completely remove the need for the (unfortunately named) `timestamp_keys` table. On the other hand, I agree that we don't want a bunch of unused keys hanging around in the signer, especially if we want to support HSMs.

cyli:
@endophage Possibly we can move the expiry and mark active functionality to the signer, if you want to eliminate the table.  So any key created on the signer will be marked for automatic deletion, and whenever we sign in new keys we compare the old keys on the root to the new keys on the root, and tell the signer to delete any old unused keys.  This doesn't prevent someone from just spamming the rotate endpoint though.  We can put an in-memory rate limiter in all the servers, but then we run into the inconsistency if someone load balances a bunch of servers together.

In a sense, the "timestamp_keys" table (yes, unfortunately named) is just a shared cache of rate-limiting info.

endophage:
Yeah, I'm wracking my brain trying to come up with a better way to handle the whole pending aspect. I'm almost on the side of rotating server keys requires an immediate publish. It already has to be an online operation, maybe we should just force the new key to be published immediately.

cyli:
@endophage As in we immediately delete the key that's there and create a new one every time rotate is called?  So nothing else can be signed with the old key?

endophage:
I was thinking more we can do the comparison between old and new root file and delete based on that. It still leaves us open to abuse from people that write their own client, but won't allow somebody to abuse us (at least, not in the sense they can create lots of pending keys that get left sitting around) by just calling `key rotate` over and over.

cyli:
@endophage Can't they just curl the endpoint, over and over?

endophage:
@cyli yeah, if they get a token and pass it along in the curl :-/

cyli:
Also, if we eliminate the table, won't we break the existing API for GET-ing a key (because we can't get the key for the role anymore?)  Or do you see the GET endpoint as the rotation endpoint?

cyli:
@endophage Also, what do you think about a sort of combination of ideas?  Comparing the old and new root file and deleting keys in the signer based on that.  But also keeping a table (we can name it something else) that just has the key IDs of the pending keys, and we clear them out after a day or so?  We don't even have to store the gun or role - just the IDs and expiry.  Just sort of a cleanup table.

cyli:
Closing this for now, since there is a pretty big change on how this works

Issue #483, opened 2016-01-20
Allow import/export/passwd of non-root keys via notary CLI


riyazdf:
It seems like it'd be useful to import/export keys for `targets` and potentially `snapshot` individually -- this will likely require changes to how we use our passphrase retriever as a pre-requisite since it caches the same passphrase for the two together (our concept of `tagging` key for usability).

Might also be helpful for delegation roles.

riyazdf:
Relates to https://github.com/docker/notary/pull/485

We'll need to separate `targets` and `snapshot` keys in the passphrase retriever to get this to work properly, and change the `init` flow to make this obvious (ex: ask for a separate snapshot key)

endophage:
Given there's no attempt limit on decrypting a file, I'm wondering if we can do something naive like we do for the yubikey (we try the default pin). If we've asked for a non-root password already (say for targets), we try that one for snapshot and ask for a different password on failure?

riyazdf:
From some discussion IRL: it seems like it'd be nice to separate the passphrase caching from the retriever logic, and keep the cached passphrases in a map structure so we could also cache delegation passphrases.

The steps for retrieving a passphrase would then be: first check if I have this passphrase in the cached map and try if so, then ask for passphrase input on failure.

PR #536, opened 2016-02-03
Non-root key import and export CLI

riyazdf:
Allows for importing and exporting non-root keys via notary CLI.  This also works for importing delegation keys:

`notary key import <file> --role <ROLE> --gun <GUN>`

The default role is root, so that the behavior of `notary key import <file>` remains unchanged.

export is the same but now also can export non-root keys by ID:

`notary key export <keyID> <file>`

I also made "private", "tuf_keys", and "root_keys" constants in `const.go`

Partially addresses #483 (`passwd` can be a separate PR, along with changes to passphrase retriever)

### DIFF
--- a/cmd/notary/integration_test.go
+++ b/cmd/notary/integration_test.go
@@ -674,6 +674,47 @@ func TestClientDelegationsPublishing(t *testing.T) {
 	output, err = runCommand(t, tempDir, "-s", server.URL, "list", "gun", "--roles", "targets/releases")
 	assert.NoError(t, err)
 	assert.Contains(t, output, "targets/releases")
+
+	// remove the target for this role only
+	_, err = runCommand(t, tempDir, "remove", "gun", target, "--roles", "targets/releases")
+	assert.NoError(t, err)
+
+	// publish repo
+	_, err = runCommand(t, tempDir, "-s", server.URL, "publish", "gun")
+	assert.NoError(t, err)
+
+	// Now try using the key import command to instantiate the private key with this role
+	assert.NoError(t, os.Remove(filepath.Join(keyDir, canonicalKeyID+".key")))
+	tempPrivFile, err := ioutil.TempFile("/tmp", "privfile")
+	assert.NoError(t, err)
+	defer os.Remove(tempPrivFile.Name())
+
+	// Write the private key to a file so we can import it
+	_, err = tempPrivFile.Write(privKeyBytesNoRole)
+	assert.NoError(t, err)
+	tempPrivFile.Close()
+
+	// Import the private key, associating it with our delegation role
+	_, err = runCommand(t, tempDir, "key", "import", tempPrivFile.Name(), "--role", "targets/releases")
+	assert.NoError(t, err)
+
+	// add a target using the delegation -- will only add to targets/releases
+	_, err = runCommand(t, tempDir, "add", "gun", target, tempTargetFile.Name(), "--roles", "targets/releases")
+	assert.NoError(t, err)
+
+	// list targets for targets/releases - we should see no targets until we publish
+	output, err = runCommand(t, tempDir, "-s", server.URL, "list", "gun", "--roles", "targets/releases")
+	assert.NoError(t, err)
+	assert.Contains(t, output, "No targets")
+
+	// publish repo
+	_, err = runCommand(t, tempDir, "-s", server.URL, "publish", "gun")
+	assert.NoError(t, err)
+
+	// list targets for targets/releases - we should see our target!
+	output, err = runCommand(t, tempDir, "-s", server.URL, "list", "gun", "--roles", "targets/releases")
+	assert.NoError(t, err)
+	assert.Contains(t, output, "targets/releases")
 }
 
 // Splits a string into lines, and returns any lines that are not empty (
--- a/cmd/notary/integration_test.go
+++ b/cmd/notary/integration_test.go
@@ -683,7 +683,7 @@ func TestClientDelegationsPublishing(t *testing.T) {
 	_, err = runCommand(t, tempDir, "-s", server.URL, "publish", "gun")
 	assert.NoError(t, err)
 
-	// Now try using the key import command to instantiate the private key with this role
+	// Now remove this key, and make a new file to import the delegation's key from
 	assert.NoError(t, os.Remove(filepath.Join(keyDir, canonicalKeyID+".key")))
 	tempPrivFile, err := ioutil.TempFile("/tmp", "privfile")
 	assert.NoError(t, err)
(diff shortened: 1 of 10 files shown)


diogomonica:
Behavior that I think we should support:

## If a PEM key gets imported and has a role inside of it: 
- if it's root, import it to `root_keys`
- if it's delegation, import to `tuf_keys`
- if it's targets or snapshots and `--gun` is set, import it to the right directory
- if it's targets or snapshots and `--gun` is not set, fail

## if a PEM key gets imported and has no role inside of it:
- if `--role` is not set, fail
- if `--role` is set to a delegation, import it to `tuf_keys` and rewrite the PEM to include the role
- if `--role` is set to a target or snapshot, and `--gun` is set, import it to the right directory and rewrite the PEM to include the role
- if `--role` is set to a target or snapshot, and `--gun` is not set, fail

cyli:
@diogomonica :+1: on behavior.  Maybe nice error messages on `--role` not being set like "could not infer role, and no role was specified" and "specified role does not match role in PEM headers" or some such too.  It might make it easier to just have a simpler explanation in the description.

riyazdf:
@cyli those are great suggestions for error messages, I'll work on including those along with @diogomonica's behavior ideas :+1:

riyazdf:
Removing WIP: added the aforementioned `--role` and `--gun` checks with the suggested error messages, and tests for importing and exporting

cmd/notary/integration_test.go
```diff
@@ -671,6 +671,47 @@ func TestClientDelegationsPublishing(t *testing.T) {
...
+
+	// remove the target for this role only
+	_, err = runCommand(t, tempDir, "remove", "gun", target, "--roles", "targets/releases")
+	assert.NoError(t, err)
+
+	// publish repo
+	_, err = runCommand(t, tempDir, "-s", server.URL, "publish", "gun")
+	assert.NoError(t, err)
+
+	// Now try using the key import command to instantiate the private key with this role
```
diogomonica: This comment doesn't match the code immediately following it.

riyazdf: oops, will fix :+1:

cmd/notary/keys.go
```diff
@@ -258,24 +267,43 @@ func (k *keyCommander) keysExportRoot(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	cs := cryptoservice.NewCryptoService("", ks...)
+
+	// Search for this key in all of our keystores, determine whether this key has a GUN
+	keyGun := ""
+	keyRole := ""
+	for _, store := range ks {
```
diogomonica: We might not need this loop. I believe we have a GetKey method that already searches all the keystores.

cmd/notary/keys.go
```diff
@@ -258,24 +267,43 @@ func (k *keyCommander) keysExportRoot(cmd *cobra.Command, args []string) error {
...
+		for keypath, role := range store.ListKeys() {
+			if filepath.Base(keypath) == keyID {
+				keyRole = role
+				if role != data.CanonicalRootRole {
+					dirPath := filepath.Dir(keypath)
```
diogomonica: I would remove the indentation by changing this to `if role == data.CanonicalRootRole`

cmd/notary/keys.go
```diff
@@ -258,24 +267,43 @@ func (k *keyCommander) keysExportRoot(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	cs := cryptoservice.NewCryptoService("", ks...)
+
+	// Search for this key in all of our keystores, determine whether this key has a GUN
```
diogomonica: We care about the GUN so we can create a cryptoservice? The flatkeystore will help with this, correct?

cmd/notary/keys.go
```diff
@@ -338,7 +365,56 @@ func (k *keyCommander) keysImportRoot(cmd *cobra.Command, args []string) error {
...
-	err = cs.ImportRootKey(importFile)
+	pemBytes, err := ioutil.ReadAll(importFile)
+	if err != nil {
+		return fmt.Errorf("Error reading input file: %v", err)
+	}
+
+	pemRole := trustmanager.ReadRoleFromPEM(pemBytes)
+
+	// Rewind after reading the first time
+	_, err = importFile.Seek(0, 0)
```
diogomonica: why?

cmd/notary/keys.go
```diff
@@ -338,7 +365,56 @@ func (k *keyCommander) keysImportRoot(cmd *cobra.Command, args []string) error {
...
+	if err != nil {
+		return fmt.Errorf("Error reading input file: %v", err)
+	}
+
+	if pemRole != "" && !data.ValidRole(pemRole) {
```
cyli: Non-blocking - I think I just misinterpreted the behavior somewhat that @diogomonica specified.  I just realized I had previously interpreted "delegation" to mean "either a valid delegation role, or any non-valid role".  I'm fine with imports only handling valid roles, but I was just wondering what we wanted to do for importing user keys that are not limited to specific roles?  Should we do import/export at all, or just have the user drop them in?

cyli: (Or are we going to mandate the user has a valid delegation role for them, but all delegation keys can be used anywhere?  I notice that below, only targets and snapshots keys are required to have a GUN)

cryptoservice/import_export.go
```diff
@@ -99,18 +104,45 @@ func (cs *CryptoService) ExportRootKeyReencrypt(dest io.Writer, keyID string, ne
 // It prompts for the key's passphrase to verify the data and to determine
 // the key ID.
 func (cs *CryptoService) ImportRootKey(source io.Reader) error {
+	return cs.ImportRoleKey(source, data.CanonicalRootRole, nil)
+}
+
+// ImportRoleKey imports a private key in PEM format key from an io.Reader
+// It prompts for the key's passphrase to verify the data and to determine
+// the key ID.
+func (cs *CryptoService) ImportRoleKey(source io.Reader, role string, newPassphraseRetriever passphrase.Retriever) error {
```
cyli: Should `ImportRootKey` just be changed to `ImportRoleKey` everywhere?  Since currently, `ImportRootKey` is in the interface, but `ImportRoleKey` is not.

cmd/notary/keys.go
```diff
@@ -338,7 +365,56 @@ func (k *keyCommander) keysImportRoot(cmd *cobra.Command, args []string) error {
...
+	}
+
+	// If we're importing to targets or snapshot, we need a GUN
+	if (importRole == data.CanonicalTargetsRole || importRole == data.CanonicalSnapshotRole) && k.keysImportGUN == "" {
+		return fmt.Errorf("Must specify GUN for %s key", importRole)
+	}
+
+	cs := cryptoservice.NewCryptoService(k.keysImportGUN, ks...)
+	if importRole == data.CanonicalRootRole {
+		err = cs.ImportRootKey(importFile)
```
cyli: Even if we don't replace `ImportRootKey` with `ImportRoleKey`, we can probably just call `ImportRoleKey` for root too, so maybe no need for the switch :)

cmd/notary/keys.go
riyazdf: This was to keep the filepointer at the beginning of the file for `ImportRootKey` but I think I'll follow @cyli's suggestion and refactor `ImportRoleKey`.  That should get rid of this

cmd/notary/integration_test.go
```diff
@@ -71,7 +71,7 @@ func setupServer() *httptest.Server {
 	ctx = ctxu.WithLogger(ctx, logrus.NewEntry(l))
 
 	cryptoService := cryptoservice.NewCryptoService(
-		"", trustmanager.NewKeyMemoryStore(passphrase.ConstantRetriever("pass")))
+		"", trustmanager.NewKeyMemoryStore(passphrase.ConstantRetriever(testPassphrase)))
```
cyli: :+1: Thanks for fixing these!

cmd/notary/integration_test.go
```diff
@@ -1025,6 +1066,235 @@ func TestClientKeyImportExportRootOnly(t *testing.T) {
 		t, tempDir, server.URL, "gun", target, tempFile.Name())
 }
 
+// Tests importing and exporting keys for all different roles and GUNs
+func TestClientKeyImportExportAllRoles(t *testing.T) {
```
cyli: Totally non-blocking - this is a pretty awesome and comprehensive positive test!  :D  I was wondering if it would be easier to read (and maybe save you some work) to do the same import/export work just loop over all 4 roles?  That way it'd be easier to see that you're testing importing with pem headers, and importing without pem headers, etc.

Also wondering if it'd make sense to remove the imported key after each one?  Since it imports twice (once with PEM headers and one without), that'd make sure the test wasn't wasn't somehow just reading the bytes from successful previous import.

const.go
```diff
@@ -18,4 +18,10 @@ const (
 	Sha256HexSize = 64
 	// TrustedCertsDir is the directory, under the notary repo base directory, where trusted certs are stored
 	TrustedCertsDir = "trusted_certificates"
+	// PrivDir is the directory, under the notary repo base directory, where private keys are stored
+	PrivDir = "private"
```
cyli: Yay, thanks for moving these here!

cmd/notary/integration_test.go
riyazdf: The loop sounds like a great idea, we'll have conditionals for each role type for the right subdirs but I think it'll still be better.

And yeah, also a good idea to remove the key -- I don't think the test is reading the previous import, but it'd be good to make it explicit.

cmd/notary/keys.go
riyazdf: Technically for delegation keys we don't actually check the PEM header roles when signing, so it'd be fine to have an invalid name.  I'm fine changing this, I like the flexibility of being able to use a notary command or just dropping keys in and not having to interface directly with the `tuf_keys` directory.

cmd/notary/keys.go
riyazdf: This is a great idea!  I'll look into this

cryptoservice/import_export.go
riyazdf: I'm fine with either changing to `ImportRoleKey` or just not even calling `ImportRootKey` from the `key import` command.  It might be cleaner just to change the interface, though it'll make this PR a little bigger

cmd/notary/keys.go
riyazdf: Yeah, the flatkeystore or at least getting key info by ID will help this.  I'm looking into this and hope to open another PR that can provide this functionality.

Until then we need the loop to find the gun unless we want to change the current client API for exporting keys.  We have a similar loop in `removeKeyInteractively` that could also be improved

diogomonica:
@riyazdf do you wan to merge this and then work on the refactor?

riyazdf:
@diogomonica I'm fine with merging this before the refactor

diogomonica:
Rebase needed.

riyazdf:
@diogomonica rebased!

cyli:
This LGTM, given we are refactoring to a flat keystore.

PR #541, opened 2016-02-04
Separately ask for snapshot password with passphrase retriever

### LINKED ISSUE #524: Refactor passphrase retriever to separate targets and snapshot
Relates to #483, and we should also ask for the snapshot passphrase on init.

It's worth keeping the user aware of the existence of the snapshot key, especially since it can be rotated to server side.

riyazdf:
Marking as PoC because this deviates from the behavior vendored into docker/docker.

This changes the passphrase retriever to separate the cached passwords for the `targets` and `snapshot` role, such that we cache all passwords (including for delegations) individually.  I think we should move in this direction, especially because we can enable remote snapshot key management.

However, I'm still trying to wrap my head around the consequences against pre-existing notary integrations, like in docker -- because the old passphrase retriever is being used and the environment variable `DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE` is [shared](https://github.com/docker/docker/blob/master/api/client/trust.go#L200) between targets and snapshot.  If this environment variable is set, it will be used without calling the passphrase retriever, which could be an issue if the user picks different a passphrase for targets and snapshot on init.

Closes #524.

### DIFF
--- a/passphrase/passphrase.go
+++ b/passphrase/passphrase.go
@@ -61,10 +61,12 @@ func PromptRetriever() Retriever {
 // aliasMap can be used to specify display names for TUF key aliases. If aliasMap
 // is nil, a sensible default will be used.
 func PromptRetrieverWithInOut(in io.Reader, out io.Writer, aliasMap map[string]string) Retriever {
-	userEnteredTargetsSnapshotsPass := false
-	targetsSnapshotsPass := ""
-	userEnteredRootsPass := false
-	rootsPass := ""
+	userEnteredTargetsPass := false
+	targetsPass := ""
+	userEnteredSnapshotPass := false
+	snapshotPass := ""
+	userEnteredRootPass := false
+	rootPass := ""
 
 	return func(keyName string, alias string, createNew bool, numAttempts int) (string, bool, error) {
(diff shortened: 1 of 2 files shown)


cyli:
@riyazdf We could always add another environment variable, and have that take preference over `DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE`, but use `DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE` for snapshots too if it's provided.

We do have to type in more passwords, but this change seems to make more sense now that we can change passwords for every individual key now.

endophage:
@riyazdf do we still want to add this functionality?

riyazdf:
@endophage I think it makes sense to add this PR in current form on the notary side.  As for notary's integration with docker, there's less friction now that new repos will automatically have the server manage the snapshot key so I think this is worth a second look.

As for the environment variables in docker, we still have `DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE` which can be currently used for targets, snapshot, or delegation roles

endophage:
Sounds good. Wasn't sure if there had been other changes that made this moot.

riyazdf:
closing in favor of #848

PR #542, opened 2016-02-04
add non-root passwd functionality

riyazdf:
Allows notary CLI to change passphrases for non-root keys, including delegation keys.

There might be some weird behavior with the passphrase retriever on targets/snapshot key if you only change the passphrase for one individually -- #541 would fix this.

Partially addresses #483

### DIFF
--- a/cmd/notary/keys.go
+++ b/cmd/notary/keys.go
@@ -78,8 +78,8 @@ var cmdKeyRemoveTemplate = usageTemplate{
 
 var cmdKeyPasswdTemplate = usageTemplate{
 	Use:   "passwd [ keyID ]",
-	Short: "Changes the passphrase for the root key with the given keyID.",
-	Long:  "Changes the passphrase for the root key with the given keyID.  Will require validation of the old passphrase.",
+	Short: "Changes the passphrase for the key with the given keyID.",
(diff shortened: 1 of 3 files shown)


endophage:
@riyazdf can you rebase to sort out codecov's offset weirdness

riyazdf:
@endophage rebased, codecov looks normal again :)

cmd/notary/keys.go
```diff
@@ -521,18 +521,26 @@ func (k *keyCommander) keyPassphraseChange(cmd *cobra.Command, args []string) er
 		return fmt.Errorf("invalid key ID provided: %s", keyID)
 	}
 
-	// We only allow for changing the root key, so use no gun
-	cs := cryptoservice.NewCryptoService("", ks...)
+	// Find the key's GUN by ID, in case it is a non-root key
+	var keyGUN string
+	for _, store := range ks {
```
diogomonica: Same weird API.

diogomonica:
This LGTM, but we should not forget to refactor this once we change the API. The for loop in there instead of this being a simple `.GetKey` is weird.

cyli:
LGTM :)

PR #544, opened 2016-02-04
Bump the cobra requirement, since they have some nice usage printing fixes

cyli:


cyli:
@diogomonica and @riyazdf LGTM'ed IRL

PR #552, opened 2016-02-05
Flatten keystore by adding key info map

### LINKED ISSUE #502: Refactor and flatten the KeyStore
Right now we:
- Need the `gun` of a key to be able to retrieve it.
- Cannot simply drop a key in a directory and have it be picked up by the keystore object

We should refactor the keystore in a way that still mantains the directory structure that we want (tuf keys, root keys, etc), but reads all the keys under `private/` into a flat keystore that can access any key by KeyID.

riyazdf:
This introduces a `keyInfo` map to the key store that maps key IDs to the key's role and GUN.  The underlying filestore/memorystore remains the same, but we can now call `GetKey` and `RemoveKey` without a gun for `targets` and `snapshot` keys.  

The tradeoff is that we have to initialize and load this map when creating our keystore, and keep it up to date as we add/remove keys.

Closes #502 


trustmanager/keyfilestore.go
```diff
@@ -18,12 +18,15 @@ const (
 	privDir           = "private"
 )
 
+type keyIDMap map[string]KeyInfo
```
diogomonica: Isn't this a keyInfoMap?

trustmanager/keyfilestore.go
```diff
@@ -43,10 +53,45 @@ func NewKeyFileStore(baseDir string, passphraseRetriever passphrase.Retriever) (
 		return nil, err
 	}
 	cachedKeys := make(map[string]*cachedKey)
+	keyIDMap := make(keyIDMap)
 
-	return &KeyFileStore{SimpleFileStore: *fileStore,
+	keyStore := &KeyFileStore{SimpleFileStore: *fileStore,
 		Retriever:  passphraseRetriever,
-		cachedKeys: cachedKeys}, nil
+		cachedKeys: cachedKeys,
```
diogomonica: Aren't cachedKeys and keyIDMap equivalents? Why not merge both?

diogomonica:
I'm confused by this.

I thought the goal was to create a master structure that on run would populate a structure with KeyID -> path, role, gun relationship. This structure would be what `ListKeys()` actually iterates over to return the info that it needs, and it would also be what `GetKey` ultimately uses to access the key file.

What I want to remove from everywhere is the path stuff where we do `something + key.gun`. Everyone that wants to access a key would simply use KeyMap[KeyID].path to access it, not even needing to understand if it is a root key, a delegations key or within a GUN or not.

This is what I was thinking by "Flat Keystore". The only place in the whole code where we append `gun` to a path is on `AddKey`, which already receives a gun.

riyazdf:
@diogomonica: this PR introduces that master structure with the role and gun, but not the path (though I can add that).  There is a reliance on `ListKeys()` to build this structure but I can also refactor that out so that there's a separate population step, and `ListKeys` just uses this structure.

As is, this PR will let us supply only a key ID for adding/getting/removing, but it still allows for `something + key.gun` -- which I agree we should remove.  I'll take a stab at this, and refactor further :+1:

trustmanager/keyfilestore.go
```diff
@@ -43,10 +53,45 @@ func NewKeyFileStore(baseDir string, passphraseRetriever passphrase.Retriever) (
...
+	keyStore.loadKeyIDInfo()
+
+	return keyStore, nil
+}
+
+func generateKeyInfoMap(fullIDToRole map[string]string) map[string]KeyInfo {
+	keyInfoMap := make(map[string]KeyInfo)
+	for keyIDAndGun, role := range fullIDToRole {
+		keyGun := getGunFromFullID(keyIDAndGun)
+		keyID := filepath.Base(keyIDAndGun)
```
endophage: I haven't double checked `listkeys` but do these paths no longer have the `_<role>` on the end if it's an old style filename?

riyazdf: `listkeys` will handle the old-style filename and clean up the path by the time the code gets here

riyazdf:
removing WIP.  This code lets us get rid of needing to know GUNs when retrieving keys (see the removed loops in `crypto_service.go`, `keys.go`, and `prettyprint.go`) in exchange for added code when adding, removing, and especially importing keys.  The underlying file structure of keys is unchanged in this PR.

Also, since we map from ID to `KeyInfo`, we can't reuse base role keys across multiple roles or guns.  This isn't a problem with delegation keys because we can simply leave them the `tuf_keys` subdirectory without any gun.

I also added a `AddKey` api to the cryptoservice so that it could interact with the underlying key info map when adding keys.

diogomonica:
Rebase needed.

cmd/notary/keys.go
```diff
@@ -540,7 +531,11 @@ func (k *keyCommander) keyPassphraseChange(cmd *cobra.Command, args []string) er
 	// unlocking passphrase and reusing that.
 	passChangeRetriever := k.getRetriever()
 	keyStore, err := trustmanager.NewKeyFileStore(config.GetString("trust_dir"), passChangeRetriever)
-	err = keyStore.AddKey(filepath.Join(keyGUN, keyID), role, privKey)
+	keyInfo, err := cs.GetKeyInfo(keyID)
+	if err != nil {
+		return err
+	}
+	err = keyStore.AddKey(filepath.Join(keyInfo.Gun, keyID), role, privKey)
```
diogomonica: Maybe we can change `AddKey` to receive a `KeyInfo` instead?

client/client_test.go
```diff
@@ -391,15 +391,15 @@ func assertRepoHasExpectedKeys(t *testing.T, repo *NotaryRepository,
...
+		if keyInfo.Role == data.CanonicalRootRole {
 			assert.Equal(t, rootKeyID, keyID, "Unexpected root key ID")
 		}
 		// just to ensure the content of the key files created are valid
 		_, r, err := ks.GetKey(keyID)
 		assert.NoError(t, err)
-		assert.Equal(t, role, r)
-		roles[role] = true
+		assert.Equal(t, keyInfo.Role, r)
+		roles[keyInfo.Role] = true
```
diogomonica: can keyInfo ever be nil?

cmd/notary/keys.go
```diff
@@ -464,10 +449,10 @@ func removeKeyInteractively(keyStores []trustmanager.KeyStore, keyID string,
 	var storesByIndex []trustmanager.KeyStore
 
 	for _, store := range keyStores {
-		for keypath, role := range store.ListKeys() {
+		for keypath, keyInfo := range store.ListKeys() {
 			if filepath.Base(keypath) == keyID {
 				foundKeys = append(foundKeys,
-					[]string{keypath, role, store.Name()})
+					[]string{keypath, keyInfo.Role, store.Name()})
```
diogomonica: can keyInfo ever be nil?

cmd/notary/keys.go
diogomonica: ping

cryptoservice/crypto_service.go
```diff
@@ -105,13 +99,33 @@ func (cs *CryptoService) GetKey(keyID string) data.PublicKey {
...
+}
+
+// AddKey adds a private key to a specified role.
+// The GUN is inferred from the cryptoservice itself for non-root roles
+func (cs *CryptoService) AddKey(role string, key data.PrivateKey) (err error) {
+	keyID := key.ID()
+	if role != data.CanonicalRootRole {
+		keyID = filepath.Join(cs.gun, key.ID())
+	}
+	for _, ks := range cs.keyStores {
```
diogomonica: Might be worth to check if the key already exists?

trustmanager/keyfilestore.go
```diff
@@ -38,10 +49,96 @@ func NewKeyFileStore(baseDir string, passphraseRetriever passphrase.Retriever) (
...
+		} else {
+			keyIDAndGun = strings.TrimPrefix(keyPath, notary.NonRootKeysSubdir+"/")
+		}
+
+		// Separate the ID and GUN (can be empty) from the filepath
+		keyIDAndGun = strings.TrimSpace(keyIDAndGun)
+		keyGun := getGunFromFullID(keyIDAndGun)
+		keyID := filepath.Base(keyIDAndGun)
+
+		// If the key does not have a _, we'll attempt to
```
diogomonica: As opposed to read it as what?

trustmanager/keyfilestore.go
```diff
@@ -38,10 +49,96 @@ func NewKeyFileStore(baseDir string, passphraseRetriever passphrase.Retriever) (
...
+				continue
+			}
+			block, _ := pem.Decode(d)
+			if block == nil {
+				continue
+			}
+			if role, ok := block.Headers["role"]; ok {
+				keyRole = role
+			}
+		} else {
```
diogomonica: Can we open an issue to remove this for notary 0.3?

trustmanager/keyfilestore.go
```diff
@@ -54,31 +151,58 @@ func (s *KeyFileStore) Name() string {
...
+	err := removeKey(s, s.cachedKeys, name)
+	if err != nil {
+		return err
+	}
+	// Remove this key from our keyInfo map if we removed from our filesystem
+	if _, ok := s.keyInfoMap[name]; ok {
+		delete(s.keyInfoMap, name)
+	} else {
+		// This might be of the form GUN/ID  - try to delete without the gun
+		delete(s.keyInfoMap, filepath.Base(name))
```
diogomonica: Why would we ever have the name include the gun inside of `keyInfoMap`?

trustmanager/keyfilestore.go
```diff
@@ -54,31 +151,58 @@ func (s *KeyFileStore) Name() string {
...
 // RemoveKey removes the key from the keyfilestore
 func (s *KeyFileStore) RemoveKey(name string) error {
 	s.Lock()
 	defer s.Unlock()
-	return removeKey(s, s.cachedKeys, name)
+	// If this is a bare key ID without the gun, prepend the gun so the filestore lookup succeeds
+	if keyInfo, ok := s.keyInfoMap[name]; ok {
+		name = filepath.Join(keyInfo.Gun, name)
+	}
+	err := removeKey(s, s.cachedKeys, name)
```
diogomonica: Maybe we merge `keyInfoMap` with the `keyCache`?

cmd/notary/keys.go
riyazdf: I think that's a good idea. I'm currently working on what the API change would look like for each usage -- so far it seems that importing will have the biggest internal changes, but it should be good.

trustmanager/keyfilestore.go
diogomonica: Actually that might not make a lot of sense. Not sure if there is a good way of doing it.

trustmanager/keyfilestore.go
```diff
@@ -147,7 +310,44 @@ func (s *KeyMemoryStore) ExportKey(name string) ([]byte, error) {
...
+	err := importKey(s, s.Retriever, s.cachedKeys, alias, pemBytes)
+	if err != nil {
+		return err
+	}
+	keyID, keyInfo := getImportedKeyInfo(pemBytes, alias)
+	s.keyInfoMap[keyID] = keyInfo
+	return nil
+}
+
+func getImportedKeyInfo(pemBytes []byte, filename string) (string, KeyInfo) {
```
diogomonica: This is weird. Why would we have an import function that receives both `pemBytes` and the `filename`?

tuf/signed/interface.go
```diff
@@ -15,6 +15,9 @@ type KeyService interface {
 	// future features
 	Create(role, algorithm string) (data.PublicKey, error)
 
+	// AddKey adds a private key to the specified role
+	AddKey(role string, key data.PrivateKey) error
```
diogomonica: If the method is called `AddKey`, the key should probably be the first param.

client/client_test.go
riyazdf: No, since each `keyInfo` in `ListKeys` must be tied to a keyID, and when we remove any `KeyInfo` the whole entry is deleted from `keyInfoMap`

cmd/notary/keys.go
riyazdf: Same as above

cryptoservice/crypto_service.go
riyazdf: Good idea, will do

trustmanager/keyfilestore.go
riyazdf: Sorry, this comment should be clearer -- this is for reading the role from the PEM headers vs. legacy filenames

trustmanager/keyfilestore.go
riyazdf: Great idea!  Done https://github.com/docker/notary/issues/569

trustmanager/keyfilestore.go
riyazdf: `keyInfoMap` never includes the gun -- this was a defensive check in case the user passed a name of `gun/ID` to `RemoveKey()`, we're just deleting `ID` from the `keyInfoMap`.

tuf/signed/interface.go
riyazdf: Agreed, I'll swap the params

cryptoservice/crypto_service.go
riyazdf: Done!

trustmanager/keyfilestore.go
riyazdf: Done!

trustmanager/keyfilestore.go
riyazdf: this is an un-exported helper for getting the `KeyInfo` from the import arguments, but doesn't actually import the key itself

riyazdf:
@endophage @cyli PTAL? :)

trustmanager/keyfilestore.go
```diff
@@ -51,34 +148,64 @@ func (s *KeyFileStore) Name() string {
...
 }
 
 // RemoveKey removes the key from the keyfilestore
 func (s *KeyFileStore) RemoveKey(name string) error {
 	s.Lock()
 	defer s.Unlock()
-	return removeKey(s, s.cachedKeys, name)
+	// If this is a bare key ID without the gun, prepend the gun so the filestore lookup succeeds
+	if keyInfo, ok := s.keyInfoMap[name]; ok {
+		name = filepath.Join(keyInfo.Gun, name)
```
cyli: Sorry, I may have just missed it, but is there a test that this remove works on keys with the file name format `keyID_role.key`?

cyli: NM, I forgot that `removeKey` does that

trustmanager/keyfilestore.go
```diff
@@ -88,8 +215,15 @@ func (s *KeyFileStore) ExportKey(name string) ([]byte, error) {
 
 // ImportKey imports the private key in the encrypted bytes into the keystore
 // with the given key ID and alias.
+// This is only used for root, so no need to touch the keyInfoMap
```
cyli: Is this comment true after the #536 was merged?  It looks like the `keyInfoMap` is being set regardless of alias.

trustmanager/keyfilestore.go
```diff
@@ -51,34 +148,64 @@ func (s *KeyFileStore) Name() string {
...
+		name = filepath.Join(keyInfo.Gun, name)
+	}
 	return getKey(s, s.Retriever, s.cachedKeys, name)
 }
 
 // ListKeys returns a list of unique PublicKeys present on the KeyFileStore.
-func (s *KeyFileStore) ListKeys() map[string]string {
-	return listKeys(s)
+func (s *KeyFileStore) ListKeys() map[string]KeyInfo {
+	return s.keyInfoMap
```
cyli: If we're directly returning `keyInfoMap`, will this allow users to edit the state after calling `ListKeys`?  Should we return a copy instead?

trustmanager/keyfilestore.go
```diff
@@ -147,7 +316,44 @@ func (s *KeyMemoryStore) ExportKey(name string) ([]byte, error) {
...
+	}
+	keyID, keyInfo := getImportedKeyInfo(pemBytes, alias)
+	s.keyInfoMap[keyID] = keyInfo
+	return nil
+}
+
+func getImportedKeyInfo(pemBytes []byte, filename string) (string, KeyInfo) {
+	var keyID, gun, role string
+	// Try to read the role and gun from the filepath and PEM headers
+	if filepath.HasPrefix(filename, "root_keys") {
```
cyli: Can we use the be `notary.RootKeysSubdir` constant instead here, and below?  (Likewise for the `NonRootKeysSubdir`?

cyli: Non-blocking:  this logic seems very similar to the one used in the loop in `generateKeyInfoMap` - is it possible for that to call into this function, perhaps?

cryptoservice/import_export_test.go
```diff
@@ -199,7 +199,7 @@ func TestImportExportGUN(t *testing.T) {
 		if alias == data.CanonicalRootRole {
 			continue
 		}
-		relKeyPath := filepath.Join("tuf_keys", privKeyName+".key")
```
cyli: Non-blocking nitpick:  can we also use the `notary.NonRootKeysSubdir` constant here and below?

cryptoservice/crypto_service.go
cyli: I am probably just sleepy, but it looks like if the key already exists in this key store, and matches the one we are trying to add, we just move onto the next keystore to try to add it there?  Should we just bail and count it as a successful job if the key already exists in our preferred keystore?  (Can we also add a test for this case?)

trustmanager/keyfilestore.go
riyazdf: ah good catch, yes this should be removed after the merge

trustmanager/keyfilestore.go
riyazdf: Good idea, `ListKeys` should never be used to modify state so I'll change this!

cryptoservice/import_export_test.go
riyazdf: good idea, will do!

trustmanager/keyfilestore.go
riyazdf: definitely agreed on using the notary consts, I'll also see what I can do with `generateKeyInfoMap`

cryptoservice/crypto_service.go
riyazdf: You were right: I was continuing to add to other keystores earlier, but I think bailing is the right behavior.  Done

cryptoservice/import_export_test.go
riyazdf: Done!

trustmanager/keyfilestore.go
```diff
@@ -38,10 +49,65 @@ func NewKeyFileStore(baseDir string, passphraseRetriever passphrase.Retriever) (
...
+	keyStore.loadKeyInfo()
+	return keyStore, nil
+}
+
+func generateKeyInfoMap(s LimitedFileStore) map[string]KeyInfo {
+	keyInfoMap := make(map[string]KeyInfo)
+	for _, keyPath := range s.ListFiles() {
+		d, err := s.Get(keyPath)
+		if err != nil {
+			fmt.Println("HERE!!!")
```
cyli: Extra println :)

cryptoservice/crypto_service_test.go
```diff
@@ -300,6 +300,40 @@ func (c CryptoServiceTester) TestListFromMultipleKeystores(t *testing.T) {
...
+	cryptoService.keyStores = append(cryptoService.keyStores,
+		trustmanager.NewKeyMemoryStore(passphraseRetriever))
+
+	privKey, err := trustmanager.GenerateECDSAKey(rand.Reader)
+	assert.NoError(t, err)
+
+	// Add the key to the targets role
+	assert.NoError(t, cryptoService.AddKey(privKey, data.CanonicalTargetsRole))
+
+	// Check that we added the key and its info to both keystores
```
cyli: Do we really want to add a key to all the key stores?  `Create` adds it to the first keystore it can - is there a use case for different behavior in this case?

riyazdf: You're right -- `GetKey`/`GetPrivateKey` should work as long as the key is stored in just one store.  After thinking about this, I don't think there is a use case so I'll change behavior to be consistent with `Create`

cyli: (It looks like if the key already exists in one key store, then we break, but if doesn't exist and we successfully add it, it keeps on adding to all the keystores)

cyli: Also apologies, this isn't getting run.  :|  It needs to be added to the `testCryptoService` function below

trustmanager/keyfilestore.go
cyli: Since this is now used for both generateKeyInfoMap and for importing, should this be renamed to something like keyInfoFromPEM or some such? Also, it looks like sometimes a role may not be able to be inferred if there is no role header in the PEM and the key path is not legacy-named. Should this return an error, if so? It seems like we should not be able to import any keys where we cannot infer the role (or we should write the role to the PEM headers on import).

Also previously generateKeyMap skipped keys with no role - as getImportedKeyInfo is currently used, it looks like it'd store the key IDs of keys with no role. Are we ok with that?

cryptoservice/crypto_service_test.go
riyazdf: ah, I was running the test individually, but that makes sense -- it's been added now :)

I've also updated the behavior for `AddKey` to behave as we've discussed: checking all keystores for existence before performing any adds, and exiting after the successful first keystore add

trustmanager/keyfilestore.go
riyazdf: Agreed with the rename.

After discussing IRL, it seems that we could benefit from removing `ImportRootKey` and refactoring the keystore `ImportKey` API to have a different signature: `ImportKey(pemBytes []byte, role, gun string)`.  This will make the code much easier to understand -- there's currently a lot of weirdness with importing keys.   For example: the `alias` existing argument to `ImportKey` can be a full filepath or just `"root"`

cmd/notary/integration_test.go
```diff
@@ -803,7 +803,7 @@ func assertNumKeys(t *testing.T, tempDir string, numRoot, numSigning int,
 		_, err := os.Stat(filepath.Join(
 			tempDir, "private", "root_keys", rootKeyID+".key"))
 		// os.IsExist checks to see if the error is because a file already
-		// exist, and hence doesn't actually the right funciton to use here
+		// exists, and hence it isn't actually the right function to use here
```
cyli: Heh, thanks for fixing these typos!

trustmanager/keyfilestore.go
```diff
@@ -87,9 +207,17 @@ func (s *KeyFileStore) ExportKey(name string) ([]byte, error) {
 }
 
 // ImportKey imports the private key in the encrypted bytes into the keystore
-// with the given key ID and alias.
-func (s *KeyFileStore) ImportKey(pemBytes []byte, alias string) error {
-	return importKey(s, s.Retriever, s.cachedKeys, alias, pemBytes)
+// with the given key ID, role, and gun.
+func (s *KeyFileStore) ImportKey(pemBytes []byte, role, gun string) error {
+	if data.IsDelegation(role) {
```
cyli: `KeyFilestore.AddKey` sets the gun to empty `keyInfo.Role == data.CanonicalRootRole || data.IsDelegation(keyInfo.Role) || !data.ValidRole(keyInfo.Role)` - would the same check make sense here?

trustmanager/keyfilestore.go
```diff
@@ -146,8 +318,30 @@ func (s *KeyMemoryStore) ExportKey(name string) ([]byte, error) {
 
 // ImportKey imports the private key in the encrypted bytes into the keystore
 // with the given key ID and alias.
-func (s *KeyMemoryStore) ImportKey(pemBytes []byte, alias string) error {
-	return importKey(s, s.Retriever, s.cachedKeys, alias, pemBytes)
+func (s *KeyMemoryStore) ImportKey(pemBytes []byte, role, gun string) error {
+	if data.IsDelegation(role) {
```
cyli: Similarly, `KeyMemoryStore.AddKey` checks for `keyInfo.Role == data.CanonicalRootRole || data.IsDelegation(keyInfo.Role) || !data.ValidRole(keyInfo.Role)` and sets the gun to "" - should this use the same check?

cyli:
Love the new Import refactor - it makes much more sense now.  LGTM aside from the above.

trustmanager/keyfilestore.go
riyazdf: Ah yes, will update!

trustmanager/keyfilestore.go
riyazdf: done!

trustmanager/keyfilestore.go
riyazdf: Ah yes, good catch.  Done!

cmd/notary/keys.go
```diff
@@ -268,27 +268,12 @@ func (k *keyCommander) keysExport(cmd *cobra.Command, args []string) error {
...
-					continue
-				}
-				dirPath := filepath.Dir(keypath)
-				if dirPath != "." { // no gun
-					keyGun = dirPath
-				}
-				break
-			}
-		}
+	cs := cryptoservice.NewCryptoService("", ks...)
```
endophage: Should we consider changing the CryptoService to not need the GUN anymore? It seems like this PR might obsolete its use?

cyli: I think that might necessitate a change to `CryptoService.Create` to take a gun, rather than store the gun on the `CryptoService` itself, but that might be a good thing to move towards.  Possibly in another PR?

riyazdf: Yeah we would need to include a gun for both `CryptoService.Create` and `CryptoService.AddKey`.  I'm leaning toward making this change in another PR, since it will affect a lot of our repo code?

endophage: Sounds good!

cmd/notary/keys.go
```diff
@@ -599,7 +575,11 @@ func (k *keyCommander) keyPassphraseChange(cmd *cobra.Command, args []string) er
 	// unlocking passphrase and reusing that.
 	passChangeRetriever := k.getRetriever()
 	keyStore, err := trustmanager.NewKeyFileStore(config.GetString("trust_dir"), passChangeRetriever)
-	err = keyStore.AddKey(filepath.Join(keyGUN, keyID), role, privKey)
+	keyInfo, err := cs.GetKeyInfo(keyID)
+	if err != nil {
+		return err
+	}
+	err = keyStore.AddKey(privKey, keyInfo)
```
endophage: Should we ensure we're adding the key back to the same keystore we found it in? Easiest way to do that would be to iterate keystores, create a CryptoService with only one key store at a time and attempt to Get/Add via the CryptoService.

cryptoservice/crypto_service.go
```diff
@@ -74,23 +66,17 @@ func (cs *CryptoService) Create(role, algorithm string) (data.PublicKey, error)
...
-			if err == nil {
-				return
-			}
-			switch err.(type) {
-			case trustmanager.ErrPasswordInvalid, trustmanager.ErrAttemptsExceeded:
-				return
-			default:
-				continue
-			}
+		k, role, err = ks.GetKey(keyID)
```
endophage: Because all the vars are declared, we can do 
```
if k, role, err = ks.GetKey(keyID); err == nil {
    return
}
```

cryptoservice/crypto_service.go
```diff
@@ -105,12 +91,42 @@ func (cs *CryptoService) GetKey(keyID string) data.PublicKey {
...
-		for _, keyPath := range keyPaths {
-			ks.RemoveKey(keyPath)
+		ks.RemoveKey(keyID)
+	}
+	return // returns whatever the final values were
+}
+
+// AddKey adds a private key to a specified role.
+// The GUN is inferred from the cryptoservice itself for non-root roles
+func (cs *CryptoService) AddKey(key data.PrivateKey, role string) (err error) {
```
endophage: for consistency with existing functions (Create), can we pass the role first.

cryptoservice/import_export.go
```diff
@@ -100,49 +101,22 @@ func (cs *CryptoService) ExportKeyReencrypt(dest io.Writer, keyID string, newPas
...
-	if err != nil {
-		return err
-	}
-	return cs.ImportRoleKey(pemBytes, data.CanonicalRootRole, nil)
-}
-
 // ImportRoleKey imports a private key in PEM format key from a byte array
 // It prompts for the key's passphrase to verify the data and to determine
 // the key ID.
 func (cs *CryptoService) ImportRoleKey(pemBytes []byte, role string, newPassphraseRetriever passphrase.Retriever) error {
```
endophage: Is there any way we can kill this function altogether and replace it with AddKey? Would be much clearer if there were only one way to add a key.

trustmanager/keyfilestore.go
```diff
@@ -51,34 +141,64 @@ func (s *KeyFileStore) Name() string {
 }
 
 // AddKey stores the contents of a PEM-encoded private key as a PEM block
-func (s *KeyFileStore) AddKey(name, role string, privKey data.PrivateKey) error {
+func (s *KeyFileStore) AddKey(privKey data.PrivateKey, keyInfo KeyInfo) error {
```
endophage: Can we put KeyInfo first? I guess my thing with these kind of storage related functions is I really like the arg order to be (identifier, data)

trustmanager/keyfilestore.go
```diff
@@ -51,34 +141,64 @@ func (s *KeyFileStore) Name() string {
...
+func (s *KeyFileStore) ListKeys() map[string]KeyInfo {
+	return copyKeyInfoMap(s.keyInfoMap)
 }
 
 // RemoveKey removes the key from the keyfilestore
 func (s *KeyFileStore) RemoveKey(name string) error {
 	s.Lock()
 	defer s.Unlock()
-	return removeKey(s, s.cachedKeys, name)
+	// If this is a bare key ID without the gun, prepend the gun so the filestore lookup succeeds
```
endophage: Would we still have the gun in the `name`?

trustmanager/keyfilestore.go
```diff
@@ -51,34 +141,64 @@ func (s *KeyFileStore) Name() string {
...
+		// This might be of the form GUN/ID  - try to delete without the gun
+		delete(s.keyInfoMap, filepath.Base(name))
+	}
+	return nil
 }
 
-// ExportKey exportes the encrypted bytes from the keystore and writes it to
+// ExportKey exports the encrypted bytes from the keystore and writes it to
 // dest.
 func (s *KeyFileStore) ExportKey(name string) ([]byte, error) {
```
endophage: At this level, Export and Import should definitely be replaced by GetKey and AddKey and we should make the interface consistent. Either all the methods in the KeyStore operate on []byte, or they all operate on PrivateKey/PublicKey.

riyazdf:
@endophage and I met up for a design session.  It seems most fruitful that we completely overhaul the Cryptoservice and Keystore as follows:

- Cryptoservice operates only on typed keys, Keystore operates directly on byte arrays
- Cryptoservice keeps a mapping of key ID to role, gun, and keystores containing the key
    - Cryptoservice handles all role and gun bookkeeping for keys
- Keystore keeps a mapping of key ID to key path
- Remove Import/ExportKey from both Cryptoservice and Keystore.  The only way to add a key should be through a call to `AddKey`, and `GetKey` should be used for exporting

This is a big change from the current codebase (and from the code in this PR) but this should really tighten and clear up the code further.  Please comment with any design suggestions!

endophage:
:+1: This is going to be really good when done.

cmd/notary/keys.go
riyazdf: I'm overhauling cryptoservice and keystore with the changes below, and in this refactor I'll also remove the `gun` from cryptoservice -- awesome idea :)

cryptoservice/crypto_service.go
riyazdf: :+1:

cryptoservice/import_export.go
riyazdf: yes we can!  Will do.

trustmanager/keyfilestore.go
riyazdf: I don't think we ever would, but in any case we would fail the key info lookup here.

trustmanager/keyfilestore.go
riyazdf: done!

cryptoservice/import_export.go
riyazdf: done!

cryptoservice/crypto_service.go
riyazdf: done!

cmd/notary/keys.go
riyazdf: Good catch!  This is tricky, I've pushed a way of doing this but it isn't very clean because we have to make a new keystore with a new passphrase retriever so that we overwrite the previous key

trustmanager/keyfilestore.go
```diff
@@ -109,45 +231,89 @@ func (s *KeyMemoryStore) Name() string {
...
 	}
 	return keyBytes, nil
 }
 
-// ImportKey imports the private key in the encrypted bytes into the keystore
-// with the given key ID and alias.
-func (s *KeyMemoryStore) ImportKey(pemBytes []byte, alias string) error {
-	return importKey(s, s.Retriever, s.cachedKeys, alias, pemBytes)
+// KeyInfoFromPEM attempts to get a keyID and KeyInfo from the filename and PEM bytes of a key
+func KeyInfoFromPEM(pemBytes []byte, filename string) (string, KeyInfo) {
```
cyli: I'm wondering if this should return an error, also, if `pem.Decode` fails?  If so, `generateKeyInfo` can skip that file.  Right now on master, if I drop some random file named `stuff.key` into the key directory, it is ignored because I guess it's invalid.  I think this would make it so that it gets listed as a key with ID `stuff` and no GUN or role?

trustmanager/keyfilestore.go
```diff
@@ -109,45 +231,89 @@ func (s *KeyMemoryStore) Name() string {
...
+	if keyInfo, ok := s.keyInfoMap[keyID]; ok {
+		keyID = filepath.Join(keyInfo.Gun, keyID)
+	}
+	err := removeKey(s, s.cachedKeys, keyID)
+	if err != nil {
+		return err
+	}
+	// Remove this key from our keyInfo map if we removed from our filesystem
+	if _, ok := s.keyInfoMap[keyID]; ok {
+		delete(s.keyInfoMap, keyID)
```
cyli: Non-blocking - since `keyInfoMap` only contains the key ID as keys, and not the GUN/ID, you might be able to avoid the `if` and just calling `delete(s.keyInfoMap, filepath.Base(keyID))` since `filepath.Base(keyID)` just returns the same value for `keyID` if `keyID` doesn't have a "GUN/" attached.

trustmanager/keyfilestore.go
```diff
@@ -109,45 +231,89 @@ func (s *KeyMemoryStore) Name() string {
...
+	} else {
+		// This might be of the form GUN/ID  - try to delete without the gun
+		delete(s.keyInfoMap, filepath.Base(keyID))
+	}
+	return nil
 }
 
-// ExportKey exportes the encrypted bytes from the keystore and writes it to
+// ExportKey exports the encrypted bytes from the keystore and writes it to
```
cyli: Thanks for fixing this typo!  :)  It looks like at some point this function got changed though and it no longer takes a destination, but the comment was never updated.  Would you mind deleting the "and writes it to dest" part?

cryptoservice/crypto_service_test.go
```diff
@@ -300,6 +300,53 @@ func (c CryptoServiceTester) TestListFromMultipleKeystores(t *testing.T) {
 	}
 }
 
+// asserts that adding a key adds to all keystores
```
cyli: Should this comment be "adds to the first keystore"?

cmd/notary/prettyprint_test.go
```diff
@@ -103,12 +103,12 @@ func TestPrettyPrintRootAndSigningKeys(t *testing.T) {
...
 	// add keys to the key stores
-	assert.NoError(t, keyStores[0].AddKey(keys[0].ID(), root, keys[0]))
-	assert.NoError(t, keyStores[1].AddKey(keys[0].ID(), root, keys[0]))
-	assert.NoError(t, keyStores[0].AddKey(strings.Repeat("a/", 30)+keys[0].ID(), "targets", keys[0]))
-	assert.NoError(t, keyStores[1].AddKey("short/gun/"+keys[0].ID(), "snapshot", keys[0]))
-	assert.NoError(t, keyStores[0].AddKey(keys[1].ID(), "targets/a", keys[1]))
-	assert.NoError(t, keyStores[0].AddKey(keys[2].ID(), "invalidRole", keys[2]))
+	assert.NoError(t, keyStores[0].AddKey(trustmanager.KeyInfo{Role: root, Gun: ""}, keys[0]))
+	assert.NoError(t, keyStores[1].AddKey(trustmanager.KeyInfo{Role: root, Gun: ""}, keys[0]))
+	assert.NoError(t, keyStores[0].AddKey(trustmanager.KeyInfo{Role: "targets", Gun: strings.Repeat("/a", 30)}, keys[1]))
```
cyli: Apologies, this was my bad for just using the quoted string, but can we change these to use the constants instead?

trustmanager/keyfilestore.go
riyazdf: `pem.Decode` returns the first block it finds and the "rest" as a second return -- I think the only validation we could do here is if we're able to get a key ID with valid size and a role from the PEM or filename.  I can make this an error condition, what do you think?

cryptoservice/crypto_service.go
```diff
@@ -105,12 +89,42 @@ func (cs *CryptoService) GetKey(keyID string) data.PublicKey {
...
+	for _, store := range cs.keyStores {
+		if info, err := store.GetKeyInfo(keyID); err == nil {
+			return info, nil
+		}
+	}
+	return trustmanager.KeyInfo{}, fmt.Errorf("Could not find info for keyID %s", keyID)
+}
+
 // RemoveKey deletes a key by ID
 func (cs *CryptoService) RemoveKey(keyID string) (err error) {
```
cyli: Non-blocking - it looks like this never returns an error - should we just remove the return value?

trustmanager/keyfilestore.go
cyli: Ah oops, I had misread the `pem.Decode`.  Hmm... should it maybe return an error if the block is nil, or some such?  Just so we don't include that key?  Just because it currently skips over any file that isn't really a key, and I'm not sure if anything else would break if someone drops a messed up file the directory and we treat it as a key.  If nothing does, then this doesn't matter very much other than a UI thing.

cryptoservice/crypto_service.go
riyazdf: it seems we use the error in `NotarySigner`, which implements the same interface

trustmanager/keyfilestore.go
riyazdf: good idea, done!

cmd/notary/keys.go
```diff
@@ -580,30 +579,43 @@ func (k *keyCommander) keyPassphraseChange(cmd *cobra.Command, args []string) er
...
 	}
-
 	// Must use a different passphrase retriever to avoid caching the
 	// unlocking passphrase and reusing that.
 	passChangeRetriever := k.getRetriever()
-	keyStore, err := trustmanager.NewKeyFileStore(config.GetString("trust_dir"), passChangeRetriever)
-	err = keyStore.AddKey(filepath.Join(keyGUN, keyID), role, privKey)
+	var addingKeyStore trustmanager.KeyStore
+	switch foundKeyStore.Name() {
+	case "yubikey":
```
cyli: Totally non-blocking, and just because you previously mentioned you didn't like switching on the name very much - what do you think about having the keystore return constant types sort of like key algorithms are constants?  Then here we can switch based on the constants.  

(this may not be worth doing at all, since we may want to refactor all this at a later point anyway)

cmd/notary/keys.go
```diff
@@ -322,11 +307,11 @@ func (k *keyCommander) keysRestore(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	ks, err := k.getKeyStores(config, true)
+	ks, err := k.getKeyStores(config, true, true)
```
cyli: Non-blocking - just wondering if this should actually create a backup - `keysImport` does not, and if you're restoring you already have a backup in terms of a zip file.

trustmanager/yubikey/yubikeystore.go
```diff
@@ -639,15 +639,15 @@ func (s *YubiKeyStore) ListKeys() map[string]string {
...
-func (s *YubiKeyStore) AddKey(keyID, role string, privKey data.PrivateKey) error {
-	added, err := s.addKey(keyID, role, privKey)
+func (s *YubiKeyStore) AddKey(keyInfo trustmanager.KeyInfo, privKey data.PrivateKey) error {
+	added, err := s.addKey(privKey.ID(), keyInfo.Role, privKey)
 	if err != nil {
 		return err
 	}
-	if added {
-		err = s.backupStore.AddKey(privKey.ID(), role, privKey)
+	if added && s.backupStore != nil {
```
cyli: :+1: Awesome to not always have to write to a backup store!  This makes importing much nicer!

cyli: Should we add a test where add key when `backupStore` is nil succeeds?

cryptoservice/crypto_service.go
cyli: Looking over the `NotarySigner` code, it might be more consistent if the deletion was successful whether or not the key exists in the DB?  In that case, it may not have to return an error.   But I'm happy for that to just happen in the future if it's needed - this PR is fine as is :)

cyli:
Other than some non-blocking nitpicks above, LGTM!

trustmanager/yubikey/yubikeystore.go
riyazdf: great idea, done!

cmd/notary/keys.go
riyazdf: good catch, fixed :)

cryptoservice/crypto_service.go
riyazdf: I'm ok with that behavior but I agree that it might be better in a future PR if we decide we want it.  Currently the `DeleteKey` in `rpc_api.go` returns different messages depending on the error so I'm not as sure how this change would affect other parts of the codebase

cmd/notary/keys.go
cyli: Note: this was originally on `keyRestore` :(  I must have commented on an out of date diff or something - got moved.

endophage:
Needs a rebase. Otherwise LGTM!

cyli:
YAAAAAY this is finally making it in!  \o/

PR #553, opened 2016-02-08
WIP: Client rotate remote key

cyli:
This supercedes https://github.com/docker/notary/pull/529.

This adds the functionality for the client to rotate remote keys, including the timestamp key.

In order to do so, it must sign a request with the root key, in order to prove that it has the root key, and the rotation is published right away.

Part of addressing #347

### DIFF
--- a/client/backwards_compatibility_test.go
+++ b/client/backwards_compatibility_test.go
@@ -12,7 +12,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/docker/notary/client/changelist"
 	"github.com/docker/notary/passphrase"
 	"github.com/docker/notary/tuf/data"
 	"github.com/docker/notary/tuf/store"
@@ -108,11 +107,7 @@ func Test0Dot1RepoFormat(t *testing.T) {
 	require.NoError(t, repo.fileStore.RemoveMeta(data.CanonicalTimestampRole))
 
 	// rotate the timestamp key, since the server doesn't have that one
-	timestampPubKey, err := getRemoteKey(ts.URL, gun, data.CanonicalTimestampRole, http.DefaultTransport)
-	require.NoError(t, err)
-	require.NoError(
-		t, repo.rootFileKeyChange(data.CanonicalTimestampRole, changelist.ActionCreate, timestampPubKey))
-
+	require.NoError(t, repo.RotateKey(data.CanonicalTimestampRole, true))
 	require.NoError(t, repo.Publish())
 
 	targets, err = repo.ListTargets()
--- a/client/backwards_compatibility_test.go
+++ b/client/backwards_compatibility_test.go
@@ -12,6 +12,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/docker/notary/client/changelist"
 	"github.com/docker/notary/passphrase"
 	"github.com/docker/notary/tuf/data"
 	"github.com/docker/notary/tuf/store"
@@ -106,8 +107,13 @@ func Test0Dot1RepoFormat(t *testing.T) {
 	// one and try to create a new one from scratch, which will be the wrong version
 	require.NoError(t, repo.fileStore.RemoveMeta(data.CanonicalTimestampRole))
 
-	// rotate the timestamp key, since the server doesn't have that one
-	require.NoError(t, repo.RotateKey(data.CanonicalTimestampRole, true))
+	// rotate the timestamp key, since the server doesn't have that one (we can't just
+	// call RotateKey because the server will verify that it is signed by the root, and
+	// the server doesn't know about this one)
+	timestampPubKey, err := getRemoteKey(ts.URL, gun, data.CanonicalTimestampRole, http.DefaultTransport)
+	require.NoError(t, err)
+	require.NoError(
+		t, repo.rootFileKeyChange(data.CanonicalTimestampRole, changelist.ActionCreate, timestampPubKey))
 	require.NoError(t, repo.Publish())
 
 	targets, err = repo.ListTargets()
--- a/client/backwards_compatibility_test.go
+++ b/client/backwards_compatibility_test.go
@@ -12,7 +12,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/docker/notary/client/changelist"
 	"github.com/docker/notary/passphrase"
 	"github.com/docker/notary/tuf/data"
 	"github.com/docker/notary/tuf/store"
@@ -107,14 +106,9 @@ func Test0Dot1RepoFormat(t *testing.T) {
 	// one and try to create a new one from scratch, which will be the wrong version
 	require.NoError(t, repo.fileStore.RemoveMeta(data.CanonicalTimestampRole))
 
-	// rotate the timestamp key, since the server doesn't have that one (we can't just
-	// call RotateKey because the server will verify that it is signed by the root, and
-	// the server doesn't know about this one)
-	timestampPubKey, err := getRemoteKey(ts.URL, gun, data.CanonicalTimestampRole, http.DefaultTransport)
-	require.NoError(t, err)
-	require.NoError(
-		t, repo.rootFileKeyChange(data.CanonicalTimestampRole, changelist.ActionCreate, timestampPubKey))
-	require.NoError(t, repo.Publish())
+	// We need to rotate the timestamp key, since the server doesn't have the one
+	// specified by the metadata.
+	require.NoError(t, repo.RotateKey(data.CanonicalTimestampRole, true))
 
 	targets, err = repo.ListTargets()
 	require.NoError(t, err)
--- a/client/client.go
+++ b/client/client.go
@@ -843,6 +843,8 @@ func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool) error {
 		return fmt.Errorf(
 			"notary does not currently support rotating the %s key", role)
 	}
+	// We currently support locally managing targets keys, remotely managing
+	// timestamp keys, and locally or remotely managing timestamp keys.
 	if serverManagesKey && role == data.CanonicalTargetsRole {
 		return ErrInvalidRemoteRole{Role: data.CanonicalTargetsRole}
 	}
@@ -855,31 +857,32 @@ func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool) error {
 		errFmtString string
 		err          error
 	)
+	// Key rotation is an offline operation unless we are rotating a remote key, in which case we
+	// need to ensure the rotation can actually happen: the user needs to have the root key, and
+	// we want to publish right away so that the remote key doesn't expire
 	if serverManagesKey {
-		if err := r.bootstrapRepo(); err != nil {
+		if err = r.bootstrapRepo(); err != nil {
+			if _, ok := err.(store.ErrMetaNotFound); ok {
+				err = ErrRepoNotInitialized{}
+			}
+			logrus.Debug("Repository should be valid and initialized before rotating a remote key:", err)
 			return err
 		}
 
 		// get root keys - since we bootstrapped the repo, a valid root should always specify
 		// the root role
-		rootRole, ok := r.tufRepo.Root.Signed.Roles[data.CanonicalRootRole]
-		if !ok {
-			return fmt.Errorf("update resulted in an invalid state with an invalid root")
-		}
-		rootKeys := make([]data.PublicKey, 0, len(rootRole.KeyIDs))
-		for _, kid := range rootRole.KeyIDs {
-			k, ok := r.tufRepo.Root.Signed.Keys[kid]
-			if ok {
-				// this should always be the case, as key IDs need to correspond to keys in the
-				// key list, but just in case: skip if it's not
-				rootKeys = append(rootKeys, k)
-			}
+		rootKeys, err := r.tufRepo.SigningKeysForRole(data.CanonicalRootRole)
+		if err != nil {
+			logrus.Debug("The root is invalid, because it does not contain any signing keys")
+			return err
 		}
 
 		errFmtString = "unable to rotate remote key: %s"
 		remote, err := getRemoteStore(r.baseURL, r.gun, r.roundTrip)
 		if err == nil {
 			pubKey, err = remote.RotateKey(role, r.CryptoService, rootKeys...)
+		} else if _, ok := err.(signed.ErrNoKeys); ok {
+			err = fmt.Errorf("root signing key unavailable so unable to rotate key anyway")
 		}
 	} else {
 		pubKey, err = r.CryptoService.Create(role, data.ECDSAKey)
--- a/client/client.go
+++ b/client/client.go
@@ -55,6 +55,17 @@ func (err ErrInvalidRemoteRole) Error() string {
 		"notary does not support the server managing the %s key", err.Role)
 }
 
+// ErrInvalidLocalRole is returned when the client wants to manage
+// an unsupported key type
+type ErrInvalidLocalRole struct {
+	Role string
+}
+
+func (err ErrInvalidLocalRole) Error() string {
+	return fmt.Sprintf(
+		"notary does not support the client managing the %s key", err.Role)
+}
+
 // ErrRepositoryNotExist is returned when an action is taken on a remote
 // repository that doesn't exist
 type ErrRepositoryNotExist struct {
@@ -829,25 +840,32 @@ func (r *NotaryRepository) validateRoot(rootJSON []byte) (*data.SignedRoot, erro
 // creates and adds one new key or delegates managing the key to the server.
 // These changes are staged in a changelist until publish is called.
 func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool) error {
-	if role == data.CanonicalRootRole || role == data.CanonicalTimestampRole {
+	if role == data.CanonicalRootRole {
 		return fmt.Errorf(
 			"notary does not currently support rotating the %s key", role)
 	}
 	if serverManagesKey && role == data.CanonicalTargetsRole {
 		return ErrInvalidRemoteRole{Role: data.CanonicalTargetsRole}
 	}
+	if !serverManagesKey && role == data.CanonicalTimestampRole {
+		return ErrInvalidLocalRole{Role: data.CanonicalTargetsRole}
+	}
 
 	var (
-		pubKey data.PublicKey
-		err    error
+		pubKey       data.PublicKey
+		errFmtString string
+		err          error
 	)
 	if serverManagesKey {
-		pubKey, err = getRemoteKey(r.baseURL, r.gun, role, r.roundTrip)
+		pubKey, err = getRemoteKey(r.baseURL, r.gun, role, r.roundTrip, true)
+		errFmtString = "server unable to rotate key: %s"
 	} else {
 		pubKey, err = r.CryptoService.Create(role, data.ECDSAKey)
+		errFmtString = "unable to generate key: %s"
 	}
+
 	if err != nil {
-		return err
+		return fmt.Errof(errFmtString, err)
 	}
 
 	return r.rootFileKeyChange(role, changelist.ActionCreate, pubKey)
--- a/client/client.go
+++ b/client/client.go
@@ -522,11 +522,10 @@ func (r *NotaryRepository) Publish() error {
 			initialPublish = true
 		} else {
 			// We could not update, so we cannot publish.
-			logrus.Error("Could not publish Repository: ", err.Error())
+			logrus.Error("Could not publish Repository since we could not update: ", err.Error())
 			return err
 		}
 	}
-
 	cl, err := r.GetChangelist()
 	if err != nil {
 		return err
@@ -619,7 +618,7 @@ func (r *NotaryRepository) Publish() error {
 // to load metadata for all roles.  Since server snapshots are supported,
 // if the snapshot metadata fails to load, that's ok.
 // This can also be unified with some cache reading tools from tuf/client.
-// This assumes that bootstrapRepo is only used by Publish()
+// This assumes that bootstrapRepo is only used by Publish() or RotateKey()
 func (r *NotaryRepository) bootstrapRepo() error {
 	kdb := keys.NewDB()
 	tufRepo := tuf.NewRepo(kdb, r.CryptoService)
@@ -848,7 +847,7 @@ func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool) error {
 		return ErrInvalidRemoteRole{Role: data.CanonicalTargetsRole}
 	}
 	if !serverManagesKey && role == data.CanonicalTimestampRole {
-		return ErrInvalidLocalRole{Role: data.CanonicalTargetsRole}
+		return ErrInvalidLocalRole{Role: data.CanonicalTimestampRole}
 	}
 
 	var (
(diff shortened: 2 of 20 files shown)


client/backwards_compatibility_test.go
```diff
@@ -107,12 +107,13 @@ func Test0Dot1RepoFormat(t *testing.T) {
 	// one and try to create a new one from scratch, which will be the wrong version
 	require.NoError(t, repo.fileStore.RemoveMeta(data.CanonicalTimestampRole))
 
-	// rotate the timestamp key, since the server doesn't have that one
+	// rotate the timestamp key, since the server doesn't have that one (we can't just
```
diogomonica: Confusing comment.

client/client.go
```diff
@@ -829,28 +839,65 @@ func (r *NotaryRepository) validateRoot(rootJSON []byte) (*data.SignedRoot, erro
...
 func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool) error {
-	if role == data.CanonicalRootRole || role == data.CanonicalTimestampRole {
+	if role == data.CanonicalRootRole {
 		return fmt.Errorf(
 			"notary does not currently support rotating the %s key", role)
 	}
 	if serverManagesKey && role == data.CanonicalTargetsRole {
 		return ErrInvalidRemoteRole{Role: data.CanonicalTargetsRole}
 	}
+	if !serverManagesKey && role == data.CanonicalTimestampRole {
```
diogomonica: We should probably add a comment before these ifs describing that these are our current assumptions: Targets is always local and timestamps is always remote.

client/client.go
```diff
@@ -829,28 +839,65 @@ func (r *NotaryRepository) validateRoot(rootJSON []byte) (*data.SignedRoot, erro
...
 	)
 	if serverManagesKey {
-		pubKey, err = getRemoteKey(r.baseURL, r.gun, role, r.roundTrip)
+		if err := r.bootstrapRepo(); err != nil {
+			return err
+		}
+
+		// get root keys - since we bootstrapped the repo, a valid root should always specify
+		// the root role
+		rootRole, ok := r.tufRepo.Root.Signed.Roles[data.CanonicalRootRole]
```
diogomonica: not sure if `rootRole` is a good name for a list of root keys.

diogomonica: What about having a `r.getRootKeys()`

cyli: I was thinking about adding a `GetSigningKeys` to `tuf.Repo`, which already has logic similar to this for getting signing keys for signing a particular role.

client/client.go
```diff
@@ -829,28 +839,65 @@ func (r *NotaryRepository) validateRoot(rootJSON []byte) (*data.SignedRoot, erro
...
+		if !ok {
+			return fmt.Errorf("update resulted in an invalid state with an invalid root")
```
diogomonica: Let's turn this into a method that returns `err` with something more descriptive.

client/client.go
```diff
@@ -829,28 +839,65 @@ func (r *NotaryRepository) validateRoot(rootJSON []byte) (*data.SignedRoot, erro
...
+		}
+		rootKeys := make([]data.PublicKey, 0, len(rootRole.KeyIDs))
```
diogomonica: This should all be replaced by `getRootKeys()`

client/client.go
```diff
@@ -829,28 +839,65 @@ func (r *NotaryRepository) validateRoot(rootJSON []byte) (*data.SignedRoot, erro
...
 	var (
-		pubKey data.PublicKey
-		err    error
+		pubKey       data.PublicKey
+		errFmtString string
+		err          error
 	)
 	if serverManagesKey {
-		pubKey, err = getRemoteKey(r.baseURL, r.gun, role, r.roundTrip)
+		if err := r.bootstrapRepo(); err != nil {
```
diogomonica: why `:=`?

client/client.go
```diff
@@ -829,28 +839,65 @@ func (r *NotaryRepository) validateRoot(rootJSON []byte) (*data.SignedRoot, erro
 // creates and adds one new key or delegates managing the key to the server.
 // These changes are staged in a changelist until publish is called.
 func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool) error {
```
diogomonica: This method could probably use some `Debug` output.

client/client.go
cyli: This error isn't propagated to the temporary variable anyway, since it gets returned immediately.  I'm happy to make it `=` though.

client/client.go
```diff
@@ -829,28 +839,65 @@ func (r *NotaryRepository) validateRoot(rootJSON []byte) (*data.SignedRoot, erro
...
 	if err != nil {
+		return fmt.Errorf(errFmtString, err)
+	}
+
+	if err = r.rootFileKeyChange(role, changelist.ActionCreate, pubKey); err != nil {
 		return err
 	}
 
-	return r.rootFileKeyChange(role, changelist.ActionCreate, pubKey)
+	if serverManagesKey {
```
diogomonica: Comment to the effect: `Key rotation is an offline operation unless we are rotating a remote key, in which case we need to ensure the rotation actually happens`

cmd/notary/keys.go
```diff
@@ -361,13 +361,11 @@ func (k *keyCommander) keysRotate(cmd *cobra.Command, args []string) error {
 		rolesToRotate = []string{data.CanonicalSnapshotRole}
 	case data.CanonicalTargetsRole:
 		rolesToRotate = []string{data.CanonicalTargetsRole}
+	case data.CanonicalTimestampRole:
```
diogomonica: Does it still make sense to keep the default behavior for `rotate` with no arguments? (targets and snapshot?)

cyli: I think only if we care about backwards compatibility - if we don't, I'm fine with key rotation to require a role.  (Also, does it make sense to allow the user to manage their own timestamp key?  They can really screw themselves over, but if they're managing everything...)

cyli: Although maybe changing how CLI key rotation works should be in a different PR?

server/handlers/default.go
```diff
@@ -143,70 +143,95 @@ func DeleteHandler(ctx context.Context, w http.ResponseWriter, r *http.Request)
...
 	}
-
-	logger := ctxu.GetLoggerWithField(ctx, gun, "gun")
 
 	s := ctx.Value("metaStore")
 	store, ok := s.(storage.MetaStore)
 	if !ok || store == nil {
-		logger.Error("500 GET storage not configured")
-		return errors.ErrNoStorage.WithDetail(nil)
+		return nil, errors.ErrNoStorage.WithDetail(nil)
```
diogomonica: Is it useful here to return some detail in case a user gets the error and pastes it to us?

server/handlers/default.go
```diff
@@ -143,70 +143,95 @@ func DeleteHandler(ctx context.Context, w http.ResponseWriter, r *http.Request)
...
 	}
+
 	c := ctx.Value("cryptoService")
 	crypto, ok := c.(signed.CryptoService)
 	if !ok || crypto == nil {
-		logger.Error("500 GET crypto service not configured")
-		return errors.ErrNoCryptoService.WithDetail(nil)
+		return nil, errors.ErrNoCryptoService.WithDetail(nil)
```
diogomonica: Same here. Even if its an opaque error it might be useful to distinguish cryptoservice errors in the wild (vs storage errors, for example)

client/client.go
cyli: Done!

client/client.go
cyli: I added one Debug, and otherwise just tried to return better errors.  Does it look reasonable atm?

cmd/notary/keys.go
endophage: I feel like we should change it, but I'm ok with it being a secondary PR

client/client.go
```diff
@@ -55,6 +55,17 @@ func (err ErrInvalidRemoteRole) Error() string {
...
 
+// ErrInvalidLocalRole is returned when the client wants to manage
+// an unsupported key type
+type ErrInvalidLocalRole struct {
+	Role string
+}
+
+func (err ErrInvalidLocalRole) Error() string {
+	return fmt.Sprintf(
+		"notary does not support the client managing the %s key", err.Role)
```
endophage: Should the messaging be stronger? "notary does not permit ..."? Don't want people filing issues for things we never intend to support.

client/client.go
```diff
@@ -829,28 +839,68 @@ func (r *NotaryRepository) validateRoot(rootJSON []byte) (*data.SignedRoot, erro
 // creates and adds one new key or delegates managing the key to the server.
 // These changes are staged in a changelist until publish is called.
 func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool) error {
-	if role == data.CanonicalRootRole || role == data.CanonicalTimestampRole {
+	if role == data.CanonicalRootRole {
 		return fmt.Errorf(
 			"notary does not currently support rotating the %s key", role)
 	}
+	// We currently support locally managing targets keys, remotely managing
+	// timestamp keys, and locally or remotely managing timestamp keys.
```
endophage: "locally or remotely managing _snapshot_ keys"

client/client.go
```diff
@@ -829,28 +839,68 @@ func (r *NotaryRepository) validateRoot(rootJSON []byte) (*data.SignedRoot, erro
 // creates and adds one new key or delegates managing the key to the server.
 // These changes are staged in a changelist until publish is called.
 func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool) error {
-	if role == data.CanonicalRootRole || role == data.CanonicalTimestampRole {
+	if role == data.CanonicalRootRole {
```
endophage: I'd be inclined to turn these into a `switch role`

tuf/store/offlinestore.go
```diff
@@ -38,7 +41,12 @@ func (es OfflineStore) RemoveMeta(name string) error {
 }
 
 // GetKey returns ErrOffline
-func (es OfflineStore) GetKey(role string) ([]byte, error) {
+func (es OfflineStore) GetKey(role string) (data.PublicKey, error) {
+	return nil, err
```
endophage: Looking at this again, I shouldn't have use the name `err`. It's not breaking anything now but it's a package scope var. Do you mind updating it to `errOffline` or something else less generic?

tuf/store/interfaces.go
```diff
@@ -12,7 +17,8 @@ type MetadataStore interface {
 
 // PublicKeyStore must be implemented by a key service
 type PublicKeyStore interface {
-	GetKey(role string) ([]byte, error)
+	GetKey(role string) (data.PublicKey, error)
+	RotateKey(role string, cs signed.CryptoService, roots ...data.PublicKey) (data.PublicKey, error)
```
endophage: The stores are meant to be reasonably low level and it feels like they're losing cohesion when they start referencing `PublicKey` and `CryptoService`.

client/client.go
```diff
@@ -829,28 +839,68 @@ func (r *NotaryRepository) validateRoot(rootJSON []byte) (*data.SignedRoot, erro
...
+		rootKeys, err := r.tufRepo.SigningKeysForRole(data.CanonicalRootRole)
+		if err != nil {
+			logrus.Debug("The root is invalid, because it does not contain any signing keys")
+			return err
+		}
+
+		errFmtString = "unable to rotate remote key: %s"
+		remote, err := getRemoteStore(r.baseURL, r.gun, r.roundTrip)
+		if err == nil {
+			pubKey, err = remote.RotateKey(role, r.CryptoService, rootKeys...)
```
endophage: I think the signing functionality currently contained in `remote.RotateKey` should be happening here and the stores should continue to receive and return on `[]byte`

const.go
```diff
@@ -18,4 +18,7 @@ const (
 	Sha256HexSize = 64
 	// TrustedCertsDir is the directory, under the notary repo base directory, where trusted certs are stored
 	TrustedCertsDir = "trusted_certificates"
+
+	// HTTPStatusTooManyRequests is the http status 429 - will be an exported constant in http library in Go 1.6
+	HTTPStatusTooManyRequests = 429
```
endophage: :-/ I'm disappointed that net/http doesn't already have a const for this. Looks like the RFC (6585) still hasn't been accepted after almost 4 years.

server/errors/errors.go
```diff
@@ -83,8 +84,14 @@ var (
...
 		Value:          "NO_KEYALGORITHM",
 		Message:        "The server does not have a key algorithm configured.",
-		Description:    "No key algorihtm has been configured for the server and it has been asked to perform an operation that requires generation.",
+		Description:    "No key algorithm has been configured for the server and it has been asked to perform an operation that requires generation.",
 		HTTPStatusCode: http.StatusInternalServerError,
 	})
+	ErrCannotRotateKey = errcode.Register(errGroup, errcode.ErrorDescriptor{
+		Value:          "CANNOT_ROTATE_KEY",
+		Message:        "Key has already been rotated recently.",
+		Description:    "The key has been rotated too recently, and cannot be rotated again at this time.",
```
endophage: This reads oddly to me. Maybe "Cannot rotate, last key rotation too recent."

server/errors/errors.go
```diff
@@ -83,8 +84,14 @@ var (
 	ErrNoKeyAlgorithm = errcode.Register(errGroup, errcode.ErrorDescriptor{
 		Value:          "NO_KEYALGORITHM",
 		Message:        "The server does not have a key algorithm configured.",
-		Description:    "No key algorihtm has been configured for the server and it has been asked to perform an operation that requires generation.",
+		Description:    "No key algorithm has been configured for the server and it has been asked to perform an operation that requires generation.",
 		HTTPStatusCode: http.StatusInternalServerError,
 	})
+	ErrCannotRotateKey = errcode.Register(errGroup, errcode.ErrorDescriptor{
```
endophage: Given there are lots of reasons a key rotation can fail, should the name be more specific like `ErrKeyRotationRateLimited`?

server/handlers/default.go
```diff
@@ -143,70 +143,95 @@ func DeleteHandler(ctx context.Context, w http.ResponseWriter, r *http.Request)
...
+	switch s.role { // parseKeyParams ensures it's only timestamp or snapshot
 	case data.CanonicalTimestampRole:
-		key, err = timestamp.GetOrCreateTimestampKey(gun, store, crypto, keyAlgorithm)
+		key, err = timestamp.GetOrCreateTimestampKey(s.gun, s.store, s.crypto, s.keyAlgo)
 	case data.CanonicalSnapshotRole:
-		key, err = snapshot.GetOrCreateSnapshotKey(gun, store, crypto, keyAlgorithm)
-	default:
-		logger.Errorf("400 GET %s key: %v", role, err)
-		return errors.ErrInvalidRole.WithDetail(role)
+		key, err = snapshot.GetOrCreateSnapshotKey(s.gun, s.store, s.crypto, s.keyAlgo)
```
endophage: The default case seems to have dropped off? Should be maintained. There's nothing stopped a client (or even a curl request) from attempting to rotate the root, targets, or delegation keys.

endophage: Ah, I see you moved it above to line 179. I'm not sure exactly what the implementation of server signing will look like but I feel like it's simpler to keep it here in the switch. Then manipulating which roles can be server side is better contained in this one place and adding/removing cases is sufficient to change which keys the server holds, and there isn't a conditional elsewhere to worry about.

server/handlers/default.go
```diff
@@ -143,70 +143,95 @@ func DeleteHandler(ctx context.Context, w http.ResponseWriter, r *http.Request)
...
-		return errors.ErrNoCryptoService.WithDetail(nil)
+		return nil, errors.ErrNoCryptoService.WithDetail("crypto service not configured")
 	}
+
 	algo := ctx.Value("keyAlgorithm")
 	keyAlgo, ok := algo.(string)
-	if !ok || keyAlgo == "" {
-		logger.Error("500 GET key algorithm not configured")
-		return errors.ErrNoKeyAlgorithm.WithDetail(nil)
+	if !ok || keyAlgo != data.ECDSAKey && keyAlgo != data.RSAKey && keyAlgo != data.ED25519Key {
```
endophage: I'd just stick with the original check and let the signer tell us if the requested algo is unsupported. This is just adding another place we have to remember to change if we tweak the signer. Especially as at any given time our signer actually only supports one of the algos.

endophage:
I don't see where the server is checking the client's signed request. Also seems like a partial server implementation that currently just returns the existing key (if there is one). That's fine if the server piece is coming separately but I'd like to see the signed request checking in this PR given that functionality is landing here.

client/client.go
diogomonica: I'm of the opposite opinion :D

client/client.go
```diff
@@ -829,28 +839,68 @@ func (r *NotaryRepository) validateRoot(rootJSON []byte) (*data.SignedRoot, erro
...
+		} else if _, ok := err.(signed.ErrNoKeys); ok {
+			err = fmt.Errorf("root signing key unavailable so unable to rotate key anyway")
```
diogomonica: remove "anyway".

tuf/store/httpstore.go
```diff
@@ -270,13 +275,54 @@ func (s HTTPStore) buildURL(uri string) (*url.URL, error) {
...
+func (s HTTPStore) requestKey(role string, body io.Reader) (data.PublicKey, error) {
 	url, err := s.buildKeyURL(role)
 	if err != nil {
 		return nil, err
 	}
-	req, err := http.NewRequest("GET", url.String(), nil)
+
+	method := "GET"
+	resource := role + " key"
+	if body != nil {
```
diogomonica: Not sure if we should have implicit behavior like this.

tuf/store/httpstore.go
```diff
@@ -270,13 +275,54 @@ func (s HTTPStore) buildURL(uri string) (*url.URL, error) {
...
+	req := &data.SignedCommon{
+		Type:    role,
+		Expires: time.Now().Add(5 * time.Minute),
+	}
+	reqJSON, err := json.Marshal(req)
+	if err != nil {
+		return nil, err
+	}
+	signedReq := &data.Signed{Signed: reqJSON}
+	if err := signed.Sign(cs, signedReq, roots...); err != nil {
```
diogomonica: @endophage signature happening here.

cmd/notary/keys.go
cyli: Filed https://github.com/docker/notary/issues/554 for it.  Basically am trying to keep this PR under 1k lines :D

tuf/store/interfaces.go
cyli: I agree that I wasn't super happy the arguments for `RotateKey` - perhaps a `KeyRotationHTTPStore` (or just `HTTPStore`) can take the `CryptoService`, etc., and can generate a key rotation request based on that?  Then the `RotateKey` parameters would just be the role as a string.

How do you feel about these functions returning `data.PublicKey` though instead of `[]byte`?  For metatdata especially, it makes sense for a remote store interface to return bytes, since we have to calculate the hash of those bytes.   But since no validation is done on the key bytes other than to make sure they're unmarshallable as keys, it seems anything that wants to use this interface ultimately cares about the keys, not the bytes themselves.  So it seems to make sense for the interface to return `data.PublicKey`?

cyli: NM, sorry, just reread https://github.com/docker/notary/pull/553#discussion_r52268867, which says they should return `[]byte` :)  

Hmm... If we were using GRPC, would you still say that `GetKey` and `RotateKey` should return `[]bytes` and not `data.PublicKey`?  For instance I'd think that `GetMeta` and `SetMeta` should still take and return `[]bytes`, since that's what TUF operates on (to do checksumming, etc.)

tuf/store/httpstore.go
endophage: yeah... this feels like completely the wrong place to do this. The stores were very cohesive before. Now suddenly they're responsible for signing and json marshalling? This is how we end up with confusing code where it's not clear what happens where.

server/handlers/default.go
cyli: Good point

client/client.go
riyazdf: I'd agree with @endophage for stronger messaging, since we similar language for the root rotation error message and we intend to support that

client/client_test.go
```diff
@@ -21,6 +21,7 @@ import (
 	"github.com/Sirupsen/logrus"
 	ctxu "github.com/docker/distribution/context"
 	"github.com/docker/go/canonical/json"
+	"github.com/gorilla/mux"
```
riyazdf: thanks for updating this :+1:

client/client_test.go
```diff
@@ -2567,6 +2570,52 @@ func testRotateKeySuccess(t *testing.T, serverManagesSnapshotInit bool,
...
+// entirely
+func TestRemoteRotationNonRateLimitError(t *testing.T) {
+	ts, _, _ := simpleTestServer(t)
+	defer ts.Close()
+
+	repo, _ := initializeRepo(t, data.ECDSAKey, "docker.com/notary", ts.URL, true)
+	defer os.RemoveAll(repo.baseDir)
+
+	// simpleTestServer has no rotate key endpoint, so this should fail
+	assert.Error(t, repo.RotateKey(data.CanonicalTimestampRole, true))
```
riyazdf: Can we assert the type of error here?

cmd/notary/keys_test.go
```diff
@@ -227,11 +233,10 @@ func TestRemoveMultikeysRemoveOnlyChosenKey(t *testing.T) {
 	}
 }
 
-// Non-roles, root, and timestamp can't be rotated
+// Non-roles, and root can't be rotated
```
riyazdf: I think it'd be good to also document delegation key behavior here

server/errors/errors.go
riyazdf: +1 for @endophage's suggestion

tuf/store/httpstore.go
cyli: I guess it depends what you think of the signed request as.  If it should be part of any rotate key API, no matter if we implement a GRPC remote API, an in-memory API, etc, that all requires a signed request, and (and maybe in fact clients should never be able to rotate any key, even locally, without having the root key on hand?), then I agree it should probably go into the client.

If we are thinking of this as just something our HTTP server requires, as in it's part of the server's API and not a generic `RemoteStore` API, then it should probably live in the client for the notary server, which seems to be `HTTPStore`, since that's the thing which knows how to speak the server API.

server/handlers/default.go
cyli: I don't mind moving it back, but I was going to have a `rotateKeyHandler`, which would need to perform the same check, so if we want it in a single location, should it be in `parseKeys`?  `https://github.com/docker/notary/pull/534/files#diff-b60e09f129d59867e23c3c6e1e94319cR159` was eventually supposed to be how this would look.

I can also just move it back to `getKeyHandler` for now, and when/if that change lands, move it to `parseKeyParams`

tuf/store/interfaces.go
endophage: If we were using GRPC there would have to be translation between the wire types (GRPC would require us to define our own protos), and the TUF types. The manually created part of the server and signer code would do that translation and sits a level higher than the "store" code.

So to answer you first question, I would definitely be happier to see something wrapped around the HTTPStore that could do the signing/validation.

tuf/store/httpstore.go
endophage: Yeah, the other store methods take or return a complete body (ok, for SetMulti it takes a map of part name to body). This method has to actually construct a body from the parameters it's passed.

If the signing was in an HTTP header or something more generic that could be included on any request, I'd feel better about putting it into the store.

server/handlers/default.go
endophage: Switches without a default make me look twice, and are more concerning when dealing with user input. Having the default here makes it obvious what the valid input is for this flow. Having the check in parseKeyParams leaves it open to dropping off if somebody isn't paying attention.

tuf/store/interfaces.go
cyli: Hm, so I can see HTTPStore being just `[]bytes`, but do you think the interface should also just be `[]bytes`?  There can be something else between `HTTPStore` and the client, and `HTTPStore` doesn't have to definitely implement the interface.  Alternately, maybe `HTTPStore` can just implement the metadata part and something else that wraps it can implement the signing/unmarshalling?

cyli:
@endophage Yep I was hoping to do the server part in a separate PR, because it's huge :(  And this is already huge.  Will be happy to add the server validation of the request, though - hopefully that shouldn't add to much.

diogomonica:
@cyli rebase needed.

cyli:
@diogomonica will do - I think in order to decrease the size of this I'm going to address the CLI changes here and #554 first in a separate PR :)

cyli:
Closing in favor of https://github.com/docker/notary/pull/889

PR #555, opened 2016-02-10
Remove keyDB from signed.Verify

cyli:
The start of removing KeyDB.  This introduces a `RoleWithKeys` object, which `Verify` uses to verify signatures, instead of a `KeyDB`.

Part of #556

### DIFF
--- a/tuf/client/client.go
+++ b/tuf/client/client.go
@@ -204,8 +204,12 @@ func (c Client) verifyRoot(role string, s *data.Signed, minVersion int) error {
 	// Still need to determine if there has been a root key update and
 	// confirm signature with new root key
 	logrus.Debug("verifying root with existing keys")
-	err := signed.Verify(s, role, minVersion, c.keysDB)
+	rootRole, err := c.local.GetRole(role)
 	if err != nil {
+		logrus.Debug("no previous root role loaded")
+		return err
+	}
+	if err = signed.Verify(s, rootRole, minVersion); err != nil {
 		logrus.Debug("root did not verify with existing keys")
 		return err
 	}
@@ -227,7 +231,12 @@ func (c Client) verifyRoot(role string, s *data.Signed, minVersion int) error {
 	// TODO(endophage): be more intelligent and only re-verify if we detect
 	//                  there has been a change in root keys
 	logrus.Debug("verifying root with updated keys")
-	err = signed.Verify(s, role, minVersion, c.keysDB)
+	rootRole, err = c.local.GetRole(role)
+	if err != nil {
+		logrus.Debug("root role with new keys not loaded")
+		return err
+	}
+	err = signed.Verify(s, rootRole, minVersion)
 	if err != nil {
 		logrus.Debug("root did not verify with new keys")
 		return err
@@ -267,7 +276,7 @@ func (c *Client) downloadTimestamp() error {
 	// from remote, only using the cache one if we couldn't reach remote.
 	raw, s, err := c.downloadSigned(role, notary.MaxTimestampSize, nil)
 	if err == nil {
-		ts, err = c.verifyTimestamp(s, version, c.keysDB)
+		ts, err = c.verifyTimestamp(s, version)
 		if err == nil {
 			logrus.Debug("successfully verified downloaded timestamp")
 			c.cache.SetMeta(role, raw)
@@ -282,7 +291,7 @@ func (c *Client) downloadTimestamp() error {
 	}
 	logrus.Debug(err.Error())
 	logrus.Warn("Error while downloading remote metadata, using cached timestamp - this might not be the latest version available remotely")
-	ts, err = c.verifyTimestamp(old, version, c.keysDB)
+	ts, err = c.verifyTimestamp(old, version)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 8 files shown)


tuf/signed/verify.go
```diff
@@ -57,18 +56,18 @@ func VerifyRoot(s *data.Signed, minVersion int, keys map[string]data.PublicKey)
...
 // Verify checks the signatures and metadata (expiry, version) for the signed role
 // data
-func Verify(s *data.Signed, role string, minVersion int, db *keys.KeyDB) error {
-	if err := verifyMeta(s, role, minVersion); err != nil {
+func Verify(s *data.Signed, role *data.RoleWithKeys, minVersion int) error {
+	if err := verifyMeta(s, role.Name, minVersion); err != nil {
 		return err
 	}
-	return VerifySignatures(s, role, db)
+	return VerifySignatures(s, role)
```
diogomonica: When would we ever call `VerifySignatures()` without calling `verifyMeta` first? Shouldn't `verifyMeta` be a part of `VerifySignatures`?

diogomonica:
This looks good. I like the construct, and see the path to remove the keyDB. Can't wait for the next PR.

LGTM.

tuf/signed/verify.go
cyli: It was previously called by some unused code, which I deleted.  I thought `Verify` would probably go away at some point because verifying the metadata would be moved to `tuf.Repo.CheckAndSet_*`, and those would then call `VerifySignatures` and we can remove `Verify` entirely.

I'm happy to begin the merging process now, but it seems like it would be temporarily strangely named (since `VerifySignatures` would be verifying the metadata as well as the signatures)

cyli: I can also add a comment to `verifyMeta` and `Verify` saying that they'll probably be removed soon, too.  Either way works for me.

tuf/client/client.go
```diff
@@ -204,8 +204,12 @@ func (c Client) verifyRoot(role string, s *data.Signed, minVersion int) error {
 	// Still need to determine if there has been a root key update and
```
riyazdf: We should update the comment in the line above this one, which references `c.keysDB`

tuf/signed/verify.go
```diff
@@ -57,18 +56,18 @@ func VerifyRoot(s *data.Signed, minVersion int, keys map[string]data.PublicKey)
 			continue
 		}
 		// threshold of 1 so return on first success
-		return verifyMeta(s, "root", minVersion)
+		return verifyMeta(s, data.CanonicalRootRole, minVersion)
```
riyazdf: :+1: thanks for fixing this

tuf/signed/verify.go
riyazdf: I agree with phasing out `Verify`, it may be a little more difficult to delete `verifyMeta` since we also use it in `VerifyRoot`.  I'm fine with doing this now or in a separate PR, whichever you're more comfortable with.

tuf/signed/verify.go
```diff
@@ -153,28 +149,3 @@ func VerifySignatures(s *data.Signed, role string, db *keys.KeyDB) error {
...
-func UnmarshalTrusted(b []byte, v interface{}, role string, db *keys.KeyDB) error {
-	s := &data.Signed{}
-	if err := json.Unmarshal(b, s); err != nil {
-		return err
-	}
-	if err := VerifySignatures(s, role, db); err != nil {
-		return err
-	}
-	return json.Unmarshal(s.Signed, v)
-}
```
riyazdf: this is awesome

tuf/signed/verify_test.go
```diff
@@ -24,17 +23,15 @@ func TestRoleNoKeys(t *testing.T) {
 		nil,
 	)
 	assert.NoError(t, err)
-	db := keys.NewDB()
-	assert.NoError(t, err)
-	err = db.AddRole(r)
-	assert.NoError(t, err)
+	roleWithKeys := &data.RoleWithKeys{Role: *r, Keys: map[string]data.PublicKey{k.ID(): k}}
```
riyazdf: nit: I think we can also use `data.Keys` as a type here

tuf/tuf.go
```diff
@@ -771,3 +771,19 @@ func (tr Repo) sign(signedData *data.Signed, role data.Role) (*data.Signed, erro
 	}
 	return signedData, nil
 }
+
+// GetRole returns a RoleWithKeys object, given a role name.
+func (tr Repo) GetRole(role string) (*data.RoleWithKeys, error) {
```
riyazdf: nit (especially since this may change), but it might be nice to name this `GetRoleWithKeys`

riyazdf:
this is great, LGTM after comments

tuf/signed/verify.go
diogomonica: Let's add this as a comment in the corresponding issue @cyli. Just so we don't forget.

tuf/signed/verify_test.go
cyli: Definitely!  I keep forgetting about these collection types :D  Thanks!

tuf/signed/verify.go
cyli: Done!  https://github.com/docker/notary/issues/561#issuecomment-182972043

tuf/client/client.go
cyli: Done!  :)  (edit in the line below)

PR #566, opened 2016-02-11
Role types with keys, remove keyDB

### LINKED ISSUE #556: Remove KeyDB
- Creating GetBaseRoles, GetDelegatedRole
- New Role types that carry keys

riyazdf:
Introduces `BaseRole` and `DelegationRole` types, which include the public keys for the roles.  This should help us move in the direction toward deprecating keysDB.

I also added `GetBaseRole` and `GetDelegationRole` to tufRepo, which reads from TUF metadata directly instead of from the keysDB.  There is some validation involved, particularly with role names and paths.

`BaseRole` and `DelegationRole` also implement a `TufRole` interface, to make it easy to pass them around.  I didn't remove `RootRole` or `Role` for now, since we can use those for direct serialization/deserialization with TUF metadata.

Update: with David's commit, this PR fully removes keyDB

Closes #556


tuf/tuf.go
```diff
@@ -173,6 +173,132 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
...
 }
 
+// GetBaseRole gets a base role from this repo's metadata
+func (tr *Repo) GetBaseRole(name string) (data.BaseRole, error) {
+	if !data.IsBaseRole(name) {
+		return data.BaseRole{}, fmt.Errorf("Invalid base role name specified: %s", name)
+	}
+	roleData, ok := tr.Root.Signed.Roles[name]
+	if !ok {
+		return data.BaseRole{}, fmt.Errorf("TUF data for %s does not exist in root metadata", name)
```
cyli: Non-blocking - I'm not sure if this is the right error, but would `ErrNotLoaded` be useful for this?  We seem to currently be using it for returning an error if the actual `SignedX` stuff isn't loaded, but the error message says "role".

tuf/tuf.go
```diff
@@ -173,6 +173,132 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
...
+	}
+	// Get all public keys for the base role from TUF metadata
+	keyIDs := roleData.KeyIDs
+	pubKeys := make(map[string]data.PublicKey)
+	for _, keyID := range keyIDs {
+		pubKey, ok := tr.Root.Signed.Keys[keyID]
+		if !ok {
+			return data.BaseRole{}, fmt.Errorf("key %s does not exist in root metadata", keyID)
```
cyli: Nitpick: Could we add the role to the error message?  Maybe something like `"key with ID %s, specified as a signing key for role %s, was not found in root metadata"`?

tuf/tuf.go
```diff
@@ -173,6 +173,132 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
 	return nil
 }
 
+// GetBaseRole gets a base role from this repo's metadata
+func (tr *Repo) GetBaseRole(name string) (data.BaseRole, error) {
+	if !data.IsBaseRole(name) {
+		return data.BaseRole{}, fmt.Errorf("Invalid base role name specified: %s", name)
```
cyli: Non-blocking - I'm not sure if this applies here, but `NotaryRepository` uses `data.ErrInvalidRole` whenever a role is supplied with an invalid name - not sure if that'd be a useful error type to return here and in `GetDelegationRole`

tuf/tuf.go
cyli: Similarly for `GetDelegationRole`

tuf/tuf.go
```diff
@@ -173,6 +173,132 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
...
+			Keys:      pubKeys,
+			Threshold: foundRole.Threshold,
+		},
+		Paths:            foundRole.Paths,
+		PathHashPrefixes: foundRole.PathHashPrefixes,
+	}, nil
+}
+
+// Returns true if the delegationPaths are prefixed by parentPaths
+func validateDelegationPathPrefixes(parentPaths, delegationPaths []string) bool {
```
cyli: :+1: Awesome, this will be super useful, thanks for fixing the path prefixing stuff too!

tuf/tuf.go
riyazdf: agreed!  Will update

tuf/tuf.go
riyazdf: I think it's a fitting error, good idea :)

tuf/tuf.go
riyazdf: great idea, will do!

tuf/tuf.go
```diff
@@ -173,6 +173,132 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
 	return nil
 }
 
+// GetBaseRole gets a base role from this repo's metadata
+func (tr *Repo) GetBaseRole(name string) (data.BaseRole, error) {
+	if !data.IsBaseRole(name) {
+		return data.BaseRole{}, fmt.Errorf("Invalid base role name specified: %s", name)
+	}
+	roleData, ok := tr.Root.Signed.Roles[name]
```
cyli: Should we check to see if `tr.Root == `nil` first?  Otherwise this might panic.  Similarly for `GetDelegationRole`

tuf/data/interface.go
```diff
@@ -0,0 +1,28 @@
+package data
+
+// TufRole is an interface for dealing with TUF roles internally
+type TufRole interface {
+	// Returns true for BaseRole
+	IsBaseRole() bool
```
diogomonica: This seems weird. Shouldn't we simply rely on casting to verify if a Role is a of a specific type? Seems like the go way, and would remove `IsBaseRole()` and `IsDelegationRole()` here and in BaseRole and DelegationRole.

endophage: I'm with @diogomonica on this one. c.f. how public key types work.

tuf/data/roles.go
```diff
@@ -68,6 +68,16 @@ func ValidRole(name string) bool {
 	return false
 }
 
+// IsBaseRole checks if a role name is valid base role
+func IsBaseRole(name string) bool {
```
diogomonica: I would rather have a serialization method of a String to a Role Type that includes this logic.

cyli:
Would be awesome to see the following test cases:
- invalid path hash prefix
- testing that empty paths are valid subpaths (currently empty path prefixes are tested)
- calling `GetBaseRole` with a delegation role or invalid role
- calling `GetDelegationRole` with a base role or invalid role

Non-blocking:  I was also wondering if we do want to validate all the paths, parentage, etc. when getting the role?  If we make the repo modifiable with only `CheckAndSet_*` functions, and all the structure is unexported, it seems like the validation would happen there?  I totally get that we're not doing that yet, though, and it makes sense to do checking in the current in-between state.

endophage:
The `TUFRole` interface is probably a little too restrictive. The only time I see it being used is during signing/verification when we need to get the keys. In all other cases I think we have sufficient context that we should be using the concrete types.

tuf/data/interface.go
riyazdf: good point about casting being more of a golang way of doing things

I think the only place we may want this shared interface other than signing (where we have good amount of context, as @endophage commented) is when dealing with targets/delegations since a good amount of that code is intertwined, but casting should be able to work fine.  I'll update this

tuf/tuf.go
riyazdf: Weird that this got collapsed, but definitely -- great catch.  Will update

tuf/tuf_test.go
```diff
@@ -953,3 +953,143 @@ func TestGetAllRoles(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	testKey2, err := ed25519.Create("targets/test/b", data.ED25519Key)
+	assert.NoError(t, err)
+	// Now we add a delegation with a path that is not prefixed by its parent delegation
+	role, err = data.NewRole(
+		"targets/test/b", 1, []string{testKey2.ID()}, []string{"invalidpath"}, []string{})
+	assert.NoError(t, err)
+
+	err = repo.UpdateDelegations(role, data.KeyList{testKey2})
```
cyli: Non-blocking question:  This PR may not be the one to do it in - but were we going to have `UpdateDelegations` fail if the paths for the role are not subpaths of the current role?  If so, these tests would probably not work.

riyazdf:
@cyli thanks for the test ideas, I definitely think those would be great to add -- I'll let you know when they've made it into the PR.

As for validation -- since we're in an in-between-state I think it's a little hard to say, but I think we had originally envisioned for some validation on these "Get" functions.  I'm open to opinions, but I'd rather just have the validation in this PR than take it out from the codebase entirely

tuf/tuf_test.go
riyazdf: ah this is related to what I just commented -- yeah I think it'd be better to have this in a separate PR.

It's actually an interesting case because this method doesn't use a `Set...` from the tuf repo.  This function could probably benefit from being refactored to use a `CheckAndSet...` function

cyli:
@riyazdf Sorry, no, I didn't mean we should remove it from this PR!  I was just wondering if the intention was to move these into the `CheckAndSet_*` functions later.

riyazdf:
@cyli ah yes, yeah we should definitely keep it and use it to `CheckAndSet_*` later :+1:

endophage:
@cyli @riyazdf re. validation in `CheckAndSet___`, it's valid for a role to contain content it shouldn't (mainly targets it's not allowed to sign for). It's an expected scenario when the delegation or a parent is changed such that it can no longer sign a particular path. Those files should still pass validation but when searching for a target (or listing targets), the invalid targets should be excluded.

cyli:
@endophage ah, right, I forgot about the case where the parent was changed to be more restrictive.  thanks!

tuf/data/roles.go
riyazdf: I'm not sure if I understand exactly what you mean -- like a function that takes a string and outputs the corresponding Role type (or error)?

riyazdf:
@cyli added tests!

@endophage ah yes, I'll update `validateDelegationPathPrefixes` to respect this behaivor

tuf/data/roles.go
```diff
@@ -101,6 +113,111 @@ type Role struct {
 	Email            string   `json:"email,omitempty"`
 }
 
+// BaseRole is an internal representation of a root/targets/snapshot/timestamp role, with its public keys included
+type BaseRole struct {
```
diogomonica: Did we decide that we wanted one Role type for root/targets/snapshot/timestamp?

tuf/data/roles.go
```diff
@@ -101,6 +113,111 @@ type Role struct {
...
+	Keys      map[string]PublicKey `json:"keys"`
+	Name      string               `json:"name"`
+	Threshold int                  `json:"threshold"`
+}
+
+// ListKeys retrieves the public keys valid for this role
+func (b BaseRole) ListKeys() KeyList {
```
diogomonica: What was the rationale for `KeyList` vs `[]PublicKey?`

tuf/data/roles.go
```diff
@@ -101,6 +113,111 @@ type Role struct {
...
+// DelegationRole is an internal representation of a delegation role, with its public keys included
+type DelegationRole struct {
+	BaseRole
+	Paths            []string `json:"paths,omitempty"`
+	PathHashPrefixes []string `json:"path_hash_prefixes,omitempty"`
+}
+
+// ListKeys retrieves the public keys valid for this role
+func (d DelegationRole) ListKeys() KeyList {
+	keys := KeyList{}
```
diogomonica: worth it to do a `listKeys()` to reuse between both `ListKeys()`?

tuf/data/roles.go
```diff
@@ -101,6 +113,111 @@ type Role struct {
...
+	for _, key := range d.Keys {
+		keys = append(keys, key)
+	}
+	return keys
+}
+
+// ListKeyIDs retrieves the list of key IDs valid for this role
```
diogomonica: Same comment as above.

tuf/data/roles.go
riyazdf: After talking with @cyli and @endophage, I think the plan is to leave these as-is for this PR.  These will be an intermediate step before we introduce individual root/targets/snapshot/timestamp roles with the tree walker

tuf/data/roles.go
riyazdf: `KeyList` is defined as `type KeyList []PublicKey` in `keys.go`, so it seemed nice to reuse

tuf/data/roles.go
riyazdf: great idea.  Will do

tuf/tuf_test.go
```diff
@@ -953,3 +953,245 @@ func TestGetAllRoles(t *testing.T) {
...
+	testKey1, err := ed25519.Create("targets/test", data.ED25519Key)
+	assert.NoError(t, err)
+	role, err := data.NewRole(
+		"targets/test", 1, []string{testKey1.ID()}, []string{}, []string{"pathhash", "anotherpathhash"})
+	assert.NoError(t, err)
+
+	err = repo.UpdateDelegations(role, data.KeyList{testKey1})
+	assert.NoError(t, err)
+
+	testKey2, err := ed25519.Create("targets/test/b", data.ED25519Key)
```
cyli: Do we want to assert here that `target/tests`'s Role, as returned by `GetDelegationRole` has path hashes "pathhash" and "anotherpathpash"?

tuf/tuf_test.go
```diff
@@ -953,3 +953,245 @@ func TestGetAllRoles(t *testing.T) {
...
+	testKey1, err := ed25519.Create("targets/test", data.ED25519Key)
+	assert.NoError(t, err)
+	role, err := data.NewRole(
+		"targets/test", 1, []string{testKey1.ID()}, []string{"path", "anotherpath"}, []string{})
+	assert.NoError(t, err)
+
+	err = repo.UpdateDelegations(role, data.KeyList{testKey1})
+	assert.NoError(t, err)
+
+	testKey2, err := ed25519.Create("targets/test/b", data.ED25519Key)
```
cyli: Do we want to assert here that `targets/test`'s Role, as returned by `GetDelegationRole` has paths "path" and "anotherpath"?  (to as an assertion that that we don't filter the top level delegation paths at all).

tuf/data/roles.go
```diff
@@ -101,6 +113,111 @@ type Role struct {
...
+
+// DelegationRole is an internal representation of a delegation role, with its public keys included
+type DelegationRole struct {
+	BaseRole
+	Paths            []string `json:"paths,omitempty"`
+	PathHashPrefixes []string `json:"path_hash_prefixes,omitempty"`
+}
+
+// ListKeys retrieves the public keys valid for this role
+func (d DelegationRole) ListKeys() KeyList {
```
cyli: I'm think these function definitions necessary?   Since `DelegationRole` embeds `BaseRole`, calling `d.ListKeys()` [should invoke the embedded `BaseRole`'s `ListKeys()`](https://golang.org/doc/effective_go.html#embedding), which is the correct behavior for `DelegationRole` since the keys are on `BaseRole` anyway.

tuf/tuf.go
```diff
@@ -173,6 +173,112 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
...
+			// If we found the role, we can exit the loop
+			if delgRole.Name == name {
+				foundRole = delgRole
+				break
+			}
+
+			// If this is a parent role, keep traversing
+			whiteListedPaths = delgRole.Paths
+			whiteListedPathHashes = delgRole.PathHashPrefixes
+			delegationRoles = append(delegationRoles, tr.Targets[delgRole.Name].Signed.Delegations.Roles...)
```
cyli: We check that the parent role's metadata must exist in later, on [https://github.com/docker/notary/pull/566/files#diff-b7b7bae92e6be55133dab8ee3256e014R260](line 260), but shouldn't we also check here?  If the metadata doesn't exist for the role, will this panic?

tuf/tuf.go
```diff
@@ -173,6 +173,112 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
...
+	}
+	// Traverse target metadata, down to delegation itself
+	// Get all public keys for the base role from TUF metadata
+	signedTargetData, ok := tr.Targets[data.CanonicalTargetsRole]
+	if !ok {
+		return data.DelegationRole{}, fmt.Errorf("TUF data for targets does not exist in targets metadata", name)
+	}
+
+	// Keep track of paths and path hash prefixes to restrict as we traverse the delegations tree, targets implicitly has ""
+	whiteListedPaths := []string{""}
```
cyli: It looks like we are pruning paths that shouldn't exist, as we go?  This means that the in-memory data structure contain only the valid paths for the role, which makes sense if we want to list targets, etc.  But would this mean that if a target was added, or any other changes were made, to the role, then a possibly unintended paths/paths-prefix change would be published when the user publishes whatever their change is?  Since when we publish we re-serialize this in-memory structure.  Is that behavior that we want?

tuf/data/roles.go
cyli: :+1: This is an awesome place to put these utilities instead of the interface :D

tuf/data/roles.go
```diff
@@ -101,6 +113,111 @@ type Role struct {
...
+		Paths:            RestrictDelegationPathPrefixes(d.Paths, child.Paths),
+		PathHashPrefixes: RestrictDelegationPathPrefixes(d.PathHashPrefixes, child.PathHashPrefixes),
+	}, nil
+}
+
+// IsParentOf returns whether the passed in delegation role is the direct child of this role,
+// determined by delegation name.
+// Ex: targets/a is a direct parent of targets/a/b, but targets/a is not a direct parent of targets/a/b/c
+func (d DelegationRole) IsParentOf(child DelegationRole) bool {
+	// Check that the child has the prefix
```
cyli: Totally non-blocking:  would `return path.Dir(child.Name) == d.Name` work?

tuf/tuf_test.go
riyazdf: great idea, done :)

tuf/tuf_test.go
riyazdf: done!

tuf/data/roles.go
endophage: Weren't we going to get rid of this in favour of type checking: `_, ok := r.(BaseRole)`

tuf/data/roles.go
```diff
@@ -69,6 +69,16 @@ func ValidRole(name string) bool {
...
+	for _, v := range BaseRoles {
+		if name == v {
+			return true
+		}
+	}
+	return false
+}
+
 // IsDelegation checks if the role is a delegation or a root role
 func IsDelegation(role string) bool {
```
endophage: Now that we have a specific type for delegated roles, we should remove this and replace with a type check too.

tuf/data/roles.go
endophage: Further design work: we're sticking with `BaseRole` and `DelegationRole`, although in our PR we may rename `DelegationRole` to `TargetRole` and the actual targets role will be represented using that with a `Path: [""]` (can sign all paths). That makes walking of all the targets files uniform.

tuf/data/roles.go
endophage: `KeyList` exists purely for deserialization purposes because `PublicKey` is an interface so we can't deserialize directly to it. Instead we can define a custom unmarshaller on `KeyList` that is able to translate each incoming key into the correct type and assign it into the underlying `[]PublicKey`

tuf/data/roles.go
```diff
@@ -101,6 +113,94 @@ type Role struct {
...
+	keyIDs := []string{}
+	for id := range keyMap {
+		keyIDs = append(keyIDs, id)
+	}
+	return keyIDs
+}
+
+// RestrictChild restricts the paths and path hash prefixes for the passed in delegation role,
+// returning a copy of the role with validated paths as if it was a direct child
+func (d DelegationRole) RestrictChild(child DelegationRole) (DelegationRole, error) {
```
endophage: Can probably just be `Restrict`. Makes the function signature (and possibly callers) read better: `parent.Restrict(child)`

tuf/data/roles.go
riyazdf: Maybe a rename is more fitting?  I agree that we'll use type checking, but this method shouldn't be used check a role itself.  I'm think we should keep it for input validation for delegation names before we create or operate on typed roles, like on notary CLI.

tuf/data/roles.go
riyazdf: great idea, will do

riyazdf:
in the latest commit, I was able to replace `RoleWithKeys` with `GetBaseRole`/`GetDelegationRole`, such that we always traverse TUF metadata in the repo when fetching roles for verification instead of using the keyDB.  This should move us really close to getting rid of keyDB entirely.

server/storage/memory.go
```diff
@@ -84,7 +84,9 @@ func (st *MemStorage) GetCurrent(gun, role string) (data []byte, err error) {
 func (st *MemStorage) GetChecksum(gun, role, checksum string) (data []byte, err error) {
 	st.lock.Lock()
 	defer st.lock.Unlock()
-	data, ok := st.checksums[gun][checksum]
+	id := entryKey(gun, role)
+	_, ok := st.tufMeta[id]
```
cyli: Should there be an `ok` check here, or should it be assigned to a different variable?  It looks like the value would just get overwritten by the next line.

riyazdf: I'll clean this, you're right that we don't need this code

tuf/signed/verify_test.go
```diff
@@ -15,15 +15,8 @@ func TestRoleNoKeys(t *testing.T) {
 	cs := NewEd25519()
 	k, err := cs.Create("root", data.ED25519Key)
 	assert.NoError(t, err)
-	r, err := data.NewRole(
```
cyli: \o/ so much deletion!  :D

cyli:
LGTM!  Thanks for replacing `RoleWithKeys` everywhere!!

endophage:
I'm going to be pushing additional commits to this PR to get the KeyDB completely out of the system. I'm wary of merging a half removed KeyDB as it leaves too many possibilities to do something wrong.

endophage:
@riyazdf @cyli @diogomonica @NathanMcCauley apologies for how large this PR is now, but it is marginally net negative :-)

Things I've done in my additional commit:

- Fully remove KeyDB, which uncovered some interesting side effects we were relying on (in tests) and some bugs related to changed error types in Riyaz's initial work that hadn't been encountered yet due to some pieces relying on KeyDB still, and other pieces using the new GetBaseRole and GetDelegationRole code.
- Removed the PathHashPrefix stuff while I was in there fiddling with roles. I don't see us ever exposing it to users because it's far too difficult to explain to most people and the problem it solves (sharding large delegations) can be solved in other ways.

client/helpers_test.go
```diff
@@ -662,136 +662,8 @@ func TestApplyTargetsChangeInvalidType(t *testing.T) {
 	assert.Error(t, err)
 }
 
-// A delegated role MUST NOT have both Paths and PathHashPrefixes defined.
-// These next 2 tests check that attempting to edit an existing role to
-// create an invalid role errors in both possible combinations.
-func TestApplyTargetsDelegationConflictPathsPrefixes(t *testing.T) {
```
endophage: This test is no longer relevant as PathHasPrefixes have been removed and this tests that we error when adding hash prefixes to a role that already has paths.

client/helpers_test.go
```diff
@@ -662,136 +662,8 @@ func TestApplyTargetsChangeInvalidType(t *testing.T) {
...
-		changelist.TypeTargetsDelegation,
-		"",
-		tdJSON,
-	)
-
-	err = applyTargetsChange(repo, ch)
-	assert.Error(t, err)
-}
-
-func TestApplyTargetsDelegationConflictPrefixesPaths(t *testing.T) {
```
endophage: Test is no longer relevant as it tests adding Paths to a role that already has hash prefixes.

client/helpers_test.go
```diff
@@ -662,136 +662,8 @@ func TestApplyTargetsChangeInvalidType(t *testing.T) {
...
-		changelist.TypeTargetsDelegation,
-		"",
-		tdJSON,
-	)
-
-	err = applyTargetsChange(repo, ch)
-	assert.Error(t, err)
-}
-
-func TestApplyTargetsDelegationCreateInvalid(t *testing.T) {
```
endophage: Test is no longer relevant as it tests creating a role that has both paths and hash prefixes.

tuf/data/roles.go
```diff
@@ -86,31 +86,139 @@ func IsDelegation(role string) bool {
...
+// Eventually should only be used for immediately before and after serialization/deserialization
 type Role struct {
 	RootRole
-	Name             string   `json:"name"`
-	Paths            []string `json:"paths,omitempty"`
-	PathHashPrefixes []string `json:"path_hash_prefixes,omitempty"`
-	Email            string   `json:"email,omitempty"`
+	Name  string   `json:"name"`
+	Paths []string `json:"paths,omitempty"`
+	Email string   `json:"email,omitempty"`
```
cyli: Non-blocking - while we're removing stuff, should we just remove this too?  It doesn't seem like it's part of the spec?

server/handlers/validation_test.go
```diff
@@ -19,19 +18,18 @@ import (
...
 )
 
-func copyTimestampKey(t *testing.T, fromKeyDB *keys.KeyDB,
+func copyTimestampKey(t *testing.T, fromRepo *tuf.Repo,
 	toStore storage.MetaStore, gun string) {
 
-	role := fromKeyDB.GetRole(data.CanonicalTimestampRole)
+	role, err := fromRepo.GetBaseRole(data.CanonicalTimestampRole)
+	assert.NoError(t, err)
 	assert.NotNil(t, role, "No timestamp role in the KeyDB")
```
riyazdf: we should change the message here to not reference keydb

tuf/client/client_test.go
```diff
@@ -14,33 +14,35 @@ import (
...
+	originalRoot, err := data.NewRoot(
+		map[string]data.PublicKey{rootKey.ID(): rootKey},
+		map[string]*data.RootRole{"root": &rootRole.RootRole},
+		false,
+	)
+
+	repo.Root = originalRoot
 
 	// Generate new key and role. These will appear in the root.json
 	// but will not be added to the keyDB.
```
riyazdf: we should remove the keydb reference here

tuf/client/client_test.go
```diff
@@ -75,27 +77,30 @@ func TestRotation(t *testing.T) {
...
```
riyazdf: remove keydb reference

tuf/client/client_test.go
```diff
@@ -136,27 +141,30 @@ func TestRotationNewSigMissing(t *testing.T) {
...
```
riyazdf: same as above

tuf/tuf.go
```diff
@@ -417,27 +493,6 @@ func (tr *Repo) InitTimestamp() error {
 // the keys and roles in the KeyDB, and sets the Repo.Root field
```
riyazdf: remove keydb reference

tuf/tuf.go
```diff
@@ -460,12 +515,6 @@ func (tr *Repo) SetSnapshot(s *data.SignedSnapshot) error {
 // reads the delegated roles and keys into the KeyDB, and sets the
```
riyazdf: remove keydb

tuf/client/client.go
```diff
@@ -21,16 +20,14 @@ import (
 type Client struct {
 	local  *tuf.Repo
 	remote store.RemoteStore
-	keysDB *keys.KeyDB
 	cache  store.MetadataStore
 }
 
 // NewClient initialized a Client with the given repo, remote source of content, key database, and cache
```
cyli: Should we remove the reference to the key database here?

tuf/data/roles.go
```diff
@@ -86,31 +86,139 @@ func IsDelegation(role string) bool {
 		isClean
 }
 
+// BaseRole is an internal representation of a root/targets/snapshot/timestamp role, with its public keys included
+type BaseRole struct {
+	Keys      map[string]PublicKey `json:"keys"`
```
cyli: Non-blocking question - are these ever going to be serialized?  Should we leave out the JSON tags to make it clear that this is a storage struct, not a serialization struct?

tuf/data/roles.go
```diff
@@ -86,31 +86,139 @@ func IsDelegation(role string) bool {
...
-		return nil, ErrInvalidRole{
-			Role:   name,
-			Reason: "roles may not have both Paths and PathHashPrefixes",
-		}
-	}
+func NewRole(name string, threshold int, keyIDs, paths []string) (*Role, error) {
 	if IsDelegation(name) {
-		if len(paths) == 0 && len(pathHashPrefixes) == 0 {
+		if len(paths) == 0 {
 			logrus.Debugf("role %s with no Paths and no PathHashPrefixes will never be able to publish content until one or more are added", name)
```
cyli: Should we remove the reference to `PathHashPrefixes` in this comment?

tuf/data/roles.go
endophage: Yes, we should take those out. I'll do that now.

tuf/data/roles.go
endophage: Huh... yes we should, I swear I deleted that already. Maybe a similar comment existed elsewhere. Removing

tuf/data/targets.go
```diff
@@ -59,22 +57,11 @@ func (t SignedTargets) GetMeta(path string) *FileMeta {
 // order to determine priority.
 func (t SignedTargets) GetDelegations(path string) []*Role {
```
cyli: Non-blocking - I think nothing actually calls this - can this just be removed or will we need it for a future use?

tuf/data/roles.go
endophage: @NathanMcCauley @diogomonica I think we added this so a user/contact email could be added against a delegation then never used it. I'm not sure how useful that is and may bring a lot of "information discovery" type security reports. Do you have any objections to removing it?

tuf/keys/db.go
```diff
@@ -1,78 +0,0 @@
-package keys
```
cyli: \o/

tuf/data/roles.go
diogomonica: We should probably support at least an optional delegation description field of some kind.

cyli: Would it make sense to add in a different optional description field when we add support for setting it/displaying it?  Taking out the field in the struct would cause any problems if the field happened to be in the JSON at a later point.

tuf/testutils/repo.go
```diff
@@ -43,23 +42,34 @@ func createKey(cs signed.CryptoService, gun, role string) (data.PublicKey, error
 // EmptyRepo creates an in memory key database, crypto service
```
cyli: Should we remove the reference to the key database since we already mention a crypto service?

certs/certs_test.go
```diff
@@ -280,7 +280,7 @@ func testValidateSuccessfulRootRotation(t *testing.T, keyAlg, rootKeyType string
 	origRootKey := data.NewPublicKey(rootKeyType, origRootPEMCert)
 	replRootKey := data.NewPublicKey(rootKeyType, replRootPEMCert)
 
-	rootRole, err := data.NewRole("root", 1, []string{replRootKey.ID()}, nil, nil)
+	rootRole, err := data.NewRole("root", 1, []string{replRootKey.ID()}, nil)
```
diogomonica: `data.CanonicalRootRole`?

tuf/data/targets.go
endophage: Possibly with some updates I'm intending to use it for the tree walking.

tuf/data/roles.go
```diff
@@ -86,32 +86,140 @@ func IsDelegation(role string) bool {
...
+
+func listKeys(keyMap map[string]PublicKey) KeyList {
+	keys := KeyList{}
+	for _, key := range keyMap {
+		keys = append(keys, key)
+	}
+	return keys
+}
+
+func listKeyIDs(keyMap map[string]PublicKey) []string {
```
diogomonica: I vote for this to return a sorted list.

tuf/data/roles.go
```diff
@@ -86,32 +86,140 @@ func IsDelegation(role string) bool {
...
+	return listKeyIDs(b.Keys)
+}
+
+// DelegationRole is an internal representation of a delegation role, with its public keys included
+type DelegationRole struct {
+	BaseRole
+	Paths []string
+}
+
+func listKeys(keyMap map[string]PublicKey) KeyList {
```
diogomonica: I vote for this to return a sorted list.

tuf/data/roles.go
```diff
@@ -86,32 +86,140 @@ func IsDelegation(role string) bool {
...
+// IsParentOf returns whether the passed in delegation role is the direct child of this role,
+// determined by delegation name.
+// Ex: targets/a is a direct parent of targets/a/b, but targets/a is not a direct parent of targets/a/b/c
+func (d DelegationRole) IsParentOf(child DelegationRole) bool {
+	return path.Dir(child.Name) == d.Name
+}
+
+// CheckPaths checks if a given path is valid for the role
+func (d DelegationRole) CheckPaths(path string) bool {
+	return checkPaths(path, d.Paths)
```
diogomonica: Do we use `checkPaths()` more than once? If not, collapse (I couldn't find it).

tuf/data/roles.go
endophage: The only place sorting is relevant is during CLI output but this is used in a bunch of places. Seems like the wrong place to sort.

tuf/data/roles.go
endophage: Same comment as above.

tuf/data/roles.go
endophage: the `Role` type also calls through to it. Move refactoring needed to fully separate the usage of role types but that wasn't in scope for this PR and didn't have the same error prone implications as the the half removal of KeyDB.

endophage: p.s. that refactoring will come naturally as part of the state machine and changing how we walk delegations.

diogomonica: Found it. Nevermind.

tuf/tuf.go
```diff
@@ -316,53 +413,32 @@ func (tr *Repo) DeleteDelegation(role data.Role) error {
...
-	if err := tr.InitSnapshot(); err != nil {
-		return err
-	}
-	return tr.InitTimestamp()
-}
-
 // InitRoot initializes an empty root file with the 4 core roles based
-// on the current content of th ekey db
-func (tr *Repo) InitRoot(consistent bool) error {
+// on the current content of the key db
```
cyli: There is another key db reference in the comment here

tuf/tuf.go
```diff
@@ -121,8 +107,11 @@ func (tr *Repo) AddBaseKeys(role string, keys ...data.PublicKey) error {
 
 // ReplaceBaseKeys is used to replace all keys for the given role with the new keys
 func (tr *Repo) ReplaceBaseKeys(role string, keys ...data.PublicKey) error {
-	r := tr.keysDB.GetRole(role)
-	err := tr.RemoveBaseKeys(role, r.KeyIDs...)
+	r, err := tr.GetBaseRole(role)
+	if err != nil {
+		return err
+	}
+	err = tr.RemoveBaseKeys(role, r.ListKeyIDs()...)
```
diogomonica: This method is called `ReplaceBaseKeys` but it calls `td.RemoveBaseKeys`.

endophage: Yes, because it replaces the existing base keys with the provided ones. A few lines below it calls AddBaseKeys

tuf/tuf.go
```diff
@@ -173,23 +162,133 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
 	return nil
 }
 
+// GetBaseRole gets a base role from this repo's metadata
+func (tr *Repo) GetBaseRole(name string) (data.BaseRole, error) {
+	if !data.ValidRole(name) {
+		return data.BaseRole{}, data.ErrInvalidRole{Role: name, Reason: "invalid base role name"}
+	}
+	if tr.Root == nil {
```
diogomonica: Should we also check for `td.Root.Signed == nil`?

tuf/tuf.go
```diff
@@ -173,23 +162,133 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
...
+		return data.BaseRole{}, ErrNotLoaded{data.CanonicalRootRole}
+	}
+	roleData, ok := tr.Root.Signed.Roles[name]
+	if !ok {
+		return data.BaseRole{}, data.ErrInvalidRole{Role: name, Reason: "role not found in root file"}
+	}
+	// Get all public keys for the base role from TUF metadata
+	keyIDs := roleData.KeyIDs
```
diogomonica: This whole ensures that we return only keys that are included in `roleData.KeyIDs`, why? Do we not trust `tr.Root.Signed.Keys`? Because if we do, couldn't we simply return:

```golang
return data.BaseRole{
 	Name:      name,
 	Keys:      tr.Root.Signed.Keys,
        Threshold: roleData.Threshold,
 }, nil
```

tuf/tuf.go
```diff
@@ -173,23 +162,133 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
...
+
+// GetDelegationRole gets a delegation role from this repo's metadata, walking from the targets role down to the delegation itself
+func (tr *Repo) GetDelegationRole(name string) (data.DelegationRole, error) {
+	if !data.IsDelegation(name) {
+		return data.DelegationRole{}, data.ErrInvalidRole{Role: name, Reason: "invalid delegation name"}
+	}
+	if tr.Root == nil {
+		return data.DelegationRole{}, ErrNotLoaded{data.CanonicalRootRole}
+	}
+	_, ok := tr.Root.Signed.Roles[data.CanonicalTargetsRole]
```
diogomonica: Same comment as above. Should `tr.Root.Signed` checked for nil?

tuf/tuf.go
```diff
@@ -173,23 +162,133 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
...
+	if !ok {
+		return data.DelegationRole{}, ErrNotLoaded{data.CanonicalTargetsRole}
+	}
+	// Traverse target metadata, down to delegation itself
+	// Get all public keys for the base role from TUF metadata
+	signedTargetData, ok := tr.Targets[data.CanonicalTargetsRole]
```
diogomonica: Should we check if targets is nil?

tuf/tuf.go
```diff
@@ -576,14 +635,19 @@ func (tr *Repo) AddTargets(role string, targets data.Files) (data.Files, error)
 		}
 	}
 
-	// VerifyCanSign already makes sure this is not nil
-	r := tr.keysDB.GetRole(role)
+	var r data.DelegationRole
+	if role != data.CanonicalTargetsRole {
+		// we only call r.CheckPaths if the role is not "targets"
```
cyli: :+1: thanks for the helpful comment!

tuf/tuf.go
endophage: `td.Root.Signed` cannot be nil because it's not a pointer.

tuf/tuf.go
endophage: `tr.Root.Signed.Keys` contains all the keys for all the roles. It is always expected to contains keys that are not in `roleData.KeyIDs` (but will be present in a different role's `KeyIDs`.

tuf/tuf.go
endophage: Do you mean `tr.Targets` or `signedTargetData`. If the appropriate constructor has been called, `tr.Targets` will never be `nil` because it's always initialized to an empty map. For `signedTargetData`, we're using the `ok` check to determine if it has been set. It could have been set to nil, but we would have massively messed up somewhere else if that was the case.

tuf/tuf.go
```diff
@@ -173,23 +162,133 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
...
+	var foundRole *data.Role
+	for len(delegationRoles) > 0 {
+		delgRole := delegationRoles[0]
+		delegationRoles = delegationRoles[1:]
+
+		// If this role is delegated above or is our desired role, traverse it
+		if delgRole.Name == name || strings.HasPrefix(name, delgRole.Name+"/") {
+
+			// If we found the role, we can exit the loop
+			if delgRole.Name == name {
```
diogomonica: Why not move this if outside, and change the if in line 222 to only be `if strings.HasPrefix(name, delgRole.Name+"/")`?

tuf/tuf.go
```diff
@@ -173,23 +162,133 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
...
+		return data.DelegationRole{}, ErrNotLoaded{data.CanonicalTargetsRole}
+	}
+	// Traverse target metadata, down to delegation itself
+	// Get all public keys for the base role from TUF metadata
+	signedTargetData, ok := tr.Targets[data.CanonicalTargetsRole]
+	if !ok {
+		return data.DelegationRole{}, ErrNotLoaded{data.CanonicalTargetsRole}
+	}
+
+	// Start with top level roles in targets
```
diogomonica: Would love a generic description of the algorithm as a comment.

tuf/tuf_test.go
```diff
@@ -941,15 +864,215 @@ func TestAddBaseKeysToRoot(t *testing.T) {
...
+	ed25519 := signed.NewEd25519()
+	repo := initRepo(t, ed25519)
+
+	_, err := repo.GetDelegationRole("invalid")
+	assert.Error(t, err)
+
+	for _, role := range data.BaseRoles {
+		_, err = repo.GetDelegationRole(role)
+		assert.Error(t, err)
+	}
```
cyli: Can we also add a check for "targets/delegation" here?  Since no delegation role was specified, this should also return `ErrInvalidRole` since it's not in the targets tree hierarchy

tuf/tuf.go
cyli: @diogomonica previously, KeyDB refused to add a role if the keys IDs were not of keys already in the KeyDB.  Calling `SetRoot` or `SetTargets` before added the keys, then added the roles they defined, so that's where the check was before.

tuf/tuf.go
```diff
@@ -173,23 +162,133 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
...
+	}
+
+	if foundRole == nil {
+		return data.DelegationRole{}, data.ErrInvalidRole{Role: name, Reason: "delegation does not exist"}
+	}
+
+	pubKeys := make(map[string]data.PublicKey)
+	parentData, ok := tr.Targets[path.Dir(foundRole.Name)]
+	if !ok {
+		return data.DelegationRole{}, ErrNotLoaded{path.Dir(path.Dir(foundRole.Name))}
```
diogomonica: Why `path.Dir(path.Dir())`? At least it deserves a comment.

tuf/tuf.go
```diff
@@ -173,23 +162,133 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
...
+		},
+		Paths: foundRole.Paths,
+	}, nil
+}
+
 // GetAllLoadedRoles returns a list of all role entries loaded in this TUF repo, could be empty
 func (tr *Repo) GetAllLoadedRoles() []*data.Role {
-	return tr.keysDB.GetAllRoles()
+	var res []*data.Role
+	for name, rr := range tr.Root.Signed.Roles {
```
diogomonica: Isn't it possible that `tr.Root` is nil? Wouldn't this panic if so?

tuf/tuf.go
cyli: also, it seems like accessing a key in an empty/nil map is fine and doesn't panic: https://play.golang.org/p/nozL0ylF0f

tuf/tuf.go
```diff
@@ -173,23 +162,133 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
...
+		res = append(res, &data.Role{
+			RootRole: *rr,
```
diogomonica: I'm getting confused as to what `RootRole` means. It seems like `RootRole` can have any of the 4 base roles.

tuf/tuf.go
endophage: Should only be one of the path.Dir's. This represents the parent role's file hasn't been loaded.

tuf/tuf.go
```diff
@@ -173,23 +163,132 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
...
+
+	pubKeys := make(map[string]data.PublicKey)
+	parentData, ok := tr.Targets[path.Dir(foundRole.Name)]
+	if !ok {
+		return data.DelegationRole{}, ErrNotLoaded{path.Dir(path.Dir(foundRole.Name))}
+	}
+	for _, keyID := range foundRole.KeyIDs {
+		pubKey, ok := parentData.Signed.Delegations.Keys[keyID]
+		if !ok {
+			return data.DelegationRole{}, ErrNotLoaded{name}
```
cyli: If this is an issue where the keys aren't in the specified key list, should this return [the same error as above in `GetBaseRole`](https://github.com/docker/notary/pull/566/files#diff-b7b7bae92e6be55133dab8ee3256e014R184)?

tuf/tuf.go
```diff
@@ -316,53 +413,32 @@ func (tr *Repo) DeleteDelegation(role data.Role) error {
...
+func (tr *Repo) InitRoot(root, timestamp, snapshot, targets data.BaseRole, consistent bool) error {
 	rootRoles := make(map[string]*data.RootRole)
 	rootKeys := make(map[string]data.PublicKey)
-	for _, r := range data.BaseRoles {
-		role := tr.keysDB.GetRole(r)
-		if role == nil {
-			return data.ErrInvalidRole{Role: data.CanonicalRootRole, Reason: "root role not initialized in key database"}
+
+	for _, r := range []data.BaseRole{root, timestamp, snapshot, targets} {
+		rootRoles[r.Name] = &data.RootRole{
```
diogomonica: Same comment as above. We're creating a `RootRole` for each of the four base roles. Is `RootRoles` the right name for this?

cyli:
Can we also add tests for:

- `GetBaseRole` of an empty repo returns `ErrNotLoaded`
- `GetBaseRole` and `GetDelegationRole` fails if there is a key ID in the role that is not in the Keys list?

tuf/tuf.go
endophage: These should both return ErrInvalidRole. It's fine to have a key present that isn't used by any role, but we should never accept a role that has KeyIDs that aren't recorded in the Keys field of the relevant file. Will change.

tuf/tuf.go
endophage: Check added.

tuf/data/roles.go
diogomonica: Sounds good.

tuf/tuf.go
endophage: We have 2 types now: `RootRole` which represents specifically the JSON structure in root.json, and `BaseRole` which is a notary object representing the fields common to all roles (both those that appear in root.json and delegations). I think we'll get a better idea of how all this plays together as we continue the refactor in further PRs so given I don't have a good alternative right now, let's see how it plays out.

endophage:
@cyli I'll add those tests :-)

endophage:
@cyli tests added at the end of tuf_test.go, PTAL

cyli:
Awesome, this PR fixes https://github.com/docker/notary/pull/572/files#diff-d7053259a2262ec0c10ce8937be84156R903 too.  

LGTM after CI, up to you if you want to remove the `Email` from `Role`.

riyazdf:
@endophage's additions and fixes LGTM after CI :+1:

riyazdf:
LGTM in real life from @endophage, @cyli, @diogomonica, merging!

PR #568, opened 2016-02-12
Server generates snapshot and timestamp immediately

### LINKED ISSUE #563: Preemptively generate snapshot + timestamp during server verification

cyli:
Sorry for the largish change.

This changes the timestamp and snapshot code generation to just try to sign a generated timestamp and snapshot, and if there aren't enough signatures, to then fail on server-managed keys, which will help us with eliminating the `timestamp_keys` table and allowing us to have multiple signing keys for a gun and role on the signer, which is part of #347.

It also changes timestamp generation to automatically generate a timestamp when an update is sent, which fixes #563.

### DIFF
--- a/server/handlers/default_test.go
+++ b/server/handlers/default_test.go
@@ -326,8 +326,8 @@ func TestAtomicUpdateValidationFailurePropagated(t *testing.T) {
 
 	repo, cs, err := testutils.EmptyRepo(gun)
 	assert.NoError(t, err)
-	copyTimestampKey(t, repo, metaStore, gun)
-	state := handlerState{store: metaStore, crypto: cs}
+
+	state := handlerState{store: metaStore, crypto: copyKeys(t, cs, data.CanonicalTimestampRole)}
 
 	r, tg, sn, ts, err := testutils.Sign(repo)
 	assert.NoError(t, err)
@@ -368,8 +368,8 @@ func TestAtomicUpdateNonValidationFailureNotPropagated(t *testing.T) {
 
 	repo, cs, err := testutils.EmptyRepo(gun)
 	assert.NoError(t, err)
-	copyTimestampKey(t, repo, metaStore, gun)
-	state := handlerState{store: &failStore{*metaStore}, crypto: cs}
+
+	state := handlerState{store: &failStore{*metaStore}, crypto: copyKeys(t, cs, data.CanonicalTimestampRole)}
 
 	r, tg, sn, ts, err := testutils.Sign(repo)
 	assert.NoError(t, err)
@@ -409,8 +409,9 @@ func TestAtomicUpdateVersionErrorPropagated(t *testing.T) {
 
 	repo, cs, err := testutils.EmptyRepo(gun)
 	assert.NoError(t, err)
-	copyTimestampKey(t, repo, metaStore, gun)
-	state := handlerState{store: &invalidVersionStore{*metaStore}, crypto: cs}
+
+	state := handlerState{
+		store: &invalidVersionStore{*metaStore}, crypto: copyKeys(t, cs, data.CanonicalTimestampRole)}
 
 	r, tg, sn, ts, err := testutils.Sign(repo)
 	assert.NoError(t, err)
--- a/server/handlers/validation.go
+++ b/server/handlers/validation.go
@@ -107,7 +107,7 @@ func validateUpdate(cs signed.CryptoService, gun string, updates []storage.MetaU
 			}
 		}
 
-		if err := validateSnapshot(snapshotRole, oldSnap, roles[snapshotRole], roles, repo); err != nil {
+		if err := loadAndValidateSnapshot(snapshotRole, oldSnap, roles[snapshotRole], roles, repo); err != nil {
 			logrus.Error("ErrBadSnapshot: ", err.Error())
 			return nil, validation.ErrBadSnapshot{Msg: err.Error()}
 		}
@@ -270,10 +270,9 @@ func generateTimestamp(gun string, repo *tuf.Repo, store storage.MetaStore) (*st
 	return metaUpdate, nil
 }
 
-// validateSnapshot validates that the given snapshot update is valid.  It also sets the new snapshot
+// loadAndValidateSnapshot validates that the given snapshot update is valid.  It also sets the new snapshot
 // on the TUF repo, if it is valid
-func validateSnapshot(role string, oldSnap *data.SignedSnapshot, snapUpdate storage.MetaUpdate, roles map[string]storage.MetaUpdate, repo *tuf.Repo) error {
-
+func loadAndValidateSnapshot(role string, oldSnap *data.SignedSnapshot, snapUpdate storage.MetaUpdate, roles map[string]storage.MetaUpdate, repo *tuf.Repo) error {
 	s := &data.Signed{}
 	err := json.Unmarshal(snapUpdate.Data, s)
 	if err != nil {
--- a/server/handlers/validation.go
+++ b/server/handlers/validation.go
@@ -240,34 +240,35 @@ func generateSnapshot(gun string, repo *tuf.Repo, store storage.MetaStore) (*sto
 func generateTimestamp(gun string, repo *tuf.Repo, store storage.MetaStore) (*storage.MetaUpdate, error) {
 	var prev *data.SignedTimestamp
 	_, currentJSON, err := store.GetCurrent(gun, data.CanonicalTimestampRole)
-	if err == nil {
+
+	switch err.(type) {
+	case nil:
 		prev = new(data.SignedTimestamp)
-		if err = json.Unmarshal(currentJSON, prev); err != nil {
+		if err := json.Unmarshal(currentJSON, prev); err != nil {
 			logrus.Error("Failed to unmarshal existing timestamp for GUN ", gun)
 			return nil, err
 		}
-	}
-
-	if _, ok := err.(storage.ErrNotFound); !ok && err != nil {
+	case storage.ErrNotFound:
+		break // this is the first timestamp ever for the repo
+	default:
 		return nil, err
 	}
 
 	metaUpdate, err := timestamp.NewTimestampUpdate(prev, repo)
-	if err != nil {
-		_, noSigs := err.(signed.ErrInsufficientSignatures)
-		_, noKeys := err.(signed.ErrNoKeys)
+
+	switch err.(type) {
+	case nil:
+		return metaUpdate, nil
+	case signed.ErrInsufficientSignatures, signed.ErrNoKeys:
 		// If we cannot sign the timestamp, then we don't have keys for the timestamp,
 		// and the client screwed up their root
-		if noSigs || noKeys {
-			return nil, validation.ErrBadRoot{
-				Msg: fmt.Sprintf("none of the following timestamp keys exist on the server: %s",
-					strings.Join(repo.Root.Signed.Roles[data.CanonicalTimestampRole].KeyIDs, ", ")),
-			}
+		return nil, validation.ErrBadRoot{
+			Msg: fmt.Sprintf("none of the following timestamp keys exist on the server: %s",
+				strings.Join(repo.Root.Signed.Roles[data.CanonicalTimestampRole].KeyIDs, ", ")),
 		}
-
+	default:
 		return nil, validation.ErrValidation{Msg: err.Error()}
 	}
-	return metaUpdate, nil
 }
 
 // loadAndValidateSnapshot validates that the given snapshot update is valid.  It also sets the new snapshot
@@ -438,25 +439,11 @@ func checkAgainstOldRoot(oldRoot []byte, newRootRole data.BaseRole, newSigned *d
 		return err
 	}
 
-	// if the set of keys has changed between the old root and new root, then a root
-	// rotation may have taken place
-	rotation := len(oldRootRole.Keys) != len(newRootRole.Keys)
-	if !rotation { // if the number of keys is the same, we need to check every key
-		for kid := range newRootRole.Keys {
-			if _, ok := oldRootRole.Keys[kid]; !ok {
-				// if there is any difference in keys, a key rotation may have
-				// occurred.
-				rotation = true
-			}
-		}
-	}
-
-	if rotation {
-		if err := signed.VerifyRoot(newSigned, oldRootRole.Threshold, oldRootRole.Keys); err != nil {
-			return validation.ErrBadRoot{Msg: fmt.Sprintf(
-				"rotation detected and new root was not signed with at least %d old keys",
-				oldRootRole.Threshold)}
-		}
+	// Always verify the new root against the old root
+	if err := signed.VerifyRoot(newSigned, oldRootRole.Threshold, oldRootRole.Keys); err != nil {
+		return validation.ErrBadRoot{Msg: fmt.Sprintf(
+			"rotation detected and new root was not signed with at least %d old keys",
+			oldRootRole.Threshold)}
 	}
 
 	return nil
--- a/server/handlers/validation.go
+++ b/server/handlers/validation.go
@@ -55,7 +55,7 @@ func validateUpdate(cs signed.CryptoService, gun string, updates []storage.MetaU
 		// against a previous root
 		if root, err = validateRoot(gun, oldRootJSON, rootUpdate.Data); err != nil {
 			logrus.Error("ErrBadRoot: ", err.Error())
-			return nil, validation.ErrBadRoot{Msg: err.Error()}
+			return nil, err
 		}
 
 		// setting root will update keys db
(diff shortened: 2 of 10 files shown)


server/handlers/default_test.go
```diff
@@ -323,10 +323,10 @@ func TestAtomicUpdateValidationFailurePropagated(t *testing.T) {
 	gun := "testGUN"
 	vars := map[string]string{"imageName": gun}
 
-	kdb, repo, cs, err := testutils.EmptyRepo(gun)
+	_, repo, cs, err := testutils.EmptyRepo(gun)
```
riyazdf: :+1: glad to see we're phasing out keydb

server/handlers/validation.go
```diff
@@ -193,71 +202,79 @@ func loadTargetsFromStore(gun, role string, repo *tuf.Repo, store storage.MetaSt
...
-		Version: repo.Snapshot.Signed.Version,
-		Data:    sgndJSON,
-	}, nil
+	return metaUpdate, nil
 }
 
+// validateSnapshot validates that the given snapshot update is valid.  It also sets the new snapshot
+// on the TUF repo, if it is valid
 func validateSnapshot(role string, oldSnap *data.SignedSnapshot, snapUpdate storage.MetaUpdate, roles map[string]storage.MetaUpdate, repo *tuf.Repo) error {
+
```
riyazdf: nit: remove empty line

server/handlers/validation.go
```diff
@@ -193,71 +202,79 @@ func loadTargetsFromStore(gun, role string, repo *tuf.Repo, store storage.MetaSt
...
-		Role:    data.CanonicalSnapshotRole,
-		Version: repo.Snapshot.Signed.Version,
-		Data:    sgndJSON,
-	}, nil
+	return metaUpdate, nil
 }
 
+// validateSnapshot validates that the given snapshot update is valid.  It also sets the new snapshot
+// on the TUF repo, if it is valid
 func validateSnapshot(role string, oldSnap *data.SignedSnapshot, snapUpdate storage.MetaUpdate, roles map[string]storage.MetaUpdate, repo *tuf.Repo) error {
```
riyazdf: it seems like we should rename this method to `validateAndLoadSnapshot` (or `loadAndValidateSnapshot` to be consistent with targets), since it is now setting snapshot on the repo

server/handlers/validation_test.go
```diff
@@ -19,21 +20,16 @@ import (
...
-	err := toStore.SetKey(gun, data.CanonicalTimestampRole, pubTimestampKey.Algorithm(),
-		pubTimestampKey.Public())
-	assert.NoError(t, err)
+func copyKeys(t *testing.T, from signed.CryptoService, roles ...string) signed.CryptoService {
+	memKeyStore := trustmanager.NewKeyMemoryStore(passphrase.ConstantRetriever("pass"))
+	for _, role := range roles {
+		for _, keyID := range from.ListKeys(role) {
+			key, _, err := from.GetPrivateKey(keyID)
+			assert.NoError(t, err)
+			memKeyStore.AddKey(path.Base(keyID), data.CanonicalTimestampRole, key)
```
riyazdf: I'm confused, why are we specifying the `timestamp` role here?

cyli: Oh oops, sorry, that was left over from before when I was trying to copy just the timestamp key.  On the server it looks like we just get the ID without caring about the role, so the tests don't seem to be breaking, but I should definitely change this, thanks!

server/snapshot/snapshot.go
```diff
@@ -87,33 +96,28 @@ func snapshotExpired(sn *data.SignedSnapshot) bool {
...
-		// owner of gun must have generated a snapshot key otherwise
-		// we won't proceed with generating everything.
-		return nil, 0, err
+// NewSnapshotUpdate produces a new snapshot and returns it as a metadata update, given the
+// previous snapshot and the TUF repo.
+func NewSnapshotUpdate(prev *data.SignedSnapshot, repo *tuf.Repo) (*storage.MetaUpdate, error) {
+	if prev != nil {
+		repo.SetSnapshot(prev) // SetSnapshot never errors
+	} else {
+		// this will only occurr if no snapshot has ever been created for the repository
```
riyazdf: typo: `occur`

endophage:
Need to rebase on top of the keydb changes

server/handlers/validation.go
```diff
@@ -402,7 +412,7 @@ func validateRoot(gun string, oldRoot, newRoot []byte, store storage.MetaStore)
 // checkRoot errors if an invalid rotation has taken place, if the
 // threshold number of signatures is invalid, if there are an invalid
 // number of roles and keys, or if the timestamp keys are invalid
-func checkRoot(oldRoot, newRoot *data.SignedRoot, timestampKey data.PublicKey) error {
+func checkRoot(oldRoot, newRoot *data.SignedRoot, gun string, store storage.MetaStore) error {
```
riyazdf: Is there any particular reason for including the gun and store as args?  We should also update the function comment about timestamp keys

cyli: Good question - I am not sure why I left this in.  :)  Will fix, thanks!

cyli: Actually, sorry, I just realized that after the rebase, a lot of this code can go.  Will refactor it out.

riyazdf: sounds good, hooray for deleting extra code :)

server/handlers/validation.go
```diff
@@ -356,158 +373,90 @@ func validateTargets(role string, roles map[string]storage.MetaUpdate, repo *tuf
...
-				// we're skipping it because it could never have
-				// been used to validate this root.
-				continue
-			}
-			oldKeys[kid] = data.NewPublicKey(k.Algorithm(), k.Public())
-		}
-
-		// super simple check for possible rotation
-		rotation = len(oldKeys) != len(newRootRole.KeyIDs)
+	oldRootRole, err := parsedOldRoot.BuildBaseRole(data.CanonicalRootRole)
```
cyli: Note that this is a slight change in behavior:  previously, if the root had key IDs that didn't belong to any key, we would have counted that as a valid old root and used it to validate the new root.  Now, because we consider that an invalid root role, we will not use the old root to validate the new root.  Should I just build the base role without checking for that?

riyazdf: I'm personally ok with this change since I can't think of a scenario in which we could have specified key IDs without having the corresponding public key using our client or server, unless another custom client/server served us the data.  In any case, I can't think of a use case to allow for this behavior.  cc @endophage

HuKeping:
@cyli What do you mean by `eliminating the timestamp_keys table` , are we plan to drop the timestamp_keys table?

cyli:
@HuKeping That is hopefully the eventual goal, to just make the signer responsible for the keys.

server/handlers/validation.go
```diff
@@ -192,71 +201,78 @@ func loadTargetsFromStore(gun, role string, repo *tuf.Repo, store storage.MetaSt
...
+			logrus.Error("Failed to unmarshal existing snapshot for GUN ", gun)
+			return nil, err
+		}
 	}
 
-	algo, keyBytes, err := store.GetKey(gun, data.CanonicalSnapshotRole)
-	if err != nil {
-		return nil, validation.ErrBadHierarchy{Msg: "could not retrieve snapshot key. client must provide snapshot"}
+	if _, ok := err.(storage.ErrNotFound); !ok && err != nil {
+		return nil, validation.ErrValidation{Msg: err.Error()}
```
riyazdf: Can we test this case somehow?  Maybe setting a bad snapshot with the swizzler or in `snapshot_test.go`?

server/handlers/validation.go
```diff
@@ -192,71 +201,78 @@ func loadTargetsFromStore(gun, role string, repo *tuf.Repo, store storage.MetaSt
...
-			return nil, validation.ErrBadSnapshot{Msg: err.Error()}
+}
+
+// generateTimestamp generates a new timestamp from the previous one in the store - this assumes all
+// the other roles have already been set on the repo, and will set the generated timestamp on the repo as well
+func generateTimestamp(gun string, repo *tuf.Repo, store storage.MetaStore) (*storage.MetaUpdate, error) {
+	var prev *data.SignedTimestamp
+	currentJSON, err := store.GetCurrent(gun, data.CanonicalTimestampRole)
+	if err == nil {
+		prev = new(data.SignedTimestamp)
```
riyazdf: Same here, maybe a test in `timestamp_test.go` with a valid previous timestamp?

server/handlers/validation.go
```diff
@@ -192,71 +201,78 @@ func loadTargetsFromStore(gun, role string, repo *tuf.Repo, store storage.MetaSt
...
+			logrus.Error("Failed to unmarshal existing timestamp for GUN ", gun)
+			return nil, err
 		}
 	}
-	sgnd, err := repo.SignSnapshot(data.DefaultExpires(data.CanonicalSnapshotRole))
-	if err != nil {
-		return nil, validation.ErrBadSnapshot{Msg: err.Error()}
+
+	if _, ok := err.(storage.ErrNotFound); !ok && err != nil {
+		return nil, validation.ErrValidation{Msg: err.Error()}
```
riyazdf: A test in `timestamp_test.go` for this case would be awesome too!

server/handlers/validation.go
```diff
@@ -356,158 +373,90 @@ func validateTargets(role string, roles map[string]storage.MetaUpdate, repo *tuf
 	return t, nil
 }
 
-func validateRoot(gun string, oldRoot, newRoot []byte, store storage.MetaStore) (
+// validateRoot returns the parsed data.SignedRoot object if the new root:
+// - is a valid root metadata object
+// - has the correct number of timestamp keys
+// - validates against the previous root's signatures (if there was a rotation)
+// - is valid against itself (signature-wise)
```
riyazdf: this comment is so helpful, thanks for adding it :+1:

server/snapshot/snapshot.go
```diff
@@ -1,11 +1,11 @@
 package snapshot
 
 import (
-	"encoding/json"
-
 	"github.com/Sirupsen/logrus"
 
+	"github.com/docker/go/canonical/json"
```
riyazdf: I think this change to canonical json should be fine, but could we test interacting with old repos to ensure there aren't any issues?  cc @endophage since he's more familiar with our canonical/json vs. encoding/json compatibilities

server/handlers/validation.go
cyli: Great suggestion!  Will add

server/handlers/validation.go
cyli: Good idea!

server/handlers/validation.go
cyli: :+1:

cyli: Done!

server/snapshot/snapshot.go
cyli: Good point!  I've now added changes to the backwards compatibility test to also rotate a snapshot, which tests that hte new snapshot gets generated ok with old data.  (https://github.com/docker/notary/pull/568/files#diff-7b61277673385a0a9b278b5bec4c9ffaR131)

Do you have any other suggestions for tests?

server/handlers/validation.go
cyli: Added the test - also realized this should probably not be wrapped in a validation error.

server/handlers/validation.go
cyli: Have added this test as well - and also realized it should not be within a validation error message :)

server/snapshot/snapshot.go
riyazdf: :+1:  awesome!  I'm pretty excited about all of the tests you've added, but I'll let you know if I think of anything else

riyazdf:
LGTM, thanks for adding more tests :)

cyli:
@riyazdf thank you for their suggestion!  :D

server/handlers/validation.go
```diff
@@ -192,71 +201,78 @@ func loadTargetsFromStore(gun, role string, repo *tuf.Repo, store storage.MetaSt
...
+}
+
+// generateTimestamp generates a new timestamp from the previous one in the store - this assumes all
+// the other roles have already been set on the repo, and will set the generated timestamp on the repo as well
+func generateTimestamp(gun string, repo *tuf.Repo, store storage.MetaStore) (*storage.MetaUpdate, error) {
+	var prev *data.SignedTimestamp
+	_, currentJSON, err := store.GetCurrent(gun, data.CanonicalTimestampRole)
+	if err == nil {
+		prev = new(data.SignedTimestamp)
+		if err = json.Unmarshal(currentJSON, prev); err != nil {
```
endophage: Can we do `err := ...` just for safety. Was getting a bit confused looking at this against line 251 just below. Would be good to be clear this `err` shouldn't be in scope outside the if.

server/handlers/validation.go
```diff
@@ -192,71 +201,78 @@ func loadTargetsFromStore(gun, role string, repo *tuf.Repo, store storage.MetaSt
...
+			logrus.Error("Failed to unmarshal existing timestamp for GUN ", gun)
+			return nil, err
 		}
 	}
-	sgnd, err := repo.SignSnapshot(data.DefaultExpires(data.CanonicalSnapshotRole))
-	if err != nil {
-		return nil, validation.ErrBadSnapshot{Msg: err.Error()}
+
+	if _, ok := err.(storage.ErrNotFound); !ok && err != nil {
```
endophage: similar clarity reason to above. Could this be an `else if`?

endophage: Actually, given there's a return here, this should be moved up above the `if err == nil` and that `if` can be removed and the content de-indented.

server/handlers/validation.go
```diff
@@ -192,71 +201,78 @@ func loadTargetsFromStore(gun, role string, repo *tuf.Repo, store storage.MetaSt
...
+		return nil, err
 	}
-	sgndJSON, err := json.Marshal(sgnd)
+
+	metaUpdate, err := timestamp.NewTimestampUpdate(prev, repo)
 	if err != nil {
-		return nil, validation.ErrBadSnapshot{Msg: err.Error()}
+		_, noSigs := err.(signed.ErrInsufficientSignatures)
```
endophage: Can use a type switch to replace everything from the `if err != nil` to the end of the func:
```
switch err.(type) {
case nil:
    return metaUpdate, nil
case signed.ErrInsufficientSignatures, signed.ErrNoKeys:
    return nil, validation.ErrBadRoot{...}
default:
    return nil, validation.ErrValidation{...}
}
```
}

server/handlers/validation.go
```diff
@@ -356,158 +373,91 @@ func validateTargets(role string, roles map[string]storage.MetaUpdate, repo *tuf
...
 		}
 	}
 
-	// ensure that at least one of the timestamp keys specified in the role
-	// actually exists
-
-	for _, keyID := range timestampKeyIDs {
-		if timestampKey.ID() == keyID {
-			return nil
+	if rotation {
```
endophage: We should always verify against old regardless of whether we detect a rotation or not. Spec requires that any new root verifies against the previous one.

server/handlers/validation_test.go
```diff
@@ -197,11 +395,53 @@ func TestValidateRootRotation(t *testing.T) {
...
 	updates := []storage.MetaUpdate{root, targets, snapshot, timestamp}
 
-	copyTimestampKey(t, repo, store, "testGUN")
-	_, err = validateUpdate(crypto, "testGUN", updates, store)
+	serverCrypto := copyKeys(t, crypto, data.CanonicalTimestampRole)
+	_, err = validateUpdate(serverCrypto, "testGUN", updates, store)
 	assert.NoError(t, err)
 }
 
+func TestInvalidRootRotation(t *testing.T) {
```
endophage: Can we add a comment as to in what way the root is invalid? It appears to be because the root is signed with only the new key and not the old key?

endophage:
LGTM but can we rebase (even though there are no conflicts) and make sure the tests pass before merging. Want to make sure my changes to signature cleanup don't cause anything to fail here.

server/handlers/validation.go
cyli: Good point, thanks!

server/handlers/validation.go
cyli: You are the master of switch expressions.  :)

cyli:
@endophage Ah good call on the rebase - there is a semantic conflict with the flatten keystore PR (this PR does a lot of creating of keys without enough parameters)

server/handlers/validation.go
cyli: I can definitely move this above the `if err == nil`, but that `if` cannot be de-indented because this handles the case where the error is not `storage.ErrNotFound`, and that `if` handles the case where there is no error at all, but we basically want to do nothing if the error was `storage.ErrNotFound`.

I can change this to a switch statement though, since it's somewhat confusing.

PR #570, opened 2016-02-14
Client.Update as a simple state machine, logically a no-op

NathanMcCauley:
Migrate Client.Update to be a simplistic state machine. More modifications to follow in followup PR where entrance/exit conditions will be modified based on the errors returned from the `Download*()` functions

### DIFF
--- a/tuf/client/client.go
+++ b/tuf/client/client.go
@@ -45,44 +45,78 @@ func (c *Client) Update() error {
 	// 3. Check if root correct against snapshot
 	//   a. If incorrect, download new root and return to 1.
 	// 4. Iteratively download and search targets and delegations to find target meta
-	logrus.Debug("updating TUF client")
-	err := c.update()
-	if err != nil {
-		logrus.Debug("Error occurred. Root will be downloaded and another update attempted")
-		if err := c.downloadRoot(); err != nil {
-			logrus.Debug("Client Update (Root):", err)
-			return err
-		}
-		// If we error again, we now have the latest root and just want to fail
-		// out as there's no expectation the problem can be resolved automatically
-		logrus.Debug("retrying TUF client update")
-		return c.update()
-	}
-	return nil
-}
 
-func (c *Client) update() error {
-	err := c.downloadTimestamp()
-	if err != nil {
-		logrus.Debugf("Client Update (Timestamp): %s", err.Error())
-		return err
+	type TufUpdateState int
+
+	const (
+		StateDownloadTimestamps TufUpdateState = iota
+		StateDownloadSnapshots
+		StateDownloadTargets
+		StateCheckRoot
+		StateDownloadRoot
+		StateUpdateComplete
+		StateUpdateFailed
+	)
+
+	var err error
+	downloadedRoot := false
+	downloadState := StateDownloadTimestamps
+	logrus.Debug("updating TUF client")
+
+	for downloadState != StateUpdateComplete && downloadState != StateUpdateFailed {
+		switch downloadState {
+		case StateDownloadTimestamps:
+			err = c.downloadTimestamp()
+			if err != nil {
+				logrus.Debugf("Client Update (Timestamp): %s", err.Error())
+				downloadState = StateDownloadRoot
+				break
+			}
+			downloadState = StateDownloadSnapshots
+		case StateDownloadSnapshots:
+			err = c.downloadSnapshot()
+			if err != nil {
+				logrus.Debugf("Client Update (Snapshot): %s", err.Error())
+				downloadState = StateDownloadRoot
+				break
+			}
+			downloadState = StateCheckRoot
+		case StateCheckRoot:
+			err = c.checkRoot()
+			if err != nil {
+				// In this instance the root has not expired based on time, but is
+				// expired based on the snapshot dictating a new root has been produced.
+				logrus.Debugf("Client Update (Checkroot):", err.Error())
+				downloadState = StateDownloadRoot
+				break
+			}
+			downloadState = StateDownloadTargets
+		case StateDownloadTargets:
+			// will always need top level targets at a minimum
+			err = c.downloadTargets(data.CanonicalTargetsRole)
+			if err != nil {
+				logrus.Debugf("Client Update (Targets): %s", err.Error())
+				downloadState = StateDownloadRoot
+				break
+			}
+			downloadState = StateUpdateComplete
+		case StateDownloadRoot:
+			if downloadedRoot {
+				downloadState = StateUpdateFailed
+				break
+			}
+			downloadedRoot = true
+			logrus.Debug("Will download Root and attempt to update again.")
+			if err = c.downloadRoot(); err != nil {
+				logrus.Debugf("Client Update (Root):", err.Error())
+				downloadState = StateUpdateFailed
+				break
+			}
+			downloadState = StateDownloadTimestamps
+		}
 	}
-	err = c.downloadSnapshot()
-	if err != nil {
-		logrus.Debugf("Client Update (Snapshot): %s", err.Error())
-		return err
-	}
-	err = c.checkRoot()
-	if err != nil {
-		// In this instance the root has not expired base on time, but is
-		// expired based on the snapshot dictating a new root has been produced.
-		logrus.Debug(err)
-		return err
-	}
-	// will always need top level targets at a minimum
-	err = c.downloadTargets("targets")
-	if err != nil {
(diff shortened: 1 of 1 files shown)


tuf/client/client.go
```diff
@@ -45,44 +45,77 @@ func (c *Client) Update() error {
 	// 3. Check if root correct against snapshot
 	//   a. If incorrect, download new root and return to 1.
 	// 4. Iteratively download and search targets and delegations to find target meta
+
+	type TufUpdateState int
+
+	const (
+		StateDownloadTimestamps TufUpdateState = iota
+		StateDownloadSnapshots  TufUpdateState = iota
```
diogomonica: You only need `TufUpdateState = iota` on the first line.

tuf/client/client.go
```diff
@@ -45,44 +45,77 @@ func (c *Client) Update() error {
...
+			if err != nil {
+				logrus.Debugf("Client Update (Snapshot): %s", err.Error())
+				downloadState = StateDownloadRoot
+				break
+			}
+			downloadState = StateCheckRoot
+		case StateCheckRoot:
+			err = c.checkRoot()
+			if err != nil {
+				// In this instance the root has not expired base on time, but is
```
diogomonica: *based

tuf/client/client.go
```diff
@@ -45,44 +45,77 @@ func (c *Client) Update() error {
...
+			downloadState = StateUpdateComplete
+		case StateDownloadRoot:
+			if downloadedRoot {
+				downloadState = StateUpdateFailed
+				break
+			}
+			downloadedRoot = true
+			logrus.Debug("Error occurred. Root will be downloaded and another update attempted")
+			if err = c.downloadRoot(); err != nil {
+				logrus.Debug("Client Update (Root):", err)
```
diogomonica: Let's be consistent on the errors and use `logrus.Debugf` here.

tuf/client/client.go
```diff
@@ -45,44 +45,77 @@ func (c *Client) Update() error {
...
+				downloadState = StateDownloadRoot
+				break
+			}
+			downloadState = StateCheckRoot
+		case StateCheckRoot:
+			err = c.checkRoot()
+			if err != nil {
+				// In this instance the root has not expired base on time, but is
+				// expired based on the snapshot dictating a new root has been produced.
+				logrus.Debug(err)
```
diogomonica: Let's be consistent on the errors, add a message and use `logrus.Debugf` here.

tuf/client/client.go
```diff
@@ -45,44 +45,77 @@ func (c *Client) Update() error {
...
+			err = c.downloadTargets("targets")
+			if err != nil {
+				logrus.Debugf("Client Update (Targets): %s", err.Error())
+				downloadState = StateDownloadRoot
+				break
+			}
+			downloadState = StateUpdateComplete
+		case StateDownloadRoot:
+			if downloadedRoot {
+				downloadState = StateUpdateFailed
```
diogomonica: Probably worth adding a debug log on this irrecoverable state.

tuf/client/client.go
```diff
@@ -45,44 +45,77 @@ func (c *Client) Update() error {
...
+		case StateDownloadSnapshots:
+			err = c.downloadSnapshot()
+			if err != nil {
+				logrus.Debugf("Client Update (Snapshot): %s", err.Error())
+				downloadState = StateDownloadRoot
+				break
+			}
+			downloadState = StateCheckRoot
+		case StateCheckRoot:
+			err = c.checkRoot()
```
diogomonica: There is a potential transition here to `StateUpdateFailed`, in case the err being return is `tuf.ErrLocalRootExpired{}`

diogomonica: Also, an error when checking the size of a cached root doesn't seem to be recoverable from:

```
if int64(len(raw)) != size {
	return fmt.Errorf("Cached root size did not match snapshot size")
}
```

diogomonica:
I think collapsing the code to one method makes it more readable, but even though the transition to a state machine solves the ugly nested if condition, it makes it slightly harder to understand if the code terminates.

Overall I'm LGTM after comments.

tuf/client/client.go
```diff
@@ -45,44 +45,77 @@ func (c *Client) Update() error {
...
-		logrus.Debug(err)
-		return err
+	for downloadState != StateUpdateComplete && downloadState != StateUpdateFailed {
+		switch downloadState {
+		case StateDownloadTimestamps:
+			err = c.downloadTimestamp()
+			if err != nil {
+				logrus.Debugf("Client Update (Timestamp): %s", err.Error())
+				downloadState = StateDownloadRoot
+				break
```
endophage: No need for the `break` statements. Go switches explicitly do not fall through. Instead of adding `break` to exit the switch, it's the default behaviour and you have to add `fallthrough` to force it to fall through.

tuf/client/client.go
```diff
@@ -45,44 +45,77 @@ func (c *Client) Update() error {
...
+				// In this instance the root has not expired base on time, but is
+				// expired based on the snapshot dictating a new root has been produced.
+				logrus.Debug(err)
+				downloadState = StateDownloadRoot
+				break
+			}
+			downloadState = StateDownloadTargets
+		case StateDownloadTargets:
+			// will always need top level targets at a minimum
+			err = c.downloadTargets("targets")
```
riyazdf: nit: we should use `data.CanonicalTargetsRole` instead of the hardcoded string

tuf/client/client.go
NathanMcCauley: There's a bunch of additional checks to think through and modify behavior in subsequent PRs. I think it makes sense for this PR to be just a refactor to structure as a state machine, and logical changes can come in a separate PR to keep them small and reviewable.

I agree that the states you mention need to be considered.

diogomonica: :+1:

tuf/client/client.go
diogomonica: Wouldn't remove this break always make `downloadState = StateDownloadSnapshots`?

NathanMcCauley: I have the `break`s only in the `if` conditions where we explicitly want to move to a different state. I put it in b/c I thought it was prettier than having `else` blocks.

riyazdf: agreed with @diogomonica, we'd need to restructure this code if we wanted to remove the `break`.  I actually like the explicit `break` here, I think it keeps the code simpler and easier to read when changing states.

tuf/client/client.go
NathanMcCauley: Yes, +1

tuf/client/client.go
riyazdf: +1 for debug log here

NathanMcCauley: Added to the exit condition at the end of the function.

tuf/client/client.go
```diff
@@ -45,44 +45,78 @@ func (c *Client) Update() error {
...
+				break
+			}
+			downloadState = StateUpdateComplete
+		case StateDownloadRoot:
+			if downloadedRoot {
+				downloadState = StateUpdateFailed
+				break
+			}
+			downloadedRoot = true
+			logrus.Debug("Error occurred. Root will be downloaded and another update attempted")
```
riyazdf: nit: can we simplify the second debug sentence, maybe something like: `Will download Root and attempt to update again`?

NathanMcCauley: sure

riyazdf:
LGTM after small rewording/adding debugs.

NathanMcCauley:
I believe all comments thus far have been addressed.

endophage:
Asking that we keep this open until @cyli and I catch up with fixing up other bits then we'll either merge and expand this (with the load cached vs download states), or use this as the basis for building up that state machine in our own branch.

Either way, as @NathanMcCauley noted IRL this morning, the `download*` functions are misnamed as they might use cache and I'm already renaming them in the state_machine branch.

tuf/client/client.go
endophage: Actually, given `break` is meaningless to go switches, I think this will break out of the `for` loop.

endophage: Nevermind, tried it out in playground. `break` does apply to switches...

NathanMcCauley:
@endophage is this still useful to keep open? If not, we can close it.

endophage:
@cyli and I were discussing the possibility of further improving the builder because it's in an "it works but it's a little convoluted" state. That's probably a precursor to any significant changes to the update flow. We can close this and reassess down the road.

PR #571, opened 2016-02-15
Change the CLI to be able to rotate server managed keys

### LINKED ISSUE #554: Change CLI key rotation to require a key type
Since we can rotate some keys to the server now.  See https://github.com/docker/notary/pull/553#discussion_r52253370

cyli:
This includes some minor changes from #553.

Fixes #554

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -46,7 +46,18 @@ type ErrInvalidRemoteRole struct {
 
 func (err ErrInvalidRemoteRole) Error() string {
 	return fmt.Sprintf(
-		"notary does not support the server managing the %s key", err.Role)
+		"notary does not permit the server managing the %s key", err.Role)
+}
+
+// ErrInvalidLocalRole is returned when the client wants to manage
+// an unsupported key type
+type ErrInvalidLocalRole struct {
+	Role string
+}
+
+func (err ErrInvalidLocalRole) Error() string {
+	return fmt.Sprintf(
+		"notary does not permit the client managing the %s key", err.Role)
 }
 
 // ErrRepositoryNotExist is returned when an action is taken on a remote
@@ -543,11 +554,10 @@ func (r *NotaryRepository) Publish() error {
 			initialPublish = true
 		} else {
 			// We could not update, so we cannot publish.
-			logrus.Error("Could not publish Repository: ", err.Error())
+			logrus.Error("Could not publish Repository since we could not update: ", err.Error())
 			return err
 		}
 	}
-
 	cl, err := r.GetChangelist()
 	if err != nil {
 		return err
@@ -640,7 +650,7 @@ func (r *NotaryRepository) Publish() error {
 // to load metadata for all roles.  Since server snapshots are supported,
 // if the snapshot metadata fails to load, that's ok.
 // This can also be unified with some cache reading tools from tuf/client.
-// This assumes that bootstrapRepo is only used by Publish()
+// This assumes that bootstrapRepo is only used by Publish() or RotateKey()
 func (r *NotaryRepository) bootstrapRepo() error {
 	tufRepo := tuf.NewRepo(r.CryptoService)
 
@@ -858,25 +868,36 @@ func (r *NotaryRepository) validateRoot(rootJSON []byte) (*data.SignedRoot, erro
 // creates and adds one new key or delegates managing the key to the server.
 // These changes are staged in a changelist until publish is called.
 func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool) error {
-	if role == data.CanonicalRootRole || role == data.CanonicalTimestampRole {
-		return fmt.Errorf(
-			"notary does not currently support rotating the %s key", role)
-	}
-	if serverManagesKey && role == data.CanonicalTargetsRole {
-		return ErrInvalidRemoteRole{Role: data.CanonicalTargetsRole}
+	switch {
+	// We currently support locally managing targets keys, remotely managing
+	// timestamp keys, and locally or remotely managing snapshot keys.
+	case role == data.CanonicalTargetsRole:
+		if serverManagesKey {
+			return ErrInvalidRemoteRole{Role: data.CanonicalTargetsRole}
+		}
+	case role == data.CanonicalTimestampRole:
+		if !serverManagesKey {
+			return ErrInvalidLocalRole{Role: data.CanonicalTimestampRole}
+		}
+	case role == data.CanonicalSnapshotRole:
+	default:
+		return fmt.Errorf("notary does not currently permit rotating the %s key", role)
 	}
 
 	var (
-		pubKey data.PublicKey
-		err    error
+		pubKey       data.PublicKey
+		err          error
+		errFmtString string
 	)
 	if serverManagesKey {
 		pubKey, err = getRemoteKey(r.baseURL, r.gun, role, r.roundTrip)
+		errFmtString = "unable to rotate remote key: %s"
 	} else {
 		pubKey, err = r.CryptoService.Create(role, data.ECDSAKey)
+		errFmtString = "unable to generate key: %s"
 	}
 	if err != nil {
-		return err
+		return fmt.Errorf(errFmtString, err)
 	}
 
 	return r.rootFileKeyChange(role, changelist.ActionCreate, pubKey)
--- a/client/client.go
+++ b/client/client.go
@@ -531,6 +531,25 @@ func (r *NotaryRepository) ListRoles() ([]RoleWithSignatures, error) {
 // Publish pushes the local changes in signed material to the remote notary-server
 // Conceptually it performs an operation similar to a `git rebase`
 func (r *NotaryRepository) Publish() error {
+	cl, err := r.GetChangelist()
+	if err != nil {
+		return err
+	}
+	if err = r.publish(cl); err != nil {
+		return err
+	}
+	if err = cl.Clear(""); err != nil {
+		// This is not a critical problem when only a single host is pushing
+		// but will cause weird behaviour if changelist cleanup is failing
+		// and there are multiple hosts writing to the repo.
+		logrus.Warn("Unable to clear changelist. You may want to manually delete the folder ", filepath.Join(r.tufRepoPath, "changelist"))
+	}
+	return nil
+}
+
+// publish pushes the changes in the given changelist to the remote notary-server
+// Conceptually it performs an operation similar to a `git rebase`
+func (r *NotaryRepository) publish(cl changelist.Changelist) error {
 	var initialPublish bool
 	// update first before publishing
 	_, err := r.Update(true)
@@ -557,10 +576,6 @@ func (r *NotaryRepository) Publish() error {
 			logrus.Error("Could not publish Repository since we could not update: ", err.Error())
 			return err
 		}
-	}
-	cl, err := r.GetChangelist()
-	if err != nil {
-		return err
 	}
 	// apply the changelist to the repo
 	err = applyChangelist(r.tufRepo, cl)
@@ -632,18 +647,7 @@ func (r *NotaryRepository) Publish() error {
 		return err
 	}
 
-	err = remote.SetMultiMeta(updatedFiles)
-	if err != nil {
-		return err
-	}
-	err = cl.Clear("")
-	if err != nil {
-		// This is not a critical problem when only a single host is pushing
-		// but will cause weird behaviour if changelist cleanup is failing
-		// and there are multiple hosts writing to the repo.
-		logrus.Warn("Unable to clear changelist. You may want to manually delete the folder ", filepath.Join(r.tufRepoPath, "changelist"))
-	}
-	return nil
+	return remote.SetMultiMeta(updatedFiles)
 }
 
 // bootstrapRepo loads the repository from the local file system.  This attempts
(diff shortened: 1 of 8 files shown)


client/client.go
```diff
@@ -52,7 +52,18 @@ type ErrInvalidRemoteRole struct {
 
 func (err ErrInvalidRemoteRole) Error() string {
 	return fmt.Sprintf(
-		"notary does not support the server managing the %s key", err.Role)
+		"notary does not permit the server managing the %s key", err.Role)
+}
+
+// ErrInvalidLocalRole is returned when the client wants to manage
+// an unsupported key type
```
riyazdf: super nit: I'd change `unsupported` to `not permitted` so that it's consistent with the err message

client/client.go
```diff
@@ -829,25 +839,36 @@ func (r *NotaryRepository) validateRoot(rootJSON []byte) (*data.SignedRoot, erro
...
+	// timestamp keys, and locally or remotely managing snapshot keys.
+	case role == data.CanonicalTargetsRole:
+		if serverManagesKey {
+			return ErrInvalidRemoteRole{Role: data.CanonicalTargetsRole}
+		}
+	case role == data.CanonicalTimestampRole:
+		if !serverManagesKey {
+			return ErrInvalidLocalRole{Role: data.CanonicalTimestampRole}
+		}
+	case role == data.CanonicalSnapshotRole:
```
riyazdf: nit: Could we move the snapshot case to the top of the switch with a break to make it more clear that the snapshot is a valid case?

client/client_test.go
```diff
@@ -2392,16 +2392,20 @@ func TestRotateKeyInvalidRole(t *testing.T) {
 	repo, _ := initializeRepo(t, data.ECDSAKey, "docker.com/notary", ts.URL, false)
 	defer os.RemoveAll(repo.baseDir)
 
-	// the equivalent of: (root, true), (root, false), (timestamp, true),
-	// (timestamp, false), (targets, true)
+	// the equivalent of: (root, true), (root, false), (timestamp, false), (targets, true)
```
riyazdf: nit: Could you also explain the `true`/`false` in another line comment?

client/client_test.go
```diff
@@ -2410,6 +2414,21 @@ func TestRotateKeyInvalidRole(t *testing.T) {
...
+// If remotely rotating key fails, the failure is propagated
+func TestRemoteRotationError(t *testing.T) {
+	ts, _, _ := simpleTestServer(t)
+
+	repo, _ := initializeRepo(t, data.ECDSAKey, "docker.com/notary", ts.URL, true)
+	defer os.RemoveAll(repo.baseDir)
+
+	ts.Close()
+
+	// server has died, so this should fail
```
riyazdf: poor server :(   Great test!

cmd/notary/keys.go
```diff
@@ -16,11 +16,12 @@ import (
 	"github.com/docker/notary/passphrase"
 	"github.com/docker/notary/trustmanager"
 
+	"io/ioutil"
```
riyazdf: super nit (and not directly related to functionality): should this import just move right underneath `"io`"?

cmd/notary/keys.go
```diff
@@ -36,9 +37,9 @@ var cmdKeyListTemplate = usageTemplate{
 }
 
 var cmdRotateKeyTemplate = usageTemplate{
-	Use:   "rotate [ GUN ]",
-	Short: "Rotate the signing (non-root) keys for the given Globally Unique Name.",
-	Long:  "Removes all the old signing (non-root) keys for the given Globally Unique Name, and generates new ones.  This only makes local changes - please use then `notary publish` to push the key rotation changes to the remote server.",
+	Use:   "rotate [ GUN ] [ key role ]",
```
riyazdf: :+1: this is awesome

cmd/notary/keys.go
```diff
@@ -127,11 +128,7 @@ func (k *keyCommander) GetCommand() *cobra.Command {
 	cmdRotateKey.Flags().BoolVarP(&k.rotateKeyServerManaged, "server-managed", "r",
 		false, "Signing and key management will be handled by the remote server. "+
 			"(no key will be generated or stored locally) "+
-			"Can only be used in conjunction with --key-type.")
-	cmdRotateKey.Flags().StringVarP(&k.rotateKeyRole, "key-type", "t", "",
-		`Key type to rotate.  Supported values: "targets", "snapshot". `+
-			`If not provided, both targets and snapshot keys will be rotated, `+
-			`and the new keys will be locally generated and stored.`)
+			"Can only be used for timestamp and snapshot roles, and MUST be used for the timestamp role.")
```
riyazdf: not sure if we should use the all-caps here.  Maybe we can reword, what do you think of this?

`Required for timestamp role, optional for snapshot role`

cmd/notary/keys.go
```diff
@@ -127,11 +128,7 @@ func (k *keyCommander) GetCommand() *cobra.Command {
 	cmdRotateKey.Flags().BoolVarP(&k.rotateKeyServerManaged, "server-managed", "r",
 		false, "Signing and key management will be handled by the remote server. "+
 			"(no key will be generated or stored locally) "+
```
riyazdf: I think it might make sense for this parenthesized segment to go into the last sentence, so that it'd look like this with the current phrasing:
`Signing and key management will be handled by the remote server (no key will be generated or stored locally.  Can only be used for timestamp and snapshot roles, and MUST be used for the timestamp role.`

cmd/notary/keys.go
```diff
@@ -406,26 +403,9 @@ func (k *keyCommander) keysImport(cmd *cobra.Command, args []string) error {
...
-	case data.CanonicalSnapshotRole:
-		rolesToRotate = []string{data.CanonicalSnapshotRole}
-	case data.CanonicalTargetsRole:
-		rolesToRotate = []string{data.CanonicalTargetsRole}
-	default:
-		return fmt.Errorf("key rotation not supported for %s keys", k.rotateKeyRole)
-	}
-	if k.rotateKeyServerManaged && rotateKeyRole != data.CanonicalSnapshotRole {
-		return fmt.Errorf(
-			"remote signing/key management is only supported for the snapshot key")
```
riyazdf: :+1: this is great, glad to get rid of this code!

cmd/notary/keys_test.go
```diff
@@ -227,12 +233,12 @@ func TestRemoveMultikeysRemoveOnlyChosenKey(t *testing.T) {
 	}
 }
 
-// Non-roles, root, and timestamp can't be rotated
+// Non-roles, root, and delegation keys can't be rotated with this command line
```
riyazdf: nit: change `this` to `the`

cmd/notary/keys_test.go
```diff
@@ -258,10 +268,22 @@ func TestRotateKeyTargetCannotBeServerManaged(t *testing.T) {
...
 	}
-	err := k.keysRotate(&cobra.Command{}, []string{"gun"})
+	err := k.keysRotate(&cobra.Command{}, []string{"gun", data.CanonicalTargetsRole})
 	assert.Error(t, err)
-	assert.Contains(t, err.Error(),
-		"remote signing/key management is only supported for the snapshot key")
+	assert.IsType(t, client.ErrInvalidRemoteRole{}, err)
+}
+
+// Cannot rotate a timestamp key and require that it is locally managed it
```
riyazdf: remove the trailing `it`

cmd/notary/keys_test.go
```diff
@@ -280,17 +302,23 @@ func TestRotateKeyNoGUN(t *testing.T) {
...
-			fmt.Fprint(w, keyJSON)
-		}))
+	// Set up server
+	ctx := context.WithValue(
+		context.Background(), "metaStore", storage.NewMemStorage())
+
+	// Do not pass one of the const KeyAlgorithms here as the value! Passing a
+	// string is in itself good test that we are handling it correctly as we
+	// will be receiving a string from the configuration.
+	ctx = context.WithValue(ctx, "keyAlgorithm", "ecdsa")
```
riyazdf: good idea!

riyazdf:
LGTM after comments!

client/client_test.go
cyli: Oh oops!  Yes, sorry, that is a bit opaque :D

client/client.go
cyli: :+1:

cmd/notary/keys.go
cyli: That's way better, thanks!

client/client_test.go
```diff
@@ -2392,16 +2392,24 @@ func TestRotateKeyInvalidRole(t *testing.T) {
 	repo, _ := initializeRepo(t, data.ECDSAKey, "docker.com/notary", ts.URL, false)
 	defer os.RemoveAll(repo.baseDir)
 
-	// the equivalent of: (root, true), (root, false), (timestamp, true),
-	// (timestamp, false), (targets, true)
+	// the equivalent of: remotely remotely rotating the root key
```
diogomonica: two "remotely"'s

cmd/notary/keys.go
```diff
@@ -36,9 +36,9 @@ var cmdKeyListTemplate = usageTemplate{
 }
 
 var cmdRotateKeyTemplate = usageTemplate{
-	Use:   "rotate [ GUN ]",
-	Short: "Rotate the signing (non-root) keys for the given Globally Unique Name.",
-	Long:  "Removes all the old signing (non-root) keys for the given Globally Unique Name, and generates new ones.  This only makes local changes - please use then `notary publish` to push the key rotation changes to the remote server.",
+	Use:   "rotate [ GUN ] [ key role ]",
+	Short: "Rotate a signing (non-root) key of the given type for the given Globally Unique Name and role.",
+	Long:  "Generates a new signing key (non-root) for the given Globally Unique Name and role.  Rotating to a server-managed key is an online-only operation: a new key is requested from the server rather than generated, and if successful, the key rotation is immediately published.  Rotating to a locally-managed key is an offline operation only: `notary publish` must be executed manually afterward to publish to the remote server.\nThe role must be one of \"snapshot\", \"targets\", or \"timestamp\".",
```
diogomonica: Maybe we should turn rotation to a local key into an online operation too?

cyli: It would certainly make the description make sense, but given that we are trying to keep notary operations as offline as possible until publish, should we do that?

Alternately, should we add a different command (rather than a flag) that's `rotate-remote` or something?

endophage: Is there any actual online aspect to rotating from a remote key to a local key? Or would we artificially be making it an online operation.

cyli: Artificially making it an online operation

cyli: Unless we want to force an immediate publish no matter what rotation

cmd/notary/keys.go
```diff
@@ -449,12 +430,11 @@ func (k *keyCommander) keysRotate(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	for _, role := range rolesToRotate {
-		if err := nRepo.RotateKey(role, k.rotateKeyServerManaged); err != nil {
-			return err
-		}
+	err = nRepo.RotateKey(rotateKeyRole, k.rotateKeyServerManaged)
+	if err == nil && k.rotateKeyServerManaged {
+		err = nRepo.Publish()
```
riyazdf: One thought that we were discussing: is it possible for publish only the key rotation change, instead of all staged changes?  I'm not sure of the best way to go about doing this without making users published all staged changes prior to this one?

endophage: We'd need to break up `Publish` into a few component steps but that wouldn't be a bad thing.

cyli: I can move the call to publish to `nRepo.RotateKey` - I was thinking about adding a private `publish(changelist)` command.  RotateKey can generate an in-memory change file that gets published using that private method.

cyli: Unless we were thinking about breaking `Publish` up into several publicly exposed component functions, and have the CLI call that.  It might make more sense to have the CLI call `Publish` rather than `RotateKey` call `Publish`.

cyli: Slack discussion:  `publish(changelist)` seems like the way to go

cmd/notary/keys.go
riyazdf: I'm of the opinion that the local and remote key rotation should follow the same UX flow, such that `notary key rotate` also publishes the change to the server

cyli: I am fine with doing this.  @endophage?

endophage: I'm fine with that. Changes to keys are more critical so should probably be propagated immediately.

client/client.go
```diff
@@ -858,37 +872,50 @@ func (r *NotaryRepository) validateRoot(rootJSON []byte) (*data.SignedRoot, erro
...
 func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool) error {
-	if role == data.CanonicalRootRole || role == data.CanonicalTimestampRole {
-		return fmt.Errorf(
-			"notary does not currently support rotating the %s key", role)
-	}
-	if serverManagesKey && role == data.CanonicalTargetsRole {
-		return ErrInvalidRemoteRole{Role: data.CanonicalTargetsRole}
+	switch {
+	// We currently support locally or remotely managing snapshot keys...
+	case role == data.CanonicalSnapshotRole:
```
endophage: Can we add an explicit break just to be clear that this case is intentionally empty.

client/client.go
```diff
@@ -858,37 +872,50 @@ func (r *NotaryRepository) validateRoot(rootJSON []byte) (*data.SignedRoot, erro
...
+	// locally managing targets keys only
+	case role == data.CanonicalTargetsRole:
+		if serverManagesKey {
```
endophage: we could move this check into the case:

```
case role == data.CanonicalTargetsRole && serverManagesKey:
    return ErrInvalidRemoteRole{...}
```

Same for the timestamp check.

endophage:
I don't have a strong opinion on collapsing the case/if statements.

LGTM

endophage:
Rebase and we can merge this :-)

docs/advanced_usage.md
```diff
@@ -90,11 +90,13 @@ subsection.
 In case of potential compromise, notary provides a CLI command for rotating keys. Currently, you can use the `notary key rotate` command to rotate the targets or snapshot keys.
 
 While the snapshot key is managed by the notary client by default, use the `notary key
```
riyazdf: nit: It seems like I missed this documentation piece -- it might be nice to also mention the Docker 1.11 default behavior for managing snapshot keys on the server here.

cyli: It looks like your documentation appears in the next paragraph.  :)  This just needed a rebase.

riyazdf: :tada: awesome

PR #572, opened 2016-02-16
Per metadata type mutation tests

cyli:
This completes a lot of the per-metadata-type tests listed in #361

### DIFF
--- a/client/client_update_test.go
+++ b/client/client_update_test.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/docker/notary/certs"
 	"github.com/docker/notary/passphrase"
+	"github.com/docker/notary/tuf"
 	"github.com/docker/notary/tuf/client"
 	"github.com/docker/notary/tuf/data"
 	"github.com/docker/notary/tuf/signed"
@@ -818,7 +819,29 @@ func TestUpdateRootRemoteCorruptedNoLocalCache(t *testing.T) {
 	if testing.Short() {
 		t.Skip("skipping test in short mode")
 	}
-	for _, testData := range waysToMessUpServer {
+
+	waysToMessUpServerRoot := waysToMessUpServer
+	for _, roleName := range data.BaseRoles {
+		waysToMessUpServerRoot = append(waysToMessUpServer,
+			swizzleExpectations{
+				desc: fmt.Sprintf("no %s keys", roleName),
+				expectErrs: []interface{}{
+					&certs.ErrValidationFail{}, signed.ErrRoleThreshold{}},
+				swizzle: func(s *testutils.MetadataSwizzler, role string) error {
+					return s.MutateRoot(func(r *data.Root) {
+						r.Roles[roleName].KeyIDs = []string{}
+					})
+				}},
+			swizzleExpectations{
+				desc:       fmt.Sprintf("no %s role", roleName),
+				expectErrs: []interface{}{tuf.ErrNotLoaded{}},
+				swizzle: func(s *testutils.MetadataSwizzler, role string) error {
+					return s.MutateRoot(func(r *data.Root) { delete(r.Roles, roleName) })
+				}},
+		)
+	}
+
+	for _, testData := range waysToMessUpServerRoot {
 		if testData.desc == "insufficient signatures" {
 			// Currently if we download the root during the bootstrap phase,
 			// we don't check for enough signatures to meet the threshold.  We
@@ -842,7 +865,32 @@ func TestUpdateRootRemoteCorruptedNoLocalCache(t *testing.T) {
 // because the fact that the timestamp hasn't changed should mean that we don't
 // have to re-download the root.
 func TestUpdateRootRemoteCorruptedCanUseLocalCache(t *testing.T) {
-	for _, testData := range waysToMessUpServer {
+	if testing.Short() {
+		t.Skip("skipping test in short mode")
+	}
+
+	waysToMessUpServerRoot := waysToMessUpServer
+	for _, roleName := range data.BaseRoles {
+		waysToMessUpServerRoot = append(waysToMessUpServer,
+			swizzleExpectations{
+				desc: fmt.Sprintf("no %s keys", roleName),
+				expectErrs: []interface{}{
+					&certs.ErrValidationFail{}, signed.ErrRoleThreshold{}},
+				swizzle: func(s *testutils.MetadataSwizzler, role string) error {
+					return s.MutateRoot(func(r *data.Root) {
+						r.Roles[roleName].KeyIDs = []string{}
+					})
+				}},
+			swizzleExpectations{
+				desc:       fmt.Sprintf("no %s role", roleName),
+				expectErrs: []interface{}{tuf.ErrNotLoaded{}},
+				swizzle: func(s *testutils.MetadataSwizzler, role string) error {
+					return s.MutateRoot(func(r *data.Root) { delete(r.Roles, roleName) })
+				}},
+		)
+	}
+
+	for _, testData := range waysToMessUpServerRoot {
 		testUpdateRemoteCorruptValidChecksum(t, updateOpts{
 			localCache: true,
 			forWrite:   false,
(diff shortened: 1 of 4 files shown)


cyli:
Possibly a semantic conflict with #573  (these tests may fail if that is merged first)

client/client_update_test.go
```diff
@@ -842,7 +865,32 @@ func TestUpdateRootRemoteCorruptedNoLocalCache(t *testing.T) {
...
+					return s.MutateRoot(func(r *data.Root) {
+						r.Roles[roleName].KeyIDs = []string{}
+					})
+				}},
+			swizzleExpectations{
+				desc:       fmt.Sprintf("no %s role", roleName),
+				expectErrs: []interface{}{tuf.ErrNotLoaded{}},
+				swizzle: func(s *testutils.MetadataSwizzler, role string) error {
+					return s.MutateRoot(func(r *data.Root) { delete(r.Roles, roleName) })
+				}},
```
riyazdf: these are great additions to the test and the one above.  Since it's the same code specifying the additional swizzler expectations, could we move these additional expectations into a helper function?

client/client_update_test.go
```diff
@@ -862,19 +910,43 @@ func TestUpdateRootRemoteCorruptedCannotUseLocalCache(t *testing.T) {
...
+		)
+	}
+
+	// if the timestamp role is missing, then the timestamp is broken,
+	// so it will revert to the prevoius timestamp
```
riyazdf: typo: `previous`

client/client_update_test.go
```diff
@@ -893,6 +965,53 @@ func TestUpdateNonRootRemoteCorruptedNoLocalCache(t *testing.T) {
 				role: role,
 			}, testData, true)
 		}
+
+		switch role {
+		case data.CanonicalSnapshotRole:
+			for _, missing := range []string{data.CanonicalRootRole, data.CanonicalTargetsRole} {
+				testData := swizzleExpectations{
+					desc: fmt.Sprintf("snapshot missing root meta checksum"),
+					// TODO: this should probably be always client.ErrMissingMeta
```
riyazdf: Out of curiosity, where is this json syntax error originating from?  It doesn't seem like this should affect the structural integrity of `data.Signed`, unless I'm missing something?

client/client_update_test.go
```diff
@@ -915,6 +1034,53 @@ func TestUpdateNonRootRemoteCorruptedCanUseLocalCache(t *testing.T) {
...
+				desc:       fmt.Sprintf("timestamp missing snapshot meta checksum"),
+				expectErrs: []interface{}{client.ErrMissingMeta{}},
+				swizzle: func(s *testutils.MetadataSwizzler, role string) error {
+					return s.MutateTimestamp(func(ts *data.Timestamp) {
+						delete(ts.Meta, data.CanonicalSnapshotRole)
+					})
+				},
+			}
+			// TODO: this should succeed, because it's an invalid timestamp metadata
+			// and hence should fall back on the the local cached timestamp
```
riyazdf: This is weird, perhaps this should resolve itself with #573?

client/client_update_test.go
cyli: Good call, thanks!  And doing so makes me realize that I was appending the wrong thing!

client/client_update_test.go
cyli: :(  Validation of the snapshot should really have failed - getting to the syntax error is kind of convoluted.  The error occurs when trying to download and load the root.

- [we get the size and checksum for root from the snapshot](https://github.com/docker/notary/blob/master/tuf/client/client.go#L135)
- there is no file metadata information, so `c.local.Snapshot.Signed.Meta[role]` would have returned the zero value and a `false` for `ok` - since it's a map of structs, it doesn't return `nil` but rather an empty `FileMeta` structure for the zero value
- we don't actually check whether `ok` is true or false, we just grab the `Length`, which is 0 on an empty `FileMeta` object, and the checksum.  An empty `FileMeta` has an empty map for `Hashes`, and getting `"sha256"` from it returns the zero value for that map, which is an empty list -I found this somewhat counter-intuitive, but [that does seem to be the case playing around with empty structs](https://play.golang.org/p/8GZm9WeBc2)
- then we download root using `downloadSigned`, specifying the size to be 0, and end up with an empty list of bytes for root, and because the checksum is empty, we don't bother checking it.
- we take this empty byte array for root and then attempt to JSON-unmarshal it, which results in a syntax error because [you can't unmarshal empty bytes](https://play.golang.org/p/htrncemSUI)

client/client_update_test.go
cyli: It's because no error occurs in the timestamp validation itself (if one does, we just used the cached timestamp) - the error fails during downloading the snapshot, since there's no checksum.

client/client_update_test.go
riyazdf: wow, thanks for tracking this down.  We should add these additional checks so that this errors much more gracefully.  Either in this PR or another one to keep it more manageable (and I can help implement)

cyli: That would be awesome, thanks!  I think some of this might eventually be obsoleted if we validate the metadata beforehand, but definitely we should

- check if the file checksum metadata on the snapshot before using it (I think we do that for downloading the snapshots and maybe the targets?)
- also fail if the sha256 checksum isn't present

If you could help with that that would be awesome!  Either here or in another PR would work.

riyazdf: I'll add to this PR :+1:

client/client_update_test.go
```diff
@@ -811,6 +811,36 @@ var waysToMessUpServer = []swizzleExpectations{
 		}},
 }
 
+var _waysToMessUpServerRoot []swizzleExpectations
+
+// A getter that also sets once through - we can just add these, but it's very easier to do it
```
endophage: This comments seems very jumbled :-P

client/client_update_test.go
```diff
@@ -811,6 +811,36 @@ var waysToMessUpServer = []swizzleExpectations{
 		}},
 }
 
+var _waysToMessUpServerRoot []swizzleExpectations
```
endophage: as `waysToMessUpServerRoot` returns this var, it seems like it should be declared inside that function?

endophage: oh, I see, you're trying to cache the result effectively, nevermind.

riyazdf:
these are great tests, LGTM (for everything except my commit)

tuf/testutils/swizzler.go
```diff
@@ -532,3 +537,141 @@ func (m *MetadataSwizzler) UpdateTimestampHash() error {
...
+// MutateRoot takes a function that mutates the root metadata - once done, it
+// serializes the root again
+func (m *MetadataSwizzler) MutateRoot(mutate func(*data.Root)) error {
+	signedThing, err := signedFromStore(m.MetadataCache, data.CanonicalRootRole)
+	if err != nil {
+		return err
+	}
+
+	var root data.Root
+	if err := json.Unmarshal(signedThing.Signed, &root); err != nil {
```
endophage: Is it intentional to avoid the `*FromSigned` functions?

tuf/testutils/swizzler.go
```diff
@@ -532,3 +537,141 @@ func (m *MetadataSwizzler) UpdateTimestampHash() error {
...
+
+	// get the original keys, in case the mutation messes with the signing keys
+	oldPubKeys, err := getPubKeys(m.CryptoService, signedThing, data.CanonicalRootRole)
+	if err != nil {
+		return err
+	}
+
+	mutate(&root)
+
+	rootBytes, err := json.MarshalCanonical(root)
```
endophage: Also `*.ToSigned`

tuf/testutils/swizzler.go
cyli: Yes, since those should eventually validate that the metadata is correct, and these functions can produce structurally incorrect metadata.

tuf/testutils/swizzler.go
cyli: Could you elaborate?  Wouldn't that produce a Signed object whose `Signed` field was canonically marshalled?

tuf/testutils/swizzler.go
cyli: Actually I can probably just grab the bytes from the store and directly unmarshal it to a `data.SignedRoot` object, if that would be more clear, since in this case we don't need to go through the signed stage

tuf/testutils/swizzler.go
cyli: Oh, sorry, yes, I can just replace the existing one :)  Changing

tuf/testutils/swizzler.go
cyli: Oh sorry, wait, I forgot, I use it get grab the old root keys out.

endophage:
LGTM pending CI

PR #573, opened 2016-02-16
Adds extra validation when calling `XFromSigned` functions

cyli:
To make sure that the per-metadata type validation always happens

### DIFF
--- a/tuf/client/client.go
+++ b/tuf/client/client.go
@@ -417,7 +417,7 @@ func (c *Client) downloadTargets(role string) error {
 			logrus.Error("Error getting targets file:", err)
 			return err
 		}
-		t, err := data.TargetsFromSigned(s)
+		t, err := data.TargetsFromSigned(s, role)
 		if err != nil {
 			return err
 		}
@@ -481,7 +481,7 @@ func (c Client) getTargetsFile(role string, snapshotMeta data.Files, consistent 
 		}
 		err := json.Unmarshal(raw, old)
 		if err == nil {
-			targ, err := data.TargetsFromSigned(old)
+			targ, err := data.TargetsFromSigned(old, role)
 			if err == nil {
 				version = targ.Signed.Version
 			} else {
--- a/tuf/client/client.go
+++ b/tuf/client/client.go
@@ -318,12 +318,12 @@ func (c *Client) downloadSnapshot() error {
 	logrus.Debug("Downloading Snapshot...")
 	role := data.CanonicalSnapshotRole
 	if c.local.Timestamp == nil {
-		return ErrMissingMeta{role: "snapshot"}
+		return data.ErrMissingMeta{Role: "snapshot"}
 	}
 	size := c.local.Timestamp.Signed.Meta[role].Length
 	expectedSha256, ok := c.local.Timestamp.Signed.Meta[role].Hashes["sha256"]
 	if !ok {
-		return ErrMissingMeta{role: "snapshot"}
+		return data.ErrMissingMeta{Role: "snapshot"}
 	}
 
 	var download bool
@@ -402,14 +402,14 @@ func (c *Client) downloadTargets(role string) error {
 			return err
 		}
 		if c.local.Snapshot == nil {
-			return ErrMissingMeta{role: role}
+			return data.ErrMissingMeta{Role: role}
 		}
 		snap := c.local.Snapshot.Signed
 		root := c.local.Root.Signed
 
 		s, err := c.getTargetsFile(role, snap.Meta, root.ConsistentSnapshot)
 		if err != nil {
-			if _, ok := err.(ErrMissingMeta); ok && role != data.CanonicalTargetsRole {
+			if _, ok := err.(data.ErrMissingMeta); ok && role != data.CanonicalTargetsRole {
 				// if the role meta hasn't been published,
 				// that's ok, continue
 				continue
@@ -458,11 +458,11 @@ func (c Client) getTargetsFile(role string, snapshotMeta data.Files, consistent 
 	// require role exists in snapshots
 	roleMeta, ok := snapshotMeta[role]
 	if !ok {
-		return nil, ErrMissingMeta{role: role}
+		return nil, data.ErrMissingMeta{Role: role}
 	}
 	expectedSha256, ok := snapshotMeta[role].Hashes["sha256"]
 	if !ok {
-		return nil, ErrMissingMeta{role: role}
+		return nil, data.ErrMissingMeta{Role: role}
 	}
 
 	// try to get meta file from content addressed cache
--- a/tuf/client/client.go
+++ b/tuf/client/client.go
@@ -318,7 +318,7 @@ func (c *Client) downloadSnapshot() error {
 	logrus.Debug("Downloading Snapshot...")
 	role := data.CanonicalSnapshotRole
 	if c.local.Timestamp == nil {
-		return data.ErrMissingMeta{Role: "snapshot"}
+		return tuf.ErrNotLoaded{Role: data.CanonicalTimestampRole}
 	}
 	size := c.local.Timestamp.Signed.Meta[role].Length
 	expectedSha256, ok := c.local.Timestamp.Signed.Meta[role].Hashes["sha256"]
@@ -402,7 +402,7 @@ func (c *Client) downloadTargets(role string) error {
 			return err
 		}
 		if c.local.Snapshot == nil {
-			return data.ErrMissingMeta{Role: role}
+			return tuf.ErrNotLoaded{Role: data.CanonicalSnapshotRole}
 		}
 		snap := c.local.Snapshot.Signed
 		root := c.local.Root.Signed
--- a/tuf/data/errors.go
+++ b/tuf/data/errors.go
@@ -2,14 +2,14 @@ package data
 
 import "fmt"
 
-// ErrInvalidMeta is the error to be returned when metadata is invalid
-type ErrInvalidMeta struct {
-	Role string
-	Msg  string
+// ErrInvalidMetadata is the error to be returned when metadata is invalid
+type ErrInvalidMetadata struct {
+	role string
+	msg  string
 }
 
-func (e ErrInvalidMeta) Error() string {
-	return fmt.Sprintf("%s type metadata invalid: %s", e.Role, e.Msg)
+func (e ErrInvalidMetadata) Error() string {
+	return fmt.Sprintf("%s type metadata invalid: %s", e.role, e.msg)
 }
 
 // ErrMissingMeta - couldn't find the FileMeta object for a role or target
--- a/tuf/data/errors.go
+++ b/tuf/data/errors.go
@@ -2,6 +2,16 @@ package data
 
 import "fmt"
 
+// ErrInvalidMeta is the error to be returned when metadata is invalid
+type ErrInvalidMeta struct {
+	Role string
+	Msg  string
+}
+
+func (e ErrInvalidMeta) Error() string {
+	return fmt.Sprintf("%s type metadata invalid: %s", e.Role, e.Msg)
+}
+
 // ErrMissingMeta - couldn't find the FileMeta object for a role or target
 type ErrMissingMeta struct {
 	Role string
--- a/tuf/data/errors.go
+++ b/tuf/data/errors.go
@@ -1,0 +1,12 @@
+package data
+
+import "fmt"
+
+// ErrMissingMeta - couldn't find the FileMeta object for a role or target
+type ErrMissingMeta struct {
+	Role string
+}
+
+func (e ErrMissingMeta) Error() string {
+	return fmt.Sprintf("tuf: sha256 checksum required for %s", e.Role)
+}
--- a/tuf/data/roles.go
+++ b/tuf/data/roles.go
@@ -214,18 +214,6 @@ type Role struct {
 	Paths []string `json:"paths,omitempty"`
 }
 
-func isValidRootRoleStructure(r RootRole, validKeys Keys) error {
-	if r.Threshold < 1 {
-		return fmt.Errorf("invalid threshold: %v ", r.Threshold)
-	}
-	for _, keyID := range r.KeyIDs {
-		if _, ok := validKeys[keyID]; !ok {
-			return fmt.Errorf("key ID %s specified without corresponding key", keyID)
-		}
-	}
-	return nil
-}
-
 // NewRole creates a new Role object from the given parameters
 func NewRole(name string, threshold int, keyIDs, paths []string) (*Role, error) {
 	if IsDelegation(name) {
--- a/tuf/data/roles.go
+++ b/tuf/data/roles.go
@@ -214,6 +214,18 @@ type Role struct {
 	Paths []string `json:"paths,omitempty"`
 }
 
+func isValidRootRoleStructure(r RootRole, validKeys Keys) error {
+	if r.Threshold < 1 {
+		return fmt.Errorf("invalid threshold: %v ", r.Threshold)
+	}
+	for _, keyID := range r.KeyIDs {
+		if _, ok := validKeys[keyID]; !ok {
+			return fmt.Errorf("key ID %s specified without corresponding key", keyID)
+		}
+	}
+	return nil
+}
+
 // NewRole creates a new Role object from the given parameters
 func NewRole(name string, threshold int, keyIDs, paths []string) (*Role, error) {
 	if IsDelegation(name) {
--- a/tuf/data/root.go
+++ b/tuf/data/root.go
@@ -33,12 +33,9 @@ func isValidRootStructure(r Root) error {
 		return ErrInvalidMetadata{
 			role: CanonicalRootRole, msg: fmt.Sprintf("expected type %s, not %s", expectedType, r.Type)}
 	}
-	if len(r.Roles) < 4 {
-		return ErrInvalidMetadata{role: CanonicalRootRole, msg: "does not have all required roles"}
-	} else if len(r.Roles) > 4 {
-		return ErrInvalidMetadata{role: CanonicalRootRole, msg: "specifies too many roles"}
-	}
 
+	// all the base roles MUST appear in the root.json - other roles are allowed,
+	// but other than the mirror role (not currently supported) are out of spec
 	for _, roleName := range BaseRoles {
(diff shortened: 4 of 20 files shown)


tuf/client/client.go
```diff
@@ -402,7 +402,7 @@ func (c *Client) downloadTargets(role string) error {
 			return err
 		}
 		if c.local.Snapshot == nil {
-			return ErrMissingMeta{role: role}
+			return data.ErrMissingMeta{Role: role}
```
riyazdf: It seems like the error here should be referencing `Role: "snapshot"`

tuf/data/errors.go
```diff
@@ -0,0 +1,22 @@
...
+}
+
+// ErrMissingMeta - couldn't find the FileMeta object for a role or target
+type ErrMissingMeta struct {
+	Role string
+}
+
+func (e ErrMissingMeta) Error() string {
+	return fmt.Sprintf("tuf: sha256 checksum required for %s", e.Role)
+}
```
riyazdf: thanks for moving these errors, this makes a lot more sense :+1:

tuf/data/root.go
```diff
@@ -23,14 +24,53 @@ type Root struct {
 	ConsistentSnapshot bool                 `json:"consistent_snapshot"`
 }
 
+// isValidRoot returns an error, or nil, depending on whether the content of the struct
+// is valid for root metadata.  This does not check signatures or expiry, just that
+// the metadata content is valid.
+func isValidRoot(r Root) error {
```
riyazdf: your comment is really helpful, but I'm wondering if we should also make it more explicit in the function name that we are only validating structure and not all content?  Maybe, `isValidRootStructure`?

tuf/data/root_test.go
```diff
@@ -96,3 +123,83 @@ func TestMarshalJSONMarshallingErrorsPropagated(t *testing.T) {
...
+		root.Signed.Roles["extraneous"] = &RootRole{KeyIDs: []string{"key3"}, Threshold: 1}
+		_, err = rootToSignedAndBack(t, root)
+		require.IsType(t, ErrInvalidMeta{}, err)
+	}
+
+	// Add an extra role without messing with the others
+	root := validRootTemplate()
+	root.Signed.Roles["extraneous"] = &RootRole{KeyIDs: []string{"key3"}, Threshold: 1}
+	_, err = rootToSignedAndBack(t, root)
+	require.IsType(t, ErrInvalidMeta{}, err)
```
riyazdf: This is an awesome test!

tuf/data/snapshot.go
```diff
@@ -23,6 +24,27 @@ type Snapshot struct {
 	Meta    Files     `json:"meta"`
 }
 
+// isValidSnapshot returns an error, or nil, depending on whether the content of the struct
+// is valid for snapshot metadata.  This does not check signatures or expiry, just that
+// the metadata content is valid.
+func isValidSnapshot(s Snapshot) error {
```
riyazdf: Same comment as on `isValidRoot`.

tuf/data/snapshot.go
```diff
@@ -23,6 +24,27 @@ type Snapshot struct {
...
+	expectedType := TUFTypes[CanonicalSnapshotRole]
+	if s.Type != expectedType {
+		return ErrInvalidMeta{
+			Role: CanonicalSnapshotRole, Msg: fmt.Sprintf("expected type %s, not %s", expectedType, s.Type)}
+	}
+
+	for _, role := range []string{CanonicalRootRole, CanonicalTargetsRole} {
```
riyazdf: I'm wondering if we should do any validation on delegation roles that may appear in the `snapshot.json` -- we may not know the delegation names but maybe we could still iterate through the roles to check that they have checksum information?

tuf/data/snapshot_test.go
```diff
@@ -1,28 +1,175 @@
...
+func validSnapshotTemplate() *SignedSnapshot {
+	return &SignedSnapshot{
 		Signed: Snapshot{
-			Type:    TUFTypes["snapshot"],
-			Version: 0,
-			Expires: DefaultExpires("snapshot"),
-			Meta: Files{
+			Type: "Snapshot", Version: 1, Expires: time.Now(), Meta: Files{
 				CanonicalRootRole:    FileMeta{},
 				CanonicalTargetsRole: FileMeta{},
```
riyazdf: Could we add a test for a snapshot template with delegation roles?

tuf/data/targets.go
```diff
@@ -23,6 +25,38 @@ type Targets struct {
...
+// is valid for targets metadata.  This does not check signatures or expiry, just that
+// the metadata content is valid.
+func isValidTargets(t Targets, roleName string) error {
+	// even if it's a delegated role, the metadata type is "Targets"
+	expectedType := TUFTypes[CanonicalTargetsRole]
+	if t.Type != expectedType {
+		return ErrInvalidMeta{
+			Role: roleName, Msg: fmt.Sprintf("expected type %s, not %s", expectedType, t.Type)}
+	}
+
```
riyazdf: Should we validate `roleName` as well, before iterating through the delegation roles?

tuf/data/snapshot.go
cyli: Oh you mean validate that these aren't empty file metas?

tuf/data/targets.go
cyli: Oh good point, thanks!

tuf/data/targets_test.go
```diff
@@ -0,0 +1,204 @@
...
+		if IsDelegation(roleName) {
+			delgRole.Name = path.Join(CanonicalTargetsRole, "z")
+			s, err := targets.ToSigned()
+			require.NoError(t, err)
+			_, err = TargetsFromSigned(s, roleName)
+			require.Error(t, err)
+			require.IsType(t, ErrInvalidMeta{}, err)
+		}
+	}
+}
```
riyazdf: Could we add a test with an invalid `roleName` passed to `TargetsFromSigned`?

tuf/data/snapshot.go
riyazdf: Yes, I think we should validate that:
 - all roles in `snapshot.json` have checksum information
 - root and targets metas must exist
 - all other listed roles must be valid delegation names

tuf/data/targets.go
```diff
@@ -23,6 +25,38 @@ type Targets struct {
 	Delegations Delegations `json:"delegations,omitempty"`
 }
 
+// isValidTargets returns an error, or nil, depending on whether the content of the struct
+// is valid for targets metadata.  This does not check signatures or expiry, just that
+// the metadata content is valid.
+func isValidTargets(t Targets, roleName string) error {
```
riyazdf: Same comment as on `isValidRoot`.

tuf/data/timestamp.go
```diff
@@ -22,6 +23,22 @@ type Timestamp struct {
 	Meta    Files     `json:"meta"`
 }
 
+// isValidTimestamp returns an error, or nil, depending on whether the content of the struct
+// is valid for timestamp metadata.  This does not check signatures or expiry, just that
+// the metadata content is valid.
+func isValidTimestamp(t Timestamp) error {
```
riyazdf: Same comment as on `isValidRoot`.

tuf/data/snapshot.go
riyazdf: Actually after discussion with @cyli: we may not want to validate names of roles other than root and targets, in case we want to chain previous snapshots.

tuf/data/root.go
cyli: Good suggestion, thanks!

tuf/client/client.go
cyli: :|  it should probably be some kind state error, like that snapshot should have been loaded by now, but yes `Role: "snapshot"` would definitely be better than now, thanks :)

tuf/data/snapshot.go
cyli: Some discussion with @endophage: the lack of hash information in the checksum can be verified during the checksum validation, since the user might be using a role/target to demarcate something other than that a file exists, and hence the checksum may not be relevant?

riyazdf: I agree we can verify later, but I guess I don't understand why or when a user would place arbitrary roles/targets in the snapshot.json -- is there a concrete use case that we should support?

cyli: Not sure - @endophage?

tuf/client/client.go
cyli: Oh, hm... looks like it's used when we can't get the file meta for a file?

cyli: How do you feel about `tuf.ErrNotLoaded` instead?

riyazdf: I agree that `tuf.ErrNotLoaded` is more fitting here :+1:

tuf/data/snapshot_test.go
cyli: Good idea!  I've added a delegation role to the default template.

tuf/data/targets.go
cyli: Done!

tuf/data/targets_test.go
cyli: Done!

tuf/data/snapshot_test.go
riyazdf: :+1:

tuf/data/snapshot.go
cyli: I guess technically the checksum use any kind of hash algorithm, and I guess just checking that the hash isn't empty isn't that useful of a check for us?   

Actually, I think I'm going to add `data.FileMeta.Validate(content []byte, filename string) error` in this PR, that will return a checksum mismatch if there is no checksum that we support, and otherwise will validate every kind of checksum that we do support.

tuf/data/root.go
```diff
@@ -23,14 +24,53 @@ type Root struct {
...
+	}
+	if len(r.Roles) < 4 {
+		return ErrInvalidMeta{Role: CanonicalRootRole, Msg: "does not have all required roles"}
+	} else if len(r.Roles) > 4 {
+		return ErrInvalidMeta{Role: CanonicalRootRole, Msg: "specifies too many roles"}
+	}
+
+	for _, roleName := range BaseRoles {
+		roleObj, ok := r.Roles[roleName]
+		if !ok || roleObj == nil {
```
endophage: I'm thinking this should just check `!ok` then we should have a separate "role validating" function that take the `roleObj` and the key list and it'll do the rest of the checks. That function could then be used as a sub-check for inspecting delegated roles too.

tuf/data/errors.go
```diff
@@ -0,0 +1,22 @@
+package data
+
+import "fmt"
+
+// ErrInvalidMeta is the error to be returned when metadata is invalid
+type ErrInvalidMeta struct {
+	Role string
```
endophage: Any reason to export these fields?

tuf/data/root.go
cyli: I think we'd have to check if the role is nil, because that's a special case error for root roles, but I will factor out the rest of the key checks into a helper function.

tuf/data/snapshot.go
endophage: There shouldn't be arbirtary data in the snapshot, but we shouldn't fail immediately if the user's specific requirements don't encounter the invalid role. Just trying to ensure we're as robust as possible.

cyli: @endophage Shouldn't we validate that the root and targets (maybe not others) have a "sha256" checksum, though?   If it's not there, we can't do consistent downloads, even if they've provided other checksums.

tuf/data/errors.go
```diff
@@ -0,0 +1,22 @@
...
+	msg  string
+}
+
+func (e ErrInvalidMetadata) Error() string {
+	return fmt.Sprintf("%s type metadata invalid: %s", e.role, e.msg)
+}
+
+// ErrMissingMeta - couldn't find the FileMeta object for a role or target
+type ErrMissingMeta struct {
+	Role string
```
endophage: any reason to export?

tuf/data/roles.go
```diff
@@ -214,6 +214,18 @@ type Role struct {
 	Paths []string `json:"paths,omitempty"`
 }
 
+func isValidRootRoleStructure(r RootRole, validKeys Keys) error {
```
endophage: I know we've talked about it before, did we decide we should enforce that there must be enough keys to cover the Threshold?

tuf/data/root.go
```diff
@@ -23,14 +24,45 @@ type Root struct {
...
+// isValidRootStructure returns an error, or nil, depending on whether the content of the struct
+// is valid for root metadata.  This does not check signatures or expiry, just that
+// the metadata content is valid.
+func isValidRootStructure(r Root) error {
+	expectedType := TUFTypes[CanonicalRootRole]
+	if r.Type != expectedType {
+		return ErrInvalidMetadata{
+			role: CanonicalRootRole, msg: fmt.Sprintf("expected type %s, not %s", expectedType, r.Type)}
+	}
+	if len(r.Roles) < 4 {
```
endophage: Unnecessary given we iterate through the BaseRoles and require all of them to be present and valid.

tuf/data/root.go
```diff
@@ -23,14 +24,45 @@ type Root struct {
...
+		return ErrInvalidMetadata{role: CanonicalRootRole, msg: "does not have all required roles"}
+	} else if len(r.Roles) > 4 {
```
endophage: It's OK to have additional roles, particularly as we may add the mirrors role at some point.

endophage:
Wondering if we can/need/want to test duplicate role entries in the root roles map: https://play.golang.org/p/pXGr_lvDMY

Also thinking we should add key ID verification on incoming json. Most reliable place to do it would be in tuf/data/keys.go when we unmarshal but that adds another random piece of verification in another random place.

riyazdf:
@endophage not sure if I'm missing something, but in your example the map itself only keeps the latest entry for `foo` https://play.golang.org/p/lKyKXRj96G

tuf/data/root.go
```diff
@@ -62,7 +94,7 @@ func (r SignedRoot) ToSigned() (*Signed, error) {
 }
 
 // MarshalJSON returns the serialized form of SignedRoot as bytes
-func (r SignedRoot) MarshalJSON() ([]byte, error) {
+func (r *SignedRoot) MarshalJSON() ([]byte, error) {
```
endophage: Shouldn't have a pointer receiver. Marshalling shouldn't modify the original.

tuf/data/root.go
```diff
@@ -42,7 +74,7 @@ func NewRoot(keys map[string]PublicKey, roles map[string]*RootRole, consistent b
 }
 
 // ToSigned partially serializes a SignedRoot for further signing
-func (r SignedRoot) ToSigned() (*Signed, error) {
+func (r *SignedRoot) ToSigned() (*Signed, error) {
```
endophage: Shouldn't have a pointer receiver. No modification happening here.

endophage:
@riyaz exactly. I don't know there's a way to meaningfully exploit that hence the "can/need/want". I know distribution had an exploitable instance of it in the legacy, obsoleted, signing system.

tuf/data/root.go
cyli: Oh, ok, forgot about that one

tuf/data/roles.go
cyli: I was thinking about that, but I wasn't sure which way we decided - is there ever a use case for removing all keys and invalidating the role without deleting it?

endophage: Ah yep. I'm ok with leaving threshold checks until signature validation, just remembered we'd had the discussion and couldn't remember if we'd decided to change anything.

riyazdf:
@endophage okay I see what you're saying -- we wouldn't be able to verify at the `isValidRootStructure` level.  I agree with you that it isn't exploitable since this should be mitigated in the current code this when verifying signatures, right?  We could write a custom unmarshaller to check for duplicate keys but I'm not convinced by the complexity.

endophage:
Probably not worth while. As long as there's consensus and understanding I'm happy to just say "last appearance of the key wins" during unmarshalling. The signing and checksumming should mitigate any problems.

tuf/data/errors.go
cyli: I moved this from `tuf/client/errors.go`, and the `tuf/client/client.go` still returns this type of error.  Should I just create a constructor for the error instead?

tuf/data/snapshot.go
riyazdf: revisiting this conversation: I think we should keep the sha256 checksum check for root and targets -- especially due to its importance for consistent downloads.  I'm good with the current functionality of `isValidSnapshotStructure` on the latest commit of this PR

riyazdf:
@endophage: agreed, I'm fine with current behavior here.

I think the only potential unresolved bit on this PR is `isValidSnapshotStructure`, but I agree with the current implementation since it keeps consistent with `isValidTimestampStructure`'s checking and ensures we have metadata for targets and root.

LGTM

tuf/data/errors.go
endophage: Nah, I'm just being pedantic. We need to overhaul our errors and make sure they're defined and returned in the right places. Packages importing and creating eachother's errors is probably a sign that we need to look at our code organization a bit more.

tuf/data/root.go
```diff
@@ -23,14 +24,42 @@ type Root struct {
...
+	// all the base roles MUST appear in the root.json - other roles are allowed,
+	// but other than the mirror role (not currently supported) are out of spec
+	for _, roleName := range BaseRoles {
+		roleObj, ok := r.Roles[roleName]
+		if !ok || roleObj == nil {
+			return ErrInvalidMetadata{
+				role: CanonicalRootRole, msg: fmt.Sprintf("missing %s role specification", roleName)}
+		}
+		if err := isValidRootRoleStructure(*roleObj, r.Keys); err != nil {
+			return ErrInvalidMetadata{
```
endophage: Maybe have `isValidRootRoleStructure` return this error so that we don't have to create it here and on `isValidTargetsStructure`?

endophage:
IRL decision with @cyli and @riyazdf: we're going to leave `isValidSnapshotStructure` as written for now.

LGTM

PR #575, opened 2016-02-17
Update integration test to exclude yubikeys from importing/exporting non-root keys

riyazdf:
It seems like jenkins got into a weird state and didn't actually test this code with yubikeys until now. Without this PR, this test will fail when using a yubikey because it cannot export keys, nor can it import non-root keys.  `TestClientKeyImportExportRootOnly` still tests that importing a root key on a yubikey functions as expected.

@cyli and I investigated the jenkins build configuration, and we've fixed it to ensure it pulls the latest commit off master.


cyli:
LGTM!  Thank you for fixing so fast!

PR #577, opened 2016-02-18
WalkTargets on tuf repo for targets and delegations

### LINKED ISSUE #560: Tree walking for notary delegation
This would be something os.Walk style.

Depends on #556

riyazdf:
Defines a `WalkTargets` that walks the targets and delegations structure in a uniform way, and we can pass closures to perform special visit functions to list targets, get targets by name, and get delegation roles.  Also used for adding/removing targets and delegation roles.

This PR also removes unused code that would also find targets or get target meta by walking the targets structure.

Closes #560

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -391,18 +391,19 @@ func (r *NotaryRepository) ListTargets(roles ...string) ([]*TargetWithRole, erro
 	targets := make(map[string]*TargetWithRole)
 	for _, role := range roles {
 		// Define a visitor function to populate the targets map in priority order
-		listVisitorFunc := func(tgt *data.SignedTargets, roleName string) error {
+		listVisitorFunc := func(tgt *data.SignedTargets, validRole data.DelegationRole) error {
 			if tgt == nil {
 				return tuf.ErrContinueWalk{}
 			}
 			// We found targets so we should try to add them to our targets map
 			for targetName, targetMeta := range tgt.Signed.Targets {
 				// Follow the priority by not overriding previously set targets
-				if _, ok := targets[targetName]; ok {
+				// and check that this path is valid with this role
+				if _, ok := targets[targetName]; ok || !validRole.CheckPaths(targetName) {
 					continue
 				}
 				targets[targetName] =
-					&TargetWithRole{Target: Target{Name: targetName, Hashes: targetMeta.Hashes, Length: targetMeta.Length}, Role: roleName}
+					&TargetWithRole{Target: Target{Name: targetName, Hashes: targetMeta.Hashes, Length: targetMeta.Length}, Role: validRole.Name}
 			}
 			return tuf.ErrContinueWalk{}
 		}
@@ -437,15 +438,15 @@ func (r *NotaryRepository) GetTargetByName(name string, roles ...string) (*Targe
 	var resultRoleName string
 	for _, role := range roles {
 		// Define a visitor function to find the specified target
-		getTargetVisitorFunc := func(tgt *data.SignedTargets, roleName string) error {
+		getTargetVisitorFunc := func(tgt *data.SignedTargets, validRole data.DelegationRole) error {
 			var ok bool
 			if tgt == nil {
 				return tuf.ErrContinueWalk{}
 			}
-			// We found the target so we should stop our walk
-			// and set the resultMeta and resultRoleName variables
+			// We found the target and validated path compatiblity in our walk,
+			// so we should stop our walk and set the resultMeta and resultRoleName variables
 			if resultMeta, ok = tgt.Signed.Targets[name]; ok {
-				resultRoleName = roleName
+				resultRoleName = validRole.Name
 				return tuf.ErrStopWalk{}
 			}
 			return tuf.ErrContinueWalk{}
--- a/client/client.go
+++ b/client/client.go
@@ -23,6 +23,7 @@ import (
 	"github.com/docker/notary/tuf/data"
 	"github.com/docker/notary/tuf/signed"
 	"github.com/docker/notary/tuf/store"
+	"github.com/docker/notary/tuf/utils"
 )
 
 func init() {
@@ -390,11 +391,11 @@ func (r *NotaryRepository) ListTargets(roles ...string) ([]*TargetWithRole, erro
 	}
 	targets := make(map[string]*TargetWithRole)
 	for _, role := range roles {
+		// Define an array of roles to skip for this walk (see IMPORTANT comment above)
+		skipRoles := utils.StrSliceRemove(roles, role)
+
 		// Define a visitor function to populate the targets map in priority order
-		listVisitorFunc := func(tgt *data.SignedTargets, validRole data.DelegationRole) error {
-			if tgt == nil {
-				return tuf.ErrContinueWalk{}
-			}
+		listVisitorFunc := func(tgt *data.SignedTargets, validRole data.DelegationRole) interface{} {
 			// We found targets so we should try to add them to our targets map
 			for targetName, targetMeta := range tgt.Signed.Targets {
 				// Follow the priority by not overriding previously set targets
@@ -405,9 +406,9 @@ func (r *NotaryRepository) ListTargets(roles ...string) ([]*TargetWithRole, erro
 				targets[targetName] =
 					&TargetWithRole{Target: Target{Name: targetName, Hashes: targetMeta.Hashes, Length: targetMeta.Length}, Role: validRole.Name}
 			}
-			return tuf.ErrContinueWalk{}
+			return nil
 		}
-		r.tufRepo.WalkTargets("", role, listVisitorFunc)
+		r.tufRepo.WalkTargets("", role, listVisitorFunc, skipRoles...)
 	}
 
 	var targetList []*TargetWithRole
@@ -436,23 +437,27 @@ func (r *NotaryRepository) GetTargetByName(name string, roles ...string) (*Targe
 	}
 	var resultMeta data.FileMeta
 	var resultRoleName string
+	var foundTarget bool
 	for _, role := range roles {
+		// Define an array of roles to skip for this walk (see IMPORTANT comment above)
+		skipRoles := utils.StrSliceRemove(roles, role)
+
 		// Define a visitor function to find the specified target
-		getTargetVisitorFunc := func(tgt *data.SignedTargets, validRole data.DelegationRole) error {
-			var ok bool
+		getTargetVisitorFunc := func(tgt *data.SignedTargets, validRole data.DelegationRole) interface{} {
 			if tgt == nil {
-				return tuf.ErrContinueWalk{}
+				return nil
 			}
 			// We found the target and validated path compatibility in our walk,
 			// so we should stop our walk and set the resultMeta and resultRoleName variables
-			if resultMeta, ok = tgt.Signed.Targets[name]; ok {
+			if resultMeta, foundTarget = tgt.Signed.Targets[name]; foundTarget {
 				resultRoleName = validRole.Name
-				return tuf.ErrStopWalk{}
+				return tuf.StopWalk{}
 			}
-			return tuf.ErrContinueWalk{}
+			return nil
 		}
-		err = r.tufRepo.WalkTargets(name, role, getTargetVisitorFunc)
-		if err == nil {
+		err = r.tufRepo.WalkTargets(name, role, getTargetVisitorFunc, skipRoles...)
+		// Check that we didn't error, and that we assigned to our target
+		if err == nil && foundTarget {
 			return &TargetWithRole{Target: Target{Name: name, Hashes: resultMeta.Hashes, Length: resultMeta.Length}, Role: resultRoleName}, nil
 		}
 	}
--- a/client/client.go
+++ b/client/client.go
@@ -390,10 +390,23 @@ func (r *NotaryRepository) ListTargets(roles ...string) ([]*TargetWithRole, erro
 	}
 	targets := make(map[string]*TargetWithRole)
 	for _, role := range roles {
-		// we don't need to do anything special with removing role from
-		// roles because listSubtree always processes role and only excludes
-		// descendant delegations that appear in roles.
-		r.listSubtree(targets, role, roles...)
+		// Define a visitor function to populate the targets map in priority order
+		listVisitorFunc := func(tgt *data.SignedTargets, roleName string) error {
+			if tgt == nil {
+				return tuf.ErrContinueWalk{}
+			}
+			// We found targets so we should try to add them to our targets map
+			for targetName, targetMeta := range tgt.Signed.Targets {
+				// Follow the priority by not overriding previously set targets
+				if _, ok := targets[targetName]; ok {
+					continue
+				}
+				targets[targetName] =
+					&TargetWithRole{Target: Target{Name: targetName, Hashes: targetMeta.Hashes, Length: targetMeta.Length}, Role: roleName}
+			}
+			return tuf.ErrContinueWalk{}
+		}
+		r.tufRepo.WalkTargets("", role, listVisitorFunc)
 	}
 
 	var targetList []*TargetWithRole
@@ -404,34 +417,6 @@ func (r *NotaryRepository) ListTargets(roles ...string) ([]*TargetWithRole, erro
 	return targetList, nil
 }
 
-func (r *NotaryRepository) listSubtree(targets map[string]*TargetWithRole, role string, exclude ...string) {
-	excl := make(map[string]bool)
-	for _, r := range exclude {
-		excl[r] = true
-	}
-	roles := []string{role}
-	for len(roles) > 0 {
-		role = roles[0]
-		roles = roles[1:]
-		tgts, ok := r.tufRepo.Targets[role]
-		if !ok {
-			// not every role has to exist
-			continue
-		}
-		for name, meta := range tgts.Signed.Targets {
-			if _, ok := targets[name]; !ok {
-				targets[name] = &TargetWithRole{
-					Target: Target{Name: name, Hashes: meta.Hashes, Length: meta.Length}, Role: role}
-			}
-		}
-		for _, d := range tgts.Signed.Delegations.Roles {
-			if !excl[d.Name] {
-				roles = append(roles, d.Name)
-			}
-		}
-	}
-}
-
 // GetTargetByName returns a target given a name. If no roles are passed
 // it uses the targets role and does a search of the entire delegation
 // graph, finding the first entry in a breadth first search of the delegations.
@@ -440,7 +425,7 @@ func (r *NotaryRepository) listSubtree(targets map[string]*TargetWithRole, role 
 // will be returned
 // See the IMPORTANT section on ListTargets above. Those roles also apply here.
 func (r *NotaryRepository) GetTargetByName(name string, roles ...string) (*TargetWithRole, error) {
-	c, err := r.Update(false)
+	_, err := r.Update(false)
 	if err != nil {
 		return nil, err
 	}
@@ -448,11 +433,26 @@ func (r *NotaryRepository) GetTargetByName(name string, roles ...string) (*Targe
 	if len(roles) == 0 {
 		roles = append(roles, data.CanonicalTargetsRole)
 	}
+	var resultMeta data.FileMeta
+	var resultRoleName string
 	for _, role := range roles {
-		meta, foundRole := c.TargetMeta(role, name, roles...)
-		if meta != nil {
-			return &TargetWithRole{
-				Target: Target{Name: name, Hashes: meta.Hashes, Length: meta.Length}, Role: foundRole}, nil
+		// Define a visitor function to find the specified target
+		getTargetVisitorFunc := func(tgt *data.SignedTargets, roleName string) error {
+			var ok bool
+			if tgt == nil {
+				return tuf.ErrContinueWalk{}
+			}
+			// We found the target so we should stop our walk
+			// and set the resultMeta and resultRoleName variables
+			if resultMeta, ok = tgt.Signed.Targets[name]; ok {
+				resultRoleName = roleName
+				return tuf.ErrStopWalk{}
+			}
+			return tuf.ErrContinueWalk{}
+		}
+		err = r.tufRepo.WalkTargets(name, role, getTargetVisitorFunc)
+		if err == nil {
+			return &TargetWithRole{Target: Target{Name: name, Hashes: resultMeta.Hashes, Length: resultMeta.Length}, Role: resultRoleName}, nil
 		}
 	}
 	return nil, fmt.Errorf("No trust data for %s", name)
--- a/client/client.go
+++ b/client/client.go
@@ -443,7 +443,7 @@ func (r *NotaryRepository) GetTargetByName(name string, roles ...string) (*Targe
 			if tgt == nil {
 				return tuf.ErrContinueWalk{}
 			}
-			// We found the target and validated path compatiblity in our walk,
+			// We found the target and validated path compatibility in our walk,
 			// so we should stop our walk and set the resultMeta and resultRoleName variables
 			if resultMeta, ok = tgt.Signed.Targets[name]; ok {
 				resultRoleName = validRole.Name
--- a/client/helpers.go
+++ b/client/helpers.go
@@ -79,11 +79,13 @@ func changeTargetsDelegation(repo *tuf.Repo, c changelist.Change) error {
 		if err != nil {
 			return err
 		}
-		r, _, err := repo.GetDelegation(c.Scope())
-		if _, ok := err.(data.ErrNoSuchRole); err != nil && !ok {
-			// error that wasn't ErrNoSuchRole
+		_, err = repo.GetDelegationRole(c.Scope())
+		if _, ok := err.(data.ErrInvalidRole); err != nil && !ok {
+			// error that wasn't ErrInvalidRole
 			return err
 		}
+
+		r, _, _ := repo.GetDelegation(c.Scope())
 		if err == nil {
 			// role existed, attempt to merge paths and keys
 			if err := r.AddPaths(td.AddPaths); err != nil {
@@ -103,10 +105,11 @@ func changeTargetsDelegation(repo *tuf.Repo, c changelist.Change) error {
 		if err != nil {
 			return err
 		}
-		r, keys, err := repo.GetDelegation(c.Scope())
+		_, err = repo.GetDelegationRole(c.Scope())
 		if err != nil {
 			return err
 		}
+		r, keys, _ := repo.GetDelegation(c.Scope())
 
 		// We need to translate the keys from canonical ID to TUF ID for compatibility
 		canonicalToTUFID := make(map[string]string)
@@ -125,8 +128,7 @@ func changeTargetsDelegation(repo *tuf.Repo, c changelist.Change) error {
 
 		// If we specify the only keys left delete the role, else just delete specified keys
 		if strings.Join(r.KeyIDs, ";") == strings.Join(removeTUFKeyIDs, ";") && len(td.AddKeys) == 0 {
-			r := data.Role{Name: c.Scope()}
-			return repo.DeleteDelegation(r)
+			return repo.DeleteDelegation(c.Scope())
 		}
 		// if we aren't deleting and the role exists, merge
 		if err := r.AddPaths(td.AddPaths); err != nil {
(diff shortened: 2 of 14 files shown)


tuf/tuf.go
```diff
@@ -49,6 +49,20 @@ func (err ErrNotLoaded) Error() string {
 	return fmt.Sprintf("%s role has not been loaded", err.role)
 }
 
+// ErrStopWalk - used by visitor functions to signal WalkTargets to stop walking
+type ErrStopWalk struct{}
```
riyazdf: This may not be necessary, since this PR as is will also stop walking on returning `nil`.  When @endophage and I discussed the design of the walker it seemed useful to have different states for "finishing" a walk and "aborting" a walk but the use case hasn't presented itself so far.  I can delete this or leave this in, depending on what we think.

endophage:
This is looking good so far. Try adding some of the write functions to make sure it works for those too :-)

tuf/tuf.go
```diff
@@ -564,34 +578,6 @@ func (tr Repo) TargetDelegations(role, path, pathHex string) []*data.Role {
 	return roles
 }
 
-// FindTarget attempts to find the target represented by the given
-// path by starting at the top targets file and traversing
-// appropriate delegations until the first entry is found or it
-// runs out of locations to search.
-// N.B. Multiple entries may exist in different delegated roles
-//      for the same target. Only the first one encountered is returned.
-func (tr Repo) FindTarget(path string) *data.FileMeta {
```
riyazdf: I couldn't find any uses of this code

tuf/tuf.go
endophage: We can assess when we're closer to merging. It may become obvious as we add the additional uses of the walker whether this is necessary or not.

endophage:
@NathanMcCauley the `WalkTargets` function in here is what I was intending in that issue that mentions `os.Walk`.

tuf/tuf.go
endophage: :+1:

riyazdf:
@endophage sounds good, I'll also update `GetDelegationRole` to use the walker directly

tuf/client/client.go
```diff
@@ -530,40 +529,3 @@ func (c Client) getTargetsFile(role string, snapshotMeta data.Files, consistent
 	}
 	return s, nil
 }
-
-// TargetMeta ensures the repo is up to date. It assumes downloadTargets
-// has already downloaded all delegated roles
-func (c Client) TargetMeta(role, path string, excludeRoles ...string) (*data.FileMeta, string) {
```
cyli: Yay!!  :+1: Glad to see all this go :D

tuf/tuf_test.go
```diff
@@ -271,68 +248,13 @@ func TestUpdateDelegationsNotEnoughKeys(t *testing.T) {
 	assert.False(t, ok, "no targets file should be created since delegation failed")
 }
 
-func TestUpdateDelegationsReplaceRole(t *testing.T) {
```
riyazdf: Since `UpdateDelegations` doesn't take a pointer to a `data.Role`, this scenario is no longer possible

tuf/tuf_test.go
```diff
@@ -988,25 +868,20 @@ func TestGetDelegationRolesInvalidPaths(t *testing.T) {
...
+	// Now we add a delegation with a path that is not prefixed by its parent delegation, the invalid path won't be added
+	err = repo.UpdateDelegations("targets/test/b", changelist.TufDelegation{AddKeys: []data.PublicKey{testKey2}, AddPaths: []string{"invalidpath"}, NewThreshold: 1})
 	assert.NoError(t, err)
 
-	// Getting this delegation does not actually restrict paths, unless we use the RestrictChild method
+	// Updating the delegation will restrict paths
 	delgRole, err := repo.GetDelegationRole("targets/test/b")
 	assert.NoError(t, err)
-	assert.Contains(t, delgRole.Paths, "invalidpath")
+	assert.NotContains(t, delgRole.Paths, "invalidpath")
```
riyazdf: Updating delegations on a repo validates paths now, so the `invalidpath` won't be added to the role in the repo

tuf/tuf.go
```diff
@@ -210,32 +225,49 @@ func (tr *Repo) GetDelegationRole(name string) (data.DelegationRole, error) {
...
+	var foundRole *data.DelegationRole
+	buildDelegationRoleVisitor := func(tgt *data.SignedTargets, validRole data.DelegationRole) error {
+		if tgt == nil {
+			return ErrContinueWalk{}
+		}
+
+		// Try to find the delegation and build a DelegationRole structure
+		for _, role := range tgt.Signed.Delegations.Roles {
+			if role.Name == name {
+				pubKeys := make(map[string]data.PublicKey)
```
cyli: Non-blocking - since this seems to be shared with the code in `SignedTargets.buildDelegationRoles`, would it make sense to pull this out into a helper function (at least the `BaseRole` part)?  It could also be used by `tuf.Repo.GetBaseRole`, etc.

tuf/tuf.go
```diff
@@ -296,114 +302,102 @@ func (tr *Repo) GetAllLoadedRoles() []*data.Role {
...
-	for _, k := range keys {
-		if !utils.StrSliceContains(role.KeyIDs, k.ID()) {
-			role.KeyIDs = append(role.KeyIDs, k.ID())
+	// Walk to parent, and either set or update this delegation
+	// Ensure all updates are valid
+	delegationUpdateVisitor := func(tgt *data.SignedTargets, validRole data.DelegationRole) error {
+		var err error
+		var createdRole bool
+		// Validate the changes underneath this restricted validRole for adding paths
+		changes.AddPaths = data.RestrictDelegationPathPrefixes(validRole.Paths, changes.AddPaths)
```
cyli: Non-blocking question: do we want to actually restrict the paths when creating the delegation?  If a user is requesting a paths change, this may be a surprise when the paths they specify are changed.  Would we want to just reject the change instead?

riyazdf: Good point, rejecting the change would be less surprising.  I'll change this unless there are objections :+1:

tuf/tuf.go
```diff
@@ -296,114 +302,102 @@ func (tr *Repo) GetAllLoadedRoles() []*data.Role {
...
 		return err
 	}
 
 	// check the parent role's metadata
-	p, ok := tr.Targets[parent]
+	_, ok := tr.Targets[parent]
 	if !ok { // the parent targetfile may not exist yet - if not, then create it
 		var err error
-		p, err = tr.InitTargets(parent)
+		_, err = tr.InitTargets(parent)
```
cyli: If we call `InitTargets` here, the created targets metadata will be marked as dirty.  If later, as we walk the tree, we figure out that the delegation is invalid (not within the target/delegations role hierarchy), this metadata may still be published.  Currently, walk continues if the metadata doesn't exist, but I'm wondering if it should just pass the nil pointer to the walk functions, so they can create a new metadata object and assign it if necessary?  It would prevent us from having to do a create before we actually walk the tree.

(Alternately, we can set `p.Dirty = false` here)

cyli: Oh nm, I guess `VerifyCanSign` will fail if the parent isn't a valid role, and if we're assuming we can't start the repo in an invalid state (metadata for a role that is invalid), this will never happen.  I think I still slightly favor the walking function creating the metadata, but it's definitely not causing a problem here.

tuf/tuf.go
cyli: Not relevant to this PR, but it'd be awesome to in the future roll the verify can sign functionality into the update visitor, so we don't have to walk the tree twice (once to get the parent role to check if it can sign, once to update the parent role)

tuf/tuf.go
cyli: Not relevant to this PR, but it'd be awesome to in the future roll the `VerifyCanSign` functionality into the update visitor, so we don't have to walk the tree twice (once to get the parent role to check if it can sign, once to update the parent role)

tuf/tuf.go
```diff
@@ -632,47 +598,118 @@ func (tr *Repo) VerifyCanSign(roleName string) error {
...
+
+		// Check the role metadata
+		signedTgt, ok := tr.Targets[role.Name]
+		if !ok {
+			// The role meta doesn't exist in the repo so continue onward
+			continue
+		}
+
+		// We're at a prefix of the desired role subtree, so add its delegation role children and continue walking
+		if strings.HasPrefix(rolePath, role.Name+"/") {
```
cyli: :+1: This is awesome!  so we don't continue walking if we're already at the role we want :D

tuf/tuf.go
riyazdf: Agreed, I'll do my best to factor out the shared code :)

tuf/tuf.go
```diff
@@ -632,47 +598,118 @@ func (tr *Repo) VerifyCanSign(roleName string) error {
...
+
+		}
+	}
+	return nil
+}
+
+// helper function that returns whether the candidateChild role name is an ancestor or equal to the candidateAncestor role name
+// Will return true if given an empty candidateAncestor role name
+func isAncestorRole(candidateChild, candidateAncestor string) bool {
+	return candidateAncestor == "" || candidateAncestor == candidateChild || strings.HasPrefix(candidateChild, candidateAncestor+"/")
```
cyli: Should the arguments to the `strings.HasPrefix` part of this be reversed?  Since if we are at role `targets` with `rolePath = "targets/releases"`, we want to check if `targets/releases` has the prefix `targets/`.  (Although we seem to never reach this point, since [we already check for that condition and skip, if so, above](https://github.com/docker/notary/pull/577/files#diff-b7b7bae92e6be55133dab8ee3256e014R633)

riyazdf: The `strings.HasPrefix` check is used for traversing the subtree of our desired role, in case we want to traverse past the role we're looking for and into its own delegations (like for `ListTargets`).  The first condition is for traversing the tree ancestor chain to our desired role, and this check is to visit that desired role and its subtree

tuf/tuf.go
riyazdf: Yes, definitely agreed with `VerifyCanSign` being part of the visitor.  I'm a little hesitant to include it in this PR since it's already fairly large with both the targets and delegations walking, but I can take a stab at it here if we'd like to also include it in this PR.

Also, good find on `InitTargets`!  I'm of the opinion that the tree walking should not create intermediate roles so that we keep the walking logic as simple as possible (it'd add complexity to the parent walking), but if others would also prefer the tree walking to create intermediate metadata I can change this. This is also relevant to `AddTargets` cc @endophage @diogomonica

tuf/tuf.go
cyli: Oh ok, that makes more sense, thanks!  :)

tuf/tuf.go
```diff
@@ -624,47 +566,120 @@ func (tr *Repo) VerifyCanSign(roleName string) error {
...
+		if strings.HasPrefix(rolePath, role.Name+"/") {
+			roles = append(roles, signedTgt.GetValidDelegations(role)...)
+			continue
+		}
+
+		// Determine whether to visit this role or not:
+		// If the paths validate against the specified targetPath and the rolePath is empty or is in the subtree
+		if isValidPath(targetPath, role) && isAncestorRole(role.Name, rolePath) {
+			// If we had matching path or role name, visit this target and determine whether or not to keep walking
+			err := visitTarget(signedTgt, role)
```
riyazdf: rename to `visitTargets`

tuf/tuf.go
```diff
@@ -624,47 +566,120 @@ func (tr *Repo) VerifyCanSign(roleName string) error {
...
+			switch err.(type) {
+			case ErrStopWalk:
```
riyazdf: Remove 'Err' prefix

tuf/tuf.go
```diff
@@ -624,47 +566,120 @@ func (tr *Repo) VerifyCanSign(roleName string) error {
...
+				// If the visitor function signalled a stop, return nil to finish the walk
+				return nil
+			case ErrContinueWalk:
```
riyazdf: Investigate if 'nil' can replace 'ErrContinueWalk'

tuf/tuf.go
```diff
@@ -624,47 +566,120 @@ func (tr *Repo) VerifyCanSign(roleName string) error {
 	return signed.ErrNoKeys{KeyIDs: role.ListKeyIDs()}
 }
 
+// used for walking the targets/delegations tree, potentially modifying the underlying SignedTargets for the repo
+type walkVisitorFunc func(*data.SignedTargets, data.DelegationRole) error
```
riyazdf: Change `error` return to `interface`, to define behavior with/without errors

client/client.go
```diff
@@ -390,10 +391,26 @@ func (r *NotaryRepository) ListTargets(roles ...string) ([]*TargetWithRole, erro
 	}
 	targets := make(map[string]*TargetWithRole)
 	for _, role := range roles {
-		// we don't need to do anything special with removing role from
-		// roles because listSubtree always processes role and only excludes
-		// descendant delegations that appear in roles.
-		r.listSubtree(targets, role, roles...)
+		// Define a visitor function to populate the targets map in priority order
+		listVisitorFunc := func(tgt *data.SignedTargets, validRole data.DelegationRole) error {
+			if tgt == nil {
```
riyazdf: Wrap this checking in another visitor function

client/client.go
```diff
@@ -390,10 +391,26 @@ func (r *NotaryRepository) ListTargets(roles ...string) ([]*TargetWithRole, erro
...
+				return tuf.ErrContinueWalk{}
+			}
+			// We found targets so we should try to add them to our targets map
+			for targetName, targetMeta := range tgt.Signed.Targets {
+				// Follow the priority by not overriding previously set targets
+				// and check that this path is valid with this role
+				// Also check that we should not exclude this target (see IMPORTANT comment above)
+				shouldExclude := utils.StrSliceContains(roles, validRole.Name) && validRole.Name != role
```
riyazdf: Move exclusion logic to `WalkTargets`

client/client.go
```diff
@@ -448,11 +437,28 @@ func (r *NotaryRepository) GetTargetByName(name string, roles ...string) (*Targe
...
+			// so we should stop our walk and set the resultMeta and resultRoleName variables
+			// Also check that we should not exclude this target (see IMPORTANT comment above)
+			shouldExclude := utils.StrSliceContains(roles, validRole.Name) && validRole.Name != role
+			if resultMeta, ok = tgt.Signed.Targets[name]; ok || shouldExclude {
+				resultRoleName = validRole.Name
+				return tuf.ErrStopWalk{}
+			}
+			return tuf.ErrContinueWalk{}
+		}
+		err = r.tufRepo.WalkTargets(name, role, getTargetVisitorFunc)
```
riyazdf: When the error is not nil, we don't need to do anything with it because we intended to walk the entire targets tree, though we should also check if the target we "found" is not nil

riyazdf: Also add test for `GetTargetByName` when target does not exist

client/helpers.go
```diff
@@ -79,38 +80,34 @@ func changeTargetsDelegation(repo *tuf.Repo, c changelist.Change) error {
 		if err != nil {
 			return err
 		}
-		r, _, err := repo.GetDelegation(c.Scope())
-		if _, ok := err.(data.ErrNoSuchRole); err != nil && !ok {
-			// error that wasn't ErrNoSuchRole
+		_, err = repo.GetDelegationRole(c.Scope())
+		if _, ok := err.(data.ErrInvalidRole); err != nil && !ok {
+			// error that wasn't ErrInvalidRole
```
riyazdf: Delete this, handled in UpdateDelegations

client/helpers.go
```diff
@@ -79,38 +80,34 @@ func changeTargetsDelegation(repo *tuf.Repo, c changelist.Change) error {
...
+		delgRole, err := repo.GetDelegationRole(c.Scope())
 		if err != nil {
 			return err
 		}
 
 		// We need to translate the keys from canonical ID to TUF ID for compatibility
 		canonicalToTUFID := make(map[string]string)
-		for tufID, pubKey := range keys {
+		for tufID, pubKey := range delgRole.Keys {
 			canonicalID, err := utils.CanonicalKeyID(pubKey)
```
riyazdf: move this to `UpdateDelegations`

tuf/tuf.go
```diff
@@ -294,114 +272,104 @@ func (tr *Repo) GetAllLoadedRoles() []*data.Role {
...
+		var delgRole *data.Role
+		for _, role := range tgt.Signed.Delegations.Roles {
+			if role.Name == roleName {
+				delgRole = role
+				delgRole.AddPaths(changes.AddPaths)
+				delgRole.RemovePaths(changes.RemovePaths)
+				if changes.ClearAllPaths {
+					delgRole.Paths = []string{}
+				}
+				delgRole.RemoveKeys(changes.RemoveKeys)
```
riyazdf: `break` after this line.

tuf/tuf.go
```diff
@@ -294,114 +272,104 @@ func (tr *Repo) GetAllLoadedRoles() []*data.Role {
...
+		// Validate the changes underneath this restricted validRole for adding paths, reject invalid path additions
+		if len(changes.AddPaths) != len(data.RestrictDelegationPathPrefixes(validRole.Paths, changes.AddPaths)) {
+			return data.ErrInvalidRole{Role: roleName, Reason: "invalid paths to add to role"}
+		}
+		// Try to find the delegation and amend it using our changelist
+		var delgRole *data.Role
+		for _, role := range tgt.Signed.Delegations.Roles {
+			if role.Name == roleName {
+				delgRole = role
+				delgRole.AddPaths(changes.AddPaths)
```
riyazdf: switch with precedence for `RemovePaths`

tuf/tuf.go
```diff
@@ -294,114 +272,104 @@ func (tr *Repo) GetAllLoadedRoles() []*data.Role {
...
+		}
+		// Add the key IDs to the role and the keys themselves to the parent
+		for _, k := range changes.AddKeys {
+			if !utils.StrSliceContains(delgRole.KeyIDs, k.ID()) {
+				delgRole.KeyIDs = append(delgRole.KeyIDs, k.ID())
+			}
+			tgt.Signed.Delegations.Keys[k.ID()] = k
+		}
+		// Make sure we have a valid role still
+		if len(delgRole.KeyIDs) < delgRole.Threshold {
```
riyazdf: Make copies/new roles to perform modifications on, prior to running this step or modifying `tgt.Signed.Delegations.Keys`.  If this check succeeds, assign the copy back to the metadata

tuf/tuf.go
```diff
@@ -294,114 +272,104 @@ func (tr *Repo) GetAllLoadedRoles() []*data.Role {
...
 	}
 
-	foundAt := utils.FindRoleIndex(p.Signed.Delegations.Roles, role.Name)
-
-	if foundAt >= 0 {
-		p.Signed.Delegations.Roles[foundAt] = role
-	} else {
-		p.Signed.Delegations.Roles = append(p.Signed.Delegations.Roles, role)
+	// Walk to the parent of this delegation, since that is where its role metadata exists
+	err := tr.WalkTargets("", parent, delegationUpdateVisitor)
```
riyazdf: Add comment: we don't have to verify that the walker reached its desired role in this scenario since we've already done a walk earlier to the parent role

tuf/tuf.go
```diff
@@ -624,47 +566,120 @@ func (tr *Repo) VerifyCanSign(roleName string) error {
...
+// helper function that returns whether the delegation Role is valid against the given path
+// Will return true if given an empty candidatePath
+func isValidPath(candidatePath string, delgRole data.DelegationRole) bool {
+	return candidatePath == "" || delgRole.CheckPaths(candidatePath)
+}
+
 // AddTargets will attempt to add the given targets specifically to
 // the directed role. If the metadata for the role doesn't exist yet,
 // AddTargets will create one.
 func (tr *Repo) AddTargets(role string, targets data.Files) (data.Files, error) {
```
riyazdf: Modify to accept only one target at a time

tuf/testutils/repo.go
```diff
@@ -80,11 +81,7 @@ func EmptyRepo(gun string, delegationRoles ...string) (*tuf.Repo, signed.CryptoS
 		if err != nil {
 			return nil, nil, err
 		}
-		role, err := data.NewRole(delgName, 1, []string{}, []string{""})
-		if err != nil {
-			return nil, nil, err
-		}
-		if err := r.UpdateDelegations(role, []data.PublicKey{delgKey}); err != nil {
+		if err := r.UpdateDelegations(delgName, changelist.TufDelegation{AddKeys: []data.PublicKey{delgKey}, AddPaths: []string{""}, NewThreshold: 1, NewName: delgName}); err != nil {
```
riyazdf: Investigate if NewName is necessary

cyli:
LGTM aside from non-blocking.

cyli:
(oh oops :(  I Just realized I commented on the commit instead of the diff - apologies)

client/helpers.go
```diff
@@ -79,38 +80,34 @@ func changeTargetsDelegation(repo *tuf.Repo, c changelist.Change) error {
 		if err != nil {
 			return err
 		}
-		r, _, err := repo.GetDelegation(c.Scope())
-		if _, ok := err.(data.ErrNoSuchRole); err != nil && !ok {
-			// error that wasn't ErrNoSuchRole
+		_, err = repo.GetDelegationRole(c.Scope())
+		if _, ok := err.(data.ErrInvalidRole); err != nil && !ok {
```
endophage: I'm thinking we should have a different error to indicate the path to the parent was correct but the role didn't exist when we got there. `ErrInvalidRole` could indicate a bunch of problems other than non-existence.

endophage: Also thought we were going to move this inside UpdateDelegations as the walk to the parent is the same either way.

client/helpers.go
```diff
@@ -79,38 +80,34 @@ func changeTargetsDelegation(repo *tuf.Repo, c changelist.Change) error {
...
-			}
-			return repo.UpdateDelegations(r, td.AddKeys)
-		}
-		// create brand new role
-		r, err = td.ToNewRole(c.Scope())
-		if err != nil {
-			return err
+			// role existed, merge paths and keys to this copy of the delegation role
+			addTD := changelist.TufDelegation{AddPaths: td.AddPaths, AddKeys: td.AddKeys}
+			return repo.UpdateDelegations(c.Scope(), addTD)
```
endophage: The more I look at it the more I dislike passing the change object to the repo. I feel like we're setting ourselves up for creating circular imports we'll have to fix down the road. It might make sense, as verbose as it is, to pass the constituent parts of the TufDelegation object as individual args to UpdateDelegations.

client/helpers.go
riyazdf: Ah yes, the walk in UpdateDelegations already handles this so I removed it.

As for errors, it seems like `ErrNoSuchRole` is a good fit?

tuf/tuf.go
```diff
@@ -294,114 +257,122 @@ func (tr *Repo) GetAllLoadedRoles() []*data.Role {
...
+		for _, k := range changes.AddKeys {
+			if !utils.StrSliceContains(delgRole.KeyIDs, k.ID()) {
+				delgRole.KeyIDs = append(delgRole.KeyIDs, k.ID())
+			}
+		}
+		// Make sure we have a valid role still
+		if len(delgRole.KeyIDs) < delgRole.Threshold {
+			return data.ErrInvalidRole{Role: roleName, Reason: "insufficient keys to meet threshold"}
+		}
+		// If valid, we should edit the actual SignedTargets to match our copy
```
endophage: Can we add a comment in CAPS just above this noting that the remainder of this closure is not allowed to error as it is performing final modifications to the SignedTargets object and anything that might error _must_ be done first.

tuf/tuf.go
```diff
@@ -624,47 +569,120 @@ func (tr *Repo) VerifyCanSign(roleName string) error {
...
+			continue
+		}
+
+		// Determine whether to visit this role or not:
+		// If the paths validate against the specified targetPath and the rolePath is empty or is in the subtree
+		// Also check if we are choosing to skip visiting this role on this walk (see ListTargets and GetTargetByName priority)
+		if isValidPath(targetPath, role) && isAncestorRole(role.Name, rolePath) && !utils.StrSliceContains(skipRoles, role.Name) {
+			// If we had matching path or role name, visit this target and determine whether or not to keep walking
+			res := visitTargets(signedTgt, role)
+			switch res.(type) {
```
endophage: This switch is so pretty now.

client/helpers.go
riyazdf: I see what you're saying -- my only concern is that it will be really verbose to handle all of the args unless we break down UpdateDelegations, and breaking it down might further fragment our validation or make `changeTargetsDelegation` more complicated.  It also seems like the changelist package is currently pretty light on imports.  I'm happy to give it a shot if you think it'd be better to pass the changelist as unpacked arguments, though I'm thinking it might be better to keep the changelist itself as a compact argument to keep the API cleaner.   cc @cyli for additional input

tuf/tuf.go
riyazdf: \o/

tuf/tuf.go
riyazdf: will do!

tuf/tuf.go
```diff
@@ -294,114 +257,122 @@ func (tr *Repo) GetAllLoadedRoles() []*data.Role {
...
+					RootRole: data.RootRole{
+						KeyIDs:    keyIDCopy,
+						Threshold: role.Threshold,
+					},
+					Name:  role.Name,
+					Paths: pathsCopy,
+				}
+				delgRole.RemovePaths(changes.RemovePaths)
+				delgRole.AddPaths(changes.AddPaths)
+				if changes.ClearAllPaths {
```
cyli: Non-blocking: Should `ClearAllPaths` take precedence over AddPaths?  Since it's sort of remove-like functionality.

riyazdf: Great point, yes.  Will change.

tuf/tuf.go
```diff
@@ -624,47 +569,120 @@ func (tr *Repo) VerifyCanSign(roleName string) error {
...
+			switch res.(type) {
+			case StopWalk:
+				// If the visitor function signalled a stop, return nil to finish the walk
+				return nil
+			case nil:
+				// If the visitor function signalled to continue, add this role's delegation to the walk
+				roles = append(roles, signedTgt.GetValidDelegations(role)...)
+			case error:
+				// Propagate any errors from the visitor
+				return res.(error)
```
endophage: If you do `switch typedRes := res.(type)` there's no need to do the cast here, you can just `return typedRes` instead. It's magic :-D (https://golang.org/ref/spec#Type_switches)

riyazdf: this is amazing! Great idea, will do :)

tuf/tuf.go
endophage: skipRoles needs to be taken into account here too.

riyazdf: @endophage I think because we aren't visiting the roles here, we shouldn't check `skipRoles`

For example, if we're listing targets for `targets/releases` and also specified `targets` as a second-priority role to list for, checking `skipRoles` here would make us skip our prioritized first traversal down to `targets/releases`

client/helpers.go
cyli: I was debating whether to mention this before - it did seem like a little bit of an abstraction break for this to take a change object, and it be the ONLY place a change object is taken.  But having a huge list of arguments also seems kinda not great, so I am torn between the two.  One way to avoid having the changelist be imported here and to avoid taking a large number of arguments is to construct `tuf`'s own change object, and have there be a translation.  Or to have the `TufDelegation` object just be an embed of the `tuf` update-delegation object. :|  Which also isn't great, because then we have two objects that are essentially the same thing.

The other thing I was thinking was that maybe we want to eventually move all our `applyChange*` type functions into the `tuf`package, and maybe have that be the interface to update a TUF repo?  In which case maybe taking change objects would be less problematic?  (Although maybe we'd still want a layer in between, dunno)

tuf/tuf.go
endophage: Hmmm, I see what you're saying. Yeah, you're right.

client/helpers.go
endophage: Let's discuss first thing in the morning. This doesn't necessarily need to be a blocker on merging but I want a plan/decision on it before we merge otherwise it'll just fall into the "things to do later" bucket and not get looked at again until something breaks.

endophage:
I'm salivating over that test coverage bump :-D

client/helpers.go
riyazdf: Agreed.  I like the idea of `applyChange*` that Ying suggested but whatever we agree on we should definitely plan and decide together before moving on.

cyli:
@endophage YEEEES

cyli:
LGTM!

endophage:
LGTM :-)

@riyazdf you can have the honours or merging this gem of a PR.

PR #578, opened 2016-02-18
updates and bugfixes for go 1.6

endophage:
The Makefile check is just annoying so I removed it. We should just add a "requirements" section on the main README.

There has been a change to how encoding/json treats embedded structs. It's loosely described here: https://golang.org/doc/go1.6#reflect The TL;DR is that json now only traverses embedded structs where the struct is exported in its own right (so its name must start with an upper-case letter). 

N.B. Embedded structs themselves do not appear in serialized JSON, they are (conceptually) replaced by their encapsulated fields. Therefore this change does not have compatibility implications and is valid for Go 1.6 and prior versions. 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,14 +42,6 @@ GO_VERSION = $(shell go version | awk '{print $$3}')
 .DELETE_ON_ERROR: cover
 .DEFAULT: default
 
-go_version:
-ifeq (,$(findstring go1.5.,$(GO_VERSION)))
-	$(error Requires go version 1.5.x - found $(GO_VERSION))
-else
-	@echo
-endif
-
-
 all: AUTHORS clean fmt vet fmt lint build test binaries
 
 AUTHORS: .git/HEAD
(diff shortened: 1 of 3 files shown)


endophage:
@cyli any idea what's going on here. Seems like something may have changed regarding testing for 1.6?

riyazdf:
@endophage I've had some weirdness like this happen to me before -- rebasing and pushing again seemed to fix but it seems like your branch is up to date

cyli:
I asked the codecov folks, and apparently this is related to what information circleci vs jenkins pushes to codecov:

```
branch  | pr  | build  
----------+-----+--------
 pull/578 | 578 | 1967.0
 pull/578 | 578 | 1967.1
 go1.6    |     | 286
 go1.6    |     | 
 pull/578 | 578 |
```

was for the commit that was strange and broken.  I have no idea why the PR number was sent from Circle - the last time this weirdness happened it was because the PR number was not available in the Circle environment, and but was on Jenkins (so I modified the Jenkins job to send nothing as the PR number).   Maybe Circle changed their API, in which case I can just undo that change I made in Jenkins.

Also, codecov folks say that this discrepancy shouldn't be a problem in the next release - they are working on a fix so mismatched tables like would be fine in the future.

(I got this table from their support btw - I couldn't find how to get this info from the codecov site, but if you look at the circleci jobs, as well as the jenkins job, at the end of the test output there should be a thing about what's being uploaded to codecov, and they include the query parameters, which for the last test was:  `package=bash-5ca7b8c&branch=pull/578&commit=e2707957682b32b862ddb31cc565e513abe4d902&build=1968.0&build_url=&slug=docker/notary&service=circleci&pr=578&job=`)

riyazdf:
LGTM

cyli:
LGTM

PR #579, opened 2016-02-19
remove path hash prefix from TargetsDelegations

riyazdf:
I noticed a stray and unused `pathHex` that we should remove, since we no longer support path hash prefixes

### DIFF
--- a/tuf/client/client.go
+++ b/tuf/client/client.go
@@ -3,7 +3,6 @@ package client
 import (
 	"bytes"
 	"crypto/sha256"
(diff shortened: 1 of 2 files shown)


endophage:
Thanks for catching this. Not sure what's up with circle/jenkins/codecov today...

riyazdf:
@endophage seems to be fixed now, after just pushing the same commit again...strange

cyli:
LGTM!

endophage:
LGTM now that CI is fixed

PR #580, opened 2016-02-19
Bottleneck TUF Repo validation by using a builder to create one in the client and server

### LINKED ISSUE #52: Add check of cert when uploading root.json
We should check that the name in the cert matches the provided GUN when a user uploads a root.json.

### LINKED ISSUE #181: Coalesce client and server side validations
https://github.com/docker/notary/pull/178 introduces a bunch of good server-side validations. Some of these would also make sense as client-side validations as well.

### LINKED ISSUE #558: Merge validate root functions

### LINKED ISSUE #561: Move all verification into setters on repo
All the verifications should be moved down to the setters on the repository. The only thing we have to do at the store level is the size verification.

### LINKED ISSUE #599: Consolidate Server + Client Validation
As much as this is possible.

cyli:
This is based on #573 

~~Update: I've rebased, but the tuf client tests are still completely broken - I think I just want to move some of them (like the root rotation) to the builder, and possibly the client update tests will cover these?~~

~~Also, some notes:~~

- ~~the code around bootstrapping with previous root keys vs pinned trust is kind of ugly - maybe it should take another interface that deals with either pinned trust OR previous root keys?  That way it would not have to take a cert store~~

- ~~currently the download functionality just reaches into the repo and get the snapshot - I'm wondering if we should wrap the builder in a `consistent builder` or something that gives the checksum and the role of whatever needs to be loaded next?~~

Ok, rebased everything.  No longer creates invalid repos.  A lot of builder functionality is covered by the client tests and server validation tests, I would like to add the following builder tests:

- [x] error tests when something is loaded before root, or a delegation is loaded before targets (e.g. order of operations for all the functions in the builder)
- [x] where/when things are added to the cache of data to be checksummed, and when they are cleaned out
- [x] more tests around `GetConsistentInfo`, `ConsistentInfo`, and `GetLoadedVersion`

Fixes #558 
Fixes #561 
Fixes #599 
Fixes #181 (I think might be a dupe of #599)

Doesn't actually fix #52, but once we remove cert-store and all roots are validated using (maybe empty) trust pinning and maybe previous roots, that issue should be closed as well.

Future improvements:
- server should load and check the previous versions of metadata?
- removing the cert store (actually a larger change than I thought, so should probably go into a future PR)
- using the walker for downloading (`tuf/client/Client` does its own walking right now)
- move `bootstrapClient` over into `tuf/client/Client` (so that client just serves to download a repo given a cache, and can maybe be used to download different versions of metadata)

Caveats:
- Not super happy with having `GenerateSnapshot` and `GenerateTimestamp`, but if we're having the builder be the only thing that sets metadata, maybe that would make sense?  Maybe the server should just call `Finish()` when the server and targets/meta are done, and then set the previous snapshot/timestamp itself and sign?  We'd basically need those same functions, just maybe not on the builder.


tuf/builder.go
```diff
@@ -0,0 +1,402 @@
...
+// ErrBuildFailed is returned when any functions are called on RepoBuilder, and it
+// is already failed building and will not accept any other data
+type ErrBuildFailed struct{}
+
+func (e ErrBuildFailed) Error() string {
+	return "the builder has failed building and cannot accept any more input or produce any more output"
+}
+
+// ErrInvalidBuilderInput is returned when RepoBuilder.Load is called
+// with the wrong type of metadata for thes tate that it's in
```
riyazdf: nit: typo "the state"

tuf/builder.go
```diff
@@ -0,0 +1,402 @@
...
+	LoadSnapshot(content []byte, minVersion int) error
+	LoadTimestamp(content []byte, minVersion int) error
+	LoadTargets(content []byte, minVersion int) error
+	LoadDelegation(role string, content []byte, minVersion int) error
+	Finish() (*Repo, error)
+	BootstrapNewBuilder() RepoBuilder
+	GetRepo() *Repo
+}
+
+// NewRepoBuilder is the only way to get a pre-built RepoBuilder
```
riyazdf: I really like the way this is designed :+1:

tuf/builder.go
```diff
@@ -0,0 +1,402 @@
...
+		gun:                  gun,
+		certStore:            certStore,
+		cs:                   cs,
+		loadedNotChecksummed: make(map[string][]byte),
+	}
+}
+
+type repoBuilder struct {
+	finished bool
+	failed   bool
```
riyazdf: documenting IRL discussion with @cyli : I'm not sure if we want these flags to be state on the repoBuilder itself, or if we need separate cases for `failed` and `finished`.  It would also simplify the switch statements in the `RepoBuilder.Load*` functions

tuf/client/client.go
```diff
@@ -46,493 +42,197 @@ func (c *Client) Update() error {
...
-	var raw []byte
-	if download {
-		// use consistent download if we have the checksum.
-		raw, s, err = c.downloadSigned(role, size, expectedSha256)
-		if err != nil {
-			return err
-		}
-	} else {
-		logrus.Debug("using cached root")
-		s = old
```
riyazdf: it's awesome to see this get abstracted away :+1:

riyazdf:
to your notes:

- on bootstrapping root: the current code (https://github.com/docker/notary/pull/580/files#diff-f899aa8e7233ba3ea46223ec99b46714R180) isn't too unwieldy.  I'd prefer to keep it as clear and strongly typed as possible, but if you think we can design an interface that can take either certs or keys and still accomplish this I think it'd be nice (but not imperative) to have.

- on `consistent builder`: this sounds like it might be generally worth refactoring the builder to keep more explicit state about which stage it is in the building process (similar to https://github.com/docker/notary/pull/570)

tuf/client/client.go
```diff
@@ -46,493 +42,197 @@ func (c *Client) Update() error {
 	err := c.update()
 	if err != nil {
 		logrus.Debug("Error occurred. Root will be downloaded and another update attempted")
-		if err := c.downloadRoot(); err != nil {
+		logrus.Debug("Resetting the TUF builder...")
+
+		repoSoFar := c.builder.GetRepo()
+		c.builder = c.builder.BootstrapNewBuilder()
```
NathanMcCauley: Why is it necessary to overwrite the builder here?

cyli: Do you mean as opposed to bootstrapping a new client entirely?

NathanMcCauley: Trying to figure out why the existing builder couldn't be used.

I see now that we essentially want to restart the build process at this point, because we have more information. (That appears to be the point of `BootstrapNewBuilder`, as far as I can tell)

cyli: Ah ok, basically we wanted to throw out any existing information we have and restart building a `tuf.Repo` object all over again fresh, so we can't just accidentally overwrite some previous data and end up with a half new half old repo.

client/client.go
```diff
@@ -774,19 +775,26 @@ func (r *NotaryRepository) Update(forWrite bool) (*tufclient.Client, error) {
...
-
 	if cachedRootErr == nil {
-		signedRoot, cachedRootErr = r.validateRoot(rootJSON)
+		cachedRootErr = b.Load(data.CanonicalRootRole, rootJSON, version)
+		if cachedRootErr == nil {
+			successfullyBootstrapped = true
+		} else {
+			// try to get the version if we can
+			signedMeta := &data.SignedMeta{}
+			if err := json.Unmarshal(rootJSON, signedMeta); err == nil {
```
NathanMcCauley: Not sure if we should be consuming this file if we haven't first verified it.

cyli: Good point - it definitely failed validation at this point - I guess we really want to say "validate everything except expiry" and load the version otherwise?

client/client.go
```diff
@@ -766,7 +766,8 @@ func (r *NotaryRepository) Update(forWrite bool) (*tufclient.Client, error) {
 		}
 		return nil, err
 	}
-	return c, nil
+	r.tufRepo = repo
+	return nil, nil
```
riyazdf: Should we change the function signature to only return `error`, since we never return a pointer to a tuf client now?

client/client.go
riyazdf: Agreed that we should only consume the version if the only verification error was the expiry check

tuf/builder.go
```diff
@@ -0,0 +1,402 @@
...
+	"github.com/Sirupsen/logrus"
+	"github.com/docker/notary/certs"
+	"github.com/docker/notary/trustmanager"
+	"github.com/docker/notary/tuf/data"
+	"github.com/docker/notary/tuf/signed"
+)
+
+// ErrBuildDone is returned when any functions are called on RepoBuilder, and it
+// is already finished building
+type ErrBuildDone struct{}
```
riyazdf: non-blocking: seems like this could just be a type instead of an error -- similar to `StopWalk` in `tuf.go`?  It's less clear whether a generic type makes sense here because we have more error cases, but I'm also not sure if the "done building" state should be messaged as an error either.

tuf/builder.go
```diff
@@ -0,0 +1,402 @@
...
+		return err
+	}
+
+	signedTimestamp, err := data.TimestampFromSigned(signedObj)
+	if err != nil {
+		return err
+	}
+
+	rb.repo.SetTimestamp(signedTimestamp)
+	return rb.validateChecksums(roleName, signedTimestamp.Signed.Meta)
```
riyazdf: Unless I'm missing something - should we `SetTimestamp` after validating checksums?  I guess either way the code is structured we'd get a failed builder, but I think it'd be better to not set any invalid state in the builder's repo if possible

cyli: I originally had it that way, but for snapshot, we need to set the failed snapshot because otherwise we won't be able to bootstrap a new builder (since the snapshot wouldn't ever be set).  Because if the checksums are wrong, we need to download a new root based on the snapshot that we downloaded.  :|

Another way to do this would be to have the `tuf/client` ALSO keep all the metadata that it's downloaded, so it has a cache of metadata separate from the builder so it can bootstrap a new builder (although it would also need to store the certificate validator, etc. that is right now created in `client/client` in order to make a completely new builder).

cyli: Unfortunately there is a lot of overlap between the consistent downloader states and the builder states - I wasn't sure whether each should have its own metadata to maintain state.  Which would make sense - the downloader would know if it's in the snapshot state and can then use the snapshot to get the checksums to download stuff.

riyazdf: ah yes.  I'd rather not duplicate metadata in another cache in `tuf/client` (nightmares of keyDB), so I'm good with this current flow

client/client.go
cyli: Addressed in https://github.com/docker/notary/pull/668

cyli:
This is probably ready for a look - I'm trying to split out as much as I can (currently tuf/client tests are disabled - I'm working on removing redundant parts and moving the non-redundant ones to `client_update_tests` instead), and it can use some more test coverage for the builder itself.

client/client.go
```diff
@@ -659,47 +658,28 @@ func (r *NotaryRepository) publish(cl changelist.Changelist) error {
...
 
-	snapshotJSON, err := r.fileStore.GetMeta(data.CanonicalSnapshotRole, -1)
-	if err == nil {
-		snapshot := &data.SignedSnapshot{}
-		err = json.Unmarshal(snapshotJSON, snapshot)
+	for _, role := range data.BaseRoles {
+		jsonBytes, err := r.fileStore.GetMeta(role, -1)
 		if err != nil {
+			if _, ok := err.(store.ErrMetaNotFound); ok &&
+				role == data.CanonicalSnapshotRole || role == data.CanonicalTimestampRole {
```
riyazdf: I really like the way this code cleaned up, but I think it might be helpful to leave a comment about what it means to hit this code for snapshot and timestamp and how snapshot vs. timestamp are potentially different cases

client/client.go
```diff
@@ -801,26 +783,37 @@ func (r *NotaryRepository) Update(forWrite bool) error {
 // Returns a tufclient.Client for the remote server, which may not be actually
 // operational (if the URL is invalid but a root.json is cached).
 func (r *NotaryRepository) bootstrapClient(checkInitialized bool) (*tufclient.Client, error) {
-	var (
-		rootJSON   []byte
-		err        error
-		signedRoot *data.SignedRoot
-	)
+	version := 0
```
riyazdf: nit: can we rename this variable to `minVersion`?

client/client.go
cyli: Good point, thanks!

client/client_update_test.go
```diff
@@ -160,10 +154,6 @@ func TestUpdateSucceedsEvenIfCannotWriteExistingRepo(t *testing.T) {
 			repo.fileStore = &unwritableStore{MetadataStore: origFileStore, roleToNotWrite: role}
 			err := repo.Update(forWrite)
 
-			if role == data.CanonicalRootRole {
-				require.Error(t, err) // because checkRoot loads root from cache to check hashes
```
riyazdf: 👍 awesome that we don't need to do this awkward checking in our tests anymore

tuf/builder.go
```diff
@@ -0,0 +1,707 @@
...
+	return utils.ConsistentName(c.RoleName, c.fileMeta.Hashes[notary.SHA256])
+}
+
+// Length returns the expected length of the role as per this consistent
+// information - if no checksum information is known, the size is -1.
+func (c ConsistentInfo) Length() int64 {
+	if c.ChecksumKnown() {
+		return c.fileMeta.Length
+	}
+	return -1
```
riyazdf: we can do this in a followup PR, but I'm wondering if it'd be useful to have a const for `-1` to represent unknown or max size that we can use here and for other `GetMeta` operations

tuf/builder.go
```diff
@@ -0,0 +1,707 @@
...
+
+func (rb *repoBuilder) Finish() (*Repo, error) {
+	return rb.repo, nil
+}
+
+func (rb *repoBuilder) BootstrapNewBuilder() RepoBuilder {
+	var rootRole *data.BaseRole
+	if rb.repo.Root != nil {
+		roleObj, err := rb.repo.GetBaseRole(data.CanonicalRootRole)
+		// this should always be true, since it was already validated, otherwise something
```
riyazdf: nit: should replace "always be true" with "never error"?

tuf/builder.go
```diff
@@ -0,0 +1,707 @@
...
+	default:
+		if tgts, ok := rb.repo.Targets[roleName]; ok {
+			return tgts.Signed.Version
+		}
+	}
+
+	return 0
+}
+
+// GetConsistentInfo returns the consistent name and size of a role, if it is known,
```
riyazdf: nit: this function doesn't return a size

riyazdf: Actually sorry -- it does but it might be helpful to explain that it's embedded in a `ConsistentInfo` struct :)

tuf/builder.go
```diff
@@ -0,0 +1,707 @@
...
+	case data.CanonicalSnapshotRole:
+		return rb.repo.Snapshot != nil
+	case data.CanonicalTimestampRole:
+		return rb.repo.Timestamp != nil
+	default:
+		return rb.repo.Targets[roleName] != nil
+	}
+}
+
+// GetLoadedVersion returns the metadata version, if it is loaded, or 0 otherwise
```
riyazdf: This seems like it should be ok, but I wanted a sanity check -- since anything that's loaded must be signed, the version must be strictly > 0 correct?

tuf/builder.go
```diff
@@ -0,0 +1,707 @@
...
+	}
+
+	// since the snapshot was generated using the root and targets data that
+	// that have been loaded, remove all of them from rb.loadedNotChecksummed
+	for tgtName := range rb.repo.Targets {
+		delete(rb.loadedNotChecksummed, tgtName)
+	}
+	delete(rb.loadedNotChecksummed, data.CanonicalRootRole)
+
+	// cache the snapshot bytes so we can validate hte checksum in case a timestamp
```
riyazdf: nit typo: `the`

client/client.go
riyazdf: Also cc: @endophage  - unless I'm missing something, it seems that we can be stricter about version checks and enforce that all versions are >= 1 right?

tuf/builder.go
```diff
@@ -0,0 +1,707 @@
...
+
+	sgndJSON, err := json.Marshal(sgnd)
+	if err != nil {
+		rb.repo.Timestamp = nil
+		return nil, 0, err
+	}
+
+	// since the timestamp was generated using the snapshot that has been loaded,
+	// remove it from rb.loadedNotChecksummed
+	delete(rb.loadedNotChecksummed, data.CanonicalSnapshotRole)
```
riyazdf: It might be helpful to document pre and post-conditions on `loadedNotChecksummed` since we require certain roles to be loaded before calling functions and also expect certain roles to be removed after execution of the function

tuf/builder.go
cyli: Yes I think so - I originally was going to return -1, since I thought 0 was a valid version, but in our download data our default version seems to be 0.

tuf/builder.go
```diff
@@ -0,0 +1,707 @@
...
+	if rb.repo.Root != nil {
+		roleObj, err := rb.repo.GetBaseRole(data.CanonicalRootRole)
+		// this should always be true, since it was already validated, otherwise something
+		// is very wrong and we should not bootstrap with this root
+		if err == nil {
+			rootRole = &roleObj
+		}
+	}
+
+	return &repoBuilderWrapper{RepoBuilder: &repoBuilder{
```
riyazdf: I think we discussed this already IRL but I think we're missing `trustpin` here

tuf/client/client.go
```diff
@@ -44,503 +42,187 @@ func (c *Client) Update() error {
...
-		return nil, nil, err
+		return nil, err
 	}
 
-	if expectedHashes != nil {
-		if err := data.CheckHashes(raw, expectedHashes); err != nil {
-			return nil, nil, ErrChecksumMismatch{role: role}
-		}
-	}
+	// we know it unmarshals fine
```
riyazdf: to clarify: this is because we already have `ConsistentInfo` and didn't error on the load above?

tuf/builder.go
cyli: Added, thank you!

tuf/builder.go
cyli: Added, thanks!

tuf/builder.go
cyli: That's a good idea!  Probably a separate PR though, to avoid making this any bigger.  :)  https://github.com/docker/notary/issues/690

client/client.go
```diff
@@ -801,26 +785,37 @@ func (r *NotaryRepository) Update(forWrite bool) error {
 // Returns a tufclient.Client for the remote server, which may not be actually
 // operational (if the URL is invalid but a root.json is cached).
 func (r *NotaryRepository) bootstrapClient(checkInitialized bool) (*tufclient.Client, error) {
-	var (
-		rootJSON   []byte
-		err        error
-		signedRoot *data.SignedRoot
-	)
+	minVersion := 0
```
cyli: [@riyaz's previous comment](https://github.com/docker/notary/pull/580#discussion_r60459534) got collapsed, so repeating it here:

> Also cc: @endophage - unless I'm missing something, it seems that we can be stricter about version checks and enforce that all versions are >= 1 right?

tuf/builder.go
```diff
@@ -198,7 +199,7 @@ func (rb *repoBuilder) GetLoadedVersion(roleName string) int {
 		}
 	}
 
-	return 0
+	return 1
```
riyazdf: nit: it seems like it'd be helpful to make a `MinMetadataVersion` or similar constant since we reuse `1` throughout the code

tuf/builder.go
```diff
@@ -0,0 +1,741 @@
...
+	sgndJSON, err := json.Marshal(sgnd)
+	if err != nil {
+		rb.repo.Snapshot = nil
+		return nil, 0, err
+	}
+
+	// loadedNotChecksummed should currently contain the root awaiting checksumming,
+	// since it has to have been loaded.  Since the snapshot was generated using
+	// the root and targets data (there may not be any) that that have been loaded,
+	// remove all of them from rb.loadedNotChecksummed
```
riyazdf: these comments about the pre/post conditions on `loadedNotChecksummed` are really helpful, thanks for adding them in 👍

tuf/builder.go
```diff
@@ -0,0 +1,741 @@
...
+		return nil, 0, err
+	}
+
+	sgndJSON, err := json.Marshal(sgnd)
+	if err != nil {
+		rb.repo.Timestamp = nil
+		return nil, 0, err
+	}
+
+	// The snapshot should have been loaded (and not checksumemd, since a timestamp
```
riyazdf: nit typo: `checksummed`

riyazdf:
this is amazing work, thank you for pushing through all of this!  LGTM!!! 🎉 

For other reviewers: I prefer having the `loadedNotChecksummed` map even though it's extra state in exchange for ensuring we never hold an invalid repo itself in the builder

riyazdf:
review updates LGTM!

endophage:
LGTM! Thanks for taking this on and carrying it to completion, this is going to lead to more goodness :-)

PR #585, opened 2016-02-22
check validity of children during download

endophage:
I originally also added a check to prevent overwriting of targets objects in tuf.Repo.SetTargets but this broke operation as we current re-use the repo object. Adding the check during download should be sufficient to ensure we never write an incorrect child as a second call to SetTargets with an existing delegation will set a validated object.


riyazdf:
LGTM

cyli:
LGTM

PR #588, opened 2016-02-24
Fix server signed snapshot expiry/regeneration

endophage:
Two related fixes:

- The server was not setting the longer snapshot expiry time consistent with the client. 
- When generating a timestamp it was also retrieving the snapshot directly from the database and only validating the checksum still matched what was in the timestamp. Due to the addition of consistent downloads, this mean a new snapshot never got generated (when the server was responsible for this). It is necessary for `GetOrCreateTimestamp` to call `GetOrCreateSnapshot` to ensure a new snapshot is generated as and when required

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -27,13 +27,7 @@ import (
 )
 
 func init() {
-	data.SetDefaultExpiryTimes(
-		map[string]int{
-			"root":     3650,
-			"targets":  1095,
-			"snapshot": 1095,
-		},
-	)
+	data.SetDefaultExpiryTimes(notary.NotaryDefaultExpiries)
 }
 
 // ErrRepoNotInitialized is returned when trying to publish an uninitialized
--- a/const.go
+++ b/const.go
@@ -1,5 +1,9 @@
 package notary
 
+import (
+	"time"
+)
+
 // application wide constants
 const (
 	// MaxDownloadSize is the maximum size we'll download for metadata if no limit is given
@@ -24,4 +28,23 @@ const (
 	RootKeysSubdir = "root_keys"
 	// NonRootKeysSubdir is the subdirectory under PrivDir where non-root private keys are stored
 	NonRootKeysSubdir = "tuf_keys"
+
+	// Day is a duration of one day
+	Day  = 24 * time.Hour
+	Year = 365 * Day
+
+	// NotaryRootExpiry is the duration representing the expiry time of the Root role
+	NotaryRootExpiry      = 10 * Year
+	NotaryTargetsExpiry   = 3 * Year
+	NotarySnapshotExpiry  = 3 * Year
+	NotaryTimestampExpiry = 14 * Day
 )
+
+// NotaryDefaultExpiries is the construct used to configure the default expiry times of
+// the various role files.
+var NotaryDefaultExpiries = map[string]time.Duration{
+	"root":      NotaryRootExpiry,
+	"targets":   NotaryTargetsExpiry,
+	"snapshot":  NotarySnapshotExpiry,
+	"timestamp": NotaryTimestampExpiry,
+}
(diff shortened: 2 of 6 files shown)


client/client.go
```diff
@@ -29,9 +29,10 @@ import (
 func init() {
 	data.SetDefaultExpiryTimes(
 		map[string]int{
-			"root":     3650,
```
NathanMcCauley: Why aren't these values b/w client and server?

riyazdf: Agreed, we could potentially put them in `const.go` to ensure they're shared consistently

endophage: Do you mean why aren't they shared? They should be, I'll put them in const.go

diogomonica: Let's also add a comment on the unit they represent.

riyazdf:
Should we also make these expiries different or consistent? https://github.com/endophage/notary/blob/fix_snapshot_expiry/tuf/data/types.go#L175

endophage:
@riyazdf those ones are the TUF recommended defaults. I'd say they should be left as is.

riyazdf:
@endophage makes sense.  Could we add a comment explaining how they're recommended by TUF? 

Also recording IRL discussion with @NathanMcCauley and @endophage to change hardcoded numbers to expressions over `time` constants.  Ex: `365 * 24 * time.Hour` instead of `365`.

endophage:
@NathanMcCauley Apparently `time.Add(duration)` does not account for leap years, which might be an issue with short expiry timestamps (particularly if/when we reduce the expiry window). `time.AddDate()` does do this accounting so is safer. (https://github.com/docker/notary/pull/312#discussion_r46729318)

With that in mind, if we want to keep using `AddDate` we can't do the `365 * 24 * time.Hour` format because the integer this produces is the number of nanoseconds in 365 days, and AddDate needs integer numbers of years, months, and days.

endophage:
@NathanMcCauley my indifference of time.Duration vs an integer number of days was swayed by the fact I actually suggested a use case for very short expiries earlier today (testing in stage with super rapidly expiring data).

const.go
```diff
@@ -24,4 +28,23 @@ const (
...
+
+	// Day is a duration of one day
+	Day  = 24 * time.Hour
+	Year = 365 * Day
+
+	// NotaryRootExpiry is the duration representing the expiry time of the Root role
+	NotaryRootExpiry      = 10 * Year
+	NotaryTargetsExpiry   = 3 * Year
+	NotarySnapshotExpiry  = 3 * Year
+	NotaryTimestampExpiry = 14 * Day
```
riyazdf: the consolidation and format of these constants look great now :+1:

riyazdf:
LGTM pending CI

endophage:
@cyli @diogomonica @NathanMcCauley can I get another LGTM?

cyli:
LGTM

client/client.go
```diff
@@ -27,13 +27,7 @@ import (
...
 
 func init() {
-	data.SetDefaultExpiryTimes(
-		map[string]int{
-			"root":     3650,
-			"targets":  1095,
-			"snapshot": 1095,
-		},
-	)
+	data.SetDefaultExpiryTimes(notary.NotaryDefaultExpiries)
```
diogomonica: what about `NotaryDefaultExpiryTimes`? The word Expiries is weird. But hey, that is just me.

tuf/data/types.go
```diff
@@ -171,16 +172,16 @@ func NewDelegations() *Delegations {
...
 }
 
-// defines number of days in which something should expire
-var defaultExpiryTimes = map[string]int{
-	CanonicalRootRole:      365,
-	CanonicalTargetsRole:   90,
-	CanonicalSnapshotRole:  7,
-	CanonicalTimestampRole: 1,
+// These values are recommended TUF expiry times.
+var defaultExpiryTimes = map[string]time.Duration{
```
diogomonica: Why do we have two defaults, one for TUF and one for Notary? Feels like asking for trouble in case we forget to change it.

endophage: The TUF ones are the recommended times we should be aiming for (particularly with server signed things). The Notary ones are our overrides.

PR #589, opened 2016-02-24
print IDs instead of key itself

riyazdf:
Changes to print the canonical key IDs when adding a delegation, instead of the public key bytes.


cyli:
LGTM!

endophage:
LGTM

PR #590, opened 2016-02-24
Move yubikey import role check to avoid excessive passphrase prompting

riyazdf:
When importing to a yubikeystore, we should reject non-root keys upfront before prompting for passphrases or attempting to read key bytes.


endophage:
We really need to fix that whole `keyPath` inconsistency... but yeah, LGTM on this. 

We should target Notary 0.4 (or whatever release we do after 0.3) for putting non-root roles in hardware, then this check disappears.

endophage:
@cyli @diogomonica can we get another LGTM on this?

cyli:
LGTM

PR #593, opened 2016-02-25
Add SHA-512

### LINKED ISSUE #412: Add SHA-384/512
Based on [NSA advice](https://www.nsa.gov/ia/programs/suiteb_cryptography/index.shtml) we should be using SHA-384 at a minimum. We should add this in addition to the existing SHA-256 (generating multiple hashes for each target/meta file) so that old clients continue to operate correctly.

HuKeping:
Closes #412 

This patch set will generate multiple hashes for each target/meta file:
- [PATCH 1/8] Create constants for sha256 and sha512
- [PATCH 2/8] Add some helper functions
- [PATCH 3/8] Add sha512 when creating target, snapshot and timestamp
- [PATCH 4/8] Add sha512 when updating and also make the change on test
- [PATCH 5/8] Add sha512 check on CLI command
- [PATCH 6/8] Add sha512 check when downloading TUF roles and update the relevant test.
- [PATCH 7/8] Updating the server side
- [PATCH 8/8] Some test for the added functions of this patch set

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -140,7 +140,7 @@ func NewTarget(targetName string, targetPath string) (*Target, error) {
 		return nil, err
 	}
 
-	meta, err := data.NewFileMeta(bytes.NewBuffer(b))
+	meta, err := data.NewFileMeta(bytes.NewBuffer(b), data.NotaryDefaultHashes...)
 	if err != nil {
 		return nil, err
 	}
--- a/cmd/notary/tuf.go
+++ b/cmd/notary/tuf.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"bufio"
-	"crypto/sha256"
 	"fmt"
 	"io/ioutil"
 	"net"
@@ -11,8 +10,6 @@ import (
 	"os"
 	"strings"
 	"time"
-
-	"crypto/subtle"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/docker/distribution/registry/client/auth"
@@ -385,13 +382,10 @@ func (t *tufCommander) tufVerify(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("error retrieving target by name:%s, error:%v", targetName, err)
 	}
 
-	// Create hasher and hash data
-	stdinHash := sha256.Sum256(payload)
-	serverHash := target.Hashes["sha256"]
-
-	if subtle.ConstantTimeCompare(stdinHash[:], serverHash) == 0 {
-		return fmt.Errorf("notary: data not present in the trusted collection")
+	if err := data.CheckHashes(payload, target.Hashes); err != nil {
+		return fmt.Errorf("data not present in the trusted collection, %v", err)
 	}
+
 	_, _ = os.Stdout.Write(payload)
 	return nil
 }
--- a/const.go
+++ b/const.go
@@ -20,6 +20,10 @@ const (
 	PubCertPerms = 0755
 	// Sha256HexSize is how big a Sha256 hex is in number of characters
 	Sha256HexSize = 64
+	// SHA256 is the name of SHA256 hash algorithm
+	SHA256 = "sha256"
+	// SHA512 is the name of SHA512 hash algorithm
+	SHA512 = "sha512"
 	// TrustedCertsDir is the directory, under the notary repo base directory, where trusted certs are stored
 	TrustedCertsDir = "trusted_certificates"
 	// PrivDir is the directory, under the notary repo base directory, where private keys are stored
--- a/server/timestamp/timestamp.go
+++ b/server/timestamp/timestamp.go
@@ -97,11 +97,7 @@ func timestampExpired(ts *data.SignedTimestamp) bool {
 func snapshotExpired(ts *data.SignedTimestamp, snapshot []byte) bool {
 	// If this check failed, it means the current snapshot was not exactly what we expect
 	// via the timestamp. So we can consider it to be "expired."
-	if err := data.CheckHashes(snapshot, ts.Signed.Meta["snapshot"].Hashes); err != nil {
-		return true
-	}
-
-	return false
+	return data.CheckHashes(snapshot, ts.Signed.Meta[data.CanonicalSnapshotRole].Hashes) != nil
 }
 
 // CreateTimestamp creates a new timestamp. If a prev timestamp is provided, it
--- a/server/timestamp/timestamp.go
+++ b/server/timestamp/timestamp.go
@@ -1,8 +1,6 @@
 package timestamp
 
 import (
-	"bytes"
-
 	"github.com/docker/go/canonical/json"
 	"github.com/docker/notary/tuf/data"
 	"github.com/docker/notary/tuf/signed"
@@ -95,16 +93,15 @@ func timestampExpired(ts *data.SignedTimestamp) bool {
 	return signed.IsExpired(ts.Signed.Expires)
 }
 
+// snapshotExpired verifies the checksum(s) for the given snapshot using metadata from the timestamp
 func snapshotExpired(ts *data.SignedTimestamp, snapshot []byte) bool {
-	meta, err := data.NewFileMeta(bytes.NewReader(snapshot), "sha256")
-	if err != nil {
-		// if we can't generate FileMeta from the current snapshot, we should
-		// continue to serve the old timestamp if it isn't time expired
-		// because we won't be able to generate a new one.
-		return false
+	// If this check failed, it means the current snapshot was not exactly what we expect
+	// via the timestamp. So we can consider it to be "expired."
+	if err := data.CheckHashes(snapshot, ts.Signed.Meta["snapshot"].Hashes); err != nil {
+		return true
 	}
-	hash := meta.Hashes["sha256"]
-	return !bytes.Equal(hash, ts.Signed.Meta["snapshot"].Hashes["sha256"])
+
+	return false
 }
 
 // CreateTimestamp creates a new timestamp. If a prev timestamp is provided, it
--- a/tuf/client/client.go
+++ b/tuf/client/client.go
@@ -453,7 +453,7 @@ func (c *Client) downloadSigned(role string, size int64, expectedHashes data.Has
 		return nil, nil, err
 	}
 
-	if len(expectedHashes) != 0 {
+	if expectedHashes != nil {
 		if err := data.CheckHashes(raw, expectedHashes); err != nil {
 			return nil, nil, ErrChecksumMismatch{role: role}
 		}
--- a/tuf/client/client.go
+++ b/tuf/client/client.go
@@ -1,8 +1,6 @@
 package client
 
 import (
-	"bytes"
-	"crypto/sha256"
 	"encoding/json"
 	"fmt"
 	"path"
@@ -92,16 +90,16 @@ func (c *Client) update() error {
 func (c Client) checkRoot() error {
 	role := data.CanonicalRootRole
 	size := c.local.Snapshot.Signed.Meta[role].Length
-	hashSha256 := c.local.Snapshot.Signed.Meta[role].Hashes["sha256"]
+
+	expectedHashes := c.local.Snapshot.Signed.Meta[role].Hashes
 
 	raw, err := c.cache.GetMeta("root", size)
 	if err != nil {
 		return err
 	}
 
-	hash := sha256.Sum256(raw)
-	if !bytes.Equal(hash[:], hashSha256) {
-		return fmt.Errorf("Cached root sha256 did not match snapshot root sha256")
+	if err := data.CheckHashes(raw, expectedHashes); err != nil {
+		return fmt.Errorf("Cached root hashes did not match snapshot root hashes")
 	}
 
 	if int64(len(raw)) != size {
@@ -127,11 +125,19 @@ func (c *Client) downloadRoot() error {
 	// We can't read an exact size for the root metadata without risking getting stuck in the TUF update cycle
 	// since it's possible that downloading timestamp/snapshot metadata may fail due to a signature mismatch
 	var size int64 = -1
-	var expectedSha256 []byte
+
+	// We could not expect what the "snapshot" meta has specified.
+	//
+	// In some old clients, there is only the "sha256",
+	// but both "sha256" and "sha512" in the newer ones.
+	//
+	// And possibly more in the future.
+	var expectedHashes data.Hashes
+
 	if c.local.Snapshot != nil {
 		if prevRootMeta, ok := c.local.Snapshot.Signed.Meta[role]; ok {
 			size = prevRootMeta.Length
-			expectedSha256 = prevRootMeta.Hashes["sha256"]
+			expectedHashes = prevRootMeta.Hashes
 		}
 	}
 
@@ -144,8 +150,9 @@ func (c *Client) downloadRoot() error {
 	old := &data.Signed{}
 	version := 0
 
-	if expectedSha256 != nil {
-		// can only trust cache if we have an expected sha256 to trust
+	// Due to the same reason, we don't really know how many hashes are there.
+	if len(expectedHashes) != 0 {
+		// can only trust cache if we have an expected sha256(for example) to trust
 		cachedRoot, err = c.cache.GetMeta(role, size)
 	}
 
@@ -153,11 +160,11 @@ func (c *Client) downloadRoot() error {
 		logrus.Debug("didn't find a cached root, must download")
 		download = true
 	} else {
-		hash := sha256.Sum256(cachedRoot)
-		if !bytes.Equal(hash[:], expectedSha256) {
+		if err := data.CheckHashes(cachedRoot, expectedHashes); err != nil {
 			logrus.Debug("cached root's hash didn't match expected, must download")
 			download = true
 		}
+
 		err := json.Unmarshal(cachedRoot, old)
 		if err == nil {
 			root, err := data.RootFromSigned(old)
@@ -176,7 +183,7 @@ func (c *Client) downloadRoot() error {
 	var raw []byte
 	if download {
 		// use consistent download if we have the checksum.
-		raw, s, err = c.downloadSigned(role, size, expectedSha256)
+		raw, s, err = c.downloadSigned(role, size, expectedHashes)
 		if err != nil {
 			return err
 		}
@@ -322,8 +329,8 @@ func (c *Client) downloadSnapshot() error {
 		return tuf.ErrNotLoaded{Role: data.CanonicalTimestampRole}
 	}
 	size := c.local.Timestamp.Signed.Meta[role].Length
-	expectedSha256, ok := c.local.Timestamp.Signed.Meta[role].Hashes["sha256"]
-	if !ok {
+	expectedHashes := c.local.Timestamp.Signed.Meta[role].Hashes
+	if len(expectedHashes) == 0 {
 		return data.ErrMissingMeta{Role: "snapshot"}
 	}
 
@@ -336,11 +343,11 @@ func (c *Client) downloadSnapshot() error {
 		download = true
 	} else {
 		// file may have been tampered with on disk. Always check the hash!
-		genHash := sha256.Sum256(raw)
-		if !bytes.Equal(genHash[:], expectedSha256) {
+		if err := data.CheckHashes(raw, expectedHashes); err != nil {
 			logrus.Debug("hash of snapshot in cache did not match expected hash, must download")
 			download = true
 		}
+
 		err := json.Unmarshal(raw, old)
 		if err == nil {
 			snap, err := data.SnapshotFromSigned(old)
@@ -357,7 +364,7 @@ func (c *Client) downloadSnapshot() error {
 	}
 	var s *data.Signed
 	if download {
-		raw, s, err = c.downloadSigned(role, size, expectedSha256)
+		raw, s, err = c.downloadSigned(role, size, expectedHashes)
 		if err != nil {
 			return err
 		}
@@ -439,18 +446,19 @@ func (c *Client) downloadTargets(role string) error {
 	return nil
 }
 
-func (c *Client) downloadSigned(role string, size int64, expectedSha256 []byte) ([]byte, *data.Signed, error) {
-	rolePath := utils.ConsistentName(role, expectedSha256)
+func (c *Client) downloadSigned(role string, size int64, expectedHashes data.Hashes) ([]byte, *data.Signed, error) {
+	rolePath := utils.ConsistentName(role, expectedHashes["sha256"])
 	raw, err := c.remote.GetMeta(rolePath, size)
 	if err != nil {
 		return nil, nil, err
 	}
-	if expectedSha256 != nil {
-		genHash := sha256.Sum256(raw)
-		if !bytes.Equal(genHash[:], expectedSha256) {
+
+	if len(expectedHashes) != 0 {
+		if err := data.CheckHashes(raw, expectedHashes); err != nil {
 			return nil, nil, ErrChecksumMismatch{role: role}
 		}
 	}
+
 	s := &data.Signed{}
 	err = json.Unmarshal(raw, s)
 	if err != nil {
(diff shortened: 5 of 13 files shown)


tuf/tuf.go
```diff
@@ -760,7 +760,7 @@ func (tr *Repo) UpdateSnapshot(role string, s *data.Signed) error {
 	if err != nil {
 		return err
 	}
-	meta, err := data.NewFileMeta(bytes.NewReader(jsonData), "sha256")
+	meta, err := data.NewFileMeta(bytes.NewReader(jsonData), "sha256", "sha512")
```
HuKeping: Actually there are two way to implement this.

One is check the present hash algorithm first and then only update
what we have.

The other is update/add both sha256 and sha512 no matter whether we
have the hash of sha512 or not.

Personally I prefer the latter, for it brings much less change of the 
code and will also not affect the validate of the old clients.

endophage: I concur with your preference.

riyazdf: Yes agreed with the latter choice, we should keep compatible with old clients

cmd/notary/tuf.go
```diff
@@ -385,17 +386,35 @@ func (t *tufCommander) tufVerify(cmd *cobra.Command, args []string) error {
...
+
 	_, _ = os.Stdout.Write(payload)
 	return nil
 }
 
+func checkHash(payload []byte, hashes data.Hashes) error {
+	// k, v indicate the hash algorithm and the corresponding value
+	for k, v := range hashes {
+		switch k {
+		case "sha256":
```
riyazdf: nit: it'd be great if "sha256" and "sha512" could be constants in `const.go` (we could refactor `NewFileMeta` too)

endophage: :+1:

HuKeping: Please have a check if the constant name is OK

tuf/client/client.go
```diff
@@ -197,6 +207,27 @@ func (c *Client) downloadRoot() error {
 	return nil
 }
 
+func checkHashes(payload []byte, hashes data.Hashes) error {
```
HuKeping: Will use a common function to replace this

cmd/notary/tuf.go
```diff
@@ -385,17 +387,35 @@ func (t *tufCommander) tufVerify(cmd *cobra.Command, args []string) error {
...
-		return fmt.Errorf("notary: data not present in the trusted collection")
+	if err := checkHashes(payload, target.Hashes); err != nil {
+		return err
 	}
+
 	_, _ = os.Stdout.Write(payload)
 	return nil
 }
 
+func checkHashes(payload []byte, hashes data.Hashes) error {
```
HuKeping: Will use a common function to replace this

tuf/client/client_test.go
```diff
@@ -250,15 +251,25 @@ func TestChecksumMismatch(t *testing.T) {
...
+	origHashes := GetHashes(orig)
+
 	remoteStorage.SetMeta("targets", orig)
 
-	_, _, err = client.downloadSigned("targets", int64(len(orig)), origSha256[:])
+	_, _, err = client.downloadSigned("targets", int64(len(orig)), origHashes)
 	assert.IsType(t, ErrChecksumMismatch{}, err)
 }
 
+func GetHashes(raw []byte) data.Hashes {
```
HuKeping: It is ugly here, will use a common function for all the other code who want to generate the `hashes`

tuf/data/types.go
```diff
@@ -127,6 +128,68 @@ type FileMeta struct {
...
+			checksum := sha512.Sum512(payload)
+			if subtle.ConstantTimeCompare(checksum[:], v) == 0 {
+				return fmt.Errorf("%s checksum mismatched", k)
+			}
+		default:
+			return fmt.Errorf("unsupported hash algorithm")
+		}
+	}
+
+	return nil
```
cyli: This would return true if there were no hashes for the given target.  Could we add a check here to make sure at least one supported checksum hash is included in the `Hashes` object?

tuf/data/types.go
```diff
@@ -127,6 +128,68 @@ type FileMeta struct {
...
+		}
+	}
+
+	return hashes
+}
+
+// SanityCheckForHashes returns an error, or nil, depending on whether
+// the content of the hashes is valid or not. This only check that the
+// checksums are valid in case somebody is drunk.
+func SanityCheckForHashes(hashes Hashes) error {
```
riyazdf: Should we rename this function to something more descriptive about what it's checking, maybe like `CheckValidHashStructures`?

tuf/client/client.go
```diff
@@ -439,18 +446,20 @@ func (c *Client) downloadTargets(role string) error {
 	return nil
 }
 
-func (c *Client) downloadSigned(role string, size int64, expectedSha256 []byte) ([]byte, *data.Signed, error) {
-	rolePath := utils.ConsistentName(role, expectedSha256)
+func (c *Client) downloadSigned(role string, size int64, expectedHashes data.Hashes) ([]byte, *data.Signed, error) {
+	// TODO:(hukeping) not sure about this, since we havn't do the sanity check of the expectedHashes yes.
```
riyazdf: We didn't previously do the sanity check because it was possible that the hash was empty (and so we wouldn't append any hash to the rolePath).  It's expected of the caller to pass any valid hash it wants to retrieve.

We have to make sure that we keep consistent with which checksums we download, cc @endophage @cyli for any ideas/timeframe on when we should allow downloading sha512 addresses

tuf/data/types.go
endophage: I'm in agreement with @cyli. We should apply the same logic to the `SanityCheckForHashes` function, if there are no hashes, it should error.

tuf/data/types.go
```diff
@@ -127,6 +128,68 @@ type FileMeta struct {
...
+			if subtle.ConstantTimeCompare(checksum[:], v) == 0 {
+				return fmt.Errorf("%s checksum mismatched", k)
+			}
+		case notary.SHA512:
+			checksum := sha512.Sum512(payload)
+			if subtle.ConstantTimeCompare(checksum[:], v) == 0 {
+				return fmt.Errorf("%s checksum mismatched", k)
+			}
+		default:
+			return fmt.Errorf("unsupported hash algorithm")
```
endophage: This shouldn't be an error. We validate all the hashes we understand and ignore any we don't. We must validate at least one hash, and all _recognized_ hashes must validate.

tuf/client/client.go
endophage: Yep, the existing implementation was discussed as part of the consistent download work. When downloading a timestamp, and in some instances a root, we will not have `expectedHashes`.

Also, consistent download is more prescriptive than other hash usages so it will only support sha256 for the foreseeable future, hence passing the one required hash, rather than the map.

const.go
```diff
@@ -48,3 +52,10 @@ var NotaryDefaultExpiries = map[string]time.Duration{
 	"snapshot":  NotarySnapshotExpiry,
 	"timestamp": NotaryTimestampExpiry,
 }
+
+// NotaryDefaultHashes is the construct used to configure the default
+// supported hash algorithms.
+var NotaryDefaultHashes = map[string]string{
```
cyli: Since this only seems to be used to get a list of the supported hashes, should we change this to an array?  That way, instead calling `NewFileMeta(bytes.NewReader(rootJSON), "sha256", "sha512")`, we can do something like `NewFileMeta(bytes.NewReader(rootJSON), notary.SupportedHashFunctions...)`?

client/client.go
```diff
@@ -140,7 +140,7 @@ func NewTarget(targetName string, targetPath string) (*Target, error) {
 		return nil, err
 	}
 
-	meta, err := data.NewFileMeta(bytes.NewBuffer(b))
+	meta, err := data.NewFileMeta(bytes.NewBuffer(b), "sha256", "sha512")
```
cyli: Related to https://github.com/docker/notary/pull/593/files#r54474516 - if we could replace this with a constant that knows which hashes are supported, that would be great.  :)

But if you didn't want to have a single collection though, it'd still be nice to use the awesome constants that you added - `notary.SHA256` and `notary.SHA512`

const.go
riyazdf: I agree with @cyli -- this will be clearer to understand and it'd also be nice to get rid of as many `"sha256"` and `"sha512"` strings in the code as possible.

endophage: :+1:

client/client.go
endophage: Rather than defining in const.go, if we want to define a slice of supported checksums, it would be good to put it right above the `NewFileMeta` definition just so it's obvious that it's available for use with that function.

tuf/data/types.go
```diff
@@ -127,6 +128,68 @@ type FileMeta struct {
...
+			hashes[v] = checksum[:]
+		}
+	}
+
+	return hashes
+}
+
+// SanityCheckForHashes returns an error, or nil, depending on whether
+// the content of the hashes is valid or not. This only check that the
+// checksums are valid in case somebody is drunk.
```
riyazdf: also, this is a fun comment but we should probably keep comments related to the code :smile:

tuf/data/types_test.go
```diff
@@ -53,3 +53,83 @@ func TestSignatureUnmarshalJSON(t *testing.T) {
...
+	assert.NoError(t, err)
+	err = SanityCheckForHashes(hashes)
+	assert.NoError(t, err)
+
+	hashes["sha512"], err = hex.DecodeString("795d9e95db099464b6730844f28effddb010b0d5abae5d5892a6ee04deacb09c9e622f89e816458b5a1a81761278d7d3a6a7c269d9707eff8858b16c51de0315")
+	assert.NoError(t, err)
+	err = SanityCheckForHashes(hashes)
+	assert.NoError(t, err)
+
+	// Also should be succeed since only check the lenth of the checksum
```
riyazdf: typo: length

tuf/data/types.go
endophage: Agreed, but if you (@HuKeping) make it out to San Francisco, or a DockerCon, we're totally buying you a beer :-)

HuKeping:
Thanks for all the suggestions, I really appreciate that.  It seems there is still a lot of work to do but thanks for the time-zone problem, it will be ready tomorrow :laughing:

const.go
HuKeping: Great idea!

client/client.go
HuKeping: OK, that's would be much more clearer.

tuf/data/types.go
HuKeping: Done.

tuf/data/types.go
HuKeping: I add a increment count and check if it equals to 0 before the `return`

tuf/data/types.go
HuKeping: updated.

tuf/data/types.go
HuKeping: Updated.

tuf/data/types_test.go
HuKeping: Done.

tuf/data/types_test.go
```diff
@@ -53,3 +53,108 @@ func TestSignatureUnmarshalJSON(t *testing.T) {
...
 }
+
+func TestCheckHashes(t *testing.T) {
+	var err error
+	raw := []byte("Bumblebee")
+
+	// Since only provide an un-supported hash algorithm here,
+	// it should be considered as fail.
+	unSupported := make(Hashes)
+	unSupported["Arthas"] = []byte("is past away.")
```
cyli: All now hail Fordragon :)

cyli:
This is awesome, thanks @HuKeping.  Could we add one more case for TestCheckHashes where we have one correct supported hash and one incorrect supported hash?  To test that all the provided hashes are checked.

Other than that LGTM!

server/timestamp/timestamp.go
```diff
@@ -95,16 +93,15 @@ func timestampExpired(ts *data.SignedTimestamp) bool {
 	return signed.IsExpired(ts.Signed.Expires)
 }
 
+// snapshotExpired verify the checksum(s) for the given snapshot via timestamp's data field.
```
riyazdf: nit: "snapshotExpired verifies the checksum(s) for the given snapshot using metadata from the timestamp"

tuf/data/snapshot.go
```diff
@@ -39,10 +38,18 @@ func isValidSnapshotStructure(s Snapshot) error {
 		// Meta is a map of FileMeta, so if the role isn't in the map it returns
 		// an empty FileMeta, which has an empty map, and you can check on keys
 		// from an empty map.
-		if checksum, ok := s.Meta[role].Hashes["sha256"]; !ok || len(checksum) != sha256.Size {
+		//
+		// For now sha256 is must and sha512 is not.
```
riyazdf: nit: replace `must` with `required`

tuf/data/types_test.go
```diff
@@ -53,3 +53,108 @@ func TestSignatureUnmarshalJSON(t *testing.T) {
...
+
+	// The most standard one.
+	hashes["sha256"], err = hex.DecodeString("d13e2b60d74c2e6f4f449b5e536814edf9a4827f5a9f4f957fc92e77609b9c92")
+	assert.NoError(t, err)
+	hashes["sha512"], err = hex.DecodeString("f2330f50d0f3ee56cf0d7f66aad8205e0cb9972c323208ffaa914ef7b3c240ae4774b5bbd1db2ce226ee967cfa9058173a853944f9b44e2e08abca385e2b7ed4")
+	assert.NoError(t, err)
+	err = CheckHashes(raw, hashes)
+	assert.NoError(t, err)
+
+	// Expected as succeed since there are already supported hash here,
```
riyazdf: nit: replace `succeed` with `success`

tuf/data/types_test.go
```diff
@@ -53,3 +53,108 @@ func TestSignatureUnmarshalJSON(t *testing.T) {
...
+	var err error
+	hashes := make(Hashes)
+
+	// Expected to fail since there is no checksum at all.
+	err = CheckValidHashStructures(hashes)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "at least one supported hash needed")
+
+	// Expected to fail even though the checksum of sha384 is valid,
+	// because we havn't provide a supported hash algorithm yet.
```
riyazdf: cool test!

Also nit on the last phrase: `because we haven't supported that hash algorithm yet`

tuf/data/types_test.go
```diff
@@ -53,3 +53,108 @@ func TestSignatureUnmarshalJSON(t *testing.T) {
...
+	assert.NoError(t, err)
+	err = CheckValidHashStructures(hashes)
+	assert.NoError(t, err)
+
+	// Also should be succeed since only check the length of the checksum
+	hashes["sha256"], err = hex.DecodeString("01234567890a4f2307e49160fa242db6fb95f071ad81a198eeb7d770e61cd6d8")
+	err = CheckValidHashStructures(hashes)
+	assert.NoError(t, err)
+
+	// Should faild since the first '0' is missed
```
riyazdf: typo nit: replace `faild` with `failed`, and `missed` with `missing`

riyazdf:
I have some small comments but overall this is looking really good!

LGTM after comments and @cyli's test suggestion

tuf/data/types_test.go
HuKeping: :smile_cat:

tuf/data/types_test.go
HuKeping: @riyazdf `we haven't supported that hash algorithm yet` doesn't always mean to fail this test, in this case it failed because only `sha384` was provided. 

If the `sha384` comes along with a supported one, let's say `sha256`, then this case will expected as success

HuKeping:
Updated and added the test as the last case in `TestCheckHashes`

tuf/data/types_test.go
riyazdf: oops, yes you're right that it should be something more like `we haven't provided a supported hash algorithm yet (ex: sha256).`

tuf/data/types_test.go
```diff
@@ -53,3 +53,119 @@ func TestSignatureUnmarshalJSON(t *testing.T) {
...
+	// The most standard one.
+	hashes["sha256"], err = hex.DecodeString("d13e2b60d74c2e6f4f449b5e536814edf9a4827f5a9f4f957fc92e77609b9c92")
+	assert.NoError(t, err)
+	hashes["sha512"], err = hex.DecodeString("f2330f50d0f3ee56cf0d7f66aad8205e0cb9972c323208ffaa914ef7b3c240ae4774b5bbd1db2ce226ee967cfa9058173a853944f9b44e2e08abca385e2b7ed4")
+	assert.NoError(t, err)
+	err = CheckHashes(raw, hashes)
+	assert.NoError(t, err)
+
+	// Expected as success since there are already supported hash here,
+	// just ignore the unsupported one.
```
riyazdf: :+1: thanks!

tuf/data/types_test.go
HuKeping: Updated :smiley:

tuf/data/types_test.go
```diff
@@ -53,3 +53,119 @@ func TestSignatureUnmarshalJSON(t *testing.T) {
...
+	// Expected to fail because of the failure of sha512
+	// even though the sha256 is OK.
+	doubleFace := make(Hashes)
+	doubleFace["sha256"], err = hex.DecodeString("d13e2b60d74c2e6f4f449b5e536814edf9a4827f5a9f4f957fc92e77609b9c92")
+	assert.NoError(t, err)
+	doubleFace["sha512"], err = hex.DecodeString("d13e2b60d74c2e6f4f449b5e536814edf9a4827f5a9f4f957fc92e77609b9c92")
+	assert.NoError(t, err)
+	err = CheckHashes(raw, doubleFace)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "checksum mismatched")
```
riyazdf: and also thank you for adding this case :)

endophage:
We will do a team review of this on Monday (there are conflicts preventing it happening earlier) to determine if there are any security or compatibility issues. Will either merge or request final fixes then.

HuKeping:
Sure @endophage  :)

server/timestamp/timestamp.go
```diff
@@ -95,16 +93,15 @@ func timestampExpired(ts *data.SignedTimestamp) bool {
...
 func snapshotExpired(ts *data.SignedTimestamp, snapshot []byte) bool {
-	meta, err := data.NewFileMeta(bytes.NewReader(snapshot), "sha256")
-	if err != nil {
-		// if we can't generate FileMeta from the current snapshot, we should
-		// continue to serve the old timestamp if it isn't time expired
-		// because we won't be able to generate a new one.
-		return false
+	// If this check failed, it means the current snapshot was not exactly what we expect
+	// via the timestamp. So we can consider it to be "expired."
+	if err := data.CheckHashes(snapshot, ts.Signed.Meta["snapshot"].Hashes); err != nil {
```
endophage: Please replace `"snapshot"` with `data.CanonicalSnapshotRole`. We're trying to fix all the occurrences that we use literal strings for roles with the constants defined in the data package. Whenever you touch code that uses a string literal, please replace :-)

endophage: Also, this function can just be reduced to:
```
return data.CheckHashes(snapshot, ts.Signed.Meta["snapshot"].Hashes) != nil
```

tuf/data/timestamp.go
```diff
@@ -37,10 +36,17 @@ func isValidTimestampStructure(t Timestamp) error {
 	// Meta is a map of FileMeta, so if the role isn't in the map it returns
 	// an empty FileMeta, which has an empty map, and you can check on keys
 	// from an empty map.
-	if cs, ok := t.Meta[CanonicalSnapshotRole].Hashes["sha256"]; !ok || len(cs) != sha256.Size {
+	//
+	// For now sha256 is required and sha512 is not.
+	if _, ok := t.Meta[CanonicalSnapshotRole].Hashes["sha256"]; !ok {
```
cyli: Can we use the constant `notary.SHA256` here instead of the quoted string?

tuf/data/snapshot.go
```diff
@@ -39,10 +38,18 @@ func isValidSnapshotStructure(s Snapshot) error {
 		// Meta is a map of FileMeta, so if the role isn't in the map it returns
 		// an empty FileMeta, which has an empty map, and you can check on keys
 		// from an empty map.
-		if checksum, ok := s.Meta[role].Hashes["sha256"]; !ok || len(checksum) != sha256.Size {
+		//
+		// For now sha256 is required and sha512 is not.
+		if _, ok := s.Meta[role].Hashes["sha256"]; !ok {
```
cyli: Likewise, can we use the constant here?

tuf/data/types.go
```diff
@@ -127,6 +131,81 @@ type FileMeta struct {
...
+	if cnt == 0 {
+		return fmt.Errorf("at least one supported hash needed")
+	}
+
+	return nil
+}
+
+// GetSupportedHashes returns the checksums of all the supported hash algorithms
+// of the given payload
+func GetSupportedHashes(payload []byte) Hashes {
```
endophage: As this is only used in tests, can we move it to a testing package? Either a relevant `_test.go` file or the testutils package. Don't want it to be used by accident in production code.

riyazdf: Also, in addition to @endophage's comment -- could we have this function call `NewFileMeta` with `NotaryDefaultHashes` to generate the hashes, and then pull them out?  If this is used for testing only, it'd be good to use `NewFileMeta`

tuf/client/client.go
```diff
@@ -439,18 +446,19 @@ func (c *Client) downloadTargets(role string) error {
...
+	rolePath := utils.ConsistentName(role, expectedHashes["sha256"])
 	raw, err := c.remote.GetMeta(rolePath, size)
 	if err != nil {
 		return nil, nil, err
 	}
-	if expectedSha256 != nil {
-		genHash := sha256.Sum256(raw)
-		if !bytes.Equal(genHash[:], expectedSha256) {
+
+	if len(expectedHashes) != 0 {
```
cyli: Can we explicitly check for `expectedHashes == nil` here instead of the length?  If one of our other checks fail, and the hash we get for a role is there but empty, we should still run the `CheckHashes` function as opposed to bypassing it altogether.

cyli:
Apologies, we know in a lot of places you were just changing existing code, but if you happen to see anywhere in the tests or in the code where we are using the string `"targets"` or some other quoted role, instead of using a constant like `data.CanonicalTargetsRole` (similarly with "sha256" and "sha512"), would you mind changing that to use the constant instead?

HuKeping:
Thanks for all the comments, I'll update later :)

endophage:
@HuKeping Sounds good!

tuf/client/client.go
HuKeping: Sorry, I don't quite understand here. If the `expectedHashes == nil` the `CheckHashes ` will definitely fail, so why should we run it since the ` expectedHashes ` is nil?

HuKeping:
This change is a little big and for better review, I add 3 new commits to fix as the above comments, hope it was not a problem.

server/timestamp/timestamp.go
HuKeping: Thanks and commit [MISC 1/4] fixing this

tuf/data/timestamp.go
HuKeping: Thanks and commit `[MISC 2/4]` fixing this and the below one.

tuf/data/types.go
HuKeping: Thanks and commit [MISC 3/4] fixing this.

tuf/client/client.go
endophage: The problem is that `len(expectedHashes)` on a nil map is a valid operation and also returns length 0. We want a clear distinction between "I am explicitly saying there are no hashes to check" and "the snapshot/timestamp contained an empty hashes entry (probably as the result of malicious behaviour)". There are other checks that need to be done elsewhere because the hashes entry could itself be `null` in the JSON, but this particular method is generic and should distinguish the 2 above cases.

HuKeping: hmm, I see, the `hash[ ]` and `nil` should not be considered as one case.

endophage:
We want to make sure this gets in to Docker 1.11 so we're going to merge but there's one bit of cleanup we'd like to see: https://github.com/docker/notary/issues/619

It's great work and thank you for all the iterations. This is a really useful contribution that we would probably have waited much longer if you hadn't picked it up!

LGTM!

cyli:
LGTM!  Thanks @HuKeping

riyazdf:
LGTM, thanks for all of your work!!

Issue #287, opened 2015-11-14
notary publish only has client-side effect on second publish


diogomonica:
```
➜  ~  notary init docker.io/diogomonica/dc8
Root key found, using: 92dd31e2705af3c661f2cc297067f631339a668d3558890eb1131c90f1901a6e
Please touch the attached Yubikey to perform signing.
Enter username: diogomonica
Enter password:
Enter passphrase for new targets key with ID a551764 (docker.io/diogomonica/dc8):
Repeat passphrase for new targets key with ID a551764 (docker.io/diogomonica/dc8):
Please touch the attached Yubikey to perform signing.
➜  ~  notary list docker.io/diogomonica/dc8
ERRO[0002] Client Update (Timestamp): unexpected end of JSON input
ERRO[0002] client Update (Root):Unable to reach trust server at this time: 401.
* fatal: Unable to reach trust server at this time: 401.
➜  ~  notary add docker.io/diogomonica/dc8 foobar Dockerfile
Addition of target "foobar" to repository "docker.io/diogomonica/dc8" staged for next publish.
➜  ~  notary publish docker.io/diogomonica/dc8
Pushing changes to docker.io/diogomonica/dc8
Enter username: diogomonica
Enter password:
Enter passphrase for targets key with ID a551764 (docker.io/diogomonica/dc8):
d%                                                                                                                                                                                   ➜  ~  notary list docker.io/diogomonica/dc8
ERRO[0002] Client Update (Timestamp): unexpected end of JSON input
ERRO[0002] client Update (Root):Unable to reach trust server at this time: 401.
* fatal: Unable to reach trust server at this time: 401.
➜  ~  notary publish docker.io/diogomonica/dc8
Pushing changes to docker.io/diogomonica/dc8
Enter username: diogomonica
Enter password:
Enter passphrase for targets key with ID a551764 (docker.io/diogomonica/dc8):
➜  ~  notary list docker.io/diogomonica/dc8
foobar b77c77e694ae7746a15761e2a5f8f7da85f53f3bb215136ca14be1f1c5016fd1 463
➜  ~
```

endophage:
Cause analysis:

The token server will not issue a token for something that doesn't exist. Because the notary server that works with hub relies on hub's token server, the repo must be created in hub before you can do a `notary list ...`. Pushes work because the token server will always issue a push token for any name the user should have access to.

[...]

PR #594, opened 2016-02-25
Bugfix: the actions should not always be pull and push

HuKeping:
When I use the JWT as the authentication like what the [Distribution](https://github.com/docker/distribution/blob/master/docs/spec/auth/token.md) does, the same auth server refused me when I use `notary list xxx` because notary post both of the `push` and `pull` as the `ScopeAction`.

The similiar fix in docker daemon:
- https://github.com/docker/docker/pull/20382

It probably a fix of #287 

### DIFF
--- a/cmd/notary/tuf.go
+++ b/cmd/notary/tuf.go
@@ -153,7 +153,7 @@ func (t *tufCommander) tufInit(cmd *cobra.Command, args []string) error {
 	}
 	gun := args[0]
 
-	rt, err := getTransport(config, gun, false, "push", "pull")
+	rt, err := getTransport(config, gun, false)
 	if err != nil {
 		return err
 	}
@@ -198,7 +198,7 @@ func (t *tufCommander) tufList(cmd *cobra.Command, args []string) error {
 	}
 	gun := args[0]
 
-	rt, err := getTransport(config, gun, true, "pull")
+	rt, err := getTransport(config, gun, true)
 	if err != nil {
 		return err
 	}
@@ -233,7 +233,7 @@ func (t *tufCommander) tufLookup(cmd *cobra.Command, args []string) error {
 	gun := args[0]
 	targetName := args[1]
 
-	rt, err := getTransport(config, gun, true, "pull")
+	rt, err := getTransport(config, gun, true)
 	if err != nil {
 		return err
 	}
@@ -304,7 +304,7 @@ func (t *tufCommander) tufPublish(cmd *cobra.Command, args []string) error {
 
 	cmd.Println("Pushing changes to", gun)
 
-	rt, err := getTransport(config, gun, false, "push", "pull")
+	rt, err := getTransport(config, gun, false)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


cmd/notary/tuf.go
```diff
@@ -533,7 +533,7 @@ func tokenAuth(trustServerURL string, baseTransport *http.Transport, gun string,
 	}
 
 	ps := passwordStore{anonymous: readOnly}
-	tokenHandler := auth.NewTokenHandler(authTransport, ps, gun, "push", "pull")
```
cyli: It seems like we can just make the change here to just include `"pull"` when `readonly` is true, and `"push", "pull"` when `readonly` is false?  That way nothing else needs to know about what the values of the actions have to be.

endophage:
@HuKeping #287 is a slightly different problem related to the token server's implementation.

As I said over on the docker PR, this is good cleanup, but just for clarity because I know this has confused other contributors in the past, the token server is expected to issue tokens for whatever permissions are valid. So if a user only has read permissions on a repo, but askes for "push" and "pull" (a.k.a. read and write), the token server is expected to issue a "push" token and silently just not issue the "pull" token.

This is the [relevant documentation](https://github.com/docker/distribution/blob/master/docs/spec/auth/token.md), specifically:

> Once the token server has determined what access the client has to the resources requested in the scope parameter, it will take the intersection of the set of requested actions on each resource and the set of actions that the client has in fact been granted. If the client only has a subset of the requested access it must not be considered an error as it is not the responsibility of the token server to indicate authorization errors as part of this workflow.

cmd/notary/tuf.go
HuKeping: Oh I missed the `readonly` parameter and agreed your opinion.

On the other hand, do you think it's worth to drop the `readonly` and then use the `actions...` (Just find we have two solutions here)

HuKeping:
@endophage you are right, it related to the token server's implementation.

cmd/notary/tuf.go
cyli: On the one hand, the actions doesn't hide the `push` and `pull`, on the other hand it seems like then all the code would have to know about the exact auth mechanism?  It may make sense for most things to just know whether it requires read access or read and write access.

Possibly if you wanted to translate that to a constant instead (READ or READWRITE), that'd work?  I think we should abstract the fact that read access is sent as "pull" to the auth realm, and write access is sent as "push", away from most of the code, though.

HuKeping: Get it, will update later. Thanks @cyli

HuKeping:
Updated @cyli

endophage:
@HuKeping yeah, I've heard from the distribution team that Huawei has implemented a server that doesn't quite line up with the spec. I'm fine with this addition though. We really should only request the permissions we need as a matter of good practice.

cyli:
LGTM!  Thanks for the change @HuKeping

riyazdf:
I really like the way the code turned out, thanks!  LGTM

endophage:
LGTM!

riyazdf:
Jenkins: test this please

PR #601, opened 2016-02-26
Return cache control headers when returning metadata from server

### LINKED ISSUE #597: Set cache control headers
This will help with CDN cache management.

Anything with a checksum should be cached indefinitely, everything else should have relatively short cache times (a few minutes)

cyli:
This includes the `Cache-Control` header, the `Last-Modified` header, and the `ETag` header (just using sha256 for now).

Fixes #597

### DIFF
--- a/cmd/notary-server/config.go
+++ b/cmd/notary-server/config.go
@@ -11,7 +11,6 @@ import (
 	_ "github.com/docker/distribution/registry/auth/htpasswd"
 	_ "github.com/docker/distribution/registry/auth/token"
 	"github.com/docker/go-connections/tlsconfig"
-	"github.com/docker/notary/server/handlers"
 	"github.com/docker/notary/server/storage"
 	"github.com/docker/notary/signer/client"
 	"github.com/docker/notary/tuf/data"
@@ -123,7 +122,7 @@ func getTrustService(configuration *viper.Viper, sFactory signerFactory,
 	hRegister(
 		"Trust operational",
 		// If the trust service fails, the server is degraded but not
-		// exactly unheatlthy, so always return healthy and just log an
+		// exactly unhealthy, so always return healthy and just log an
 		// error.
 		func() error {
 			err := notarySigner.CheckHealth(minute)
@@ -136,30 +135,29 @@ func getTrustService(configuration *viper.Viper, sFactory signerFactory,
 	return notarySigner, keyAlgo, nil
 }
 
-// Gets the cache configuration for GET-ting metadata.  This is the max-age
-// (an integer in seconds, just like in the Cache-Control header) for consistent
-// (content-addressable) downloads and current (latest version) downloads.
-// The max-age must be between 0 and 31536000 (one year in seconds, which is
-// the recommended maximum time data is cached), else parsing will return an
-// error.  A max-age of 0 will disable caching for that type of download
-// (consistent or current).
-func getCacheConfig(configuration *viper.Viper) (*handlers.CacheControlConfig, error) {
-	cacheConfig := handlers.NewCacheControlConfig()
-	for option, setMaxAge := range map[string]func(int){
-		"current_metadata":     cacheConfig.SetCurrentCacheMaxAge,
-		"metadata_by_checksum": cacheConfig.SetConsistentCacheMaxAge,
-	} {
-		m := configuration.GetString(fmt.Sprintf("caching.max_age.%s", option))
+// Gets the cache configuration for GET-ting current and checksummed metadata
+// This is mainly the max-age (an integer in seconds, just like in the
+// Cache-Control header) for consistent (content-addressable) downloads and
+// current (latest version) downloads. The max-age must be between 0 and 31536000
+// (one year in seconds, which is the recommended maximum time data is cached),
+// else parsing will return an error.  A max-age of 0 will disable caching for
+// that type of download (consistent or current).
+func getCacheConfig(configuration *viper.Viper) (utils.CacheControlConfig, utils.CacheControlConfig, error) {
+	var cccs []utils.CacheControlConfig
+	types := []string{"current_metadata", "metadata_by_checksum"}
+
+	for _, optionName := range types {
+		m := configuration.GetString(fmt.Sprintf("caching.max_age.%s", optionName))
 		if m == "" {
 			continue
 		}
 		seconds, err := strconv.Atoi(m)
 		if err != nil || seconds < 0 || seconds > maxMaxAge {
-			return nil, fmt.Errorf(
+			return nil, nil, fmt.Errorf(
 				"must specify a cache-control max-age between 0 and %v", maxMaxAge)
 		}
 
-		setMaxAge(seconds)
+		cccs = append(cccs, utils.NewCacheControlConfig(seconds, optionName == "current_metadata"))
 	}
-	return cacheConfig, nil
+	return cccs[0], cccs[1], nil
 }
--- a/cmd/notary-server/config.go
+++ b/cmd/notary-server/config.go
@@ -1,0 +1,165 @@
+package main
+
+import (
+	"crypto/tls"
+	"fmt"
+	"strconv"
+	"time"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/docker/distribution/health"
+	_ "github.com/docker/distribution/registry/auth/htpasswd"
+	_ "github.com/docker/distribution/registry/auth/token"
+	"github.com/docker/go-connections/tlsconfig"
+	"github.com/docker/notary/server/handlers"
+	"github.com/docker/notary/server/storage"
+	"github.com/docker/notary/signer/client"
+	"github.com/docker/notary/tuf/data"
+	"github.com/docker/notary/tuf/signed"
+	"github.com/docker/notary/utils"
+	_ "github.com/go-sql-driver/mysql"
+	"github.com/spf13/viper"
+)
+
+// get the address for the HTTP server, and parses the optional TLS
+// configuration for the server - if no TLS configuration is specified,
+// TLS is not enabled.
+func getAddrAndTLSConfig(configuration *viper.Viper) (string, *tls.Config, error) {
+	httpAddr := configuration.GetString("server.http_addr")
+	if httpAddr == "" {
+		return "", nil, fmt.Errorf("http listen address required for server")
+	}
+
+	tlsConfig, err := utils.ParseServerTLS(configuration, false)
+	if err != nil {
+		return "", nil, fmt.Errorf(err.Error())
+	}
+	return httpAddr, tlsConfig, nil
+}
+
+// sets up TLS for the GRPC connection to notary-signer
+func grpcTLS(configuration *viper.Viper) (*tls.Config, error) {
+	rootCA := utils.GetPathRelativeToConfig(configuration, "trust_service.tls_ca_file")
+	clientCert := utils.GetPathRelativeToConfig(configuration, "trust_service.tls_client_cert")
+	clientKey := utils.GetPathRelativeToConfig(configuration, "trust_service.tls_client_key")
+
+	if clientCert == "" && clientKey != "" || clientCert != "" && clientKey == "" {
+		return nil, fmt.Errorf("either pass both client key and cert, or neither")
+	}
+
+	tlsConfig, err := tlsconfig.Client(tlsconfig.Options{
+		CAFile:   rootCA,
+		CertFile: clientCert,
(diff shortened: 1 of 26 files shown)


server/handlers/default.go
```diff
@@ -20,6 +25,14 @@ import (
 	"github.com/docker/notary/tuf/validation"
 )
 
+const (
+	// ConsistentCacheMaxAge is the Cache-Control header's max age for consistent downloads
+	ConsistentCacheMaxAge int = 30 * 24 * 60 * 60 // 30 days
+
+	// NonConsistentCacheMaxAge is the Cache-Control header's max age for current (non-consistent) downloads
+	NonConsistentCacheMaxAge int = 5 * 60 // five minutes
```
riyazdf: it might be nice to use a `time.Duration` type here, like we do for other times in `const.go`

riyazdf: IRL discussion with @cyli: it seems that making these cache times configurable would also be nice

server/snapshot/snapshot_test.go
```diff
@@ -144,18 +144,23 @@ func TestGetSnapshotCurrValid(t *testing.T) {
 
 	// test when db is missing the role data
 	store.UpdateCurrent("gun", storage.MetaUpdate{Role: "snapshot", Version: 0, Data: snapJSON})
-	_, err = GetOrCreateSnapshot("gun", store, crypto)
+	c1, result, err := GetOrCreateSnapshot("gun", store, crypto)
 	assert.NoError(t, err)
+	assert.True(t, bytes.Equal(snapJSON, result))
```
riyazdf: thanks for adding extra checking :+1:

server/storage/database.go
```diff
@@ -124,16 +125,17 @@ func (db *SQLStorage) UpdateMany(gun string, updates []MetaUpdate) error {
 }
 
 // GetCurrent gets a specific TUF record
-func (db *SQLStorage) GetCurrent(gun, tufRole string) ([]byte, error) {
+func (db *SQLStorage) GetCurrent(gun, tufRole string) (*time.Time, []byte, error) {
 	var row TUFFile
-	q := db.Select("data").Where(&TUFFile{Gun: gun, Role: tufRole}).Order("version desc").Limit(1).First(&row)
+	q := db.Select("created_at, data").Where(
```
riyazdf: From IRL discussion: we should use `updated_at` here

server/handlers/default.go
endophage: Yeah, configurable times would be great!

server/storage/database.go
endophage: Yep. And `created_at` is correct for GetChecksum due to the immutability aspect.

server/handlers/default.go
cyli: Do we want to allow configuration to disable cache headers altogether?

riyazdf: I'm fine either way because I'm not sure if I see a use case for not caching.  I guess if we don't make enabling/disabling  configurable, how would we handle the `0` time?

cyli: I was just going to return an error, but that's a good idea, I'll make 0 time disabling.

endophage: :+1:

cyli: Ok, working on the config now, but I think I might skip the `time.Duration`.  Cache control max age is in seconds, so you can't actually pass a smaller unit than second in the configuration (and it makes sense to have the configuration value just be the max-age value, since if users need to configure it they probably are already familiar with and have a value in mind for the max age value).

And converting from seconds to duration and then back, and then rounding and converting to an int seems maybe unnecessary if I only allow ints?

cmd/notary-server/main.go
```diff
@@ -32,6 +34,7 @@ import (
 const (
 	jsonLogFormat = "json"
 	DebugAddress  = "localhost:8080"
+	maxMaxAge     = 31536000
```
riyazdf: Could we add a comment to give some context behind this number?

cmd/notary-server/main.go
```diff
@@ -170,6 +173,29 @@ func getTrustService(configuration *viper.Viper, sFactory signerFactory,
 	return notarySigner, keyAlgo, nil
 }
 
+func getCacheConfig(configuration *viper.Viper) (*handlers.CacheControlConfig, error) {
```
riyazdf: this is awesome :+1:

server/handlers/default.go
```diff
@@ -20,6 +25,69 @@ import (
...
+	c.headerVals["current"] = value
+}
+
+// UpdateConsistentHeaders updates the given Headers object with the Cache-Control
+// headers for consistent downloads
+func (c *CacheControlConfig) UpdateConsistentHeaders(headers http.Header, lastModified time.Time) {
+	c.updateHeaders(headers, lastModified, true)
+}
+
+// UpdateCurrentHeaders updates the given Headers object with th eCache-Control
```
riyazdf: typo nit: "the"

server/handlers/default.go
```diff
@@ -20,6 +25,69 @@ import (
...
+
+	if consistent {
+		seconds = c.headerVals["consistent"]
+		cacheHeader = fmt.Sprintf("public, max-age=%v, s-maxage=%v, must-revalidate", seconds, seconds)
+	} else {
+		seconds = c.headerVals["current"]
+		cacheHeader = fmt.Sprintf("public, max-age=%v, s-maxage=%v", seconds, seconds)
+	}
+
+	if seconds > 0 {
```
riyazdf: It might be worth documenting somewhere that `0` time will mean that we don't cache and set the headers appropriately

riyazdf:
LGTM after comments!

server/handlers/default.go
cyli: Done - added this to the top level comment in `CacheControlConfig` and in the parsing in `cmd/notary-server/main.go`.

riyazdf: :+1:

cmd/notary-server/main.go
```diff
@@ -32,6 +34,10 @@ import (
 const (
 	jsonLogFormat = "json"
 	DebugAddress  = "localhost:8080"
+	// This is the generally recommended maximum age for Cache-Control headers
+	// (one year, in seconds, since one year is forever in terms of internet
+	// content)
+	maxMaxAge = 31536000
```
endophage: After the discussions about the expiries would `60 * 60 * 24 * 365` be clearer about what the value is (without reading the comment)?

cmd/notary-server/main.go
```diff
@@ -170,6 +176,36 @@ func getTrustService(configuration *viper.Viper, sFactory signerFactory,
...
 }
 
+// Gets the cache configuration for GET-ting metadata.  This is the max-age
+// (an integer in seconds, just like in the Cache-Control header) for consistent
+// (content-addressable) downloads and current (latest version) downloads.
+// The max-age must be between 0 and 31536000 (one year in seconds, which is
+// the recommended maximum time data is cached), else parsing will return an
+// error.  A max-age of 0 will disable caching for that type of download
+// (consistent or current).
+func getCacheConfig(configuration *viper.Viper) (*handlers.CacheControlConfig, error) {
```
endophage: Thinking maybe this and the other config slicing functions should maybe be moved in a `cmd/notary-server/(config|setup).go` file? There's ~150 LOC which seems like a sufficient amount to justify its own file rather than creating noise with the `main()` in here.

cmd/notary-server/main.go
```diff
@@ -170,6 +176,36 @@ func getTrustService(configuration *viper.Viper, sFactory signerFactory,
...
+	cacheConfig := handlers.NewCacheControlConfig()
+	for _, option := range []string{"current_metadata", "metadata_by_checksum"} {
```
endophage:
```
for option, func := range map[string]func(int){
    "current_metadata": cacheConfig.SetCurrentCacheMaxAge,
    "metadata_by_checksum": cacheConfig.SetConsistentCacheMaxAge,
}{
    ...
    // replace switch with just
    func(seconds)
}
```

cyli: Oh nice, will change that, thanks

endophage:
Sorry to jump in with this idea so late, but I don't think it's a huge change (if everyone even agrees). I'm wondering if it makes sense to implement cache headers as a chained handler. Then we would just have to add another wrapping handler in server.go and we'd be done. 

The cache handler would pass a wrapped version of the ResponseWriter to the next handler in the chain. That wrapped version would determine based on the functions called whether or not to set cached headers.

This might also help make it clear what the caching policy is for each endpoint as we would simply declare multiple cache handler instances configured differently and wrap the immutable endpoints with a long duration instance.

cyli:
@endophage I wanted to do that at first, but you seem to have to call `Write` last unless you want to write trailers and not headers?

```
type ResponseWriter interface {
        // Header returns the header map that will be sent by
        // WriteHeader. Changing the header after a call to
        // WriteHeader (or Write) has no effect unless the modified
        // headers were declared as trailers by setting the
        // "Trailer" header before the call to WriteHeader (see example).
        // To suppress implicit response headers, set their value to nil.
        Header() Header
...
```

endophage:
@cyli yep, our cache handler would wrap write and determine if a) headers had already been written, and if not, b) based on the status code, should headers be written, before passing through the write call to the actual ResponseWriter

cyli:
@endophage Oh, I see, so it'd do that in `Write`?

endophage:
yep, possibly in WriteHeader too. I'm not sure if calling WriteHeader blocks you from calling Header afterwards.

cyli:
@endophage I think it doesn't block, but we probably want to wrap it anyway to make sure it was either not called or called with a 200 before writing the headers.  Thanks!  This is much neater :)

riyazdf:
@endophage +1 great idea!  Should definitely consolidate the caching logic

cmd/notary-server/main.go
```diff
@@ -32,6 +23,10 @@ import (
 const (
 	jsonLogFormat = "json"
 	DebugAddress  = "localhost:8080"
+	// This is the generally recommended maximum age for Cache-Control headers
+	// (one year, in seconds, since one year is forever in terms of internet
+	// content)
+	maxMaxAge = 60 * 60 * 24 * 365
```
endophage: should this go in notary/const.go?

cyli: I can definitely move it - I wasn't sure if it should go there, since it's only used for the server config, and it's an int of seconds instead of a duration (since the header is in ints, and the config should probably be an int since that's what people are probably used to if they want to set a cache control header)

endophage:
@cyli I don't feel too strongly about moving the maxMaxAge const, we can move it later if and when we need to.

Love this PR. LATM! (Looks AWESOME To Me)

cmd/notary-server/config.go
```diff
@@ -0,0 +1,163 @@
...
+		configuration.GetString("trust_service.hostname"),
+		configuration.GetString("trust_service.port"),
+		clientTLS,
+	)
+
+	minute := 1 * time.Minute
+	hRegister(
+		"Trust operational",
+		// If the trust service fails, the server is degraded but not
+		// exactly unheatlthy, so always return healthy and just log an
```
riyazdf: typo: unhealthy

utils/http.go
```diff
@@ -94,3 +96,102 @@ func buildAccessRecords(repo string, actions ...string) []auth.Access {
...
+
+// WriteHeader stores the header before writing it, so we can tell if it's been set
+// to a non-200 status code
+func (c *cacheControlResponseWriter) WriteHeader(statusCode int) {
+	c.statusCode = statusCode
+	c.ResponseWriter.WriteHeader(statusCode)
+}
+
+// Write will set the cache headers if they haven't already been set and if the status
+// code has either not been set or
```
riyazdf: Seems like an incomplete comment, should it be completed with `...or is not 200`?

cmd/notary-server/main.go
riyazdf: I don't feel too strongly about this constant either since it's only used here, when we use it in another place we can move it to `const.go`

riyazdf:
I have some super small nit comments, but LGTM!

utils/http.go
cyli: Oops sorry, should be `...is 200` :)

riyazdf: ah...yes :+1:

endophage:
There are no documentation changes in this PR and the docs job appears to not be running. We're not waiting on it to merge.

PR #603, opened 2016-02-29
Carry 293

moxiegirl:
Carry and close #293 

* Formatting issues throughout
* Moved configuration into its own section.
* Fixed link formats 
* Rely on generated "on this page"
* Terminology in line with company standard

PR #606, opened 2016-03-01
Documentation fixes

riyazdf:
Fixes to the docs from feedback in #603


docs/reference/client-config.md
```diff
@@ -21,38 +20,25 @@ but the path to a different configuration file can be specified using the
...
+JSON keys to learn more about the configuration section corresponding to that key:
 
-In addition please see the optional password [environment
-variables](#environment-variables-optional) that the Notary client can take for
-ease of use.
 
-An example (full) clietn configuration file.
-
-```json
+<pre><code class="language-json">
```
endophage: have we confirmed the markdown compiler will handle this tag correctly?

riyazdf: yup, it looks pretty good when I generate the docs :)

screen shot 2016-03-01 at 3 10 31 pm

endophage: :+1: nice

endophage: Although, looks like it's including the initial newline at the end of the tag. Want to try moving the `{` to immediately after the opening `<code>` tag?

docs/service_architecture.md
```diff
@@ -104,53 +103,57 @@ sever, and signer:
...
-    attacks.
+    attacks, so long as the attacker ensures that the version number of their
+    malicious metadata is higher than the version number of the most recent
+    good metadata that any client may have.
 
- Note that the timestamp and snapshot keys are never compromised; once the
- Server compromise is mitigated, an attacker will not be able to generate
- valid timestamp or snapshot metadata and serve them on a malicious mirror, for
- example.
+ Note that the timestamp and snapshot keys cannot be compromised in a server-only
```
endophage: @cyli saw your comment on the previous docs PR about this. It read a bit too much like a statement in the original version, which was a little confusing. The extra context here makes the meaning clearer.

endophage:
Just the one newline to be fixed then rebase and LGTM.

docs/reference/client-config.md
riyazdf: good catch, done!

docs/service_architecture.md
cyli: Ah ok, cool thanks.  Did we want to move the threat model stuff into another doc page as you suggested in the previous PR?

riyazdf: I'm fine with this as is -- the threat model makes much more sense with the architecture explanation as prior context.

endophage: Yeah, Diogo and I also agreed that this was good to go as is. At a later point we might want to write up a more detailed and exhaustive threat model and at that point we would separate it out.

PR #607, opened 2016-03-01
Move all imports of sqlite3 to tests only

cyli:


endophage:
LGTM!

riyazdf:
LGTM

endophage:
Are there any docs updates that need to be made in relation to not supporting sqlite?

cyli:
@endophage I don't think we said in the docs that we supported sqlite3 at all, just MySQL or an in-memory store.

endophage:
:+1: 

merging then

PR #616, opened 2016-03-09
Remove delegation role fallback when applying targets changes

riyazdf:
Given the behavior proposed in https://github.com/docker/docker/pull/21046, the fallback behavior is no longer that necessary -- this PR removes the fallback logic when applying changes on targets entirely.

Docker will be able to determine the roles it will sign for from TUF metadata and private keys in its trust directory, so there is no need to silently fallback on the notary side (I'm especially worried about any potential role confusion)

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -2046,8 +2046,7 @@ func createKey(t *testing.T, repo *NotaryRepository, role string, x509 bool) dat
 // Publishing delegations works so long as the delegation parent exists by the
 // time that delegation addition change is applied.  Most of the tests for
 // applying delegation changes in in helpers_test.go (applyTargets tests), so
-// this is just a sanity test to make sure Publish calls it correctly and
-// no fallback happens.
+// this is just a sanity test to make sure Publish calls it correctly
 func TestPublishDelegations(t *testing.T) {
 	testPublishDelegations(t, true, false)
 	testPublishDelegations(t, false, false)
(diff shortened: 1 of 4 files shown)


client/client_test.go
```diff
@@ -2227,7 +2195,7 @@ func testPublishTargetsDelgationScopeNoFallbackIfNoKeys(t *testing.T, clearCache
 // not its parents.  This tests the case where the local machine knows about
 // all the roles (in fact, the role creations will be applied before the
 // targets)
-func TestPublishTargetsDelgationSuccessLocallyHasRoles(t *testing.T) {
+func TestPublishTargetsDelegationSuccessLocallyHasRoles(t *testing.T) {
```
cyli: Yay thanks for fixing this typo!

client/client_test.go
```diff
@@ -2242,7 +2210,7 @@ func TestPublishTargetsDelgationSuccessLocallyHasRoles(t *testing.T) {
 	}
 
 	// just always check signing now, we've already established we can publish
-	// delgations with and without the metadata and key cache
+	// delegations with and without the metadata and key cache
```
cyli: Also this one - I'm surprised misspell didn't catch it

client/helpers_test.go
```diff
@@ -864,8 +864,8 @@ func TestApplyChangelistTargetsToMultipleRoles(t *testing.T) {
 	assert.False(t, ok, "no change to targets/level2, so metadata not created")
 }
 
-// ApplyTargets falls back to role that exists when adding or deleting a change
-func TestApplyChangelistTargetsFallbackRoles(t *testing.T) {
+// ApplyTargets does not fall back to role that exists when adding or deleting a change to a nonexistent delegation
+func TestApplyChangelistTargetsDoesNotFallbackRoles(t *testing.T) {
```
cyli: Non-blocking nit:  If we're not doing fallback, I wonder if we should just leave the concept of fallback out of the comments and function names entirely?  It definitely makes sense to for the target to fail if the role doesn't exist (which is what happens here), and mentioning fallback might be somewhat confusing since we're not doing it anymore?

client/helpers.go
```diff
@@ -158,21 +141,16 @@ func changeTargetMeta(repo *tuf.Repo, c changelist.Change) error {
 		}
 		files := data.Files{c.Path(): *meta}
 
-		err = doWithRoleFallback(c.Scope(), func(role string) error {
-			_, e := repo.AddTargets(role, files)
-			return e
-		})
-		if err != nil {
+		// Attempt to add the target to this role, with no fallback
```
cyli: Non-blocking nit:  same here - would it make sense to just leave out the idea of fallback?

cyli:
LGTM!  Thank you so much for fixing all the misspellings of "delegation" in the test case names and comments :D!

client/helpers_test.go
riyazdf: Agreed, I think I just changed the comments for myself/the team but it would make more sense when reading the code for the first time not to even include it :+1:

endophage:
Just to confirm my understanding: it will be up to docker (or a user of the notary CLI) to explicitly specify "targets" vs "targets/a_delegation" and if the top level "targets" role is provided, we'll happily add the entry to that role?

Assuming I'm understanding that correctly, LGTM.

riyazdf:
@endophage: yes -- we'll attempt to add to whichever role is specified, and it will only succeed if docker/notary has the appropriate private keys to sign

PR #617, opened 2016-03-09
Message canonical key ids on error

riyazdf:
I noticed while testing that we aren't being consistent with which format key IDs are messaged on `ErrNoKeys`.  This PR ensures we only message canonical IDs, which is already the case in [`sign.go`](https://github.com/docker/notary/blob/4aa9239aafa4e421134daab980f0c32d45d0ed85/tuf/signed/sign.go#L49)


cyli:
LGTM!  (I wonder if `ListCanonicalKeyIDs()` might be a useful utility function for `BaseRole` in the future, for when we display delegation information, but not relevant to this PR).

Thank you for noticing and fixing this inconsistency!

endophage:
LGTM

PR #620, opened 2016-03-12
json.RawMessage needs to be a pointer 

endophage:
As discovered during hackday, there are double encoding issues if a `json.RawMessage` object is included in a larger JSON object. The official "fix" is to use a `*json.RawMessage`

N.B. we don't use the Custom field anywhere (other than our hackday project which doesn't count) and it's set to `omitempty` so it doesn't appear in any of our json right now, this is an "if we use it in the future" fix

### DIFF
--- a/tuf/data/types.go
+++ b/tuf/data/types.go
@@ -126,9 +126,9 @@ var NotaryDefaultHashes = []string{notary.SHA256, notary.SHA512}
 // FileMeta contains the size and hashes for a metadata or target file. Custom
 // data can be optionally added.
 type FileMeta struct {
-	Length int64           `json:"length"`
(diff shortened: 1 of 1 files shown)


riyazdf:
LGTM pending CI.

Here's a related stack overflow thread documenting the behavior we encountered http://stackoverflow.com/questions/24229205/marshalling-json-rawmessage-returns-base64-encoded-string

cyli:
LGTM!

HuKeping:
It's a great fix ,although we haven't use this field yet.

endophage:
@HuKeping we used it for an internal hackday project and ran into the double encoding problem :-)

PR #621, opened 2016-03-14
remove signatures that are no longer valid during signing, either bec…

endophage:
…ause the key is no longer a valid signing key for the role, or the signature is invalid.


endophage:
submitting from a notary branch to try and fix coverage reporting

PR #623, opened 2016-03-14
remove signatures that are no longer valid during signing

endophage:
either because the key is no longer a valid signing key for the role, or the signature is invalid.

### DIFF
--- a/tuf/signed/sign.go
+++ b/tuf/signed/sign.go
@@ -22,10 +22,13 @@ import (
 
 // Sign takes a data.Signed and a key, calculated and adds the signature
 // to the data.Signed
+// N.B. All public keys for a role should be passed so that this function
+//      can correctly clean up signatures that are no longer valid.
 func Sign(service CryptoService, s *data.Signed, keys ...data.PublicKey) error {
 	logrus.Debugf("sign called with %d keys", len(keys))
 	signatures := make([]data.Signature, 0, len(s.Signatures)+1)
 	signingKeyIDs := make(map[string]struct{})
+	tufIDs := make(map[string]data.PublicKey)
 	ids := make([]string, 0, len(keys))
 
 	privKeys := make(map[string]data.PrivateKey)
@@ -34,6 +37,7 @@ func Sign(service CryptoService, s *data.Signed, keys ...data.PublicKey) error {
 	for _, key := range keys {
 		canonicalID, err := utils.CanonicalKeyID(key)
 		ids = append(ids, canonicalID)
+		tufIDs[key.ID()] = key
 		if err != nil {
 			continue
 		}
@@ -78,6 +82,20 @@ func Sign(service CryptoService, s *data.Signed, keys ...data.PublicKey) error {
 			// key is in the set of key IDs for which a signature has been created
 			continue
 		}
+		var (
+			k  data.PublicKey
+			ok bool
+		)
+		if k, ok = tufIDs[sig.KeyID]; !ok {
+			// key is no longer a valid signing key
+			continue
+		}
+		if err := VerifySignature(s.Signed, sig, k); err != nil {
+			// signature is no longer valid
+			continue
+		}
+		// keep any signatures that still represent valid keys and are
+		// themselves valid
 		signatures = append(signatures, sig)
 	}
 	s.Signatures = signatures
--- a/tuf/signed/sign_test.go
+++ b/tuf/signed/sign_test.go
@@ -21,11 +21,6 @@ type FailingCryptoService struct {
 	testKey data.PrivateKey
 }
 
-func (mts *FailingCryptoService) Sign(keyIDs []string, _ []byte) ([]data.Signature, error) {
-	sigs := make([]data.Signature, 0, len(keyIDs))
-	return sigs, nil
-}
-
 func (mts *FailingCryptoService) Create(_, _ string) (data.PublicKey, error) {
 	return mts.testKey, nil
 }
@@ -36,10 +31,10 @@ func (mts *FailingCryptoService) ListKeys(role string) []string {
 
 func (mts *FailingCryptoService) ListAllKeys() map[string]string {
 	return map[string]string{
-		mts.testKey.ID(): "root",
-		mts.testKey.ID(): "targets",
-		mts.testKey.ID(): "snapshot",
-		mts.testKey.ID(): "timestamp",
+		mts.testKey.ID(): data.CanonicalRootRole,
+		mts.testKey.ID(): data.CanonicalTargetsRole,
+		mts.testKey.ID(): data.CanonicalSnapshotRole,
+		mts.testKey.ID(): data.CanonicalTimestampRole,
 	}
 }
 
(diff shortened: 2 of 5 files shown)


tuf/signed/sign_test.go
```diff
@@ -264,9 +267,87 @@ func TestSignWithX509(t *testing.T) {
...
 
 	assert.Len(t, testData.Signatures, 1)
 	assert.Equal(t, tufRSAx509Key.ID(), testData.Signatures[0].KeyID)
 }
+
+func TestSignWithOldKey(t *testing.T) {
+	cs := NewEd25519()
+	testData := data.Signed{}
+
+	key1, err := cs.Create("root", data.ED25519Key)
```
riyazdf: nit: can we change this and the other `"root"` strings to `data.CanonicalRootRole`?

endophage: ah yep, no problem, copy/paste artifacts

tuf/tuf.go
```diff
@@ -917,11 +917,7 @@ func (tr *Repo) SignTimestamp(expires time.Time) (*data.Signed, error) {
 }
 
 func (tr Repo) sign(signedData *data.Signed, role data.BaseRole) (*data.Signed, error) {
-	ks := role.ListKeys()
-	if len(ks) < 1 {
-		return nil, signed.ErrNoKeys{}
```
riyazdf: :+1:  excited that we're getting rid of this duplicate code

tuf/signed/sign_test.go
```diff
@@ -264,9 +267,87 @@ func TestSignWithX509(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	baseRole := data.NewBaseRole(
+		data.CanonicalRootRole,
+		1,
+		key1,
+	)
+	Sign(cs, &testData, baseRole)
+
```
riyazdf: Can we add a check here to validate that the last sign created a signature with key1?

tuf/signed/sign_test.go
```diff
@@ -264,9 +267,87 @@ func TestSignWithX509(t *testing.T) {
...
```
riyazdf: Same here, it'd be great if we can add some checks that this `Sign` succeeded

tuf/signed/sign_test.go
```diff
@@ -264,9 +267,87 @@ func TestSignWithX509(t *testing.T) {
...
+	key1, err := cs.Create(data.CanonicalRootRole, data.ED25519Key)
+	assert.NoError(t, err)
+
+	baseRole := data.NewBaseRole(
+		data.CanonicalRootRole,
+		1,
+		key1,
+	)
+	Sign(cs, &testData, baseRole)
+
```
riyazdf: Same here, it'd be great if we can add some checks that this `Sign` succeeded.

tuf/signed/sign_test.go
```diff
@@ -264,9 +267,87 @@ func TestSignWithX509(t *testing.T) {
...
+	// should remove key1 sig because it's out of date
+	Sign(
+		cs,
+		&testData,
+		baseRole,
+	)
+
+	assert.Len(t, testData.Signatures, 1)
+	assert.Equal(t, key2.ID(), testData.Signatures[0].KeyID)
+}
```
riyazdf: It'd be really cool if we could add tests for testing this and `TestSignWithOldKey ` against delegation roles, especially with > 2 keys

endophage: The logic within `Sign` is role independent. The only thing it looks at is `BaseRole.Keys`.

riyazdf: Yes agreed that the role doesn't matter, I was just thinking of a use case since with delegations we may have many keys for the role but only sign with one at a time.

This is more of a nit since your test cases show the functionality, I figure it'd be nice to have for illustration purposes.

endophage: `Sign` is back to just taking `data.PublicKey...` The logic for `Sign` has always been meant to be all keys for a role are always passed, and we sign with whatever keys we have. I've seen that we haven't always obeyed that logic (I had to update one of the key rotation tests), and I don't see it documented anywhere obvious. I'll update the comment on `Sign`. This tests checks that it correctly removes a signature of a key that isn't passed and other tests in this file check it correctly preserves a (still valid) signature for a key that was passed but wasn't found in the private key store.

tuf/signed/sign_test.go
```diff
@@ -270,3 +248,59 @@ func TestSignWithX509(t *testing.T) {
 	assert.Len(t, testData.Signatures, 1)
 	assert.Equal(t, tufRSAx509Key.ID(), testData.Signatures[0].KeyID)
 }
+
+func TestSignWithOldKey(t *testing.T) {
```
cyli: Nit:  the test name implies to me that this is testing that we can sign with old keys - could we rename this to `TestSignRemovesValidSignaturesByOldKeys`?

tuf/signed/sign_test.go
```diff
@@ -270,3 +248,59 @@ func TestSignWithX509(t *testing.T) {
...
+		cs,
+		&testData,
+		key2,
+	)
+
+	assert.Len(t, testData.Signatures, 1)
+	assert.Equal(t, key2.ID(), testData.Signatures[0].KeyID)
+}
+
+func TestSignWithOldSig(t *testing.T) {
```
cyli: Nitpick:  could we rename this to something like `TestSignRemovesInvalidSignatures`?

tuf/signed/verify.go
```diff
@@ -145,3 +138,18 @@ func VerifySignatures(s *data.Signed, roleData data.BaseRole) error {
...
+func VerifySignature(msg []byte, sig data.Signature, pk data.PublicKey) error {
+	// method lookup is consistent due to Unmarshal JSON doing lower case for us.
+	method := sig.Method
+	verifier, ok := Verifiers[method]
+	if !ok {
+		return fmt.Errorf("continuing b/c signing method is not supported: %s\n", sig.Method)
+	}
+
+	if err := verifier.Verify(pk, sig.Signature, msg); err != nil {
+		return fmt.Errorf("continuing b/c signature was invalid\n")
```
cyli: Non-blocking nitpick:  Maybe this could just be "signature was invalid"?  `VerifySignature` doesn't know it's being called in a loop or that the caller will ignore the error.

tuf/signed/sign_test.go
```diff
@@ -36,10 +31,10 @@ func (mts *FailingCryptoService) ListKeys(role string) []string {
 
 func (mts *FailingCryptoService) ListAllKeys() map[string]string {
 	return map[string]string{
-		mts.testKey.ID(): "root",
-		mts.testKey.ID(): "targets",
-		mts.testKey.ID(): "snapshot",
-		mts.testKey.ID(): "timestamp",
+		mts.testKey.ID(): data.CanonicalRootRole,
```
cyli: Yay thanks for all these cleanups!

tuf/signed/sign.go
```diff
@@ -78,6 +82,20 @@ func Sign(service CryptoService, s *data.Signed, keys ...data.PublicKey) error {
 			// key is in the set of key IDs for which a signature has been created
```
cyli: Non-blocking - apologies I couldn't add this comment 2 lines up - would it make sense to add a comment to the effect that we want to both clean old signatures and also validate the signatures we just produced before returning the signed object?

cyli: bah, IGNORE ME

riyazdf:
LGTM, considering we aren't supporting thresholds > 1

cyli:
LGTM - looks like this removes the invalid sigs ok.

PR #624, opened 2016-03-14
Trust pinning configuration

### LINKED ISSUE #559: Config for CAs + Certs
Dependent on #558

See: https://docs.google.com/document/d/1Skt3d0twJjtNioyCaRN8ZF-vZZDT50Eg-F4QKdRKFcM

riyazdf:
Living design document [here](https://docs.google.com/document/d/1Skt3d0twJjtNioyCaRN8ZF-vZZDT50Eg-F4QKdRKFcM/edit#heading=h.vxzt849fq0um)

Adds a `trust_pinning` section to the client config as outlined in the document.  There are three keys: `ca`, `certs`, and `disable_tofu`.  Only one of these keys is to be used when deciding on a trust pin behavior, with `certs` having the highest priority and `tofu` having the lowest (this is detailed in the doc). Using TOFUs via `disable_tofu: false` is the default if no config is specified (this is current behavior).

Closes #559


certs/certs.go
```diff
@@ -100,9 +112,86 @@ func ValidateRoot(certStore trustmanager.X509Store, root *data.Signed, gun strin
 		}
 	} else {
 		logrus.Debugf("found no currently valid root certificates for %s", gun)
+		logrus.Debugf("using trust_pinning config to bootstrap trust: %v", trustPinning)
```
NathanMcCauley: This method is too big. I think it might make sense to create an object that can validate a cert/GUN combo against a `TrustPinConfig`.

Something like:

```
trustPinChecker, err := notary.TrustPinChecker(trustPinConfig, GUN)
if err != nil {
 // blah
}
for each cert {
  trustPinChecker.checkCert(cert)
}
```

This object can encapsulate the logic about whether a given cert/GUN combo satisfies a PinningConfig. WDYT?

riyazdf: Agreed, `ValidateRoot` is giant and this control flow can be factored out.  I'll work on refactoring this.

docs/reference/client-config.md
```diff
@@ -113,6 +118,48 @@ Remote server example:
...
+The `trust_pinning` specifies how to bootstrap trust for the root of a
+Notary client's trusted collection.
+
+This section is optional, Notary will use TOFU over HTTPS by default and
+trust certificates in the downloaded root file.
+
+In this section, one can provide specific certificates to pin to, or a CA
+to pin to as a root of trust for a GUN.
+
+Only one trust pinning option can be used per GUN, and any validation failure
```
cyli: This seems to imply that if more than one option is given, the parsing the options will fail, but the code seems to prioritize one over the other.  If so, would it make sense to specify the order of preference of options here?  The design doc is actually pretty clear and that state diagram seems pretty useful - would it make sense to include it in the docs somewhere?

riyazdf: Definitely agree, I'll include the state diagram and revisit the wording here.

certs/trustpin.go
```diff
@@ -0,0 +1,89 @@
...
+		if err != nil {
+			return false
+		}
+		return t.config.Certs[t.gun] == certID
+	case ca:
+		for caGunPrefix, caFilepath := range t.config.CA {
+			if strings.HasPrefix(t.gun, caGunPrefix) {
+				// Try to add the CA cert to our certificate store,
+				// and use it to validate certs in the root.json later
+				caCert, err := trustmanager.LoadCertFromFile(caFilepath)
```
cyli: Wondering should loading certs from a file and validating happen when the trust config is created, or if the root pool from this CA should be cached so that it doesn't have to be done for each leaf cert in the root.json?

certs/certs.go
```diff
@@ -99,10 +109,33 @@ func ValidateRoot(certStore trustmanager.X509Store, root *data.Signed, gun strin
...
-		logrus.Debugf("found no currently valid root certificates for %s", gun)
+		logrus.Debugf("found no currently valid root certificates for %s, using trust_pinning config to bootstrap trust:", gun, trustPinning)
+		trustPinChecker, err := NewTrustPinChecker(trustPinning, gun)
+		if err != nil {
+			return &ErrValidationFail{Reason: err.Error()}
+		}
+
+		validPinnedCerts := []*x509.Certificate{}
+		for _, cert := range allValidCerts {
+			_, intermediateCerts := parseAllCerts(signedRoot)
```
cyli: `intermediateCerts` seems like it remains the same each time, since `parseAllCerts` takes just the `signedRoot`.  Can this just be called once, rather than for every cert in `allValidCerts`?  Unless verifying a cert given the roots and intermediates mutates those cert pools.

Also I think `allValidCerts` calls `parseAllCerts` as well, and ignores the intermediate certs - since we just seem to call it here, I'm wondering if we can refactor `validatRootLeafCerts` to take a bunch of leaf certs rather than the `data.SignedRoot`?  That way all that stuff can just be calculated in this if/else block?

certs/trustpin.go
```diff
@@ -0,0 +1,89 @@
...
+	tofus TrustPinMode = "tofu"
+	ca    TrustPinMode = "ca"
+	certs TrustPinMode = "certs"
+)
+
+// NewTrustPinChecker returns a new TrustPinChecker from a TrustPinConfig for a GUN
+func NewTrustPinChecker(trustPinConfig notary.TrustPinConfig, gun string) (TrustPinChecker, error) {
+	trustPinChecker := TrustPinChecker{config: trustPinConfig, gun: gun}
+	// Determine the mode, and if it's even valid
+	if trustPinConfig.TOFU {
```
cyli: If I'm interpreting the control flow right for trust pinning, should the check for TOFU come last?  If there is a pinned cert for this GUN, shouldn't it take precedence over TOFU?  (similarly for CA)

certs/trustpin.go
riyazdf: good idea, I can move that to the trust config initialization

certs/trustpin.go
riyazdf: good catch, I'll change the order!  Good idea on the tests too, I'll add some precedence tests :)

certs/certs.go
riyazdf: Yep definitely, I'll refactor `validateRootLeafCerts` to not need to call `parseAllCerts` - that's a great idea

cyli:
Non-blocking - Do you think we should rename the `certs` package the `trustpinning` package?  Since it's just used to validate root signatures against pinned trust, and doesn't have more general certificate functionality?

certs/trustpin.go
```diff
@@ -0,0 +1,90 @@
...
+// NewTrustPinChecker returns a new TrustPinChecker from a TrustPinConfig for a GUN
+func NewTrustPinChecker(trustPinConfig notary.TrustPinConfig, gun string) (TrustPinChecker, error) {
+	trustPinChecker := TrustPinChecker{config: trustPinConfig, gun: gun}
+	// Determine the mode, and if it's even valid
+	if _, ok := trustPinConfig.Certs[gun]; ok {
+		trustPinChecker.mode = certs
+	} else if utils.ContainsKeyPrefix(trustPinConfig.CA, gun) {
+		trustPinChecker.mode = ca
+		for caGunPrefix, caFilepath := range trustPinConfig.CA {
+			if strings.HasPrefix(gun, caGunPrefix) {
```
cyli: What happens if there is more than one CA prefix that matches the gun?  Should we pick the most specific one?  Otherwise, since it's looping over a map, it'd be random.

That might be fine, so long as we document what the user can expect the behavior to be.

certs/certs.go
```diff
@@ -243,7 +274,7 @@ func parseAllCerts(signedRoot *data.SignedRoot) (map[string]*x509.Certificate, m
 	return leafCerts, intCerts
 }
 
-// certsToRemove returns all the certifificates from oldCerts that aren't present
+// certsToRemove returns all the certificates from oldCerts that aren't present
```
cyli: Thanks for fixing these misspellings!

certs/certs.go
```diff
@@ -99,10 +110,32 @@ func ValidateRoot(certStore trustmanager.X509Store, root *data.Signed, gun strin
...
+			validPinnedCerts = append(validPinnedCerts, cert)
+		}
+		if len(validPinnedCerts) == 0 {
+			return &ErrValidationFail{Reason: "unable to match any certificates to trust_pinning config"}
+		}
+		allValidCerts = validPinnedCerts
 	}
 
 	// Validate the integrity of the new root (does it have valid signatures)
+	// Note that allValidCerts is guaranteed to be unchanged only if we had prior cert data for this GUN or enabled TOFUS
```
cyli: :+1: Thank you for calling attention to this detail!

certs/trustpin.go
riyazdf: Agreed that picking the most specific one makes sense here -- it'd also be nice to have a "global ca" under the empty prefix `""

riyazdf:
@cyli: I agree on the rename, since it's slightly confusing with the functionality in `cmd/notary/cert.go` -- done!

cmd/notary/tuf.go
```diff
@@ -546,3 +547,19 @@ func getRemoteTrustServer(config *viper.Viper) string {
...
 	return defaultServerURL
 }
+
+func getTrustPinning(config *viper.Viper) notary.TrustPinConfig {
+	// Need to parse out Certs section from config
+	certMap := config.GetStringMap("trust_pinning.certs")
+	resultCertMap := make(map[string][]string)
+	for gun, certSlice := range certMap {
+		if castedCertSlice, ok := certSlice.([]string); ok {
+			resultCertMap[gun] = castedCertSlice
```
cyli: Non-blocking - wondering if we should error if it's not a slice, so that users will figure out why things aren't getting pinned correctly sooner?

cmd/notary/main_test.go
```diff
@@ -369,3 +370,152 @@ func TestConfigFileOverridenByCmdLineFlags(t *testing.T) {
...
+	commander := &notaryCommander{
+		getRetriever: func() passphrase.Retriever { return passphrase.ConstantRetriever("pass") },
+	}
+
+	tempDir := tempDirWithConfig(t, fmt.Sprintf(`{
+		"remote_server": {
+			"url": "%s"
+		},
+		"trust_pinning": {
+		    "tofu": true
```
cyli: Should this option be `disable_tofu`?

cmd/notary/main_test.go
```diff
@@ -369,3 +370,152 @@ func TestConfigFileOverridenByCmdLineFlags(t *testing.T) {
...
+	// validate that we actually managed to connect and attempted to download the root though
+	require.Len(t, m.gotten, 1)
+	require.Equal(t, m.gotten[0], "repo.root")
+
+	tempDir2 := tempDirWithConfig(t, fmt.Sprintf(`{
+		"remote_server": {
+			"url": "%s"
+		},
+		"trust_pinning": {
+		    "tofu": false
```
cyli: Similarly here, should this option be `disable_tofu`?

cmd/notary/main_test.go
riyazdf: ah yep, great catch!  This is still passing because `disable_tofu` defaults to false and we still have TOFUs :)

cmd/notary/main_test.go
```diff
@@ -369,3 +370,152 @@ func TestConfigFileOverridenByCmdLineFlags(t *testing.T) {
...
+		"--tlscert", filepath.Clean(filepath.Join(cwd, "../../fixtures/notary-server.crt")),
+		"--tlskey", "../../fixtures/notary-server.key"})
+	cmd.SetOutput(new(bytes.Buffer)) // eat the output
+	err = cmd.Execute()
+	require.Error(t, err, "there was no repository, so list should have failed")
+	require.NotContains(t, err.Error(), "TLS", "there was no TLS error though!")
+
+	config, err = commander.parseConfig()
+	require.NoError(t, err)
+	require.Equal(t, false, config.GetBool("trust_pinning.disable_tofu"))
```
cyli: I'm wondering if it would make sense to pass the config to `getTrustPinning` here rather than to test the viper config itself?  And assert true/false on the `TrustPinConfig` object that comes out?

const.go
```diff
@@ -59,3 +59,11 @@ var NotaryDefaultExpiries = map[string]time.Duration{
 	"snapshot":  NotarySnapshotExpiry,
 	"timestamp": NotaryTimestampExpiry,
 }
+
+// TrustPinConfig represents the configuration under the trust_pinning section of the config file
+// This struct represents the preferred way to bootstrap trust for this repository
+type TrustPinConfig struct {
```
endophage: Probably the wrong place for this as it's a type, not a constant value.

cmd/notary/main_test.go
cyli: Also totally non-blocking - I'm fine with testing TLS stuff here also, but if we wanted to isolate this test to just trust pinning, we can probably just set the `configFile` filed in `notaryCommander`:

```
	tempDir := tempDirWithConfig(t, `{
		"trust_pinning": {
		    "tofu": true
		 }
	}`)
	defer os.RemoveAll(tempDir)
	commander := &notaryCommander{
		getRetriever: func() passphrase.Retriever { return passphrase.ConstantRetriever("pass") },
		configFile:   filepath.Join(tempDir, "config.json"),
	}

	// Check that tofu was set correctly
	config, err := commander.parseConfig()
	require.NoError(t, err)
	require.Equal(t, false, config.GetBool("trust_pinning.disable_tofu"))
```

trustpinning/trustpin.go
```diff
@@ -0,0 +1,97 @@
...
+		return caCheck(caRootPool), nil
+	}
+
+	if !trustPinConfig.DisableTOFU {
+		return tofusCheck, nil
+	}
+	return nil, fmt.Errorf("invalid trust pinning specified")
+}
+
+var certsCheck = func(pinnedCertIDs []string) CertChecker {
```
endophage: Ah, I see what you were doing now. I was actually thinking you'd keep the struct you had before and just return a bound method:
```
type trustPinChecker struct {
    gun           string		
    config       notary.TrustPinConfig		
    pinnedCAPool *x509.CertPool
}

func NewChecker() CertChecker {
    c := trustPinChecker{}
    if pinnedCerts, ok := trustPinConfig.Certs[gun]; ok {
        return c.certsCheck, nil
    }
    ...
}
```

cmd/notary/main_test.go
riyazdf: totally agree for both, will update!

cmd/notary/tuf.go
riyazdf: I'll refactor to have this function also return an error so we can handle this more gracefully 👍

const.go
riyazdf: agreed, will move to `trustpinning`

trustpinning/certs_test.go
```diff
@@ -205,11 +209,218 @@ func TestValidateRoot(t *testing.T) {
...
+
+	// This call to ValidateRoot should fail due to an invalid GUN (even though the cert ID is correct), and TOFUS defaults to false
+	err = ValidateRoot(certStore, &testSignedRoot, "docker.com/notary", notary.TrustPinConfig{Certs: map[string][]string{"not_a_gun": {rootPubKeyID}}, DisableTOFU: true})
+	require.Error(t, err)
+
+	// This call to ValidateRoot should fail due to an invalid cert ID, even though it's a valid key ID for targets
+	err = ValidateRoot(certStore, &testSignedRoot, "docker.com/notary", notary.TrustPinConfig{Certs: map[string][]string{"docker.com/notary": {targetsPubKeyID}}})
+	require.Error(t, err)
+
+	// This call to ValidateRoot should succeed because we fall through to TOFUS even though we have no matching GUNs under Certs
```
cyli: nit: should this be "because we have no matching GUNs under Certs"?  If we did, it wouldn't fall through, as the awesome tests above demonstrate.

trustpinning/certs_test.go
```diff
@@ -205,11 +209,218 @@ func TestValidateRoot(t *testing.T) {
...
+	invalidCAFilepath := filepath.Join(tempBaseDir, "invalid.ca")
+	require.NoError(t, ioutil.WriteFile(invalidCAFilepath, []byte("ABSOLUTELY NOT A PEM"), 0644))
+
+	// Using this invalid CA cert should fail on ValidateRoot
+	err = ValidateRoot(certStore, &testSignedRoot, "docker.com/notary", notary.TrustPinConfig{CA: map[string]string{"docker.com/notary": invalidCAFilepath}, DisableTOFU: true})
+	require.Error(t, err)
+
+	validCAFilepath := "../fixtures/root-ca.crt"
+
+	// If we pass an invalid Certs entry in addition to this valid CA entry, since Certs has priority for pinning we will fail
```
cyli: 👍  Awesome test!

cyli:
I was wondering if we could add two more test cases for CA parsing?  
- one certificate in the bundle is valid and one is expired (success)
- both certificates are expired (error)

trustpinning/certs_test.go
riyazdf: agreed, will change!

trustpinning/trustpin.go
riyazdf: that's probably easier to debug than the closure, though I like the idea of only returning the function to ensure that we don't mutate the trust pinning state by accident -- I'll give this a shot and see how it looks

riyazdf:
@cyli: great test ideas, added them to `TestValidateRootWithPinnedCA`

cmd/notary/main_test.go
```diff
@@ -369,3 +370,89 @@ func TestConfigFileOverridenByCmdLineFlags(t *testing.T) {
...
+	trustPin, err := getTrustPinning(config)
+	require.NoError(t, err)
+	require.Equal(t, false, trustPin.DisableTOFU)
+
+	tempDir = tempDirWithConfig(t, `{
+		"remote_server": {
+			"url": "%s"
+		},
+        "trust_pinning": {
```
cyli: non-blocking nitpick:  I think this block got unindented :)  But it's just a string, so doesn't really matter

trustpinning/certs.go
```diff
@@ -164,9 +196,9 @@ func ValidateRoot(certStore trustmanager.X509Store, root *data.Signed, gun strin
 // validRootLeafCerts returns a list of non-expired, non-sha1 certificates
 // found in root whose Common-Names match the provided GUN. Note that this
 // "validity" alone does not imply any measure of trust.
-func validRootLeafCerts(root *data.SignedRoot, gun string) ([]*x509.Certificate, error) {
-	// Get a list of all of the leaf certificates present in root
-	allLeafCerts, _ := parseAllCerts(root)
+// validRootLeafCerts returns a list of non-expired, non-sha1 certificates whose
```
cyli: nit:  This comment may have gotten duplicated during a rebase

riyazdf: Definitely did, thanks for the catch :)

trustpinning/certs_test.go
```diff
@@ -203,11 +208,269 @@ func TestValidateRoot(t *testing.T) {
...
+	require.NoError(t, err)
+	allExpiredCertBundle, err := trustmanager.CertChainToPEM([]*x509.Certificate{expiredCert, expiredCert2})
+	require.NoError(t, err)
+	allExpiredCertPath := filepath.Join(tempBaseDir, "all_expired_cert.pem")
+	require.NoError(t, ioutil.WriteFile(allExpiredCertPath, allExpiredCertBundle, 0644))
+	// Now only use expired certs in the bundle, we should fail
+	err = ValidateRoot(certStore, newTestSignedRoot, "notary-signer", TrustPinConfig{CA: map[string]string{"notary-signer": allExpiredCertPath}, DisableTOFU: true})
+	require.Error(t, err)
+
+	// Add an CA for a different GUN that won't validate
```
cyli: Non-blocking - the GUN seems to be the same as the GUN for `newTestSignedRoot` - it's not using the `goodRootCABundle` - should this comment be just the wrong CA for the key?  (As opposed to one with a different GUN?)

riyazdf: agreed!  I'll fix that right now.

cyli:
Thanks for the quick responses!  :D  LGTM

trustpinning/trustpin.go
endophage: It still only returned the function. It just happens to be the function bound to a specific instance of the struct (which serves like a closure)

cmd/notary/tuf.go
```diff
@@ -546,3 +587,30 @@ func getRemoteTrustServer(config *viper.Viper) string {
...
+
+func getTrustPinning(config *viper.Viper) (trustpinning.TrustPinConfig, error) {
+	var ok bool
+	// Need to parse out Certs section from config
+	certMap := config.GetStringMap("trust_pinning.certs")
+	resultCertMap := make(map[string][]string)
+	for gun, certSlice := range certMap {
+		var castedCertSlice []interface{}
+		if castedCertSlice, ok = certSlice.([]interface{}); !ok {
+			return trustpinning.TrustPinConfig{}, fmt.Errorf("invalid format for trust_pinning.certs")
```
NathanMcCauley: Why return an empty `trustpinning.TrustPinConfg{}` instead of nil?

cmd/notary/tuf.go
```diff
@@ -546,3 +587,30 @@ func getRemoteTrustServer(config *viper.Viper) string {
...
+		}
+		certsForGun := make([]string, len(castedCertSlice))
+		for idx, certIDInterface := range castedCertSlice {
+			if certID, ok := certIDInterface.(string); ok {
+				certsForGun[idx] = certID
+			} else {
+				return trustpinning.TrustPinConfig{}, fmt.Errorf("invalid format for trust_pinning.certs")
```
NathanMcCauley: same as above

cmd/notary/tuf.go
```diff
@@ -546,3 +587,30 @@ func getRemoteTrustServer(config *viper.Viper) string {
 	}
 	return defaultServerURL
 }
+
+func getTrustPinning(config *viper.Viper) (trustpinning.TrustPinConfig, error) {
```
NathanMcCauley: I don't see any tests specifically for this method -- there's a decent amount of logic here. Would you mind adding some tests that we properly parse these files and generate the expected `TrustPinConfig`s?

cryptoservice/certificate.go
```diff
@@ -28,6 +28,13 @@ func GenerateTestingCertificate(signer crypto.Signer, gun string) (*x509.Certifi
 	return generateCertificate(signer, gun, startTime, startTime.AddDate(10, 0, 0))
 }
 
+// GenerateExpiredTestingCertificate generates an expired X509 Certificate from a template, given a GUN.
+// Good enough for tests where expiration does not really matter; do not use if you care about the policy.
+func GenerateExpiredTestingCertificate(signer crypto.Signer, gun string) (*x509.Certificate, error) {
```
NathanMcCauley: Maybe it'd make sense to move this method in to one of the test files instead of here? Looks like it's only used in `trustpinning/certs_test.go`.

cmd/notary/tuf.go
riyazdf: We can't return nil with this function return signature since we return the struct itself instead of a pointer -- I can change this function to return `*trustpinning.TrustPinConfig` and a nil pointer for error cases if we prefer?

trustpinning/trustpin.go
riyazdf: Ah, yeah that makes sense -- that sounds good, I'll change that

cmd/notary/tuf.go
cyli: Just out of curiosity, is there a best practice in go about returning pointers vs objects? 

I went to this neat [go training meetup](https://github.com/ardanlabs/gotraining/) where the instructor mentioned that unless data needs to be modified that it's better to pass by value.  I think his reasoning was just optimizing for having the value be on the stack rather than the heap.  

It seems like premature optimization for us to really care about it that much, but I was just wondering if there's a counter-argument for passing pointers for just data other than being able to determine whether the user specified a zero value?

In this case we don't need to be able to determine the difference between default trust pinning (the zero value) and whether the user specified any trust pinning (for instance, if it matters whether we can tell the difference between an empty string or just an unspecified string).

cmd/notary/tuf.go
riyazdf: I have the `TestConfigFileTrustPinning` test in `main_test.go`, which tests against this method and parsing the config file itself -- I put it there because we have other configuration parsing tests in that file.  If there are any particular test cases you had in mind I can definitely add them

cryptoservice/certificate.go
riyazdf: great idea, I'll move this to a test file

cmd/notary/tuf.go
NathanMcCauley: Hmm, interesting, I'm not sure what's the better style. Haven't seen any style guides that seem to have an opinion. @endophage any opinion?

trustpinning/trustpin.go
riyazdf: done!  Let me know if you had something else in mind, but I think it definitely looks cleaner :)

trustpinning/certs_test.go
```diff
@@ -203,11 +208,274 @@ func TestValidateRoot(t *testing.T) {
...
+	validCert, err := cryptoservice.GenerateCertificate(testPrivKey3, "notary-signer", time.Now(), time.Now().AddDate(1, 0, 0))
+	require.NoError(t, err)
+	bundleWithWrongCert, err := trustmanager.CertChainToPEM([]*x509.Certificate{validCert})
+	require.NoError(t, err)
+	bundleWithWrongCertPath := filepath.Join(tempBaseDir, "bundle_with_expired_cert.pem")
+	require.NoError(t, ioutil.WriteFile(bundleWithWrongCertPath, bundleWithWrongCert, 0644))
+	err = ValidateRoot(certStore, newTestSignedRoot, "notary-signer", TrustPinConfig{CA: map[string]string{"notary-signer": bundleWithWrongCertPath}, DisableTOFU: true})
+	require.Error(t, err)
+}
+
```
riyazdf: Here's a list of the new `certs_test.go` test cases (more or less in order) to make it easier to understand the coverage:

#### TOFUS
- Failure when tofus is disabled, and no other trust pinning is enabled
- Success when (inherited from other tests) tofus is enabled by default, and will succeed in bootstrapping trust when no other trust pinning is specified

#### Certs
- Success when the only provided valid cert ID is pinned for the specified GUN
- Success when only one of the provided cert IDs is valid for the specified GUN
- Success when none of the entries in `certs` match the GUN, and we fall through to TOFUS
- Failure when an invalid cert ID is pinned against the specified GUN
- Failure when an empty cert ID is pinned against the specified GUN
- Failure when the correct cert ID is pinned against the incorrect GUN (tofus explicitly disabled)
- Failure when a different key ID (targets key) is pinned against the specified GUN

#### CA
- Success when a valid CA bundle with a single valid root is pinned to the specified GUN
- Success when a valid CA bundle with one expired and one valid root is pinned to the specified GUN
- Failure when a CA bundle of only expired content is pinned to the specified GUN
- Failure when the path for the CA pinned against the specified GUN is invalid
- Failure when there are no pinned CAs for the GUN (tofus explicitly disabled)
- Failure when invalid PEM is pinned as CA for specified GUN
- Failure when a pinned CA for the GUN does not validate against the provided signed root

trustpinning/trustpin.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	if caFilepath, err := getPinnedCAFilepathByPrefix(gun, trustPinConfig); err == nil {
+		// Try to add the CA certs from its bundle file to our certificate store,
+		// and use it to validate certs in the root.json later
+		caCerts, err := trustmanager.LoadCertBundleFromFile(caFilepath)
+		if err != nil {
+			return nil, fmt.Errorf("could not load root cert from CA path")
+		}
+		// Now only consider certificates that are direct children from this CA cert, overwriting allValidCerts
```
cyli: non-blocking nit:  We can probably delete the `overwriting allValidCerts` part of this comment

cmd/notary/tuf.go
endophage: I haven't heard a good argument either way. Pointers are more memory efficient but copies (non-pointers) are easier to reason about. As with json.RawMessage there are cases where it really matters due to Golang's addressability concepts. The go docs seem to be unopinionated.

IMO, in this context, the parsing function would be fine returning a pointer, but the caller should go on to pass the config by value to any other receivers (so the receivers can't modify the config in a way other functions will be affected by).

endophage:
Short of writing some really inane tests which would basically all hit the same case of the trust pin config being bad in various cmd/notary/*.go files, I don't think there's much we can do about the -0.01% change. This specific case certainly isn't where we should be burning time worrying about test coverage.

This is good work, the code is understandable and the tests look pretty comprehensive.

LGTM!

endophage:
Scratch that. I think there's a bug when pinning certs IDs and having a cert chain in a root public key field. Working on a test and fix.

riyazdf:
I've pushed the latest commit to work with the cert bundling test for pinning cert IDs, but I don't think we've thought too much about how to handle cert bundles in root -- it seems like many of our helper functions tend to only consider the first certificate in the chain.

Also, it isn't immensely clear how to type a certificate bundle as an object that can produce an ID if the leaf and intermediate certificates use differing public key algorithms, though it seems to make most sense to just use the leaf to decide?  I'll dig more into this tomorrow as well.

riyazdf:
team LGTM, merging now.

dnwake:
@riyazdf  Is there any way to specify the root public key in the trust_pinning section of notary.json?  That would be a lot simpler than having to specify a separate targets public key for each GUN.

endophage:
@dnwake it is the root cert ID that should be specified. Do we have a mistake in the docs somewhere that says it's the targets?

dnwake:
@endophage My apologies.  I see now that the docs do say "trust for the root of a Notary client's trusted collection.".  It would be good to explicitly say this in the detailed docs of the "certs" parameter -- I somehow managed to completely miss it :)

endophage:
Do you have the link to the docs that need to be made more explicit? Happy to fix them :-)

riyazdf:
I think @dnwake is referring to https://docs.docker.com/notary/reference/client-config/#trust-pinning-section-optional, definitely open to specifying that the cert IDs refer to root cert IDs

dnwake:
@riyazdf Yes, that is it -- sorry!

Is there any way to combine trust pinning with trust delegation?  The use case is Docker, where the pushing and pulling machines are trusted, but the Docker registry and Notary server are not.

* multiple trusted machines push to multpile GUNs
* the pushing machines all share the delegation keypair
* pushing machines use trust delegation so that they can all push updates equally to each GUN
* pulling machines want to ensure that only updates from the trusted machines are pulled.
* new GUNs may be pushed at any time

If there is some way to pin trust to the delegation key rather than the root keys (which are different for each GUN)?

Thanks!

riyazdf:
@dnwake: the trust pinning functionality in this PR is only concerned with bootstrapping trust in root metadata, so the delegation pinning configuration you describe is not in master at this time.

That said -- in Docker 1.11+, we do treat the `targets/releases` role in a special manner similar to what you've described, in which trusted pulls will pull only from the `targets/releases` role if it exists for a repo (behavior described [here](https://docs.docker.com/notary/advanced_usage/#use-delegations-with-content-trust))

dnwake:
@riyazdf Thanks!  I'll check 1.11 out, but we're probably stuck on Docker 1.10 for the time being.

I hope you don't mind if I ask another question:  the docs speak of "the root CA file with which to verify the certificates in the root file".  All the Docker-generated certs I've seen have been self-signed:  is there any way to make Docker generate certificates signed by a user-specified CA file when it pushes?

endophage:
I think we have issue(s) for it but I can't find them right now. The key import command will already let you import appropriately created certs. However we will be adding functionality (not currently prioritized), to allow you to specify a local CA, i.e. a local keypair, or possibly a remote CA (there's definitely an issue for CFSSL integration).

dnwake:
@endophage @riyazdf So, given that all the certs are created automatically by Docker / Notary and are self-signed, the "ca"  option in the trust_pinning section of config.json would have to point to a file that actually contained all of the certificates themselves?  Or am I missing something?  Thanks.

dnwake:
@endophage @riyazdf Thanks! Without the functionality you mention (giving clients the ability to specify a local CA as issuer for the certs generated for use in a first-time push), I'm having difficulty seeing how we can make meaningful use of the "ca" section in the trust-pinning config.  Am I missing something?

Cheers.

riyazdf:
@dnwake yes you're right.  It's on our roadmap -- we open issues to import root keys instead of auto-generating https://github.com/docker/notary/issues/731 as well as hooking into CFSSL to generate other certs https://github.com/docker/notary/issues/706

dnwake:
@riyazdf Thanks.  Do you have any idea of when https://github.com/docker/notary/issues/731 will be prioritized?  It may be critical for our usage of Notary.  (I see no one is currently assigned).

riyazdf:
@dnwake: definitely a priority, hard to give an exact "land" date but it's marked for 0.4 (and we have an open PR for part of the functionality!).  If you have feedback for how to best surface this functionality, please let us know on #731 and any PRs/issues linked to it

PR #626, opened 2016-03-15
Use 'require' instead of 'assert' in client and TUF client tests

cyli:


riyazdf:
Thanks for standardizing this!!  LGTM

endophage:
LGTM

PR #634, opened 2016-03-17
make cross and make integration endpoints

cyli:
This PR does the following:

1. Fix the `make cross` endpoint to be able to build linux and mac os notary client binaries (in a container)
1. Add a `make integration` endpoint that spins up a full notary service, along with a container that builds the client and runs some basic commands against the service.

I am trying to run the second part in CircleCI as part of our tests.


cyli:
Note once this is merged we have to bump the level of parallelism for notary on CircleCi to 3, but this might break existing PRs that do not have this circle file.

Makefile
```diff
@@ -13,8 +13,7 @@ endif
 CTIMEVAR=-X $(NOTARY_PKG)/version.GitCommit=$(GITCOMMIT) -X $(NOTARY_PKG)/version.NotaryVersion=$(NOTARY_VERSION)
 GO_LDFLAGS=-ldflags "-w $(CTIMEVAR)"
 GO_LDFLAGS_STATIC=-ldflags "-w $(CTIMEVAR) -extldflags -static"
-GOOSES = darwin freebsd linux
-GOARCHS = amd64
+GOOSES = darwin linux
```
riyazdf: can't help but comment on this:

cyli: :D

riyazdf:
LGTM, excited to add this to circle!

cyli:
@riyazdf Thanks so much for testing it!  :)

cyli:
@endophage If you have time to take a look at this, I'd love to have circleci start running the sanity test :D

endophage:
integration appears to be failing for network related reasons for me. I see signer come up successfullly but server fails to connect to it. Also, can we move `.build` to not be a hidden folder. Seems unnecessary to hide code.

endophage:
It appears something related to linking is broken in the latest version of docker-compose (I'm on 1.6.2). If I modify development.yml and explicitly add a `ports` entry for `7899` to the `signer` definition it all works... That shouldn't be necessary as links should take care of the containers talking to each other.

Need to install an older version of compose to confirm. Haven't had this problem (historically) with 1.6.0 and older.

cyli:
@endophage Does that mean the linking should be aded for the regular `docker-compose.yml` as well?

endophage:
Seems to just be problems with my laptop. Got it working as is. LGTM

PR #638, opened 2016-03-18
Fix Create arg order for signer

riyazdf:
This one got away in the flatten keystore PR -- here's a fix and added test to make sure we have coverage.

Also, another reason I'm super excited to have #634 since that test would've caught this :+1: 

### DIFF
--- a/signer/client/signer_trust_test.go
+++ b/signer/client/signer_trust_test.go
@@ -114,6 +114,18 @@ func TestGetPrivateKeyIfNoKey(t *testing.T) {
 	assert.Nil(t, privKey)
 }
 
+func TestCreatePrivateKey(t *testing.T) {
+	signer := setUpSigner(t, trustmanager.NewKeyMemoryStore(ret))
+	key, err := signer.Create(data.CanonicalSnapshotRole, "docker.com/notary", data.ECDSAKey)
+	assert.NoError(t, err)
+	assert.NotNil(t, key)
+	retrievedKey := signer.GetKey(key.ID())
+	assert.NotNil(t, retrievedKey)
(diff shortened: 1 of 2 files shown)


cyli:
LGTM pending CI!  Thanks for fixing so quickly!

signer/client/signer_trust_test.go
```diff
@@ -114,6 +114,18 @@ func TestGetPrivateKeyIfNoKey(t *testing.T) {
...
+func TestCreatePrivateKey(t *testing.T) {
+	signer := setUpSigner(t, trustmanager.NewKeyMemoryStore(ret))
+	key, err := signer.Create(data.CanonicalSnapshotRole, "docker.com/notary", data.ECDSAKey)
+	assert.NoError(t, err)
+	assert.NotNil(t, key)
+	retrievedKey := signer.GetKey(key.ID())
+	assert.NotNil(t, retrievedKey)
+	assert.Equal(t, key.Public(), retrievedKey.Public())
+	assert.Equal(t, key.ID(), retrievedKey.ID())
+	assert.Equal(t, key.Algorithm(), retrievedKey.Algorithm())
```
cyli: Oops, sorry, this is newish.  Do we want the expected value to be `data.ECDSAKey` here specifically, to ensure that the right algorithm got passed?  (Maybe also assertion like `assert.Equal(t, data.ECDSAKey, key.Algorithm())` up above where `retrieveKey` is assigned?)

riyazdf: Sounds good!  Will do.

endophage:
LGTM.

PR #639, opened 2016-03-18
update top level Signed.Signed to be a *json.RawMessage

endophage:

### DIFF
--- a/server/timestamp/timestamp.go
+++ b/server/timestamp/timestamp.go
@@ -130,13 +130,14 @@ func CreateTimestamp(gun string, prev *data.SignedTimestamp, snapshot []byte, st
 	if prev != nil {
 		ts.Signed.Version = prev.Signed.Version + 1
 	}
-	sgndTs, err := json.MarshalCanonical(ts.Signed)
+	var sgndTs json.RawMessage
+	sgndTs, err = json.MarshalCanonical(ts.Signed)
 	if err != nil {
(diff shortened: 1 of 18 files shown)


server/timestamp/timestamp.go
```diff
@@ -130,13 +130,14 @@ func CreateTimestamp(gun string, prev *data.SignedTimestamp, snapshot []byte, st
 	if prev != nil {
 		ts.Signed.Version = prev.Signed.Version + 1
 	}
-	sgndTs, err := json.MarshalCanonical(ts.Signed)
+	var sgndTs json.RawMessage
+	sgndTs, err = json.MarshalCanonical(ts.Signed)
```
HuKeping: Why replace the `sgndTs, err := json.MarshalCanonical(ts.Signed)` here?

endophage: If I don't do it this way, `sgndTs` has type `[]byte` and I have to do a really ugly type cast from `[]byte` to `*json.RawMessage` on the line that changed just below. This is only one line longer and at least in my opinion much cleaner.

HuKeping: Prefer this way too :+1:

riyazdf:
changes LGTM, also tested against @cyli's integration test and my own local repos (both new and old) with no issues.

cyli:
LGTM!

PR #641, opened 2016-03-18
Print one key and/or path per line when listing delegations

riyazdf:
Changes the CLI output to print each delegation key and path on its own line, rather than keeping each role to a single line with commas between keys/paths.  Sample below:

screen shot 2016-03-18 at 4 14 32 pm


cyli:
LGTM!

endophage:
LGTM, much clearer display

PR #645, opened 2016-03-22
Delegation password environment var

riyazdf:
Adds an environment variable for notary to try against delegation role passphrases, for automation and testing purposes.


endophage:
LGTM. Given scope of change and time crunch on Docker code freeze, I'm OK with merging even with minor coverage loss.

cyli:
:+1: since although it's not covered here, the docker tests that utilize this pass, and as @endophage mentioned, we are short on time

PR #646, opened 2016-03-23
update vendoring

endophage:
Started doing this for the rethinkDB work but it turned out to be a lot of noise. This updates vendoring on its own.


riyazdf:
LGTM, seems that documentation jenkins had an unrelated blip

endophage:
forcing a rebuild just to try and make docs pass (just in case) Will merge if all green

PR #647, opened 2016-03-23
Rethinkdb support for server+signer

endophage:


cmd/notary-signer/main.go
```diff
@@ -76,17 +77,41 @@ func passphraseRetriever(keyName, alias string, createNew bool, attempts int) (p
...
+				defaultAlias = configuration.GetString(defaultAliasEnv)
+			}
+
+			if defaultAlias == "" {
+				return nil, fmt.Errorf("must provide a default alias for the key DB")
+			}
+			logrus.Debug("Default Alias: ", defaultAlias)
+			s, err := keydbstore.NewKeyRethinkDBStore(passphraseRetriever, defaultAlias, sess)
+			keyStore = s
+			hRegister("DB operational", s.CheckHealth, time.Second*60)
```
NathanMcCauley: `time.Minute`?

server/storage/rethinkdb.go
```diff
@@ -0,0 +1,194 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+// GetCurrent returns the modification date and data part of the metadata for
+// the latest version of the given GUN and role.  If there is no data for
+// the given GUN and role, an error is returned.
+func (rdb RethinkDB) GetCurrent(gun, role string) (created *time.Time, data []byte, err error) {
```
NathanMcCauley: Do we actually need this function to be parameterized? Don't we only have use cases for `getCurrentTimestamp` and `getCurrentRoot`?

NathanMcCauley: Comment also applies to the GetCurrent w/in the ConsistentMetaStore.

server/storage/consistent_store.go
```diff
@@ -0,0 +1,69 @@
...
+	"encoding/hex"
+	"fmt"
+	"time"
+
+	"github.com/docker/go/canonical/json"
+	"github.com/docker/notary"
+	"github.com/docker/notary/tuf/data"
+)
+
+// ConsistentMetaStorage wraps a MetaStore in order to walk the TUF tree for GetCurrent in a consistent manner
```
NathanMcCauley: What is meant by 'Consistent' here? I was expecting it to mean 'lookup by hash' in the sense of 'consisent snapshots'. Is it something else?

riyazdf: You're correct in that it's lookup by hash after we retrieve the latest timestamp -- I couldn't think of a great name at the time, I'll try a rename :+1:

endophage: Consistent here means regardless of race conditions on read vs write or failed updates leaving artifacts in the database, you will always get the correct piece of metadata. This consistency is only relevant in the context of `GetCurrent` where selecting the `root` (or any other role) based purely on ordering the versions in the storage may return a root that can't actually be reached via a timestamp+snapshot due to race conditions and failure cases.

buildscripts/rethinkdbintegrationtest.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
```
cyli: Wondering if we should just modify `integrationtest.sh` to take a compose file argument?  This seems otherwise the same except for that parameter.

endophage: After all the complexity via code repetition in the docker build scripts that sounds like a really good idea.

docker-compose.rethink.yml
```diff
@@ -0,0 +1,87 @@
...
+        - ./fixtures/rethinkdb:/tls
+        - rdb-01-data:/var/data
+      networks:
+        rdb:
+          aliases:
+            - rdb
+            - rdb.rdb
+            - rdb-01.rdb
+      command: "--bind all --no-http-admin --server-name rdb_01 --canonical-address rdb-01.rdb --directory /var/data/rethinkdb --join rdb.rdb --driver-tls --driver-tls-key /tls/key.pem --driver-tls-cert /tls/cert.pem --cluster-tls --cluster-tls-key /tls/key.pem --cluster-tls-cert /tls/cert.pem --cluster-tls-ca /tls/ca.pem"
+    rdb-02:
```
cyli: Should we have 3 rethink DBs for HA?  (odd number)  Not as important for our integration test version, but this might be the way we recommend users run everything with a rethink backend.

buildscripts/rethinkdbintegrationtest.sh
riyazdf: definitely agreed, updated!

docker-compose.rethink.yml
riyazdf: Thanks for pointing this out, we should definitely change this to 3.  I'll also change the the integration test since I figure we should mimic our deployable setup as much as possible

utils/configuration_test.go
```diff
@@ -178,34 +178,22 @@ func TestParseInvalidStorageNoDBSource(t *testing.T) {
...
-	require.Equal(t, expected, *store)
-}
-
 // A supported backend with DB source will be successfully parsed.
-func TestParseStorageDBStore(t *testing.T) {
+func TestParseSQLStorageDBStore(t *testing.T) {
 	config := configure(`{
 		"storage": {
-			"backend": "MySQL",
+			"backend": "mysql",
```
cyli: Non-blocking - do we specifically want different casings of "mysql" (such as "MySQL") to be invalid?  I'm fine with this, although I can see people using the normal casing of MySQL (but certainly "mYSqL" would be weird).

cmd/notary-signer/main.go
```diff
@@ -74,17 +80,41 @@ func passphraseRetriever(keyName, alias string, createNew bool, attempts int) (p
...
+			health.RegisterPeriodicFunc("DB operational", s.CheckHealth, time.Minute)
+		}
+
+	case notary.MySQLBackend, notary.SQLiteBackend:
+		storeConfig, err := utils.ParseSQLStorage(configuration)
+		if err != nil {
+			return nil, err
+		}
 		defaultAlias := configuration.GetString("storage.default_alias")
 		if defaultAlias == "" {
```
cyli: Non-blocking - should we split this out into a separate helper function, since both SQL and Rethink stores need it?

utils/configuration_test.go
```diff
@@ -178,34 +178,22 @@ func TestParseInvalidStorageNoDBSource(t *testing.T) {
...
 			require.Error(t, err, fmt.Sprintf("'%s' should be an error", configJSON))
 			require.Contains(t, err.Error(),
 				fmt.Sprintf("must provide a non-empty database source for %s", backend))
 		}
 	}
 }
 
-// If a memory storage backend is specified, no DB URL is necessary for a
-// successful storage parse.
-func TestParseStorageMemoryStore(t *testing.T) {
```
cyli: Since we're parsing memory stores in the `cmd/server` and `cmd/signer` packages now, it'd be nice to move this test to each of them (e.g. that we don't need any additional params for the memory store).

cmd/notary-signer/main.go
cyli: Also, can we add a test for rethink DB needing the default alias as well?  a la `TestSetupCryptoServicesDBStoreSuccess` and `TestSetupCryptoServicesDBStoreNoDefaultAlias`?

cmd/notary-server/config.go
```diff
@@ -61,25 +63,44 @@ func grpcTLS(configuration *viper.Viper) (*tls.Config, error) {
...
-	if storeConfig.Backend == utils.MemoryBackend {
+	switch backend {
+	case notary.MemoryBackend:
 		return storage.NewMemStorage(), nil
+	case notary.MySQLBackend, notary.SQLiteBackend:
+		storeConfig, err := utils.ParseSQLStorage(configuration)
+		if err != nil {
+			return nil, err
+		}
+		s, err := storage.NewSQLStorage(storeConfig.Backend, storeConfig.Source)
```
cyli: if using `:=` here will shadow the previous scope's `var err` - also, should this check for `err == nil` before registering a health check and setting the store, like the rethink db setting down in the next case?

cmd/notary-server/config.go
```diff
@@ -61,25 +63,44 @@ func grpcTLS(configuration *viper.Viper) (*tls.Config, error) {
...
+		store = *storage.NewTUFMetaStorage(s)
+		hRegister("DB operational", s.CheckHealth, time.Second*60)
+	case notary.RethinkDBBackend:
+		var sess *gorethink.Session
+		storeConfig, err := utils.ParseRethinkDBStorage(configuration)
+		if err != nil {
+			return nil, err
+		}
+		sess, err = rethinkdb.Connection(storeConfig.CA, storeConfig.Source)
```
cyli: Can we add a negative test for what happens if we can't connect (tls cert doesn't exist, or can't connect to server)?

utils/configuration_test.go
riyazdf: If I'm understanding correctly, I think we test this in `TestSetupCryptoServicesMemoryStore` and `TestGetMemoryStore` for signer and server?  I'm actually not sure why this test was deleted, @endophage any ideas?

cyli: Ah yes, we do, sorry.  :)  Thanks!

utils/configuration.go
```diff
@@ -81,37 +84,72 @@ func ParseLogLevel(configuration *viper.Viper, defaultLevel logrus.Level) (
...
+// ParseRethinkDBStorage tries to parse out Storage from a Viper.  If backend and
+// URL are not provided, returns a nil pointer.  Storage is required (if
+// a backend is not provided, an error will be returned.)
+func ParseRethinkDBStorage(configuration *viper.Viper) (*RethinkDBStorage, error) {
+	store := RethinkDBStorage{
+		Storage: Storage{
+			Backend: configuration.GetString("storage.backend"),
+			Source:  configuration.GetString("storage.db_url"),
+		},
+		CA:      configuration.GetString("storage.tls_ca_file"),
```
cyli: Would it make sense to call `GetPathRelativeToConfig` here because we allow relative paths for all our other cert options?

cmd/notary-signer/main.go
riyazdf: Totally agree for the default alias helper, I'll update this PR with that.

As for the rethink test, I'd like to make one but I'm not sure if there's a good way to mock a rethink connection -- I'll look into this further (cc @endophage if he has any ideas), but so far haven't had much luck with this

cmd/notary-server/config.go
riyazdf: Yeah definitely, on both points for `err` -- I'll update this

fixtures/server-config.rethink.json
```diff
@@ -0,0 +1,26 @@
...
+	},
+	"logging": {
+		"level": "debug"
+	},
+	"storage": {
+		"backend": "rethinkdb",
+		"auth_key": "devauthkey",
+		"db_url": "rdb-proxy.rdb",
+		"database": "notaryserver",
+		"tls_ca_file": "/tls/ca.pem"
```
cyli: Non-blocking - would it make sense to make this a relative link instead, if we allow relative links for this?  Since a reader would have to keep in mind that this can only be used in the context of the container as specified by one of our rethink compose files.

cmd/notary-signer/main.go
cyli: Ah right, forgot the positive test issues.  Maybe we can just have the negative test and check the error text for default alias info?

utils/configuration.go
cyli: (also, possibly a test for `ParseRethinkDBStorage` that fails if any of these are not provided or empty?)

server/storage/database_test.go
```diff
@@ -31,6 +31,18 @@ func SampleTUF(version int) TUFFile {
 	}
 }
 
+func SampleCustomTUF(role, gun string, data []byte, version int) TUFFile {
```
cyli: Totally non-blocking - wondering if it'd make sense `func SampleTuf` to make use of this function?

utils/configuration.go
riyazdf: Agree for both!  I'll change it to a relative path for CA unless there are any objections from @endophage

cmd/notary-signer/main.go
riyazdf: Sounds good, will do!

server/storage/database_test.go
riyazdf: Great idea, done!

server/storage/interface_test.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+func ConsistentEmptyGetCurrentTest(t *testing.T, s MetaStore) {
+	_, byt, err := s.GetCurrent("testGUN", data.CanonicalRootRole)
+	require.Nil(t, byt)
+	require.Error(t, err, "There should be an error Getting an empty table")
+	require.IsType(t, ErrNotFound{}, err, "Should get a not found error")
+}
+
+func ConsistentMissingTSAndSnapGetCurrentTest(t *testing.T, s MetaStore) {
```
cyli: Non-blocking - since this is specifically a feature of `TUFMetaStorage` and not part of the `MetaStore` interface docs, would it make sense for this function to take a `TUFMetaStorage` instead?  That way to make it more valid, it can check that the underlying `MetaStore` does indeed have root metadata first, but `TUFMetaStorage` would fail to return it because it's missing timestamp or snapshot?

server/storage/tuf_store_test.go
```diff
@@ -0,0 +1,106 @@
...
+// to the snapshot specified in the checksum, to potentially other role metadata by checksum
+func TestTUFSQLGetCurrent(t *testing.T) {
+	tempBaseDir, err := ioutil.TempDir("", "notary-test-")
+	gormDB, tufDBStore := SetupTUFSQLite(t, tempBaseDir)
+	defer os.RemoveAll(tempBaseDir)
+	defer gormDB.Close()
+
+	ConsistentEmptyGetCurrentTest(t, tufDBStore)
+
+	// put an arbitrary piece of data in the database
```
cyli: nit: might be good to mention the metadata is not entirely arbitrary, since it's root metadata to specifically test the inability to get to it using a consistent walk

server/storage/tuf_store.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/docker/notary/storage"
+	"github.com/docker/notary/tuf/data"
+)
+
+// TUFMetaStorage wraps a MetaStore in order to walk the TUF tree for GetCurrent in a consistent manner,
+// by always starting from a current timestamp and then looking up other data by hash
+type TUFMetaStorage struct {
+	MetaStore
+	// cached metadata by checksum
+	cachedMeta map[string]*storedMeta
```
cyli: Non-blocking:  could we also use this cache to speed up a `GetChecksum` implementation for `TUFMetaStorage`?

server/storage/tuf_store_test.go
```diff
@@ -0,0 +1,106 @@
...
+	// Delete snapshot
+	query = gormDB.Delete(&snapTUF)
+	require.NoError(t, query.Error, "Deleting a row for snapshot in DB failed.")
+
+	// GetCurrent snapshot lookup should still succeed because of caching
+	ConsistentGetCurrentFoundTest(t, tufDBStore, snapTUF)
+
+	// targets and root lookup on GetCurrent should also still succeed because of caching
+	ConsistentGetCurrentFoundTest(t, tufDBStore, targetsTUF)
+	ConsistentGetCurrentFoundTest(t, tufDBStore, rootTUF)
```
cyli: Non-blocking - would it also be useful to add another orphaned root, and show that `GetCurrent` root returns the previous root even though the orphaned root is newer?

storage/rethinkdb/bootstrap.go
```diff
@@ -0,0 +1,137 @@
...
+	_, err := gorethink.DBCreate(name).RunWrite(session)
+	if err != nil {
+		if strings.Contains(err.Error(), "already exists") {
+			return nil
+		}
+
+		return err
+	}
+
+	resp, err := gorethink.DB(name).Wait().Run(session)
```
cyli: Non-blocking - the rethink docs say that Wait() is not supported by rethink 2.3 and above - should we include a comment to that effect since rethink 2.3 just came out recently?  (Not sure if/when we'd upgrade)

HuKeping: Comments is good for code review but not a good way to remind people to update this function when upgrade the rethink to a higher version.

But it won't be a big problem either, since the build/test will fail if it has not been updated when using the higher version of rethink.

cmd/notary-server/config.go
riyazdf: done!

server/storage/tuf_store.go
riyazdf: yes, done!

server/storage/tuf_store_test.go
riyazdf: great idea, done!

server/storage/rethinkdb.go
```diff
@@ -0,0 +1,205 @@
...
+		},
+	).RunWrite(rdb.sess)
+	if err != nil && gorethink.IsConflictErr(err) {
+		return &ErrOldVersion{}
+	}
+	return err
+}
+
+// UpdateMany adds multiple new metadata for the given GUN. RethinkDB does
+// not support transactions, therefore we will attempt to insert the timestamp
```
cyli: IIRC were we going to insert the timestamp last so that the timestamp ends up being the second phase of the two-phase commit?  Otherwise get the latest entries via consistent walking, we might end up getting a version that hasn't fully been committed/might have conflicts?

I may also just be missing it, as I haven't looked at the rest of the rethink code yet, but this doesn't seem to be pulling out the timestamp update to be inserted at the beginning or end.

server/storage/rethinkdb.go
cyli: We may need to continue to support `GetCurrent` for arbitrary roles in order to support older clients that don't support consistent downloads (I think Docker 1.10 does not, for instance).

cyli:
I think in order to linearize updates to our TUF files, we need [majority write acknowledgement and majority read acknowledgement](https://rethinkdb.com/docs/consistency/) (at least during server validation).

Looks like we can't specify write acknowledgement per insert, but we can specify write acknowledgement setting for a particular table, maybe during the bootstrap?  https://www.rethinkdb.com/docs/sharding-and-replication/#write-acks-and-durability.

And we can specify the read acknowledgement during a query to the table:  https://rethinkdb.com/api/javascript/table/.

server/storage/rethinkdb.go
```diff
@@ -0,0 +1,205 @@
...
+	"encoding/hex"
+	"fmt"
+	"time"
+
+	"github.com/dancannon/gorethink"
+	"github.com/docker/notary/storage/rethinkdb"
+)
+
+// RDBTUFFile is a tuf file record
+type RDBTUFFile struct {
```
cyli: totally non-blocking naming nitpick:  should this be moved into the model file, since these seem to be the data models?

endophage: I want to break up the storage package into sub packages that define specific backends. Then we'd have a `storage/rethinkdb/models.go` but for now, the current models file is gorm models and I don't want to mix them up.

endophage:
@cyli we definitely need the majority acknowledgement on write. Do you think we need it on anything other than the tuf_files table? I think everything else is keyed sufficiently uniquely (checksum based IDs) and used in such a way as to not need it.

server/storage/rethinkdb.go
cyli: Oh definitely,I just meant the `rethink_models` file (which are right now I think table definitions)

cyli:
@endophage I think probably just the `tuf_files` table is fine?  I think when we start doing remote key rotations where the signer keeps track of all the keys, we may want to consider it for the keys table too, but I think we can probably just do that later.

riyazdf:
@cyli: thanks for the reference!  `durability: hard` is being set in the reconfiguration options, and I agree that we would like `write_acks: majority` on `tuf_files` (and ideally I think `read_mode: majority` makes sense here too).

I've added this in the latest commit.  Unfortunately the gorethink API doesn't provide a particularly clean way to set `write_acks` but I've given it a try and gotten it to work with my jenkins integration testflow.

cmd/notary-server/bootstrap.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+func bootstrap(ctx context.Context) error {
+	s := ctx.Value("metaStore")
+	if s == nil {
+		return fmt.Errorf("no store set during bootstrapping")
+	}
+	store, ok := s.(storage.Bootstrapper)
+	if !ok {
+		return fmt.Errorf("Store does not support bootstrapping.")
```
HuKeping: Do we need to  use concrete error types here, so that the test like `TestBootstrap` could look for those types rather than checking strings?

cmd/notary-server/config.go
```diff
@@ -61,25 +63,43 @@ func grpcTLS(configuration *viper.Viper) (*tls.Config, error) {
...
+	switch backend {
+	case notary.MemoryBackend:
 		return storage.NewMemStorage(), nil
+	case notary.MySQLBackend, notary.SQLiteBackend:
+		storeConfig, err := utils.ParseSQLStorage(configuration)
+		if err != nil {
+			return nil, err
+		}
+		var s *storage.SQLStorage
+		s, err = storage.NewSQLStorage(storeConfig.Backend, storeConfig.Source)
```
HuKeping: why not just `s, err := ...` here

cmd/notary-server/config.go
```diff
@@ -61,25 +63,43 @@ func grpcTLS(configuration *viper.Viper) (*tls.Config, error) {
...
+		if err != nil {
+			return nil, fmt.Errorf("Error starting %s driver: %s", backend, err.Error())
+		}
+		store = *storage.NewTUFMetaStorage(s)
+		hRegister("DB operational", s.CheckHealth, time.Second*60)
```
HuKeping: Could 60s be a default constant?

cmd/notary-server/config.go
```diff
@@ -61,25 +63,43 @@ func grpcTLS(configuration *viper.Viper) (*tls.Config, error) {
...
-		return nil, err
-	}
-	logrus.Infof("Using %s backend", storeConfig.Backend)
-
-	if storeConfig.Backend == utils.MemoryBackend {
+	switch backend {
+	case notary.MemoryBackend:
 		return storage.NewMemStorage(), nil
+	case notary.MySQLBackend, notary.SQLiteBackend:
+		storeConfig, err := utils.ParseSQLStorage(configuration)
```
HuKeping: Since the function name is `ParseRethinkDBStorage` in RethinkDB case, can we use `ParseMySQLStorage` here to keep consistent

cmd/notary-server/config.go
```diff
@@ -61,25 +63,43 @@ func grpcTLS(configuration *viper.Viper) (*tls.Config, error) {
...
+		}
+		sess, err = rethinkdb.Connection(storeConfig.CA, storeConfig.Source)
+		if err != nil {
+			return nil, fmt.Errorf("Error starting %s driver: %s", backend, err.Error())
+		}
+		s := storage.NewRethinkDBStorage(storeConfig.DBName, sess)
+		store = *storage.NewTUFMetaStorage(s)
+		hRegister("DB operational", s.CheckHealth, time.Second*60)
+	default:
+		return nil, fmt.Errorf("%s not a supported storage backend", backend)
```
HuKeping: nit `%s is not a`

cmd/notary-server/config.go
endophage: Yep, should be `time.Minute`

cmd/notary-server/config.go
endophage: I needed to do this somewhere due to typing problems that resulting in `s.HealthCheck` a few lines below not resolving. Can't see where that was but it's possible other things have changed in the meantime that resolved it. Will change.

cmd/notary-server/config.go
endophage: Technically it's not parse MySQL storage. It parses any storage that's compatible with the ORM, which includes MySQL, SQLite, PostgreSQL, and I think MSSQL. We may also change ORM if anyone ever asks for Oracle support (easier to support one ORM for the whole SQL family).

server/timestamp/timestamp.go
```diff
@@ -65,28 +69,40 @@ func GetOrCreateTimestamp(gun string, store storage.MetaStore, cryptoService sig
 		return nil, nil, err
 	}
 
-	_, snapshot, err := snapshot.GetOrCreateSnapshot(gun, store, cryptoService)
+	snapshotTime, snapshot, err := snapshot.GetOrCreateSnapshot(gun, store, cryptoService)
```
cyli: Non-blocking:  since this is the only thing that calls `GetOrCreateSnapshot` now, would it make sense to pass the previous timestamp or snapshot checksum from the previous timestamp, so `GetOrCreateSnapshot` can just get the metadata by checksum rather than have to walk the timestamp again?

I don't think it affects functionality (since either `GetOrCreateSnapshot` would just get the same timestamp and walk to the right place, or it gets a newer timestamp, and this timestamp generation will fail when writing), but would save one extra database read and unmarshal.

endophage: That makes a lot of sense :+1:

cmd/notary-server/config.go
HuKeping: OK, `ParseSQLStorage` make sense then.

cyli:
👍 Awesome, thanks for addressing everything and getting this up and running!  The rethink integration script runs fine on my machine.  I'm fine with writing rethink model specific tests along with setting up CI for code coverage in another PR.

PR #648, opened 2016-03-24
Root rotation (previously root cert rotation)

cyli:
This is the rebased version + some changes from https://github.com/docker/notary/pull/267.

To summarize:  that PR added library support and CLI support for rotating the public x509 root public key in a root.json, in case it expires.  This does not rotate the root private key.

There are a couple additional commits that address some comments from the previous PR:

1.  https://github.com/docker/notary/pull/267#discussion_r51491390 and https://github.com/docker/notary/pull/267#discussion_r51491941: MultiSigning support - currently, notary attempts to support theoretical multisigning, and we would like to preserve this as we would like to implement threshold support for root keys in the near-ish future.  However, @mtrmac is correct in that currently, there is no recommended mechanism or any infrastructure in place to handle that.  So this PR punts determination of how many signatures are needed up to `tuf.Repo`'s `SignRoot`, which right now requires that the threshold signatures be met for every required role.

2. https://github.com/docker/notary/pull/267#discussion_r48772435, https://github.com/docker/notary/pull/267#discussion_r51492254, https://github.com/docker/notary/pull/267#discussion_r51492399:  Record of old keys - this maintains the old public keys in the `tuf.Repo.Root.Signed.Keys` object, but there is otherwise no way in spec to keep track of which of those keys are actually old keys, vs cruft.  This PR adds versioned root roles every time root keys are changed to the list of roles in `tuf.Repo.Root.Signed.Roles` - this technically does not violate the spec, because it doesn't change existing roles, and is not used for validation, only hints w.r.t. which keys to use for signing.  It may also be helpful in maintaining a history of root key changes).

This is probably ready for first look.  @mtrmac?


client/client.go
```diff
@@ -963,3 +984,93 @@ func (r *NotaryRepository) DeleteTrustData() error {
...
+		} else if _, ok := err.(*trustmanager.ErrNoCertificatesFound); !ok {
+			return nil, err
+		}
+	}
+	return certs, nil
+}
+
+// RotateRootCert generates a new certificate to replace oldCert and adds
+// a changelist entry to update the root role with this certificate replacement
+// shen r.Publish() is called.
```
riyazdf: typo: `when`

client/client.go
```diff
@@ -963,3 +984,93 @@ func (r *NotaryRepository) DeleteTrustData() error {
...
+
+	// Hard-coded policy: the generated certificate expires in 10 years.
+	startTime := time.Now()
+	newCert, err := cryptoservice.GenerateCertificate(privKey, r.gun, startTime, startTime.AddDate(10, 0, 0))
+	if err != nil {
+		return err
+	}
+	newCertX509PublicKey := trustmanager.CertToKey(newCert)
+	if newCertX509PublicKey == nil {
+		return fmt.Errorf("cannot use regenerated certificate?! format %s", newCert.PublicKeyAlgorithm)
```
riyazdf: I'd also be surprised by this error occurring, but maybe we shouldn't have the `?!` in the message :)

client/client_test.go
```diff
@@ -3339,3 +3340,134 @@ func TestListRoles(t *testing.T) {
...
+	_, err = userRepo.Update(false)
+	require.NoError(t, err)
+
+	// Rotate root certificate.
+	certs, err := authorRepo.ListRootCerts()
+	require.NoError(t, err)
+	var cert *x509.Certificate
+	require.Len(t, certs, 1)
+	// Just get the only certificate in there.
+	for _, c := range certs {
```
riyazdf: I'm unclear why we need this loop instead of just an indexing?

client/client_test.go
```diff
@@ -3339,3 +3340,134 @@ func TestListRoles(t *testing.T) {
...
+	// Verify that clients initialized post-rotation can use the repo, and use
+	// the new certificate immediately.
+	freshUserRepo, _ := newRepoToTestRepo(t, authorRepo, true)
+	defer os.RemoveAll(freshUserRepo.baseDir)
+	receivedTarget, err = freshUserRepo.GetTargetByName("current")
+	require.NoError(t, err)
+	require.Equal(t, newRootCertID, rootRoleCertID(t, freshUserRepo))
+	verifyOnlyTrustedCertificate(t, freshUserRepo, newRootCertID)
+	require.True(t, reflect.DeepEqual(*publishedTarget, receivedTarget.Target), "target does not match")
+	logRepoTrustRoot(t, "fresh client", freshUserRepo)
```
riyazdf: this is an awesome test :+1:

client/client_test.go
```diff
@@ -3339,3 +3340,134 @@ func TestListRoles(t *testing.T) {
...
+	// This would be nice to fix eventually (breaking down the NotaryRepository.Update
+	// / Repo.Update separation which causes this), but for now, just ensure that the
+	// second update does result in updated certificates.
+
+	// Verify that the author eventually rotates to the new certificate.
+	_, err = authorRepo.Update(false)
+	require.NoError(t, err)
+	logRepoTrustRoot(t, "author refresh 1", authorRepo)
+	require.Equal(t, newRootCertID, rootRoleCertID(t, authorRepo))
+	// verifyOnlyTrustedCertificate(t, authorRepo, newRootCertID)
```
riyazdf: nit: commented out line

cmd/notary/cert.go
```diff
@@ -25,6 +25,12 @@ var cmdCertListTemplate = usageTemplate{
 	Long:  "Lists root certificates known to notary.",
 }
 
+var cmdCertRotateTemplate = &usageTemplate{
+	Use:   "rotate [ GUN ]",
+	Short: "Rotate certificates for a role.",
```
riyazdf: should this be `"for a gun`"?  And below as well?

tuf/signed/sign.go
```diff
@@ -13,54 +13,72 @@ package signed
...
 )
 
-// Sign takes a data.Signed and a key, calculated and adds the signature
-// to the data.Signed
+// Sign takes a data.Signed and keys, calculates and adds at least
+// minSignature signatures using primaryKeys the data.Signed.  It will also
+// clean up any signatures produced by a key not in the primaryKeys or
+// validSigningKeys list.
+// (validSigningKeys can be either a superset of or a completely disjoint set from
+// primaryKeys)
```
riyazdf: Should it be noted that `primaryKeys` are still considered valid keys to produce signatures with?  It might be confusing if `primaryKeys` and `validSigningKeys` are disjoint, but both still completely valid

tuf/tuf.go
```diff
@@ -511,14 +545,15 @@ func (tr *Repo) InitTimestamp() error {
...
-	return nil
+	return tr.SetTimestamp(timestamp)
 }
 
 // SetRoot sets the Repo.Root field to the SignedRoot object.
 func (tr *Repo) SetRoot(s *data.SignedRoot) error {
 	tr.Root = s
-	return nil
+	var err error
+	tr.originalRootRole, err = tr.Root.BuildBaseRole(data.CanonicalRootRole)
```
riyazdf: I think it'd be really helpful for more information on what `originalRootRole` represents for the repo (either here or where it's first declared).  I'm not sure if another name might be more descriptive, I'll let you know if I think of something :)

cyli: That would be awesome, thanks :)

certs/certs.go
```diff
@@ -36,6 +37,18 @@ func (err ErrRootRotationFail) Error() string {
 	return fmt.Sprintf("could not rotate trust to a new trusted root: %s", err.Reason)
 }
 
+func certIDs(certs []*x509.Certificate) string {
+	ids := make([]string, 0, len(certs))
+	for _, cert := range certs {
+		id, err := trustmanager.FingerprintCert(cert)
+		if err != nil {
+			id = fmt.Sprintf("[Error %s]", err)
```
riyazdf: nit: since we're using this function for debug strings, can we maybe include `Debug` or a similar verb to the function names so it isn't accidentally called for other purposes?

client/changelist/change.go
```diff
@@ -38,8 +38,10 @@ type TufChange struct {
 // TufRootData represents a modification of the keys associated
 // with a role that appears in the root.json
 type TufRootData struct {
-	Keys     data.KeyList `json:"keys"`
-	RoleName string       `json:"role"`
+	ReplaceKeys data.KeyList `json:"keys"`
```
riyazdf: should this key also have the `omitempty` tag since we use this struct when unmarshalling in `applyRootRoleChange`?

riyazdf: Actually wait, I misread -- this is fine since this is the original `Keys` field :+1:

certs/certs.go
cyli: Renamed it `prettyFormatCertIDs`

certs/certs.go
```diff
@@ -276,5 +299,5 @@ func certsToRemove(oldCerts, newCerts []*x509.Certificate) map[string]*x509.Cert
 		}
 	}
 
-	return certsToRemove
+	return certsToRemove, nil
```
endophage: Can't comment in the unchanged middle of this function, but with the changed behaviour such that we error if `len(newCerts) == 0` should we also check between the 2 loops that `newCertMap` isn't 0 length?

client/changelist/change.go
endophage: Rather than having 3 fields, it seems like `AddKeys` and `RemoveKeys` would be sufficient with differerent actions denoting how those fields get used.

At an extreme, we could even keep the single `Keys` field and separate out ordered actions for add and remove.

client/client.go
```diff
@@ -963,3 +984,93 @@ func (r *NotaryRepository) DeleteTrustData() error {
...
+	}
+
+	privKey, _, err := r.CryptoService.GetPrivateKey(rootKeyID)
+	if err != nil {
+		return err
+	}
+
+	// Hard-coded policy: the generated certificate expires in 10 years.
+	startTime := time.Now()
+	newCert, err := cryptoservice.GenerateCertificate(privKey, r.gun, startTime, startTime.AddDate(10, 0, 0))
```
endophage: We should use the durations we declared: `startTime.Add(notary.Year * 10)`

client/client.go
```diff
@@ -963,3 +984,93 @@ func (r *NotaryRepository) DeleteTrustData() error {
...
+// (In particular, not old certificates which we have already
+// rotated from but are still using for signing to allow rollover,
+// even if they are still trusted locally.)
+func (r *NotaryRepository) ListRootCerts() ([]*x509.Certificate, error) {
+	_, err := r.Update(false)
+	if err != nil {
+		return nil, err
+	}
+
+	keyIDs := r.tufRepo.Root.Signed.Roles[data.CanonicalRootRole].KeyIDs
```
endophage: Can we use `tufRepo.GetBaseRole` now?

certs/certs.go
cyli: Good point - actually we can probably just check that the `newCertMap` isn't 0?

client/client.go
```diff
@@ -963,3 +984,93 @@ func (r *NotaryRepository) DeleteTrustData() error {
...
+
+// RotateRootCert generates a new certificate to replace oldCert and adds
+// a changelist entry to update the root role with this certificate replacement
+// when r.Publish() is called.
+func (r *NotaryRepository) RotateRootCert(oldCert *x509.Certificate) error {
+	oldCertX509PublicKey := trustmanager.CertToKey(oldCert)
+	if oldCertX509PublicKey == nil {
+		return fmt.Errorf("invalid format for root key: %s", oldCert.PublicKeyAlgorithm)
+	}
+	rootKeyID, err := trustmanager.X509PublicKeyID(oldCertX509PublicKey)
```
endophage: This returns the TUF ID rather than the canonical ID. If I remember right, `CryptoService.GetPrivateKey` requires the canonical ID.

cmd/notary/cert.go
```diff
@@ -159,6 +166,51 @@ func (c *certCommander) certRemove(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+// certRotate replaces a certificate with a new version
+func (c *certCommander) certRotate(cmd *cobra.Command, args []string) error {
```
endophage: This appears to always rotate all root keys (or possibly all those for which we hold the private key?). Is that the behaviour we want?

client/client.go
cyli: Yes.  Also, hm... actually looking at this piece, it looks like below where it checks the cert store may not work?  We can't count on the cert store to be updated/cleaned up after an `Update`, right?  Since that only happens currently on bootstrap, and won't happen if we download another root after bootstrap.

endophage: Ah yep, that true. There was something I couldn't grasp in the back of my mind telling me we couldn't rely on what's on disk and I think you've just worked it out. Also, with @riyazdf's work on cert pinning we can start looking at removing the cert store entirely and just relying on what's in root.json.

cyli: @endophage I think we still have cert pinning in place in that PR for TOFU, but yes that makes sense.

client/client.go
cyli: I believe this does return the canonical ID - [`utils.CanonicalKeyID`](https://github.com/docker/notary/blob/master/tuf/utils/util.go#L105) calls this `trustmanager.X509PublicKeyID` to get the canonical ID.  But it's still probably better to just call `utils.CanonicalKeyID` for consistency.

certs/certs.go
cyli: Done

client/client.go
endophage: Yeah, there are some further changes that need to be made but my idea was that having the root.json is equally good for TOFU because it contains all the certs we would have written to the store.

client/changelist/change.go
cyli: We probably want to keep the `keys` field for JSON unmarshalling for backwards compatibility, if someone has an old changefile lying around.  But maybe we can fix that in the unmarshaller.

I think we need to have `AddKeys` and `RemoveKeys` be separate though, because otherwise we can't add more keys or remove specific keys?

endophage: We've never done anything with root keys up to now so that actions can be treated slighty differently when dealing with that role specifically and it won't break existing clients.

cyli: Oh I meant if we remove/ignore the keys field.  Isn't this data structure used for all key rotations of base roles?  Or do you mean create a new data structure for root keys?

endophage: We need the `keys` field to exist in the JSON. I found `ReplaceKeys` a slightly confusing name: are they keys to be replaced, or keys we are using as replacements? Figured, as we could create a new action `ActionReplace` or similar and just use the `AddKeys` field (which would map to `keys` in JSON), rather than needing 3 key fields in the struct.

cyli: Ah ok, thanks for clarifying!

cmd/notary/cert.go
cyli: It seems to try to rotate all the root keys - if we don't have the private key to any of them, it seems to fail.  There probably shouldn't be like 50 root keys on a repo, so it seems like just rotating all the ones for which we hold the key seems ok?  On the other hand, since there should be so few keys on the repo, maybe this command should take the fingerprint instead?

We might want to alter `cert list` to also look at the `root.json`s, instead of just the certs on disk, and to star whichever ones we have private keys for?  That way users can figure out which cert they want to rotate.

cyli:
Feedback from group review:

- we should probably just rotate the whole root - root certs expire after 10 years, and it seems fine to just have to rotate the entire root including private key - this helps with root key compromise, also
- remove the `rootRoleDirty` flag, and just compare the original root role against the current root role
- we should always sign with the immediately previous version in the root, rather than necessarily with the original root role
- the yubikey check for `pkcs11.CKR_FUNCTION_FAILED` may not actually be correct

client/client_test.go
```diff
@@ -2704,6 +2729,130 @@ func testRotateKeySuccess(t *testing.T, serverManagesSnapshotInit bool,
...
+	require.Equal(t, newRootCertID, rootRoleCertID(t, freshUserRepo))
+	verifyOnlyTrustedCertificate(t, freshUserRepo, newRootCertID)
+	logRepoTrustRoot(t, "fresh client", freshUserRepo)
+
+	// Verify that the user initialized with the original certificate eventually
+	// rotates to the new certificate.
+	err = userRepo.Update(false)
+	require.NoError(t, err)
+	logRepoTrustRoot(t, "user refresh 1", userRepo)
+	require.Equal(t, newRootCertID, rootRoleCertID(t, userRepo))
```
endophage: We should test that making some other modification in the root.json (like rotating the targets key), causing it to be signed again still signs with both the old and new keys.

tuf/tuf.go
```diff
@@ -458,8 +491,7 @@ func (tr *Repo) InitRoot(root, timestamp, snapshot, targets data.BaseRole, consi
 	if err != nil {
 		return err
 	}
-	tr.Root = r
-	return nil
+	return tr.SetRoot(r)
```
endophage: should probably leave as direct assignment given your builder PR is going to take out these `Set___` methods.

tuf/tuf.go
```diff
@@ -786,27 +820,148 @@ func (tr *Repo) UpdateTimestamp(s *data.Signed) error {
...
+	tr.Root.Signed.Version++
+
+	// if the current role doesn't match with the latest saved role, save it
+	if !currRoot.Equals(latestSavedRole) {
+		tr.saveOldRootRole(currRoot, tr.Root.Signed.Version)
+		rolesToSignWith = append(rolesToSignWith, currRoot)
+	}
+
 	signed, err := tr.Root.ToSigned()
 	if err != nil {
```
endophage: rather than resetting, how complicated would it be to create a copy of the root and assign the whole thing over the top of `tr.Root` on success?

tuf/tuf.go
cyli: I originally thought I needed to call `SetRoot` because it sets the originalRootRole - I'm wondering now though whether `originalRootRole` should be a pointer and it should't `originalRootRole` if it's not set?  That way we don't accidentally save version 0.

tuf/tuf.go
cyli: Good idea!  Will do this instead.

tuf/tuf.go
cyli: Nm, will just set the original role in `InitRoot` too.  :)

client/client_test.go
cyli: Good idea!  Added.

endophage:
LGTM!!! In the immortal words of Queen: "It finally happened"

riyazdf:
LGTM!

PR #651, opened 2016-03-30
Update vendoring test to use new vendor structure

riyazdf:
Since we changed our Godeps structure to the `vendor` directory, this test should also be updated.  Was able to successfully run on my local OSX setup (and all tests passed against master :+1: )

### DIFF
--- a/misc/docker-integration-test.py
+++ b/misc/docker-integration-test.py
@@ -65,9 +65,10 @@ def fake_vendor():
     # based on a tag or SHA, and we want to build based on what was vendored in
     dockerfile_addition = ("\n"
         "RUN set -x && "
-        "GOPATH=$(pwd)/vendor/src/github.com/docker/notary/Godeps/_workspace:$GOPATH "
+        "export GO15VENDOREXPERIMENT=1 && "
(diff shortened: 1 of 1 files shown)


cyli:
LGTM!  :+1: Thanks for fixing this!

misc/docker-integration-test.py
```diff
@@ -65,9 +65,10 @@ def ignore_dirs(walked_dir, _):
     # based on a tag or SHA, and we want to build based on what was vendored in
     dockerfile_addition = ("\n"
         "RUN set -x && "
-        "GOPATH=$(pwd)/vendor/src/github.com/docker/notary/Godeps/_workspace:$GOPATH "
+        "export GO15VENDOREXPERIMENT=1 && "
```
endophage: If this is set, is it necessary to have the vendor dir set in the GOPATH? (on the basis we don't support Go < 1.5)

riyazdf: ah yes, it does handle vendor (and nested vendors) correctly!  Done

endophage:
Looking at the change, I'm fine with the codecov drop. LGTM

PR #653, opened 2016-03-31
WIP: support all roles in yubikey

endophage:

### DIFF
--- a/trustmanager/yubikey/yubikeystore.go
+++ b/trustmanager/yubikey/yubikeystore.go
@@ -5,14 +5,12 @@ package yubikey
 import (
 	"crypto"
 	"crypto/ecdsa"
-	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/sha256"
 	"crypto/x509"
 	"errors"
 	"fmt"
 	"io"
-	"math/big"
 	"os"
 	"time"
 
@@ -25,6 +23,7 @@ import (
 )
 
 const (
+	// defaults we'll attempt to use
 	USER_PIN    = "123456"
 	SO_USER_PIN = "010203040506070801020304050607080102030405060708"
 	numSlots    = 4 // number of slots in the yubikey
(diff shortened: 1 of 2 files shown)


trustmanager/yubikey/yubikeystore.go
```diff
@@ -686,18 +711,32 @@ func (s *YubiKeyStore) addKey(keyID, role string, privKey data.PrivateKey) (
...
-			slotID: slot,
-		}
-		return true, nil
+	switch privKey.(type) {
+	case *data.ECDSAPrivateKey:
+		err = addECDSAKey(
+			ctx,
+			session,
+			privKey,
+			slot,
```
riyazdf: I've been wondering if we should change the signatures for `addECDSAKey` and `createLoadCert` to have a `slotID` variable instead of `privKeyID` variable, WDYT?

endophage: The terminology is weird and we never settled on something. If memory serves (from discussions with the Yubikey folks 4-5 months back), in PKCS11 a "slot" is technically a device, i.e. the Yubikey itself is a PKCS11 "slot". I'm not sure what the correct term is for the space that a key actually resides in. We should find out and stop using "slot" because it'll just confuse anyone familiar with PKCS11.

riyazdf: Seems like "token"? The [pkcs11 spec](http://www.emc.com/emc-plus/rsa-labs/standards-initiatives/pkcs-11-cryptographic-token-interface-standard.htm) has a nice section for definitions on page 12 of the core specification PDF (numbered pg6 on the document)

endophage: Yeah, `token` does appear to be the correct PKCS11 name. I'll switch over to that in the PR.

HuKeping:
Hi, @endophage if you can put a new line between your commit message and the sign-off-by, it will be better when we use `git log --oneline` ;)

endophage:
There have been a lot of changes so closing this. Will preserve branch for future reference.

PR #655, opened 2016-04-01
Convert the server and signer images to use alpine

### LINKED ISSUE #635: Convert the server and signer to use alpine images (blocked on musl bugs)
See https://github.com/golang/go/issues/14476

We have to build the migrate package, which requires musl libc.

This is sort of done in https://github.com/docker/notary/compare/server-signer-alpine?expand=1, but docker-compose up fails.

In that branch, the server and signer build fine, but when starting up, we see:

```
signer_1 | Segmentation fault
signer_1 | waiting for notarymysql to come up.
server_1 | Segmentation fault
server_1 | waiting for notarymysql to come up.
```

Over and over - the segfault happens when trying to run the built `migrate` binary.

As per the issue, it seems like the fix might be in 1.6.1?

Once we upgrade to 1.6.1, if the bug has been fixed, then we can merge the branch.

UPDATE:  As @dmcgowan points out, that issue is linux/386 related, and `golang:1.6-alpine` is using the amd64 version of go.  I've opened https://github.com/golang/go/issues/14851 as a result - it may not be go 1.6, could possibly be a bug in go-sqlite3's CGO, but I can't tell :(

cyli:
Fixes #635.

Thanks to @ncopa for his solution at https://github.com/golang/go/issues/14851.

### DIFF
--- a/server.Dockerfile
+++ b/server.Dockerfile
@@ -1,12 +1,7 @@
-FROM golang:1.6.0
+FROM golang:1.6.0-alpine
 MAINTAINER David Lawrence "david.lawrence@docker.com"
 
-RUN apt-get update && apt-get install -y \
-    libltdl-dev \
-    --no-install-recommends \
-    && rm -rf /var/lib/apt/lists/*
-
-EXPOSE 4443
+RUN apk add --update git gcc libc-dev && rm -rf /var/cache/apk/*
 
 # Install DB migration tool
 RUN go get github.com/mattes/migrate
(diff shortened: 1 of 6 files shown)


server.Dockerfile
```diff
@@ -1,28 +1,25 @@
...
-RUN go get github.com/mattes/migrate
+RUN go get -ldflags "-extldflags -fno-PIC" github.com/mattes/migrate
 
 ENV NOTARYPKG github.com/docker/notary
 
 # Copy the local repo to the expected go path
 COPY . /go/src/github.com/docker/notary
 
-WORKDIR /go/src/${NOTARYPKG}
+WORKDIR /go/src/github.com/docker/notary
```
riyazdf: nit: this could still use `${NOTARYPKG}`, right?

server.Dockerfile
```diff
@@ -1,28 +1,25 @@
...
-    libltdl-dev \
-    --no-install-recommends \
-    && rm -rf /var/lib/apt/lists/*
-
-EXPOSE 4443
+RUN apk add --update git gcc libc-dev && rm -rf /var/cache/apk/*
 
 # Install DB migration tool
-RUN go get github.com/mattes/migrate
+RUN go get -ldflags "-extldflags -fno-PIC" github.com/mattes/migrate
```
riyazdf: out of curiosity, why do we need this change?

cyli: Because of https://github.com/golang/go/issues/14851#issuecomment-200915770 - without that the `migrate` binary segfaults.  :|  If the golang-alpine image gets upgraded, we can probably remove this.

riyazdf: :+1: we can maybe add a comment about how this can be removed in the future if golang-alpine is upgraded upstream?

cyli: Sounds good!

server.Dockerfile
cyli: Good point, thanks!

riyazdf:
LGTM pending CI (since `make integration` tests this setup)

endophage:
I'm having networking issues with compose again (while testing this PR) but I don't think they're related.

LGTM.

PR #656, opened 2016-04-01
Build run integration test

riyazdf:
Adds simple build and run tests (which pass when I run the script locally) to the docker versioning integration test.

Also bumps the version for docker 1.10 and updates the command for `notary key rotate`

### DIFF
--- a/misc/dockertest.py
+++ b/misc/dockertest.py
@@ -13,7 +13,7 @@ import re
 import shutil
 import subprocess
 from tempfile import mkdtemp
-from time import time
+from time import sleep, time
 import urllib
 from urlparse import urljoin
 
@@ -52,6 +52,11 @@ DEFAULT_DOCKER_CONFIG = os.path.expanduser("~/.docker")
 
 # Assumes the trust server will be run using compose if DOCKER_CONTENT_TRUST_SERVER is not specified
 DEFAULT_NOTARY_SERVER = "https://notary-server:4443"
+
+# please enter a custom trust server location if you do not wish to use a local
+# docker-compose instantiation.  If testing against Docker Hub's notary server or
+# another trust server, please also ensure that this script does not pick up incorrect TLS
+# certificates from ~/.notary/config.json by default
 TRUST_SERVER =  os.getenv('DOCKER_CONTENT_TRUST_SERVER', DEFAULT_NOTARY_SERVER)
 
 # Assumes the test will be run with `python misc/dockertest.py` from
@@ -253,6 +258,9 @@ def push(fout, docker_version, image, tag):
                      fout)
     sha = _DIGEST_REGEX.search(output).group(1)
     size = _SIZE_REGEX.search(output).group(1)
+
+    # sleep for 1s after pushing, just to let things propagate :)
+    time.sleep(1)
 
     # list
     targets = notary_list(fout, image)
--- a/misc/dockertest.py
+++ b/misc/dockertest.py
@@ -27,14 +27,13 @@ DOCKERS = {
     "1.8": "docker-1.8.3",
     "1.9": "docker-1.9.1",
     "1.10": "docker",
-    "1.10.2": "docker-1.10.2.RC1",
 }
 
 # delete any of these if you want to specify the docker binaries yourself
 DOWNLOAD_DOCKERS = {
     "1.8": ("https://get.docker.com", "docker-1.8.3"),
     "1.9": ("https://get.docker.com", "docker-1.9.1"),
-    "1.10": ("https://get.docker.com", "docker-1.10.1")
+    "1.10": ("https://get.docker.com", "docker-1.10.3")
 }
 
 # please replace with private registry if you want to test against a private
@@ -57,9 +56,6 @@ NOTARY_CLIENT = "bin/notary -c cmd/notary/config.json"
 
 # Assumes the trust server will be run using compose
 TRUST_SERVER = "https://notary-server:4443"
-
-# simple dockerfile to test building with trust
-DOCKERFILE_FIXTURE = "FROM alpine:latest\nRUN sh"
 
 # ---- setup ----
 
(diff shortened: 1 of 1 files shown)


misc/dockertest.py
```diff
@@ -271,6 +294,24 @@ def notary_list(fout, repo):
     return [line.strip().split() for line in lines[2:]]
 
 
+def test_build(fout, image, docker_version):
+    """
+    Build from a simple Dockerfile and ensure it works with DCT enabled
+    """
+    # build
+    # simple dockerfile to test building with trust
+    dockerfile = "FROM {0}:{1}\nRUN sh\n".format(image, docker_version)
```
cyli: Should this call `clear_tuf()` first?  When we were testing manually, this succeeded if the TUF data existed because it fell back on the cache, even if there was a problem with the server.  

I'm less sure about whether we care about removing the tag first, since I think trusted builds will probably check the trust server anyway even if the image exists locally - I forget if that impacted anything at all when we were testing yesterday.

misc/dockertest.py
```diff
@@ -335,6 +376,16 @@ def test_push(tempdir, docker_version, image, tag="", allow_push_failure=False,
         return return_val
 
 
+def test_run(fout, image, docker_version):
+    """
+    Runs a simple alpine container to ensure it works with DCT enabled
+    """
+    # run
+    output = run_cmd(
```
cyli: Similarly, do we want to clear all TUF data first?

misc/dockertest.py
```diff
@@ -271,6 +294,24 @@ def notary_list(fout, repo):
...
+    """
+    # build
+    # simple dockerfile to test building with trust
+    dockerfile = "FROM {0}:{1}\nRUN sh\n".format(image, docker_version)
+    tempdir_dockerfile = os.path.join(_TEMPDIR, "Dockerfile")
+    with open(tempdir_dockerfile, 'wb') as ftemp:
+      ftemp.write(dockerfile)
+
+    output = run_cmd_with_input(
+        "{0} build -".format(DOCKERS[docker_version]), tempdir_dockerfile, fout)
```
cyli: Just curious - was just wondering why `cat <file> | docker build -` as opposed to `docker build -f <file>`?

misc/dockertest.py
riyazdf: good idea, will do!

misc/dockertest.py
riyazdf: Yeah I was avoiding the build context, though it doesn't actually work for docker 1.8.3 :(

cyli:
Non-blocking: did we want to change the notary server to get the server from an environment variable, in case we want to use hub?

riyazdf:
@cyli: definitely -- I added a commit to use the value in `DOCKER_CONTENT_TRUST_SERVER` and use `https://notary-server:4443` by default if no value is set in the env var.  Is this what you had in mind?

cyli:
@riyazdf Awesome, yes, thank you.  :)  I think the server must also be passed to the notary client (maybe skipping the config file entirely), though, since the notary client uses the `cmd/notary/config.json` which points it at a local server currently.

riyazdf:
@cyli good catch!  I've added `TRUST_SERVER` as a flag to the notary client command, which should take priority over the config value

misc/dockertest.py
```diff
@@ -51,12 +50,13 @@
...
+TRUST_SERVER =  os.getenv('DOCKER_CONTENT_TRUST_SERVER', "https://notary-server:4443")
+
 # Assumes the test will be run with `python misc/dockertest.py` from
 # the root of the notary repo after binaries are built
-NOTARY_CLIENT = "bin/notary -c cmd/notary/config.json"
-
-# Assumes the trust server will be run using compose
-TRUST_SERVER = "https://notary-server:4443"
+# also overrides the notary server location if need be
+NOTARY_CLIENT = "bin/notary -c cmd/notary/config.json -s {0}".format(TRUST_SERVER)
```
cyli: I think this may cause problems because the `config.json` has SSL certs for our testing server, which would fail if applied to the `TRUST_SERVER` - I think we may need to only conditionally pass `cmd/notary/config.json` :|  Apologies for the problems this is causing!

riyazdf: ah, yes.  I've updated with a conditional for assigning to `NOTARY_CLIENT`

cyli:
LGTM, thank you for all your work on this!  Tests pass for me locally, with a couple caveats:

- One issue I ran into against docker hub is I think maybe the auth token propagation times?  Every so often, during a list immediately after a push, I get a 401, but then later when I run the list command manually it passes.  If we want to make this into a CI job, we might want to slow the script down a bit if hitting hub.

- Non-blocking:  Might be useful to note that if a user has a configuration file at `~/.notary/config.json` with custom TLS certs, TLS passing `DOCKER_CONTENT_TRUST_SERVER=https://notary.docker.io` is still broken because it uses those custom certs.  We could write an empty config file and pass that instead, but the user also needs the correct docker config file with creds, so I'm not too opposed to saying that the user should just have clean/good state instead.

riyazdf:
@cyli thank you for all of your feedback!

Yeah I'm not sure what we can do there aside from adding some sleeps between commands?  I can add conditional logic to add `sleep(10)` that we can trigger for CI.

I'm fine with adding a comment to note the TLS certs -- I don't think automatically writing an empty config file would be expected

cyli:
@riyazdf Yeah sleeps are pretty much all I can think of.  maybe just a delay of 500ms-1s between invocations to `run` would be sufficient?  Or maybe just to after `push`, since that creates the repo and then once the token propagates everything is fine.

cyli:
LGTM despite weird codecov results.

endophage:
LGTM

Re the propagation times, I've seen that before. It seems to take a little time for registry to report to the token server that the repo was actually created, and the token server will only issue a `pull` token for a repository that it thinks exists (it'll issue a `push` token for anything you would have access to based on the name).

Issue #628, opened 2016-03-16
Replace assert with require


endophage:
In all the *_test.go files, replace the use of the `assert` package with the `require` package, both from the `github.com/stretchr/testify` repo. It should just be a straight string replace `s/assert/require/`. It just makes the tests run a little cleaner as assert allows the test to continue, sometimes resulting in panics (that output a lot of noise), while require fails the test immediately.

IMPORTANT: don't submit one massive PR with every single test file changed. We do actually need to review the change. Submit one PR per Go package. That'll also help avoid as many rebase/merge problems.

HuKeping:
@endophage If you can make a list here, it would be more clear for people to pick it up.

HuKeping:
[ ] client/client_test.go
[ ] client/changelist/changelist_test.go
[ ] client/changelist/file_changelist_test.go
[ ] client/changelist/change_test.go
[ ] client/helpers_test.go
[ ] client/backwards_compatibility_test.go
[ ] client/client_pkcs11_test.go
[ ] client/client_root_validation_test.go
[ ] client/client_update_test.go
[ ] signer/client/signer_trust_test.go
[ ] signer/keydbstore/keydbstore_test.go
[ ] signer/api/api_test.go
[ ] signer/api/rpc_api_test.go
[ ] trustmanager/keyfilestore_test.go
[ ] trustmanager/yubikey/yubikeystore_test.go
[ ] trustmanager/x509filestore_test.go
[ ] trustmanager/x509utils_test.go
[ ] trustmanager/x509store_test.go
[ ] trustmanager/filestore_test.go
[ ] trustmanager/x509memstore_test.go
[ ] certs/certs_test.go
[ ] vendor/github.com/Sirupsen/logrus/hooks/sentry/sentry_test.go
[ ] vendor/github.com/Sirupsen/logrus/hooks/papertrail/papertrail_test.go
[ ] vendor/github.com/Sirupsen/logrus/hooks/bugsnag/bugsnag_test.go
[ ] vendor/github.com/Sirupsen/logrus/hooks/airbrake/airbrake_test.go
[ ] vendor/github.com/Sirupsen/logrus/hooks/syslog/syslog_test.go
[ ] vendor/github.com/Sirupsen/logrus/formatters/logstash/logstash_test.go
[ ] vendor/github.com/Sirupsen/logrus/text_formatter_test.go
[ ] vendor/github.com/Sirupsen/logrus/logrus_test.go
[...]

PR #659, opened 2016-04-04
Use require instead of assert for passphrase package

riyazdf:
Replaces just for the `passphrase` package, part of #628 


cyli:
LGTM pending CI!  :)

PR #660, opened 2016-04-04
Use require instead of assert for cmd

riyazdf:
Updates all test assertions to use require in the `cmd` package, part of #628 


cyli:
:+1: Awesome, thanks for doing these!  LGTM!

endophage:
LGTM

PR #661, opened 2016-04-04
Update to use require for cryptoservice package

riyazdf:
Using require instead of assert for the cryptoservice package, part of #628


cyli:
LGTM pending CI, thank you so much for fixing these!  It looks like the doc builder had a spurious error.  I tried to restart the job, but if that doesn't work maybe we can re-push?

riyazdf:
@cyli thanks for the retry, but I think it failed again so I've repushed!

endophage:
LGTM.

PR #664, opened 2016-04-05
Finish converting assert to require in tuf/* packages

cyli:
Part of #628


riyazdf:
LGTM!   Codecov is fine even though the hook is complaining

PR #666, opened 2016-04-06
Embed SignedCommon in Root, Snapshot, and Timestamp

cyli:
Since it's already embedded in `Targets`.

Will be useful for me for https://github.com/docker/notary/pull/580

### DIFF
--- a/tuf/data/root_test.go
+++ b/tuf/data/root_test.go
@@ -27,9 +27,11 @@ func (e errorSerializer) Unmarshal([]byte, interface{}) error {
 func validRootTemplate() *SignedRoot {
 	return &SignedRoot{
 		Signed: Root{
-			Type:    "Root",
-			Version: 1,
-			Expires: time.Now(),
+			SignedCommon: SignedCommon{
+				Type:    TUFTypes[CanonicalRootRole],
+				Version: 1,
+				Expires: time.Now(),
+			},
 			Keys: Keys{
 				"key1":  NewPublicKey(RSAKey, []byte("key1")),
 				"key2":  NewPublicKey(RSAKey, []byte("key2")),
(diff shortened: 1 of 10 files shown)


tuf/data/root_test.go
```diff
@@ -52,7 +54,7 @@ func validRootTemplate() *SignedRoot {
 }
 
 func TestRootToSignedMarshalsSignedPortionWithCanonicalJSON(t *testing.T) {
-	r := SignedRoot{Signed: Root{Type: "root", Version: 2, Expires: time.Now()}}
+	r := SignedRoot{Signed: Root{SignedCommon: SignedCommon{Type: "root", Version: 2, Expires: time.Now()}}}
```
riyazdf: nit: should this be uppercase `"Root"`?  It shouldn't affect the behavior of this test but it seems like we should keep this consistent with what is expected (there are a few other instances in this file)

tuf/data/timestamp.go
```diff
@@ -64,9 +61,11 @@ func NewTimestamp(snapshot *Signed) (*SignedTimestamp, error) {
 	return &SignedTimestamp{
 		Signatures: make([]Signature, 0),
 		Signed: Timestamp{
-			Type:    TUFTypes["timestamp"],
-			Version: 0,
-			Expires: DefaultExpires("timestamp"),
+			SignedCommon: SignedCommon{
+				Type:    TUFTypes["timestamp"],
```
riyazdf: nit: can we use the `CanonicalTimestampRole` constant here and below for `Expires:`?

riyazdf:
LGTM after comments!

tuf/data/root_test.go
cyli: Yep, thanks for noticing that!

tuf/data/timestamp.go
cyli: Yes, good idea.

endophage:
LGTM!

PR #667, opened 2016-04-06
Convert the main and development compose files to be v2

cyli:


cyli:
Oh :|  I guess circleci's docker engine is not 1.10.0 yet

PR #668, opened 2016-04-06
NotaryRepository.Update now just returns an error, rather than a client and error

cyli:
Since we don't actually use the TUF client anymore in `NotaryRepository` except during `Update`.

Addresses https://github.com/docker/notary/pull/580#discussion_r55722929.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -386,7 +386,7 @@ func (r *NotaryRepository) RemoveTarget(targetName string, roles ...string) erro
 // subtree and also the "targets/x" subtree, as we will defer parsing it until
 // we explicitly reach it in our iteration of the provided list of roles.
 func (r *NotaryRepository) ListTargets(roles ...string) ([]*TargetWithRole, error) {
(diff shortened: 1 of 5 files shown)


client/client.go
```diff
@@ -460,9 +459,8 @@ func (r *NotaryRepository) GetTargetByName(name string, roles ...string) (*Targe
 			}
 			return nil
 		}
-		err = r.tufRepo.WalkTargets(name, role, getTargetVisitorFunc, skipRoles...)
 		// Check that we didn't error, and that we assigned to our target
-		if err == nil && foundTarget {
+		if err := r.tufRepo.WalkTargets(name, role, getTargetVisitorFunc, skipRoles...); err == nil && foundTarget {
```
riyazdf: thanks for the additional cleanup! 👍

riyazdf:
thanks for following up on that comment in #580, LGTM!

endophage:
LGTM

PR #669, opened 2016-04-07
Kind of cleanup

HuKeping:
Like commit 84f5ed28d2f1eaeb8488c3c02e199a099311fbab upstream.

Move the configuration parsing for notary-server to its own file.

/CC @cyli 

### DIFF
--- a/cmd/notary-server/config.go
+++ b/cmd/notary-server/config.go
@@ -11,6 +11,7 @@ import (
 	_ "github.com/docker/distribution/registry/auth/token"
 	"github.com/docker/go-connections/tlsconfig"
 	"github.com/docker/notary"
+	"github.com/docker/notary/server"
 	"github.com/docker/notary/server/storage"
 	"github.com/docker/notary/signer/client"
 	"github.com/docker/notary/tuf/data"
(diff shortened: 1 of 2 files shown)


riyazdf:
Jenkins, test this please.

riyazdf:
change LGTM -- circle ci seems to be having issues with `go vet` recently so we'll look into getting that green

cyli:
This LGTM!  Apologies, looks like there were maybe some changes in the underlying `golang:1.6.0` image and also some changes in the latest version of oxcross, so the integration test's client container was busted.  Should be fixed though, if you wanted to rebase @HuKeping ?

HuKeping:
rebased @cyli , the error of coverage seems wired.

cyli:
please test this

cyli:
test this please

cyli:
Sorry @HuKeping - the -4.55% was because the Yubikey tests hadn't been run yet.  And I think there's a slight bug in codecov updating the status given the large amount of time difference between when the other tests were run and the yubikey tests being run.  The [code coverage details](https://codecov.io/github/docker/notary?ref=31851edf818722cdf88746689f752109d13d00c1) page says that the code coverage for this PR is at 77.07% though, which is the same as master, so I think we can just merge this anyway.

Thank you for the cleanup!

PR #670, opened 2016-04-07
Do not install go vet on the Dockerfile

cyli:
Not sure what changed, but looks like go vet just comes with the image now and also osxcross breaks unless pinned to a commit.  Copied the OSX cross stuff from docker/docker Dockerfile.

Also bumped the min OSX version to 10.9.


endophage:
I'll leave it to you to work out if godep is required, otherwise LGTM.

riyazdf:
LGTM

PR #673, opened 2016-04-11
Add the buildtags to go list ./... in the Makefile

cyli:
Also, remove `+build pkcs11 ` build tag from the signer.

Thanks to @endophage for noticing that previously, signer tests weren't be run  :(

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ GO_EXC = go
 NOTARYDIR := /go/src/github.com/docker/notary
 
 # check to be sure pkcs11 lib is always imported with a build tag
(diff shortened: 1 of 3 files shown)


endophage:
LGTM pending CI

riyazdf:
noticing a `go vet` error not related to code in this PR, but being flagged since the tags were updated -- seems that we need to update a lock to use a pointer in `notary-signer/main.go`?

cyli:
@riyazdf Thanks!  Changed to pointer :)

riyazdf:
LGTM.  Codecov is fine here with me too, though the reporting shows a slight decrease for some reason

endophage:
Yep LGTM. The codecov diff already existed, we just couldn't see it because the notary-signer package wasn't being tested.

PR #675, opened 2016-04-12
Key ID verification

riyazdf:
Adds verification to `Verify` to ensure that the key ID being used to verify signatures is actually correct against its sha256 hash


endophage:
LGTM

cyli:
LGTM!

PR #676, opened 2016-04-12
Write to file to signal success for make integration

riyazdf:
Ensures that we actually fail when one of the containers in the integration test aborts, since docker-compose may return a 0 exit code when aborting containers.

Relies on https://github.com/docker/notary/pull/674 to be green


cyli:
Thank you for fixing this!  LGTM! Tested this locally and if I make the tests fail I get an exit code of 2 after running `make integration`.

Non-blocking:  wondering if it'd help to pipe `UNATTENDED=yes OSX_VERSION_MIN=10.6 ${OSXCROSS_PATH}/build.sh` to `/dev/null` in `Dockerfile`?  Otherwise the circleci log is tends of thousands of lines and has to be downloaded, which is fine, but it'd be more immediately obvious when looking at circleci to see what goes wrong in the future with our own containers.

riyazdf:
@cyli: great idea, done!

endophage:
No logical reason for test coverage reduction. LGTM

PR #677, opened 2016-04-12
@mtrmac's cleanups, split from root/cert rotation feature

cyli:
@mtrmac fully implemented root cert (public key) rotations, which in #648 was changed to be a full root rotation as per group code review results.

Some of the cleanups are not as relevant to that feature anymore, but are still really nice, so to try to reduce the size of #648 and make it a little smaller, this PR contains some of the cleanups from that PR.

### DIFF
--- a/certs/certs.go
+++ b/certs/certs.go
@@ -4,6 +4,7 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/Sirupsen/logrus"
@@ -34,6 +35,18 @@ type ErrRootRotationFail struct {
 // by either failing to add the new root certificate, or delete the old ones
 func (err ErrRootRotationFail) Error() string {
 	return fmt.Sprintf("could not rotate trust to a new trusted root: %s", err.Reason)
+}
+
+func certIDs(certs []*x509.Certificate) string {
+	ids := make([]string, 0, len(certs))
+	for _, cert := range certs {
+		id, err := trustmanager.FingerprintCert(cert)
+		if err != nil {
+			id = fmt.Sprintf("[Error %s]", err)
+		}
+		ids = append(ids, id)
+	}
+	return strings.Join(ids, ", ")
 }
 
 /*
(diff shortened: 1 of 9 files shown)


certs/certs.go
```diff
@@ -245,14 +266,9 @@ func parseAllCerts(signedRoot *data.SignedRoot) (map[string]*x509.Certificate, m
 
 // certsToRemove returns all the certifificates from oldCerts that aren't present
```
riyazdf: typo: `certificates` (this might also be in the trust pinning PR)

certs/certs.go
```diff
@@ -264,6 +280,14 @@ func certsToRemove(oldCerts, newCerts []*x509.Certificate) map[string]*x509.Cert
 		newCertMap[certID] = struct{}{}
 	}
 
+	// We don't want to "rotate" certificates to an empty set, nor keep old certificates if the
+	// new root does not trust them.  newCerts should come from validRootLeafCerts, which refuses
```
riyazdf: It might be worth bubbling up a mention of this comment to the function doc comment above, so this method isn't called in other contexts

riyazdf:
LGTM after nit, some great new comments and helpers in here 👍

certs/certs.go
cyli: Good idea - added

signer/keydbstore/keydbstore.go
```diff
@@ -179,7 +179,7 @@ func (s *KeyDBStore) RotateKeyPassphrase(name, newPassphraseAlias string) error
 	// Retrieve the GORM private key from the database
 	dbPrivateKey := GormPrivateKey{}
 	if s.db.Where(&GormPrivateKey{KeyID: name}).First(&dbPrivateKey).RecordNotFound() {
```
endophage: just for my sanity, can we rename the `name` parameter to `keyID` because that's what it must be based on this query vs other queries on the `GormPrivateKey` type.

endophage: We can do it in a follow up commit. Looks like `name` is used in a few of the methods.

endophage:
LGTM.

PR #680, opened 2016-04-13
Update some comments

HuKeping:
Commit 1(I'm not very sure about this one since the "100MB" here seems more directly):
If we update the maximum value for "infinite" in the future, it is possible
that we forget to update this comments and thus leads misunderstanding.

Commit 2:
Change to a more readable format.


riyazdf:
Jenkins, test this please

endophage:
LGTM

riyazdf:
LGTM, thanks for clarifying this!

PR #681, opened 2016-04-14
Add note about versioning

riyazdf:
Adding a note to the docs on ensuring that the notary client and server/signer are the same version, also fixed a small typo

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Any security vulnerabilities can be reported to security@docker.com.
 # Getting started with the Notary CLI
 
 Please get the Notary Client CLI binary from [the official releases page](https://github.com/docker/notary/releases) or you can [build one yourself](#building-notary).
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -42,6 +42,7 @@ Any security vulnerabilities can be reported to security@docker.com.
 # Getting started with the Notary CLI
 
 Please get the Notary Client CLI binary from [the official releases page](https://github.com/docker/notary/releases) or you can [build one yourself](#building-notary).
+The version of Notary server and signer should match the Notary CLI to ensure feature compatibility, and all official releases are associated with GitHub tags.
```
cyli: Should we say that the server version should be >= the client version?  Newer versions of the server should also work with older versions of clients.

riyazdf: Yeah, definitely!  Updated.

endophage:
Would be good to include an example: `i.e. if you're using notary 0.2, ensure you are using a server and signer tagged with an equal or higher version than 0.2`

cyli:
LGTM!  Thank you for adding this!!

endophage:
LGTM

PR #682, opened 2016-04-14
minor cleanup of filestore initialization

endophage:


endophage:
@cyli @riyazdf completing the import/export refactor and merging KeyFileStore with KeyMemoryStore is going to be a lot more work than I want and will delay getting started on higher priority stuff. This is a minimal refactor that just cleans up a little duplicate code and starts laying out more clearly the interfaces. Lots more to be done but we can at least merge this.

Will probably have a minor testing drop just because most of the dupe code being deleted was covered so the ratio will get slightly worse.

cyli:
Should we be removing some of the other `SimpleFileStore` code that's not being currently used?  e.g. `RemoveDir` and `ListDir` don't seem to be called anywhere except from tests.  Not sure if they're necessary for the later import/export refactor and merging.

Otherwise, LGTM pending CI!

trustmanager/filestore.go
```diff
@@ -15,40 +15,33 @@ type SimpleFileStore struct {
 	perms   os.FileMode
 }
 
-// NewSimpleFileStore creates a directory with 755 permissions
-func NewSimpleFileStore(baseDir string, fileExt string) (*SimpleFileStore, error) {
+// NewFileStore creates a fully configurable file store
+func NewFileStore(baseDir, fileExt string, perms os.FileMode) (*SimpleFileStore, error) {
```
riyazdf: nit: should this method be unexported, and only be accessible from `NewSimpleFileStore` and `NewPrivateSimpleFileStore`?

riyazdf:
LGTM pending nit comment and further potential cleanup on `SimpleFileStore` code that @cyli has suggested

trustmanager/filestore.go
endophage: As the most general initializer I think it's semantically correct to export it. The other initializers are just helpful wrappers for our most common use cases.

riyazdf: Ok makes sense, this is good as is

riyazdf:
not sure how to understand the codecov drop here, since we deleted so much unused code and only one new line is potentially uncovered.  Still LGTM, merging.

PR #683, opened 2016-04-15
Some root rotation tests and CLI warnings

### LINKED ISSUE #404: Support root key rotation/replacement
… to allow migrating away from known compromised keys, or from suspected-weak algorithms.

(Related: `notary key rotate` currently rotates non-root keys, and #267 is in progress to rotate root certificates. This issue is filed for the underlying root private key pair.)

cyli:
This PR:

- adds tests to ensure that server validation or client update during root rotation do not care about old saved root roles, just whatever root role was previously pinned
- tests that root rotation validation checks that the old root role is satisfied, not that a particular key's signature has to be there based on the previous root's signatures
- warns users that they can't just throw away the root key if they rotate the root key
- adds support for root threshold validation - this causes `signed.VerifyRoot` to go away since then it would require the exact same functionality as `signed.VerifySignatures`

Closes #404 probably.

### DIFF
--- a/client/client_update_test.go
+++ b/client/client_update_test.go
@@ -1321,3 +1321,102 @@ func testUpdateRemoteKeyRotated(t *testing.T, targetsRole string) {
 	require.IsType(t, signed.ErrRoleThreshold{}, err, "expected ErrRoleThreshold when %s: got %s",
 		msg, reflect.TypeOf(err))
 }
+
+// A valid root rotation only cares about the immediately previous old root keys,
+// whether or not there are old root roles, and cares that the role is satisfied
+// (for instance if the old role has 2 keys, either of which can sign, then it
+// doesn't matter which key signs the rotation)
+func TestValidateRootRotationWithOldRole(t *testing.T) {
+	// start with a repo with a root with 2 keys, optionally signing 1
+	_, serverSwizzler := newServerSwizzler(t)
+	ts := readOnlyServer(t, serverSwizzler.MetadataCache, http.StatusNotFound, "docker.com/notary")
+	defer ts.Close()
+
+	repo := newBlankRepo(t, ts.URL)
+	defer os.RemoveAll(repo.baseDir)
+
+	pubKey1, err := testutils.CreateKey(serverSwizzler.CryptoService, "docker.com/notary", data.CanonicalRootRole)
+	require.NoError(t, err)
+	pubKey2, err := testutils.CreateKey(serverSwizzler.CryptoService, "docker.com/notary", data.CanonicalRootRole)
+	require.NoError(t, err)
+
+	// update the root metadata to have 2 keys and threshold 1, and then sign
+	// only with one of the keys
+	rootBytes, err := serverSwizzler.MetadataCache.GetMeta(data.CanonicalRootRole, -1)
+	require.NoError(t, err)
+	signedRoot := data.SignedRoot{}
+	require.NoError(t, json.Unmarshal(rootBytes, &signedRoot))
+
+	// save the old role to prove that it is not needed for client updates
+	oldVersion := fmt.Sprintf("%v.%v", data.CanonicalRootRole, signedRoot.Signed.Version)
+	signedRoot.Signed.Roles[oldVersion] = &data.RootRole{
+		Threshold: 1,
+		KeyIDs:    signedRoot.Signed.Roles[data.CanonicalRootRole].KeyIDs,
+	}
+	signedRoot.Signed.Version++
+	signedRoot.Signed.Keys[pubKey1.ID()] = pubKey1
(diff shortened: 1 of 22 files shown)


cmd/notary/keys.go
```diff
@@ -424,6 +426,28 @@ func (k *keyCommander) keysRotate(cmd *cobra.Command, args []string) error {
...
+
+		readIn := bufio.NewReader(k.input)
+
+		result, err := readIn.ReadBytes('\n')
+		if err != nil {
+			return err
+		}
+		yesno := strings.ToLower(strings.TrimSpace(string(result)))
+
+		if yesno != "yes" && yesno != "y" {
```
riyazdf: nit: can we use `askConfirm()` here instead?

cyli: Good idea!  Will make this change here and in `removeKeyInteractively`

server/handlers/validation_test.go
```diff
@@ -359,50 +359,161 @@ func TestValidateRootGetCurrentRootBroken(t *testing.T) {
...
+	require.NoError(t, repo.ReplaceBaseKeys(data.CanonicalRootRole, newRootKey2))
+
+	r, err = repo.SignRoot(data.DefaultExpires(data.CanonicalRootRole))
+	require.NoError(t, err)
+	require.Len(t, r.Signatures, 3)
+	// delete all signatures except the one which didn't sign the previous role
+	// and the new key
+	sigs := make([]data.Signature, 0, 2)
+	for _, sig := range repo.Root.Signatures {
+		if sig.KeyID == newRootID || sig.KeyID == newRootID2 {
```
riyazdf: this is an awesome test!  Just a small nit: it might be worth leaving a comment on what `newRootKey` and `newRootKey2` map to, just for future readability

client/client_update_test.go
```diff
@@ -1321,3 +1321,102 @@ func testUpdateRemoteKeyRotated(t *testing.T, targetsRole string) {
 	require.IsType(t, signed.ErrRoleThreshold{}, err, "expected ErrRoleThreshold when %s: got %s",
 		msg, reflect.TypeOf(err))
 }
+
+// A valid root rotation only cares about the immediately previous old root keys,
+// whether or not there are old root roles, and cares that the role is satisfied
+// (for instance if the old role has 2 keys, either of which can sign, then it
+// doesn't matter which key signs the rotation)
+func TestValidateRootRotationWithOldRole(t *testing.T) {
+	// start with a repo with a root with 2 keys, optionall signing 1
```
riyazdf: typo: `optionally`

server/handlers/validation_test.go
cyli: Good point, thanks!  Maybe I can rename these to `additionalRootKey` and `additionalRootKey2`?

riyazdf: That works!  We could also differentiate the two as `additionalRootKey` and `rotatedRootKey` since we only perform one real rotation?

cyli: 👍

riyazdf:
thanks for adding these, LGTM!

cyli:
Just so we make sure we don't get client issues in the future when we want to add threshold support to the CLI, I'll fix threshold checking for roots now in this PR.

client/client_update_test.go
```diff
@@ -851,13 +851,6 @@ func TestUpdateRootRemoteCorruptedNoLocalCache(t *testing.T) {
 	}
 
 	for _, testData := range waysToMessUpServerRoot() {
-		if testData.desc == "insufficient signatures" {
-			// Currently if we download the root during the bootstrap phase,
-			// we don't check for enough signatures to meet the threshold.  We
```
riyazdf: 👍 glad to see this go!

riyazdf:
thank you for adding the extra validation!  LGTM

cmd/notary/keys.go
```diff
@@ -424,6 +425,19 @@ func (k *keyCommander) keysRotate(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+
+	if rotateKeyRole == data.CanonicalRootRole {
+		cmd.Println("Warning: you are about to rotate your root key.\n" +
+			"You will still need your old key to sign future root changes\n" +
+			"so that clients who may not be behind can update.\n" +
```
endophage: This doesn't read right: "clients who may not be behind"

Not sure what the right wording is as it doesn't matter how up to date a client is, when a root rotation happens, if the old key isn't used to sign, everyone has to manually reset.

cyli: Good point, this is pretty unclear.  I was trying to imply that you'd need it forever, not just for this rotation - I'll rephrase.

endophage:
LGTM

PR #684, opened 2016-04-15
Add by hash command

### LINKED ISSUE #649: add target to notary by hash
Allow for notary to add a target solely by hash and target name (no input file needed).

We can make heavy re-use of https://github.com/riyazdf/notary/tree/conman, though we might want to make it a flag for the base add command: `notary add --hash <sha256>`

riyazdf:
Introduces a `notary addhash` command.  This command requires a GUN, target name, target byte size, and a sha256 and/or sha512 hash

Closes #649

### DIFF
--- a/cmd/notary/main_test.go
+++ b/cmd/notary/main_test.go
@@ -104,6 +104,7 @@ var exampleValidCommands = []string{
 	"status repo",
 	"publish repo",
 	"add repo v1 somefile",
+	"addhash repo targetv1 --sha256 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 10",
 	"verify repo v1",
 	"key list",
 	"key rotate repo snapshot",
--- a/cmd/notary/main_test.go
+++ b/cmd/notary/main_test.go
@@ -98,6 +98,55 @@ func TestRemoteServerCommandLineFlagOverridesConfig(t *testing.T) {
 	require.Equal(t, "http://overridden", getRemoteTrustServer(config))
 }
 
+// invalid commands for `notary addhash`
+func TestInvalidAddHashCommands(t *testing.T) {
+	tempDir := tempDirWithConfig(t, `{"remote_server": {"url": "https://myserver"}}`)
+	defer os.RemoveAll(tempDir)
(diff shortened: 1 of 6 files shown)


cyli:
LGTM!  Thanks for adding this!

cmd/notary/main_test.go
```diff
@@ -103,6 +103,7 @@ var exampleValidCommands = []string{
 	"status repo",
 	"publish repo",
 	"add repo v1 somefile",
+	"addhash repo targetv1 --sha256 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 10",
```
endophage: this works too but so that we don't have to use flags, what I was suggesting was something like distribution uses: `sha256:aaaaaa...` then we'd just split on `:`. Means a user can add any hash, even one we don't necessarily recognize which they can validate in their own client integration.

cmd/notary/tuf.go
```diff
@@ -100,6 +111,79 @@ func (t *tufCommander) AddToCommand(cmd *cobra.Command) {
...
+			return err
+		}
+		targetHash[notary.SHA512] = sha512Hash
+	}
+
+	// Manually construct the target with the given byte size and hashes
+	target := &notaryclient.Target{Name: targetName, Hashes: targetHash, Length: targetInt64Len}
+
+	// If roles is empty, we default to adding to targets
+	if err = nRepo.AddTarget(target, t.roles...); err != nil {
```
endophage: Should check `len(targetHash) > 0`. Could add the check inside `AddTarget` to make it as widely applicable as possible.

riyazdf: great idea, added!

cmd/notary/main_test.go
riyazdf: I'll look into what distribution does -- however, I'm not sure if we want this flexibility because a user typo (ex: `sha255:..`) could ruin the target entirely, and it wouldn't be noticeable until verifying after a publish.  It seems like a more usable flow to be strict up front and ensure that hashes are correctly inputted?

endophage: Yeah I agree. I was thinking about this more after commenting as this method will reduce typos.

endophage:
LGTM

PR #686, opened 2016-04-19
Update go in dockerfiles to go 1.6.1

cyli:
Since we want that security update which could affect HTTPS servers.

Also, the golang:1.6.1-alpine image includes the `no-pic` patch so we should no longer need to pass the `-fno-PIC` flag

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.6.0
+FROM golang:1.6.1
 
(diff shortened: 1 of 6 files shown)


riyazdf:
thanks for bumping the version!  Should we also update the version in the README?

riyazdf:
LGTM!

endophage:
LGTM

PR #687, opened 2016-04-20
Remove a redundant ErrChecksumMismatch error in tuf/client

cyli:
Since we already have a ErrMismatchedChecksum in tuf/data/errors.

Also, have the CheckHashes function take a role name so that the
`ErrMismatchedChecksum` error can include the role name.

This is split out from #580 to make it a little smaller.

### DIFF
--- a/tuf/data/types.go
+++ b/tuf/data/types.go
@@ -4,6 +4,7 @@ import (
 	"crypto/sha256"
 	"crypto/sha512"
 	"crypto/subtle"
+	"encoding/hex"
 	"fmt"
 	"hash"
 	"io"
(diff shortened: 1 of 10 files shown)


tuf/data/types.go
```diff
@@ -4,6 +4,8 @@ import (
 	"crypto/sha256"
 	"crypto/sha512"
 	"crypto/subtle"
+	"encoding/hex"
+
```
riyazdf: super-non-blocking nit: remove whitespace

riyazdf:
LGTM!

endophage:
LGTM but I guess we want to merge #688 first

cyli:
I think this should probably be mergeable without #688 - they're pretty much independent - this would bump #688's coverage up.

PR #688, opened 2016-04-20
Remove tuf/client tests

cyli:
Since a lot of them are redundant with tests in `client_update_test`, just add a bit more to `client_update_test` and delete `tuf/client/client_test` entirely. 

This will help with #580 since `tuf/client/client.go` changes a lot, and would require a lot of refactoring, and this means that all that can be skipped.

Also wanted to do this in its own PR because I want to make sure that coverage doesn't drop by doing so.


cyli:
Looks like the drop in coverage is due to the following:

1.  [`tuf/client/client.go` in this PR](https://codecov.io/github/docker/notary/tuf/client/client.go?ref=406b68e162c09758965a5028aa028e3175728f5a) vs [`tuf/client/client.go` in master](https://codecov.io/github/docker/notary/tuf/client/client.go?ref=4d85f964bc7809ef8db28e23f01a35c886ac08e2): the checks for if the local timestamp or snapshot is missing in the local TUF repo, because then nothing else can be downloaded - this should be moot in the builder, but I will add those tests back to `tuf/client/client_test.go` here

2. [`tuf/client/errors.go` in this PR](https://codecov.io/github/docker/notary/tuf/client/errors.go?ref=406b68e162c09758965a5028aa028e3175728f5a) vs [`tuf/client/errors.go` in master](https://codecov.io/github/docker/notary/tuf/client/errors.go?ref=4d85f964bc7809ef8db28e23f01a35c886ac08e2): `ErrChecksumMismatch` is now uncovered, which should be resolved by #687 

3. [`tuf/store/memorystore.go` in this PR](https://codecov.io/github/docker/notary/tuf/store/memorystore.go?ref=406b68e162c09758965a5028aa028e3175728f5a) vs [`tuf/store/memorystore.go` in master](https://codecov.io/github/docker/notary/tuf/store/memorystore.go?ref=4d85f964bc7809ef8db28e23f01a35c886ac08e2): the TUF client tests were the main thing using the memory store, some more lines of `GetMeta` are uncovered now - ideally we can convert the client tests to use the memory store in the future instead of the httpstore, but in the meantime, I'll add tests for the memory store too.

client/client_update_test.go
```diff
@@ -1309,6 +1311,25 @@ func testUpdateRemoteKeyRotated(t *testing.T, targetsRole string) {
...
 
+// Helper function that takes a signedRoot, and signs it with the provided keys and only these keys.
+// Then serializes this to bytes and updates the swizzler with it, and updates the snapshot and
+// timestamp too so that the update won't fail due to a checksum mismatch.
+func signSerializeAndUpdateRoot(t *testing.T, signedRoot data.SignedRoot,
+	serverSwizzler *testutils.MetadataSwizzler, keys []data.PublicKey) {
+
+	signedObj, err := signedRoot.ToSigned()
+	require.NoError(t, err)
+	// sign with the first two keys
```
riyazdf: I might be missing something, but what are the `first two` keys referring to here?  Just the provided keys?

riyazdf: Ah actually it seems that this ties in with the next test.  It seems like this comment could still be updated to 'sign with the provided keys'

tuf/client/client_test.go
```diff
@@ -1,508 +1,17 @@
...
-			},
-		},
-	}
-
-	repo.Snapshot = &snap
-
-	err = client.downloadTargets("targets")
-	require.IsType(t, ErrChecksumMismatch{}, err)
-}
-
```
riyazdf: \o/ red!

riyazdf:
this is awesome!  LGTM after nit

client/client_update_test.go
cyli: Nice catch, thanks!  It was copied from the previous test, so the comment is no longer applicable to t he general case.

endophage:
LGTM after @riyazdf comment on a comment :-P

PR #693, opened 2016-04-22
Remove certstore

riyazdf:
Use the previous root on disk to directly validate new roots on update, parsing out its own certificates in from the root metadata's `Keys` section instead of consulting a certificate store.

As a consequence, the certificate store in the notary repository (also in `trustmanager` x509 utils) and `notary cert` are no more :wave:


client/client_root_validation_test.go
```diff
@@ -34,45 +33,12 @@ func validateRootSuccessfully(t *testing.T, rootType string) {
...
-	// Initialize is supposed to have created new certificate for this repository
-	// Lets check for it and store it for later use
-	allCerts := repo.CertStore.GetCertificates()
-	require.Len(t, allCerts, 1)
-
 	fakeServerData(t, repo, mux, keys)
 
 	//
-	// Test TOFUS logic. We remove all certs and expect a new one to be added after ListTargets
+	// Test TOFUS logic.
```
cyli: Would it make sense to just delete this file?  I think we already have tests for listing targets on an empty repo in `client_tests.go`, although maybe not with an RSA key for root.

client/client_test.go
```diff
@@ -2774,15 +2747,6 @@ func rootRoleCertID(t *testing.T, repo *NotaryRepository) string {
...
-	require.NoError(t, err)
-	require.Len(t, certs, 1)
-	id, err := trustmanager.FingerprintCert(certs[0])
-	require.NoError(t, err)
-	require.Equal(t, certID, id)
-}
-
 func TestRotateRootKey(t *testing.T) {
 	ts := fullTestServer(t)
 	defer ts.Close()
```
cyli: Sorry, it's collapsed so I can't comment on the line, but there is a [very long comment](https://github.com/docker/notary/blob/remove-certstore/client/client_test.go#L2792) in this test that was added by #648 about `NotaryRepository.Update`'s handling of certs, which could probably be deleted?

cmd/notary/integration_test.go
```diff
@@ -1346,98 +1346,6 @@ func TestClientKeyImportExportAllRoles(t *testing.T) {
 	}
 }
 
-func assertNumCerts(t *testing.T, tempDir string, expectedNum int) []string {
```
cyli: \o/ So excited about all these deletes!

client/client_test.go
riyazdf: Great catch, I'll update that

client/client_root_validation_test.go
riyazdf: Yeah I'm fine with deleting this, will update

trustpinning/certs.go
```diff
@@ -100,30 +100,23 @@ func ValidateRoot(certStore trustmanager.X509Store, root *data.Signed, gun strin
...
+	// Retrieve all the trusted certificates from our previous root
+	// Note that we do not validate expiries here since our originally trusted root might have expired certs
+	allTrustedLeafCerts, allTrustedIntCerts := parseAllCerts(prevRoot)
+	trustedLeafCerts, err := validRootLeafCerts(allTrustedLeafCerts, gun, false)
 	// If we have certificates that match this specific GUN, let's make sure to
 	// use them first to validate that this new root is valid.
-	if len(trustedCerts) != 0 {
-		logrus.Debugf("found %d valid root certificates for %s: %s", len(trustedCerts), gun,
-			prettyFormatCertIDs(trustedCerts))
+	if len(trustedLeafCerts) != 0 {
```
cyli: This should not happen unless something malicious happened on disk, if there are no valid certificates in the previous root, it seems like we should error because the previous root provided is invalid.  Would it make sense to change this check to `if prevRoot != nil` and move the above `prevRoot` parsing into this if case?

trustpinning/certs.go
```diff
@@ -86,7 +86,7 @@ We shall call this: TOFUS.
 
 Validation failure at any step will result in an ErrValidationFailed error.
```
cyli: Apologies it's collapsed, but I think there was a reference to cert stores above in this comment :)

trustpinning/certs.go
riyazdf: Yeah agreed that this shouldn't happen if `prevRoot` is not nil, so I can move the parsing logic into this case 👍

trustpinning/certs.go
riyazdf: Ah yes, good catch -- I'll update the paragraph about the first step

trustpinning/certs.go
```diff
@@ -86,7 +86,7 @@ We shall call this: TOFUS.
...
 */
-func ValidateRoot(certStore trustmanager.X509Store, root *data.Signed, gun string, trustPinning TrustPinConfig) error {
+func ValidateRoot(prevRoot *data.SignedRoot, root *data.Signed, gun string, trustPinning TrustPinConfig) error {
```
cyli: Non-blocking: Wondering if it'd make sense to have ValidateRoot return the completely unmarshalled root?  `NotaryRepository.bootstrapClient`, for instance, calls `data.RootFromSigned` anyway after ValidateRoot succeeds.  This would also be useful for the builder.

trustpinning/certs_test.go
```diff
@@ -700,27 +556,53 @@ func filestoreWithTwoCerts(t *testing.T, gun, keyAlg string) (
...
-	return tempBaseDir, certStore, cryptoService, certificates
+	return tempBaseDir, cryptoService, certificates
 }
 
 func testValidateSuccessfulRootRotation(t *testing.T, keyAlg, rootKeyType string) {
 	// The gun to test
 	gun := "docker.com/notary"
 
-	tempBaseDir, certStore, cs, certificates := filestoreWithTwoCerts(t, gun, keyAlg)
+	tempBaseDir, cs, certificates := generateTwoCerts(t, gun, keyAlg)
```
cyli: Non-blocking - since we don't care about getting the certs out anymore, `testutils.CreateKey` will generate x509 keys for root keys.  Not sure, but I don't think that would cause an import cycle.

trustpinning/certs.go
riyazdf: I'm fine with ValidateRoot returning the root it just validated as a `*data.SignedRoot`, and that'll be great if it can help the builder

trustpinning/certs_test.go
riyazdf: agreed, and it doesn't cause an import cycle!  Done

trustpinning/certs.go
```diff
@@ -86,53 +87,48 @@ We shall call this: TOFUS.
...
+		allTrustedLeafCerts, allTrustedIntCerts := parseAllCerts(prevRoot)
+		trustedLeafCerts, err := validRootLeafCerts(allTrustedLeafCerts, gun, false)
+
+		// Use the certificates we found in the previous root for the GUN to verify its signatures
+		// This could potentially be an empty set, in which case we will fail to verify
+		logrus.Debugf("found %d valid root leaf certificates for %s: %s", len(trustedLeafCerts), gun,
+			prettyFormatCertIDs(trustedLeafCerts))
 		err = signed.VerifySignatures(
-			root, data.BaseRole{Keys: trustmanager.CertsToKeys(trustedCerts, allIntCerts), Threshold: 1})
+			root, data.BaseRole{Keys: trustmanager.CertsToKeys(trustedLeafCerts, allTrustedIntCerts), Threshold: 1})
```
cyli: Should the threshold be the root threshold of the previous root?

trustpinning/certs_test.go
```diff
@@ -666,11 +551,10 @@ func TestValidateSuccessfulRootRotation(t *testing.T) {
 	}
 }
 
-// Generates a X509Store in a temporary directory and returns the
-// store and certificates for two keys which have been added to the keystore.
+// Generates certificates for two keys which have been added to the keystore.
 // Also returns the temporary directory so it can be cleaned up.
-func filestoreWithTwoCerts(t *testing.T, gun, keyAlg string) (
-	string, trustmanager.X509Store, *cryptoservice.CryptoService, []*x509.Certificate) {
+func generateTwoCerts(t *testing.T, gun, keyAlg string) (
```
cyli: This function can probably be removed since we're using `CreateKey` now.

riyazdf: you're right, more red! \o/

trustpinning/certs_test.go
```diff
@@ -700,28 +576,51 @@ func filestoreWithTwoCerts(t *testing.T, gun, keyAlg string) (
...
-	tempBaseDir, certStore, cs, certificates := filestoreWithTwoCerts(t, gun, keyAlg)
+	tempBaseDir, err := ioutil.TempDir("", "notary-test-")
 	defer os.RemoveAll(tempBaseDir)
-	origRootCert := certificates[0]
-	replRootCert := certificates[1]
+	require.NoError(t, err, "failed to create a temporary directory: %s", err)
 
-	// Add the old root cert part of trustedCertificates
-	certStore.AddCert(origRootCert)
+	fileKeyStore, err := trustmanager.NewKeyFileStore(tempBaseDir, passphraseRetriever)
```
cyli: Totally non-blocking:  Just wondering if we could use a memory file store here?  Only reason would be minimally less IO during the tests.  Won't have much impact at all, though, since time taken for the client tests pretty much eclipse everything else.

riyazdf: good idea, it'll clean up the temp dir stuff as well 👍

trustpinning/certs.go
riyazdf: Agreed, will update.

cyli:
LGTM!  All the red is so beautiful!  :D  Thanks for adding this!

Just wanted to add a caveat for other reviewers that while it'd be good if we checked that previous root was validly signed (not just that it was valid metadata) before we used it for trust pinning, that functionality should  be happening in the builder, and this PR already improves upon existing behavior (where we do even less validation of the previous root object, and we don't check for signing thresholds).

trustpinning/certs.go
```diff
@@ -86,53 +87,55 @@ We shall call this: TOFUS.
...
-		logrus.Debugf("found %d valid root certificates for %s: %s", len(trustedCerts), gun,
-			prettyFormatCertIDs(trustedCerts))
+
 		err = signed.VerifySignatures(
-			root, data.BaseRole{Keys: trustmanager.CertsToKeys(trustedCerts, allIntCerts), Threshold: 1})
+			root, data.BaseRole{Keys: trustmanager.CertsToKeys(trustedLeafCerts, allTrustedIntCerts), Threshold: prevRootRoleData.Threshold})
 		if err != nil {
 			logrus.Debugf("failed to verify TUF data for: %s, %v", gun, err)
-			return &ErrValidationFail{Reason: "failed to validate data with current trusted certificates"}
+			return nil, &ErrValidationFail{Reason: "failed to validate data with current trusted certificates"}
```
cyli: Non-blocking - I was wondering if it'd make sense to use `ErrRootRotationFail` here.  It may not, since it seems to be a one-off error case, and maybe we don't want to do so for that reason.

If we don't want to use `ErrRootRotationFail` though, we can probably remove it as it was only ever used when rotating the certs in the cert store.

riyazdf: I'm fine either way for this, though it seems having the more explicit `ErrRootRotationFail` might be better for error type-checking and display.

tuf/testutils/repo.go
```diff
@@ -36,7 +36,13 @@ func CreateKey(cs signed.CryptoService, gun, role string) (data.PublicKey, error
 		if err != nil {
 			return nil, err
 		}
-		key = data.NewECDSAx509PublicKey(trustmanager.CertToPEM(cert))
+		// Keep the x509 key type consistent with the key's algorithm
+		if keyAlgorithm == data.RSAKey {
+			key = data.NewRSAx509PublicKey(trustmanager.CertToPEM(cert))
+		} else {
```
endophage: shouldn't assume if it's not RSA it is always ECDSA. Maybe use a switch with a default: error

riyazdf: I think it's technically impossible since the call to `cs.Create()` will reject any different type, but agreed that it would be nicer.

trustpinning/certs_test.go
```diff
@@ -889,3 +784,11 @@ func generateExpiredTestingCertificate(rootKey data.PrivateKey, gun string) (*x5
 	startTime := time.Now().AddDate(-10, 0, 0)
 	return cryptoservice.GenerateCertificate(rootKey, gun, startTime, startTime.AddDate(1, 0, 0))
 }
+
+func generateRootKeyIDs(r *data.SignedRoot) {
```
cyli: Non-blocking - a comment here to explain why this function is necessary (to generate data for an unexported cached field for equality testing) since I'm sure I will totally forget why this is here in a day or two :)

riyazdf: Totally agree, done!

cyli: Thanks!

cyli:
LGTM!  Thanks for all your hard work getting this awesome batch of deletions in!

endophage:
LGTM!

PR #695, opened 2016-04-25
Add const for max size -1 for GetMeta

### LINKED ISSUE #690: Add a constant for `GetMeta` with arbitrary size
We pass `-1` to `GetMeta` when we don't know the metadata size - as @riyazdf mentions in https://github.com/docker/notary/pull/580#discussion_r60469243, we should probably use a constant.  Maybe in `tuf/store/interface`?

riyazdf:
Closes #690.

### DIFF
--- a/tuf/store/filestore.go
+++ b/tuf/store/filestore.go
@@ -39,7 +39,7 @@ func (f *FilesystemStore) getPath(name string) string {
 }
 
 // GetMeta returns the meta for the given name (a role) up to size bytes
-// If size is -1, this corresponds to "infinite," but we cut off at the
+// If size is "MaxSize", this corresponds to "infinite," but we cut off at a
 // predefined threshold "notary.MaxDownloadSize".
 func (f *FilesystemStore) GetMeta(name string, size int64) ([]byte, error) {
 	meta, err := ioutil.ReadFile(f.getPath(name))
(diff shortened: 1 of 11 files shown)


tuf/store/filestore.go
```diff
@@ -49,7 +49,7 @@ func (f *FilesystemStore) GetMeta(name string, size int64) ([]byte, error) {
 		}
 		return nil, err
 	}
-	if size == -1 {
+	if size == MaxSize {
```
HuKeping: Since we have already change `-1` to the constant `MaxSize `, could we also update the comments for this function.

The old comments is try to explain what's the `-1` stand for, and I think it might be out of date now.

cyli:
LGTM!  Thanks for fixing this!

HuKeping:
+1 for the new constant "NoSizeLimit", more clear that "MaxSize"

endophage:
I'm wondering, if we're going to set this value, why not just use `notary.MaxDownloadSize` everywhere we use `store.NoSizeLimit`? Seems redundant to have to 2 constants when one is just a front for the other (i.e. `NoSizeLimit` _always_ gets translated into `MaxDownloadSize`)

riyazdf:
@endophage you're right that it's always currently translated, though I think we had discussed potentially setting different limits for different types of stores.  If we'd prefer to just keep everything consistent between stores as is, I'm fine with removing the `-1` functionality altogether

endophage:
I don't remember discussing it but that's a good point about possibly setting different limits for different stores. I can't think of any immediate use for it but let's leave as you have it and if we still don't have a use case the next time we touch this code we'll switch it out then.

endophage:
LGTM

PR #708, opened 2016-05-02
Allow for configurable notary signer db name, check allowed backends

riyazdf:
Some small cleanup to allow for a configurable notary signer db name in Rethinkdb (this matches the notary server configuration).

Also added a check to `setUpCryptoservices` to actually check against allowed backends and error out if an invalid backend is specified. 

I've also run this against the Rethinkdb integration test with successful results.


cyli:
LGTM!  Thanks for adding this in!

endophage:
LGTM

PR #719, opened 2016-05-05
Add client key and cert to rethink client tls config

riyazdf:
To connect to RethinkDB over TLS from notary server/signer

### DIFF
--- a/cmd/notary-server/config.go
+++ b/cmd/notary-server/config.go
@@ -89,7 +89,12 @@ func getStore(configuration *viper.Viper, hRegister healthRegister) (
 		if err != nil {
 			return nil, err
 		}
-		sess, err = rethinkdb.Connection(storeConfig.CA, storeConfig.Source)
+		tlsOpts := tlsconfig.Options{
+			CAFile:   storeConfig.CA,
+			CertFile: storeConfig.Cert,
+			KeyFile:  storeConfig.Key,
+		}
+		sess, err = rethinkdb.Connection(tlsOpts, storeConfig.Source)
 		if err != nil {
 			return nil, fmt.Errorf("Error starting %s driver: %s", backend, err.Error())
 		}
--- a/utils/configuration.go
+++ b/utils/configuration.go
@@ -27,7 +27,9 @@ type Storage struct {
 type RethinkDBStorage struct {
 	Storage
 	CA     string
+	Cert   string
 	DBName string
+	Key    string
 }
 
 // GetPathRelativeToConfig gets a configuration key which is a path, and if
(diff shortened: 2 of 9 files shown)


utils/configuration_test.go
```diff
@@ -269,6 +277,22 @@ func TestParseRethinkStorageDBStoreEmptyCA(t *testing.T) {
 	require.Contains(t, err.Error(), "cowardly refusal to connect to rethinkdb without a CA cert")
 }
 
+// ParseRethinkDBStorage will require a client cert or key to connect to rethink databases
```
cyli: nit: "cert and key"

utils/configuration.go
```diff
@@ -136,6 +140,11 @@ func ParseRethinkDBStorage(configuration *viper.Viper) (*RethinkDBStorage, error
 			"cowardly refusal to connect to %s without a CA cert",
 			store.Backend,
 		)
+	case store.Cert == "" && store.Key == "":
+		return nil, fmt.Errorf(
+			"cowardly refusal to connect to %s without a client cert or key",
```
cyli: nit: "cert and key"

utils/configuration.go
```diff
@@ -136,6 +140,11 @@ func ParseRethinkDBStorage(configuration *viper.Viper) (*RethinkDBStorage, error
 			"cowardly refusal to connect to %s without a CA cert",
 			store.Backend,
 		)
+	case store.Cert == "" && store.Key == "":
```
cyli: Should this be an `||`, since we need both for mutual TLS?

cmd/notary-server/config.go
```diff
@@ -89,7 +89,12 @@ func getStore(configuration *viper.Viper, hRegister healthRegister) (
 		if err != nil {
 			return nil, err
 		}
-		sess, err = rethinkdb.Connection(storeConfig.CA, storeConfig.Source)
+		tlsOpts := tlsconfig.Options{
```
cyli: Non-blocking:  I think if we add the cert and key to `TestGetStoreRethinkDBStoreConnectionFails` in `cmd/notary/main_test.go`, that might cover some of these uncovered lines, since the test is failing to parse due to not having a cert and key configured here.

cyli:
Tested locally with `buildscripts/integrationtest.sh development.rethink.yml` and it all passes!  LGTM aside from the small issues above!

utils/configuration.go
riyazdf: great catch, I'll fix this and associated comments!

cmd/notary-server/config.go
riyazdf: Good idea, done!

cyli:
👍 Awesome, thank you!

endophage:
LGTM

PR #724, opened 2016-05-06
Feature: Introduce some flags on notary verify

HuKeping:
There is no reason to give out all the payload especially when it is
huge engough, it will definitely ruin the console.

### DIFF
--- a/cmd/notary/util.go
+++ b/cmd/notary/util.go
@@ -1,0 +1,51 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+)
+
+// getPayload is a helper function to get the content used to be verified
+// either from an existing file or STDIN.
+func getPayload(t *tufCommander) ([]byte, error) {
+
+	// Reads from the given file
+	if t.input != "" {
+		// Please be noticed that ReadFile will cut off the size if it was over 1e9.
+		// Thus, if the size of the file exceeds 1GB, the over part will not be
+		// loaded into the buffer.
+		payload, err := ioutil.ReadFile(t.input)
+		if err != nil {
+			return nil, err
+		}
+		return payload, nil
+	}
+
+	// Reads all of the data on STDIN
+	payload, err := ioutil.ReadAll(os.Stdin)
+	if err != nil {
+		return nil, fmt.Errorf("Error reading content from STDIN: %v", err)
+	}
+	return payload, nil
+}
+
+// feedback is a helper function to print the payload to a file or STDOUT or keep quiet
+// due to the value of flag "quiet" and "output".
+func feedback(t *tufCommander, payload []byte) error {
+	// We only get here when everythings goes well, since the flag "quiet" was
+	// provided, we output nothing but just return.
+	if t.quiet {
+		return nil
+	}
+
+	// Flag "quiet" was not "true", that's why we get here.
+	if t.output != "" {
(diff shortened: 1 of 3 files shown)


riyazdf:
I'm not sure if we'd like a drastic change in behavior, since this change would break this use case:
`curl <binary> | notary verify | sh`

How about adding a short-form mode instead?  We could even have the short-form mode only return success or failure via error codes, and not even print any output.

endophage:
I think the desired functionality here probably deserves a whole set of flags on `notary verify` focussed on verifying a file already on disk, like `--input`, `--output`, and `--quiet` (`-i`, `-o`, and `-q` for short forms) which would make it align with a lot of other *nix tools that use stdin/stdout by default and take flags for other I/O options.

For interaction between `--output` and `--quiet` I'd expect either an error saying both flags were provided and only one can be used. Or `--quiet` takes priority and `--output` is ignored.

HuKeping:
Hmm,  a set of flags is good for that, will change.

cyli:
jenkins, test this please

endophage:
Love it. LGTM! Thank you!

cmd/notary/util.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	// Flag "quiet" was not "true", that's why we get here.
+	if t.output != "" {
+		if err := ioutil.WriteFile(t.output, payload, 0644); err != nil {
+			return err
+		}
+	}
+
+	os.Stdout.Write(payload)
```
riyazdf: it seems that if we provide a value to the `output` flag, we write both to the file as well as stdout.
To be consistent with the flag description, we can remove the error-checking in the `t.output != ""` case and just `return ioutil.WriteFile(t.output, payload, 0644)`.

I'm also ok with leaving this behavior of writing to both the output file and stdout, if we update the flag description

cmd/notary/util.go
```diff
@@ -0,0 +1,51 @@
...
+	"os"
+)
+
+// getPayload is a helper function to get the content used to be verified
+// either from an existing file or STDIN.
+func getPayload(t *tufCommander) ([]byte, error) {
+
+	// Reads from the given file
+	if t.input != "" {
+		// Please be noticed that ReadFile will cut off the size if it was over 1e9.
```
riyazdf: nit: `Please note that`

cmd/notary/util.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Error reading content from STDIN: %v", err)
+	}
+	return payload, nil
+}
+
+// feedback is a helper function to print the payload to a file or STDOUT or keep quiet
+// due to the value of flag "quiet" and "output".
+func feedback(t *tufCommander, payload []byte) error {
+	// We only get here when everythings goes well, since the flag "quiet" was
```
riyazdf: nit typo: `everything`

cmd/notary/util.go
endophage: Good catch, I totally missed that! Yeah, we should only write to the file. That will be the expected behaviour for anyone familiar with *nix

HuKeping: Thanks for noticing that! Will change.

HuKeping:
Updated

riyazdf:
Jenkins, test this please.

riyazdf:
thanks for the update, this is a great contribution!  LGTM!

PR #725, opened 2016-05-06
Tiny rework

HuKeping:
I think it's OK to use the ErrOffline{} directly, we needn't to wrap it here.

The "err" is more likely be used as "err, xxx := foo()"

### DIFF
--- a/tuf/store/offlinestore.go
+++ b/tuf/store/offlinestore.go
@@ -11,43 +11,41 @@ func (e ErrOffline) Error() string {
 	return "client is offline"
 }
 
-var err = ErrOffline{}
-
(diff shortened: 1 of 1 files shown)


cyli:
Jenkins, test this please

endophage:
It's a minor difference but the existing version is cheaper on memory and garbage collection because we only create one actual instance of the error that never has to be garbage collected.

It's somewhat idiomatic for errors with static state to be predeclared and we could probably be a bit better about it https://golang.org/pkg/net/http/#pkg-variables

HuKeping:
Agreed

endophage:
I think we're going to stick with the existing behaviour here.

PR #726, opened 2016-05-07
Increase timeout for RethinkDB (default is 0), remove extra wait

riyazdf:
The default timeout for rethinkdb when calling `term.Wait()` is [0 seconds](https://rethinkdb.com/api/javascript/wait/), this bumps it to 60 seconds -- I'd be fine making it shorter as well, but we're checking across all replicas to be ready in this scenario by default.

Also remove extra `.Wait()` after db creation

Tested using the rethinkdb integration test.


cyli:
LGTM

endophage:
LGTM

PR #729, opened 2016-05-10
Server gun prefix requirement feature

### LINKED ISSUE #608: create a server config to require a prefix for all keys
If this config is specified, any request to get or delete a GUN without the prefix should provide a specific error that says something like "this name is invalid because it lacks the prefix X".

cyli:
Fixes #608

### DIFF
--- a/cmd/notary-server/config.go
+++ b/cmd/notary-server/config.go
@@ -3,7 +3,9 @@ package main
 import (
 	"crypto/tls"
 	"fmt"
+	"path"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/Sirupsen/logrus"
@@ -21,8 +23,20 @@ import (
 	_ "github.com/go-sql-driver/mysql"
 	"github.com/spf13/viper"
 	"golang.org/x/net/context"
-	"gopkg.in/dancannon/gorethink.v2"
+	gorethink "gopkg.in/dancannon/gorethink.v2"
 )
+
+// gets the required gun prefixes accepted by this server
+func getRequiredGunPrefixes(configuration *viper.Viper) ([]string, error) {
+	prefixes := configuration.GetStringSlice("repositories.gun_prefixes")
+	for _, prefix := range prefixes {
+		p := path.Clean(strings.TrimSpace(prefix))
+		if p+"/" != prefix || strings.HasPrefix(p, "/") || strings.HasPrefix(p, "..") {
+			return nil, fmt.Errorf("Invalid GUN prefix %s", prefix)
+		}
+	}
+	return prefixes, nil
+}
 
 // get the address for the HTTP server, and parses the optional TLS
 // configuration for the server - if no TLS configuration is specified,
@@ -217,6 +231,11 @@ func parseServerConfig(configFilePath string, hRegister healthRegister) (context
 	}
 	logrus.SetLevel(lvl)
 
+	prefixes, err := getRequiredGunPrefixes(config)
+	if err != nil {
+		return nil, server.Config{}, err
+	}
+
 	// parse bugsnag config
 	bugsnagConf, err := utils.ParseBugsnag(config)
 	if err != nil {
@@ -252,6 +271,7 @@ func parseServerConfig(configFilePath string, hRegister healthRegister) (context
 		Trust:                        trust,
 		AuthMethod:                   config.GetString("auth.type"),
 		AuthOpts:                     config.Get("auth.options"),
+		RepoPrefixes:                 prefixes,
 		CurrentCacheControlConfig:    currentCache,
 		ConsistentCacheControlConfig: consistentCache,
 	}, nil
--- a/cmd/notary-server/config.go
+++ b/cmd/notary-server/config.go
@@ -32,7 +32,7 @@ func getRequiredGunPrefixes(configuration *viper.Viper) ([]string, error) {
 	for _, prefix := range prefixes {
 		p := path.Clean(strings.TrimSpace(prefix))
 		if p+"/" != prefix || strings.HasPrefix(p, "/") || strings.HasPrefix(p, "..") {
-			return nil, fmt.Errorf("Invalid GUN prefix %s", prefix)
+			return nil, fmt.Errorf("invalid GUN prefix %s", prefix)
 		}
 	}
 	return prefixes, nil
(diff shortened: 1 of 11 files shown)


endophage:
This implementation works for me. If we want the behavior to be more complicated in the future we can discuss it when we have a better idea what it should be.

endophage:
Let's add the additional config entry to direct how to use the prefixes to this PR, but only implement `"restrict"` (i.e. the current proposed behaviour).

Another PR can then implement `"trim"` (or whatever we decide to call the other option).

cyli:
@endophage:  Hmm... do you want the user to actually have to specify the `"restrict"` behavior?  Since that is the default behavior, not sure if we need another configuration option yet.

When we implement `"trim"`, can that be a boolean option that gets added to the `"repositories"` section?  Do we ever want to trim without restricting?

cyli:
Update: after IRL discussion, we determined that variable behavior can be added later with a `behavior` config option or `trim` flag or some such.  Right now we're leaving it off because it is the default behavior, and we're not sure if we want others yet.

server/errors/errors.go
```diff
@@ -83,8 +83,14 @@ var (
 	ErrNoKeyAlgorithm = errcode.Register(errGroup, errcode.ErrorDescriptor{
 		Value:          "NO_KEYALGORITHM",
 		Message:        "The server does not have a key algorithm configured.",
-		Description:    "No key algorihtm has been configured for the server and it has been asked to perform an operation that requires generation.",
+		Description:    "No key algorithm has been configured for the server and it has been asked to perform an operation that requires generation.",
```
riyazdf: thanks for the fix 👍

cmd/notary-server/config.go
```diff
@@ -21,9 +23,21 @@ import (
...
+	gorethink "gopkg.in/dancannon/gorethink.v2"
 )
 
+// gets the required gun prefixes accepted by this server
+func getRequiredGunPrefixes(configuration *viper.Viper) ([]string, error) {
+	prefixes := configuration.GetStringSlice("repositories.gun_prefixes")
+	for _, prefix := range prefixes {
+		p := path.Clean(strings.TrimSpace(prefix))
+		if p+"/" != prefix || strings.HasPrefix(p, "/") || strings.HasPrefix(p, "..") {
+			return nil, fmt.Errorf("Invalid GUN prefix %s", prefix)
```
riyazdf: nit: do we want start the error message with lowercase?

cmd/notary-server/config.go
```diff
@@ -21,9 +23,21 @@ import (
...
-	"gopkg.in/dancannon/gorethink.v2"
+	gorethink "gopkg.in/dancannon/gorethink.v2"
 )
 
+// gets the required gun prefixes accepted by this server
+func getRequiredGunPrefixes(configuration *viper.Viper) ([]string, error) {
+	prefixes := configuration.GetStringSlice("repositories.gun_prefixes")
+	for _, prefix := range prefixes {
+		p := path.Clean(strings.TrimSpace(prefix))
+		if p+"/" != prefix || strings.HasPrefix(p, "/") || strings.HasPrefix(p, "..") {
```
riyazdf: I'm wondering if we also want to enforce the leading `/` and `..` cases for GUNs if the config does not specify any prefixes?

Unless I'm missing something, notary could still create trust data for GUNs like `/gun` and `../../gun` if we never get into this loop

riyazdf: There's some pretty varied behavior for this currently in master, not sure if we want to separate fixing this into another PR.  Tested with notary from master and compose setup:

leading `/`:
```
$ bin/notary -c cmd/notary/config.json init /test/
Root key found, using: 2930402da13b3062f9f59cb3bc6f93d04ae0caaf5778318cf2de8a2eb123367b
Enter passphrase for root key with ID 2930402:
Enter passphrase for new targets key with ID 8ec423d (/test):
Repeat passphrase for new targets key with ID 8ec423d (/test):

* fatal: unable to reach trust server at this time: 301.
```

one level `..` escape:
```
$ bin/notary -c cmd/notary/config.json init ../test
Root key found, using: 2930402da13b3062f9f59cb3bc6f93d04ae0caaf5778318cf2de8a2eb123367b
Enter passphrase for root key with ID 2930402:
Enter passphrase for new targets key with ID 8fa72b4 (../test):
Repeat passphrase for new targets key with ID 8fa72b4 (../test):

* fatal: timestamp key trust data unavailable.  Has a notary repository been initialized?
```

multiple `..` escape
```
$ bin/notary -c cmd/notary/config.json init ../../test
Root key found, using: 2930402da13b3062f9f59cb3bc6f93d04ae0caaf5778318cf2de8a2eb123367b
Enter passphrase for root key with ID 2930402:
Enter passphrase for new targets key with ID c36d72b (../../test):
Repeat passphrase for new targets key with ID c36d72b (../../test):

* fatal: failed to add key to filestore: path outside file store
```

cyli: I was wondering that too - https://github.com/docker/notary/issues/742 tracks the issue, but we can totally check it here.  I think @endophage is testing out whether the Go http server automatically collapses the URL for you too (e.g. whether the Go server collapses `http://localhost:8080/foo/../../../bar` to `http://localhost:8080/bar`)

cyli: We should probably also validate the GUN when `init`-ing using the CLI, because the user can probably mess up their local file system.

cyli: Oh duh sorry they can't pollute on init :).  I guess not super important then, but may just be nice to have a nicer error message about invalid GUN though

cmd/notary-server/config.go
cyli: Yes!  Thanks, forgot that

cmd/notary-server/config.go
riyazdf: totally agreed, a separate PR makes sense

riyazdf:
LGTM after small caps nit 👍

cmd/notary-server/config.go
endophage: So those filepath escapes get cleaned out and a 301 returned at the Go http server level, we never see them. Would be good to add checks to the CLI in a different PR. Because our server code won't even see them, doesn't make sense to put that in this PR.

endophage:
LGTM

PR #744, opened 2016-05-12
CheckTargetByName for checking if all roles signed a specific target with the same name and hash

riyazdf:
Add function to client library to check if a set of roles have all signed a particular target with the same name and hash.  Note that if no roles are passed, we default to just checking the targets role (following behavior in `GetTargetByName` and `ListTargets`).

This function will check the specified roles and also its child roles if it cannot find the target.
Ex: if `targets/releases` signs `latest` but `targets` does not, calling `CheckTargetByName("latest", "targets")` will return true.

I kept the error return value in case we wanted it for debugging (if there's an unexpected error during an `Update` call vs. during a walk, or mismatching hashes), but we could remove it if we would prefer to do so.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -462,6 +462,36 @@ func (r *NotaryRepository) GetTargetByName(name string, roles ...string) (*Targe
 
 }
 
+// CheckTargetByName returns true if all passed-in roles have signed the specified target.
+// If no roles are passed, it uses the targets role.  Note that child roles are not checked.
+func (r *NotaryRepository) CheckTargetByName(name string, roles ...string) (bool, error) {
+	if err := r.Update(false); err != nil {
+		return false, err
+	}
+
+	if len(roles) == 0 {
+		roles = append(roles, data.CanonicalTargetsRole)
+	}
+	var foundTarget bool
+	for _, role := range roles {
+		// Define a visitor function to find the specified target only at the specified role
+		checkTargetVisitorFunc := func(tgt *data.SignedTargets, validRole data.DelegationRole) interface{} {
+			// we never continue once we've found the role we're interested in
+			if tgt == nil {
+				return tuf.StopWalk{}
+			}
+			// We might have found the target, so check against the target name and signal up the closure
+			_, foundTarget = tgt.Signed.Targets[name]
+			return tuf.StopWalk{}
+		}
+		// Check that we didn't error, and that we found the target for our role
+		if err := r.tufRepo.WalkTargets(name, role, checkTargetVisitorFunc); err != nil || !foundTarget {
+			return false, fmt.Errorf("could not find target for role %s: additional error output %v", role, err)
+		}
+	}
+	return true, nil
+}
+
 // GetChangelist returns the list of the repository's unpublished changes
 func (r *NotaryRepository) GetChangelist() (changelist.Changelist, error) {
 	changelistDir := filepath.Join(r.tufRepoPath, "changelist")
--- a/client/client.go
+++ b/client/client.go
@@ -463,7 +463,8 @@ func (r *NotaryRepository) GetTargetByName(name string, roles ...string) (*Targe
 }
 
 // CheckTargetByName returns true if all passed-in roles have signed the specified target.
-// If no roles are passed, it uses the targets role.  Note that child roles are not checked.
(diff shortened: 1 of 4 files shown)


client/client.go
```diff
@@ -462,6 +462,46 @@ func (r *NotaryRepository) GetTargetByName(name string, roles ...string) (*Targe
...
 
+// CheckTargetByName returns true if all passed-in roles have signed the specified target with the same hash.
+// If no roles are passed, it uses the targets role.  Note that child roles are considered if
+// we can't find the target in the specified role.
+func (r *NotaryRepository) CheckTargetByName(name string, roles ...string) (bool, error) {
+	if err := r.Update(false); err != nil {
+		return false, err
+	}
+
+	if len(roles) == 0 {
```
cyli: Totally non-blocking - wondering it's worth filtering out roles that are parents of other roles.  E.g. if you specify `targets` and also `targets/releases`, there is no need to search `targets` because if it exists in `targets/releases` we count that as also being in `targets`.

Unless we want to error because maybe the user means that someone signed it specifically into `targets/releases` and also signed it into `targets` itself?

cyli:
Thanks for adding this @riyazdf!  Could we add the following two test cases also:

1.  One of the targets has an invalid hash (missing any supported checksums)
2.  Could we test that one target has only sha256 and one has a sha512 hash?

client/client.go
riyazdf: Originally I had this function only check the specified role without looking in child roles so it was clearer, do you think that's better?

I'd be ok with filtering out roles (though it would add some complexity), but I'm not sure if we want to make the case you described above as an error case -- it seems that if a target is signed by `targets/releases`, it's been delegated to sign by the `targets` role anyway?

riyazdf:
@cyli: good test ideas, I'll add them and ping once I've finished 👍

client/client.go
cyli: Just checking the specified role seems clearer, but @endophage is correct that maybe you do care about the child roles.  

Since you can basically configure your delegation hierarchy however you want though, it's hard to guess what the exact API someone would need would be (maybe they care about checking only the specified roles). 

I'm fine with this function as is because it suits our purposes for now, but it does make some assumptions consistent with only one type of workflow, whereas a lot of the other code in client is more general.

riyazdf: No problem, I agree that it's good to discuss and think this through -- I'm totally open to changing this function if we prefer different behavior if it improves the API :)

riyazdf:
@cyli: added the tests as unit tests, since adding invalid target hashes to the repo is pretty much impossible with walker and builder (which is awesome!!)

tuf/data/types_test.go
```diff
@@ -176,3 +177,46 @@ func TestCheckValidHashStructures(t *testing.T) {
...
+	hashes1 := make(Hashes)
+	hashes2 := make(Hashes)
+
+	// Expected to fail since there are no checksums at all
+	err = CompareMultiHashes(hashes1, hashes2)
+	require.Error(t, err)
+
+	// Expected to fail even though the checksum of sha384 is valid,
+	// because we haven't provided a supported hash algorithm yet (ex: sha256).
+	hashes1["sha384"], err = hex.DecodeString("64becc3c23843942b1040ffd4743d1368d988ddf046d17d448a6e199c02c3044b425a680112b399d4dbe9b35b7ccc989")
```
cyli: Could we add one more test for if the sha384 hashes don't match?  The hashes will still evaluate equal to each other, because the sha384 hash isn't supported (should that be the case?)

riyazdf: I think the behavior should mimic `CheckHashes` -- currently that function only checks against hash algorithms that we support, but I'm open to changing that behavior (could be this PR or a separate one)

cyli: I think I'm fine with leaving the behavior as is to match `CheckHashes`.

My original thought was that since this is checking equivalency, it'd make sense that if both targets specified some other unsupported hash, that they should be the same, but you're right that it doesn't necessarily make sense since someone might be using the hash structure to store data for so me other random feature that we don't support.

cyli:
This has been superceeded by the merged #746 and #747

PR #746, opened 2016-05-13
Get all Target and Role data for a given target name

riyazdf:
Similar to #744 but more generic -- allows for retrieving all `Target` and `Role` data for a specified target name, in the form of a list of `TargetWithRole` structs


cyli:
LGTM!  The hash equivalency check from #744 might also be useful for those using this function, but I'm fine with that being a separate PR if you prefer.  :)  Thanks for adding this!  This seems more general purpose.

endophage:
LGTM pending CI

PR #747, opened 2016-05-14
Utility function for comparing multihash

riyazdf:
Separating out the multihash comparison from #744 into its own PR, as suggested by @cyli 

This PR adds a multihash comparison utility function for comparing supported hashes.  Note that like `CheckHashes`, unsupported hashes are not compared at all.

### DIFF
--- a/tuf/data/types.go
+++ b/tuf/data/types.go
@@ -161,6 +161,44 @@ func CheckHashes(payload []byte, name string, hashes Hashes) error {
 	return nil
 }
 
+// CompareMultiHashes verifies that the two Hashes passed in can represent the same data.
+// This means that both maps must have at least one key defined for which they map, and no conflicts.
+// Note that we only check supported hashes, like in CheckHashes
+func CompareMultiHashes(hashes1, hashes2 Hashes) error {
+	// First check if the two hash structures are valid
+	if err := CheckValidHashStructures(hashes1); err != nil {
+		return err
+	}
+	if err := CheckValidHashStructures(hashes2); err != nil {
+		return err
+	}
+	// Check if they have at least one matching hash, and no conflicts
+	cnt := 0
+	for _, hashAlg := range NotaryDefaultHashes {
+		hash1, ok := hashes1[hashAlg]
+		if !ok {
+			continue
+		}
+
+		hash2, ok := hashes2[hashAlg]
+		if !ok {
+			continue
+		}
+
+		if subtle.ConstantTimeCompare(hash1[:], hash2[:]) == 0 {
+			return fmt.Errorf("mismatched %s checksum", hashAlg)
+		}
+		// If we reached here, we had a match
+		cnt++
+	}
+
+	if cnt == 0 {
+		return fmt.Errorf("at least one supported and matching hash needed")
+	}
+
+	return nil
+}
+
 // CheckValidHashStructures returns an error, or nil, depending on whether
 // the content of the hashes is valid or not.
 func CheckValidHashStructures(hashes Hashes) error {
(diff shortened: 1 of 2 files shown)


tuf/data/types.go
```diff
@@ -161,6 +161,44 @@ func CheckHashes(payload []byte, name string, hashes Hashes) error {
...
+	// First check if the two hash structures are valid
+	if err := CheckValidHashStructures(hashes1); err != nil {
+		return err
+	}
+	if err := CheckValidHashStructures(hashes2); err != nil {
+		return err
+	}
+	// Check if they have at least one matching hash, and no conflicts
+	cnt := 0
+	for _, hashAlg := range []string{notary.SHA256, notary.SHA512} {
```
HuKeping: we have a `NotaryDefaultHashes`

https://github.com/docker/notary/pull/747/files#diff-de0c95529058d827a3e5143442b9de1cL125

riyazdf: ah yes, I was looking for that but I was looking in `const.go` instead of here, I'll use that 👍

tuf/data/types.go
```diff
@@ -161,6 +161,44 @@ func CheckHashes(payload []byte, name string, hashes Hashes) error {
...
+	// First check if the two hash structures are valid
+	if err := CheckValidHashStructures(hashes1); err != nil {
+		return err
+	}
+	if err := CheckValidHashStructures(hashes2); err != nil {
+		return err
+	}
+	// Check if they have at least one matching hash, and no conflicts
+	cnt := 0
+	for _, hashAlg := range NotaryDefaultHashes {
```
endophage: I wouldn't iterate default hashes in this instance because we're not regenerating them so we don't care if we recognize the algorithm or not. Instead iterate the entries in either one of the Hashes maps. The pass criteria should be to check the intersection of hash algorithms present all match in value.

riyazdf: I'm fine with changing this since we generate these Hashes maps ourselves in most cases, and are strict on the algorithms you can provide to `notary addhash` -- I'm concerned in the future that we would check over arbitrary keys that aren't necessarily hash functions.  I agree with changing it for the current code, though.

tuf/data/types.go
```diff
@@ -161,6 +161,41 @@ func CheckHashes(payload []byte, name string, hashes Hashes) error {
...
+	if err := CheckValidHashStructures(hashes1); err != nil {
+		return err
+	}
+	if err := CheckValidHashStructures(hashes2); err != nil {
+		return err
+	}
+	// Check if they have at least one matching hash, and no conflicts
+	cnt := 0
+	for hashAlg := range hashes1 {
+		hash1 := hashes1[hashAlg]
```
cyli: nonblocking:  can we use the 2-val assignment when ranging over maps rather than do an extra assignment here?

tuf/data/types.go
```diff
@@ -161,6 +161,41 @@ func CheckHashes(payload []byte, name string, hashes Hashes) error {
...
+		hash2, ok := hashes2[hashAlg]
+		if !ok {
+			continue
+		}
+
+		if subtle.ConstantTimeCompare(hash1[:], hash2[:]) == 0 {
+			return fmt.Errorf("mismatched %s checksum", hashAlg)
+		}
+		// If we reached here, we had a match
+		cnt++
```
cyli: Should we only increment if the hash algorithm is a supported algorithm?  If not, should we change the error message below?

tuf/data/types.go
riyazdf: great idea, done!

tuf/data/types.go
riyazdf: To keep consistent with @endophage's idea, I've just changed the error message

cyli: Oh wait, I'm dumb, sorry, `CheckValidHashStructures` already checks to see if there at least one supported hash.  The error message change is still good though, because if it lacked both sha256 and sha512, it would have error-ed before this point, so this error message is more appropriate.

cyli:
LGTM!  Thanks for pulling this out and adding it!

endophage:
LGTM

PR #748, opened 2016-05-14
Feature: increase/decrease logging level without restarting

### LINKED ISSUE #633: Hot configuration reload
Hi there,

Recently, I'm working on the hot configuration reload on some of our projects. I wonder if Notary need this feature as well?

One scenario is: when in production, the log should not be set as `debug`, but when something bad happens, the Ops may want some more detail via log on the debug level. Since they couldn't stop the running process, they wish to have the feature of hot reloading.

HuKeping:
Part of #633 

As discussed with the team of notary, it might not make sense to completely reload the configuration yet. 

With this patch set, we can send SIGUSR1/SIGUSR2 to increase/decrease the logging level of notary-server.

### DIFF
--- a/cmd/notary-server/config.go
+++ b/cmd/notary-server/config.go
@@ -3,11 +3,11 @@ package main
 import (
 	"crypto/tls"
 	"fmt"
+	"os"
+	"os/signal"
 	"path"
 	"strconv"
 	"strings"
-	"os"
-	"os/signal"
 	"syscall"
 	"time"
 
--- a/cmd/notary-server/config.go
+++ b/cmd/notary-server/config.go
@@ -283,44 +283,30 @@ func parseServerConfig(configFilePath string, hRegister healthRegister) (context
 	}, nil
 }
 
-func setupConfigReloadTrap(oldConfig *viper.Viper, configFile string, envPrefix string) {
+func setupSignalTrap() {
 	c := make(chan os.Signal, 1)
 	signal.Notify(c, notary.NotarySupportedSignals...)
 	go func() {
 		for {
-			reloadConfig(<-c, configFile, oldConfig, envPrefix)
+			signalHandle(<-c)
 		}
 	}()
 }
 
-// reloadConfig was designed to do all the dirty work that needed to reload
-// the config for notary-server and make the entrance-ReloadConfiguration()
-// for hot configuration reload do some generic things.
-func reloadConfig(sig os.Signal, configFile string, oldConfig *viper.Viper, envPrefix string) {
-	reload := func(config *viper.Viper) {
-		switch sig {
-		case syscall.SIGUSR1:
-			if err := utils.AdjustLogLevel(true); err != nil {
-				fmt.Printf("Attempt to increase log level failed, will remain at %s level, error: %s", logrus.GetLevel(), err)
-				return
-			}
-		case syscall.SIGUSR2:
-			if err := utils.AdjustLogLevel(false); err != nil {
-				fmt.Printf("Attempt to decrease log level failed, will remain at %s level, error: %s", logrus.GetLevel(), err)
-				return
-			}
-		case syscall.SIGHUP:
-			// We use SIGHUP to allow people to reset the logging level as their wish.
-			lvl, err := utils.ParseLogLevel(config, logrus.ErrorLevel)
-			if err != nil {
-				fmt.Printf("Attempt to reset log level failed, will remain at %s level, error: %s", logrus.GetLevel(), err)
-				return
-			}
-			logrus.SetLevel(lvl)
+// signalHandle will increase/descrease the logging level via the signal we get.
+func signalHandle(sig os.Signal) {
+	switch sig {
+	case syscall.SIGUSR1:
+		if err := utils.AdjustLogLevel(true); err != nil {
+			fmt.Printf("Attempt to increase log level failed, will remain at %s level, error: %s", logrus.GetLevel(), err)
+			return
 		}
-
-		fmt.Println("Successfully setting log level to ", logrus.GetLevel())
+	case syscall.SIGUSR2:
+		if err := utils.AdjustLogLevel(false); err != nil {
+			fmt.Printf("Attempt to decrease log level failed, will remain at %s level, error: %s", logrus.GetLevel(), err)
+			return
+		}
 	}
 
-	utils.ReloadConfiguration(oldConfig, configFile, envPrefix, reload)
+	fmt.Println("Successfully setting log level to ", logrus.GetLevel())
 }
--- a/cmd/notary-server/config.go
+++ b/cmd/notary-server/config.go
@@ -309,6 +309,14 @@ func reloadConfig(sig os.Signal, configFile string, oldConfig *viper.Viper, envP
 				fmt.Println("Decrease log level failed, will use the old one, error: ", err)
 				return
 			}
+		case syscall.SIGHUP:
+			// We use SIGHUP to allow people to reset the logging level as their wish.
+			lvl, err := utils.ParseLogLevel(config, logrus.ErrorLevel)
+			if err != nil {
+				fmt.Println("Parse log level failed, will use the old one, error: ", err)
+				return
+			}
+			logrus.SetLevel(lvl)
 		}
 
 		fmt.Println("Successfully setting log level to ", logrus.GetLevel())
--- a/cmd/notary-server/config.go
+++ b/cmd/notary-server/config.go
@@ -297,12 +297,12 @@ func signalHandle(sig os.Signal) {
 	switch sig {
 	case syscall.SIGUSR1:
 		if err := utils.AdjustLogLevel(true); err != nil {
-			fmt.Printf("Attempt to increase log level failed, will remain at %s level, error: %s", logrus.GetLevel(), err)
+			fmt.Printf("Attempt to increase log level failed, will remain at %s level, error: %s\n", logrus.GetLevel(), err)
 			return
 		}
 	case syscall.SIGUSR2:
 		if err := utils.AdjustLogLevel(false); err != nil {
-			fmt.Printf("Attempt to decrease log level failed, will remain at %s level, error: %s", logrus.GetLevel(), err)
+			fmt.Printf("Attempt to decrease log level failed, will remain at %s level, error: %s\n", logrus.GetLevel(), err)
 			return
 		}
 	}
--- a/cmd/notary-server/config.go
+++ b/cmd/notary-server/config.go
@@ -6,6 +6,9 @@ import (
 	"path"
 	"strconv"
 	"strings"
+	"os"
+	"os/signal"
+	"syscall"
 	"time"
 
 	"github.com/Sirupsen/logrus"
@@ -225,6 +228,7 @@ func parseServerConfig(configFilePath string, hRegister healthRegister) (context
 	}
 
 	ctx := context.Background()
+	ctx = context.WithValue(ctx, "config", config)
 
 	// default is error level
 	lvl, err := utils.ParseLogLevel(config, logrus.ErrorLevel)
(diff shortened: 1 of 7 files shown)


cmd/notary-server/config.go
```diff
@@ -256,3 +260,34 @@ func parseServerConfig(configFilePath string, hRegister healthRegister) (context
...
+		}
+	}()
+}
+
+func reloadConfig(sig os.Signal, configFile string, oldConfig *viper.Viper, envPrefix string) {
+	reload := func(config *viper.Viper) {
+		switch sig {
+		case syscall.SIGUSR1:
+			if err := utils.AdjustLogLevel(true); err != nil {
+				fmt.Println("Increase log level failed, will use the old one, error: ", err)
```
HuKeping: Why using `fmt.Print` not `logrus` here and below is because it will not log out anything if the log level is too low, `Panic` for example and we can't just use `logrus.Fatalf()` for it will terminate the whole process.

utils/configuration_test.go
```diff
@@ -485,3 +485,74 @@ func TestParseViperWithValidFile(t *testing.T) {
 
 	require.Equal(t, "debug", v.GetString("logging.level"))
 }
+
+func TestAdjustLogLevel(t *testing.T) {
```
afalko: I'll let this slide in this review, but you need to decompose this into a data-provider like model. You are repeating yourself all over the place in this function.

cyli:
jenkins, test this please

NathanMcCauley:
This functionality should be covered somewhere in the docs, so folks can know about it when reading through documentation.

cmd/notary-server/config.go
```diff
@@ -256,3 +260,45 @@ func parseServerConfig(configFilePath string, hRegister healthRegister) (context
...
+		for {
+			reloadConfig(<-c, configFile, oldConfig, envPrefix)
+		}
+	}()
+}
+
+// reloadConfig was designed to do all the dirty work that needed to reload
+// the config for notary-server and make the entrance-ReloadConfiguration()
+// for hot configuration reload do some generic things.
+func reloadConfig(sig os.Signal, configFile string, oldConfig *viper.Viper, envPrefix string) {
```
NathanMcCauley: This function doesn't seem to have any test coverage. @HuKeping could you add some tests for this?

cmd/notary-server/config.go
```diff
@@ -256,3 +260,45 @@ func parseServerConfig(configFilePath string, hRegister healthRegister) (context
...
+				fmt.Println("Increase log level failed, will use the old one, error: ", err)
+				return
+			}
+		case syscall.SIGUSR2:
+			if err := utils.AdjustLogLevel(false); err != nil {
+				fmt.Println("Decrease log level failed, will use the old one, error: ", err)
+				return
+			}
+		case syscall.SIGHUP:
+			// We use SIGHUB to allow people to reset the logging level as their wish.
```
NathanMcCauley: s/SIGHUB/SIGHUP/

cmd/notary-server/config.go
HuKeping: Will add.

HuKeping:
@NathanMcCauley  Will add the docs later

cyli:
jenkins, test this please

HuKeping:
docs and test ready @NathanMcCauley

HuKeping:
ping ...

riyazdf:
jenkins, test this please.

cmd/notary-server/config.go
riyazdf: Sounds good, could we also print the log level we're reverting back to?  For example: `Attempt to increase log level failed, will remain at %s level, error: ` (with the old level filling that string).  logrus has a `GetLevel` function we can use

cmd/notary-server/main_test.go
```diff
@@ -413,3 +415,34 @@ func TestSampleConfig(t *testing.T) {
...
+func TestReloadConfig(t *testing.T) {
+	f, err := os.Create("/tmp/testReloadConfig.json")
+	defer os.Remove(f.Name())
+	require.NoError(t, err)
+
+	f.WriteString(`{"logging": {"level": "info"}}`)
+
+	v := viper.New()
+	utils.SetupViper(v, "envPrefix")
+	//v.SetConfigType("json")
```
riyazdf: can we delete this line?

docs/reference/server-config.md
```diff
@@ -359,6 +361,30 @@ Example:
 	</tr>
 </table>
 
+## Hot configuration reload
+We don't support to hot reload everything at present. What we support for now is:
+- increase logging level by signal `SIGUSR1`
+- decrease logging level by signal `SIGUSR2`
+- hot reload logging level from the config file by signal `SIGHUP`
```
riyazdf: nit: can we change all three lines above to read `by signaling` instead of `by signal`?

utils/configuration.go
```diff
@@ -222,3 +222,53 @@ func ParseViper(v *viper.Viper, configFile string) error {
...
+
+	reload(newConfig)
+	return nil
+}
+
+// Reserved for future use, we can use it to validate some specific configs when necessary.
+// For example:
+// - to check if there are conflicts.
+// - to do some sanity check maybe.
+func validateConfiguration(oldConfig, newConfig *viper.Viper) error {
```
riyazdf: could we do some checking for log level here for now?  If the log-level is valid at least?  It also could be helpful to log if the config is the same or different if we're only sending signals.

cmd/notary-server/config.go
HuKeping: Good point

HuKeping:
Drop the code for reloading the whole config file, this patch set plans only to support increase/decrease logging level.

docs/reference/server-config.md
```diff
@@ -359,6 +361,29 @@ Example:
 	</tr>
 </table>
 
+## Hot logging level reload
+We don't support to completely reload the whole configuration yet at present. What we support for now is:
```
riyazdf: Can we change this to: "We don't support completely reloading..."

docs/reference/server-config.md
```diff
@@ -359,6 +361,29 @@ Example:
...
+```
+$ kill -s SIGUSR1 `ps aux | grep "notary-server -config" | grep -v "grep" | awk '{print $2}'`
+```
+
+To decrease logging level
+```
+$ kill -s SIGUSR2 `ps aux | grep "notary-server -config" | grep -v "grep" | awk '{print $2}'`
+```
+
+To reset logging level
```
riyazdf: We should remove this section since we're only using `SIGUSR1` and `SIGUSR2` right?

cmd/notary-server/config.go
```diff
@@ -278,3 +282,31 @@ func parseServerConfig(configFilePath string, hRegister healthRegister) (context
...
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, notary.NotarySupportedSignals...)
+	go func() {
+		for {
+			signalHandle(<-c)
+		}
+	}()
+}
+
+// signalHandle will increase/descrease the logging level via the signal we get.
```
riyazdf: typo: decrease

riyazdf:
jenkins, test this please

docs/reference/server-config.md
```diff
@@ -359,6 +361,29 @@ Example:
...
+## Hot logging level reload
+We don't support to completely reload the whole configuration yet at present. What we support for now is:
+- increase logging level by signaling `SIGUSR1`
+- decrease logging level by signaling `SIGUSR2`
+
+Example:
+
+To increase logging level
+```
+$ kill -s SIGUSR1 `ps aux | grep "notary-server -config" | grep -v "grep" | awk '{print $2}'`
```
cyli: Since in our docs we only give instructions on how to run the server and signer in containers, maybe we should also give an example of how to send a single into a container?

Something like `docker kill  --signal="SIGUSR1" <container_id>`?

cmd/notary-server/config.go
```diff
@@ -278,3 +282,31 @@ func parseServerConfig(configFilePath string, hRegister healthRegister) (context
...
+func signalHandle(sig os.Signal) {
+	switch sig {
+	case syscall.SIGUSR1:
+		if err := utils.AdjustLogLevel(true); err != nil {
+			fmt.Printf("Attempt to increase log level failed, will remain at %s level, error: %s", logrus.GetLevel(), err)
```
cyli: I wonder if these should be printed to stderr instead of stdout, so they don't interfere with any log aggregation (for instance, if someone wanted JSON logs, these might end up being non-JSON and cause an error with logstash).  Alternately, they could be logged via `logrus`.

cmd/notary-server/config.go
```diff
@@ -278,3 +282,31 @@ func parseServerConfig(configFilePath string, hRegister healthRegister) (context
...
+			return
+		}
+	case syscall.SIGUSR2:
+		if err := utils.AdjustLogLevel(false); err != nil {
+			fmt.Printf("Attempt to decrease log level failed, will remain at %s level, error: %s", logrus.GetLevel(), err)
+			return
+		}
+	}
+
+	fmt.Println("Successfully setting log level to ", logrus.GetLevel())
```
cyli: Similarly with this - should this be a `logrus.Debug` or `logrus.Info` call?

docs/reference/server-config.md
HuKeping: Do you think it make sense we should give out the shell command to do this ,like
```
$ kill -s SIGUSR1 `ps aux | grep "notary-server -config" | grep -v "grep" | awk '{print $2}'`
```

cmd/notary-server/config.go
```diff
@@ -225,6 +228,7 @@ func parseServerConfig(configFilePath string, hRegister healthRegister) (context
 	}
 
 	ctx := context.Background()
+	ctx = context.WithValue(ctx, "config", config)
```
HuKeping: Oh I missed this one, will clean it up.

cmd/notary-server/config.go
HuKeping: The reason why I don't use `logrus` here and blow is  from here:
https://github.com/docker/notary/pull/748#discussion-diff-63277219R279

docs/reference/server-config.md
cyli: I wouldn't be opposed to giving both, so long as we give the caveat that the kill+grep command would only work if you're just running the bare binary.  

It also probably works if you are running this on the same machine as your docker host, and if you are running the container exactly as we give it in docker-compose and not with any kind of wrapper startup script or anything.

But the kill+grep command doesn't work for me for instance because I use either docker for mac or docker machine, so giving the `docker kill` command is important I think.

cyli: Actually NM, you are correct, it looks like the `docker kill` process seems to only send it to PID1, which is the `sh -c ./migrations/migrate.sh && notary-server -config=fixtures/server-config.json`  :|  which doesn't seem to do anything.

Exec-ing in also seems to have issues for me?
```
$ docker exec notary_server_1 sh
/go/src/github.com/docker/notary # kill -s SIGUSR1 `ps aux | grep "notary-server -config" | grep -v "grep" | awk '{print $2}'`
sh: invalid number 'root'
sh: invalid number 'root'
/go/src/github.com/docker/notary # ps aux 
PID   USER     TIME   COMMAND
    1 root       0:00 sh -c ./migrations/migrate.sh && notary-server -config=fixtures/server-config.json
  115 root       0:00 notary-server -config=fixtures/server-config.json
  124 root       0:00 sh
  129 root       0:00 sh
  150 root       0:00 ps aux
```

Looks like if you exec in, the first column is the PID.

cmd/notary-server/config.go
cyli: Ah ok.  Hm... I guess we have a bunch of other things (like GRPC connection failures) which may not respect the JSON logging.  Would just be nice to get these to log in the same format as specified by the `-logf` flag, but we can maybe think of something else.

cyli: Also, do we want to add a newline to the `Printf`?  I noticed when sending signals it didn't seem to log anything until the next successful setting, and then I saw logs like:

```
Attempt to increase log level failed, will remain at debug level, error: log level can not be set higher than DebugAttempt to increase log level failed, will remain at debug level, error: log level can not be set higher than DebugAttempt to increase log level failed, will remain at debug level, error: log level can not be set higher than DebugSuccessfully setting log level to  info
```

docs/reference/server-config.md
HuKeping: Haven't noticed this before ;-(

So can we just docs it as `kill -s SIGUSR1 PID`

cyli: @HuKeping Thanks for checking!  I would be fine with documenting it as `kill -s SIGUSR1 PID` or `docker exec -i kill -s SIGUSR1 PID`, so long as we also document how to get the `PID` (e.g. `docker exec -i ps aux`) or something (and mention that it's not necessarily the PID 1 process).

cyli:
jenkins, test this please

docs/reference/server-config.md
HuKeping: OK, I'll update it later.

cyli: Thanks :)

cyli:
jenkins, test this please

HuKeping:
Some error on CI system

cyli:
jenkins, test this please

HuKeping:
rebased

cyli:
jenkins, test this please

HuKeping:
ping ...

cyli:
Sorry for the lack of response @HuKeping.  We are all prepping for DockerCon.  :)  This LGTM, thank you so much for your incredible patience and all your hard work on this feature!

riyazdf:
@HuKeping thank you for reworking this PR and your patience!   LGTM 👍

PR #751, opened 2016-05-18
Fix the re-think compose file so that all the rdb nodes join the same cluster

cyli:
Previously I was running into issues where the server and signer would bootstrap fine, but couldn't connect, so updated the wait options to include "all_replicas_ready", which should be the default already.  But for some reason this seemed to fix the issue for me.  Or it was intermittent, who knows.

But not all the RDB replicas joined the same cluster, when looking at the admin UI - this fixes it for me.


endophage:
LGTM

PR #754, opened 2016-05-23
convert docker-compose.yml and development.yml to compose v2

andrewhsu:
Converted the remaining two docker compose files over to the v2 format as suggested in #753 in hopes that the dns server in docker would resolve the hostnames of the containers instead of going off-box for hostname resolution. Unfortunately, this does not work, so I had to add the `GODEBUG=netdns=cgo` environment variable to the compose files.


andrewhsu:
Ugh..i'll create the PR from a branch from the latest from master.

PR #756, opened 2016-05-24
Run rethink integration tests

cyli:
Bump circle's docker version up to 1.10 so we can use compose v2 files.  Attempting to run rethink tests as part of our integration tests, but still trying to figure out what's going wrong.

We are encountering https://github.com/docker/compose/issues/3432 on CircleCI.


endophage:
LGTM

PR #757, opened 2016-05-25
update distribution dependency

endophage:


cmd/notary-server/main_test.go
```diff
@@ -95,6 +96,13 @@ func TestGetAddrAndTLSConfigWithClientTLS(t *testing.T) {
 	require.NotNil(t, tlsConf.ClientCAs)
 }
 
+func fakeRegisterer(callCount *int) healthRegister {
```
cyli: Awesome, thanks for pulling this out!  👍

cyli:
I think this may need the registry/client package vendor-ed in too - right now I think we just have client/auth and other subpackages

utils/http.go
```diff
@@ -66,12 +65,13 @@ func (root *rootHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		if authCtx, err = root.auth.Authorized(ctx, access...); err != nil {
 			if challenge, ok := err.(auth.Challenge); ok {
 				// Let the challenge write the response.
-				challenge.ServeHTTP(w, r)
+				challenge.SetHeaders(w)
 
-				w.WriteHeader(http.StatusUnauthorized)
+				if err := errcode.ServeJSON(w, errcode.ErrorCodeUnauthorized.WithDetail(access)); err != nil {
```
cyli: If we aren't going to do anything if the error isn't nil, should we just call it without assigning the `err` variable?

cyli:
Other than the above, LGTM!

PR #761, opened 2016-06-01
Update NOTARY_VERSION

riyazdf:
Bump from 0.2 to 0.3 -- I'll also draft a release checklist for us similar to [docker/docker's](https://github.com/docker/docker/blob/4a7bd7eaef0038b006cf23adface84081e5aab3b/project/RELEASE-CHECKLIST.md) for our own flow


cyli:
:ship it:

endophage:
LGTM.

In future I think we can let version bumps go through with one OK. We all know what it should be so if it needs to be bumped, we shouldn't gate it on our normal 2 LGTMs

PR #763, opened 2016-06-02
Error logging for POSTs to notary server

riyazdf:
Adds error-level logging for 500 status codes during an update call to notary server, which can be caught by bugsnag.


cyli:
Can we actually log the errors themselves too for 500's?  Other than that LGTM :)

riyazdf:
@cyli: great idea, done :)

cyli:
LGTM!  Thanks for adding the logging!

PR #770, opened 2016-06-08
Provide json marshalling/unmarshalling for the rethink TUF object models

cyli:
Signed-off-by: Ying Li <ying.li@docker.com>

PR #772, opened 2016-06-08
Add JSON marshalling and unmarshalling specific model types to rethink Table object

cyli:
This is a replacement for #770.


riyazdf:
Can we add failure cases for the json unmarshaling for each table type?  Otherwise, LGTM!

cyli:
@riyazdf Ah yep, thanks, good point.  Updating!

riyazdf:
👍 thanks for updating, LGTM!

PR #777, opened 2016-06-09
Bump rethink version due to vulnerabilites in 2.3.0 image

cyli:
See https://hub.docker.com/r/jlhawn/rethinkdb/tags/2.3.0/

### DIFF
--- a/development.rethink.yml
+++ b/development.rethink.yml
@@ -35,7 +35,7 @@ services:
       depends_on:
         - rdb-proxy
     rdb-01:
-      image: jlhawn/rethinkdb:2.3.0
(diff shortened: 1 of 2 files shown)


riyazdf:
LGTM pending CI!

riyazdf:
Leaving a note from reading the rethink [changelogs](https://github.com/rethinkdb/rethinkdb/releases/tag/v2.3.4): while we won't require any migrations, we must ensure that 2.3.0 rethink nodes aren't mixed with 2.3.4 nodes -- however this shouldn't be an issue for the notary-server and notary-signer setup.

PR #780, opened 2016-06-10
Change target info API from list to map return

riyazdf:
I think it makes more sense from an API perspective to have this function return a mapping from roles to the target data associated with that role for easier existence checking and comparison.


cyli:
LGTM!

PR #785, opened 2016-06-14
Bump the golang-alpine image version due to CVEs

cyli:


riyazdf:
thanks for bumping this!  LGTM

cyli:
:|  jenkins seems brokenish, and codecov is waiting on it, but I think we're fine since these are just for the Dockerfiles.

Issue #792, opened 2016-06-17
go1.7 vet dislikes MemoryFileStore


gtank:
```
$ go version
go version go1.7beta1 linux/amd64
$ make
pkcs11 import was not found anywhere without a build tag, yay
+ clean
+ fmt
+ vet
trustmanager/keyfilestore.go:213: literal copies lock value from *memStore: trustmanager.MemoryFileStore
Makefile:94: recipe for target 'vet' failed
make: *** [vet] Error 1
```

The sync.Mutex in MemoryFileStore might be unnecessary anyway, since KeyMemoryStore provides its own locking.

PR #793, opened 2016-06-17
remove literal copy of sync.Mutex from memory store

gtank:
Attempted fix of #792. The locks seem to be covered already by the KeyMemoryStore, which is the only place that uses MemoryFileStore.

I think the go vet issue at hand is https://github.com/golang/go/issues/13675#issuecomment-165798626

### DIFF
--- a/trustmanager/keyfilestore.go
+++ b/trustmanager/keyfilestore.go
@@ -210,10 +210,11 @@ func NewKeyMemoryStore(passphraseRetriever notary.PassRetriever) *KeyMemoryStore
 
 	keyInfoMap := make(keyInfoMap)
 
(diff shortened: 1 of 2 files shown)


cyli:
jenkins, test this please

cyli:
Thanks lots for fixing these vetting issues @gtank!  Should make it easier to upgrade to 1.7 when it's released.  :)  

LGTM! I have no problems with this - these are mainly used for testing, and I don't foresee needing multiple `KeyMemoryStore` accessing the same `MemoryFileStore` concurrently.

riyazdf:
thank you @gtank, LGTM!

PR #798, opened 2016-06-22
Add info logging for 400s, error for 500s, adjust timestamp error level on GetCurrent

### LINKED ISSUE #766: Fix error responses from server so they get logged
We're logging the error from serving an error. We should always log the error that was served to ensure reporting systems get triggered.

Use an appropriate log level. Anything in the 5XX range should get logged as an error. 4XX should probably be info.

riyazdf:
Adds info-level logging for 4XX errors and error-level for 5XX in the server.  Also lowers errors on retrieving timestamps to debug-level.

Closes #766

### DIFF
--- a/server/handlers/default.go
+++ b/server/handlers/default.go
@@ -7,7 +7,6 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/Sirupsen/logrus"
 	ctxu "github.com/docker/distribution/context"
 	"github.com/gorilla/mux"
 	"golang.org/x/net/context"
@@ -149,7 +148,7 @@ func getHandler(ctx context.Context, w http.ResponseWriter, r *http.Request, var
 		// of time.
 		utils.SetLastModifiedHeader(w.Header(), *lastModified)
 	} else {
-		logrus.Warnf("Got bytes out for %s's %s (checksum: %s), but missing lastModified date",
+		logger.Warnf("Got bytes out for %s's %s (checksum: %s), but missing lastModified date",
 			gun, tufRole, checksum)
 	}
 
@@ -159,15 +158,15 @@ func getHandler(ctx context.Context, w http.ResponseWriter, r *http.Request, var
 
 // DeleteHandler deletes all data for a GUN. A 200 responses indicates success.
 func DeleteHandler(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
-	s := ctx.Value("metaStore")
-	store, ok := s.(storage.MetaStore)
-	if !ok {
-		logrus.Error("500 DELETE repository: no storage exists")
-		return errors.ErrNoStorage.WithDetail(nil)
-	}
 	vars := mux.Vars(r)
 	gun := vars["imageName"]
 	logger := ctxu.GetLoggerWithField(ctx, gun, "gun")
+	s := ctx.Value("metaStore")
+	store, ok := s.(storage.MetaStore)
+	if !ok {
+		logger.Error("500 DELETE repository: no storage exists")
+		return errors.ErrNoStorage.WithDetail(nil)
+	}
 	err := store.Delete(gun)
 	if err != nil {
 		logger.Error("500 DELETE repository")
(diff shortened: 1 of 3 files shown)


cyli:
@riyazdf Thanks so much for adding these!  I was wondering actually, though, if it'd make sense to bottleneck the logging all in one place (e.g. at `util/http.go` where if there's an error when calling the handler?).

Since all of these errors, if they are of a errcode type, can get the HTTP status code, we can determine what log level to log at.

That way we can make sure to get all of them at once.

riyazdf:
@cyli: great idea.  I've added that logic in addition to the previous logging - I'm ok with logging in both places so we have more context in the logs (for codepath, http request type, etc)

utils/http.go
```diff
@@ -79,6 +79,16 @@ func (root *rootHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		ctx = authCtx
 	}
 	if err := root.handler(ctx, w, r); err != nil {
+		if httpErr, ok := err.(errcode.ErrorCoder); ok {
+			// info level logging for non-5XX http errors
+			httpErrCode := httpErr.ErrorCode().Descriptor().HTTPStatusCode
+			if (httpErrCode < http.StatusOK || httpErrCode >= http.StatusMultipleChoices) && httpErrCode < http.StatusInternalServerError {
+				logrus.Info(httpErr)
```
cyli: Can we call `log.Info` here and `log.Error` below instead of directly using `logrus`, so we can get associate the logs with the request context too?

cyli:
Thanks @riyazdf!  Other than the above nitpick, LGTM!

utils/http.go
riyazdf: ah good catch.  I'll change the `logrus` call below too 👍

server/handlers/default.go
```diff
@@ -152,6 +162,7 @@ func DeleteHandler(ctx context.Context, w http.ResponseWriter, r *http.Request)
 	s := ctx.Value("metaStore")
 	store, ok := s.(storage.MetaStore)
 	if !ok {
+		logrus.Error("500 DELETE repository: no storage exists")
```
endophage: Should this be `logger.Error`? (would need to pull the logger out of the context like other handlers)

server/handlers/default.go
```diff
@@ -176,15 +187,18 @@ func GetKeyHandler(ctx context.Context, w http.ResponseWriter, r *http.Request)
 func getKeyHandler(ctx context.Context, w http.ResponseWriter, r *http.Request, vars map[string]string) error {
 	gun, ok := vars["imageName"]
 	if !ok || gun == "" {
+		logrus.Info("400 GET no gun in request")
```
endophage: `logger.Info`? (and move line 194 up?)

riyazdf: I don't think we can for this one because we don't have the gun yet, but totally for the other `logrus.Error` (and I think I see a `logrus.Warnf` that can get fixed)

utils/http.go
```diff
@@ -79,9 +78,19 @@ func (root *rootHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		ctx = authCtx
 	}
 	if err := root.handler(ctx, w, r); err != nil {
+		if httpErr, ok := err.(errcode.ErrorCoder); ok {
+			// info level logging for non-5XX http errors
+			httpErrCode := httpErr.ErrorCode().Descriptor().HTTPStatusCode
+			if (httpErrCode < http.StatusOK || httpErrCode >= http.StatusMultipleChoices) && httpErrCode < http.StatusInternalServerError {
```
endophage: We shouldn't have an `httpErr` if the status is `200` but it we do it would probably want to be logged at `info` level. Seems like we could just collapse this could be simplified to:

```
if httpErrCode >= http.StatusInternalServerError {
    log.Error(httpErr)
} else {
    log.Info(httpErr)
}
```

riyazdf: good point, added!

endophage:
LGTM

Issue #731, opened 2016-05-10
Add flags to `--init` that allow for importing existing root key/cert


endophage:
Suggested flags are `--rootcert` and `--rootkey`. We can be restrictive about the input formats we accept.

dnwake:
This functionality is crucial to make trust-pinning a useful feature in our usage scenario (where both pushing and pulling clients are trusted , but the Notary server itself is not).

dnwake:
Experimental pull request here: https://github.com/docker/notary/pull/813

endophage:
Keys is in, we're punting the certs side to the next release. Putting it in the 1.0 backlog for now, will likely get pulled in from there.

cyc115:
We are currently working on an implementation for `--rootcert` and key rotation.

PR #801, opened 2016-06-23
Support passing in root keys on repo initialization

ecordell:
This addresses #731 by adding a `--rootkey` flag to `init`.

Example:
```
$ notary init my/collection --rootkey=keys/root.key
```
resulting root.json:
```json
{
	"signed": {
		"_type": "Root",
		"consistent_snapshot": false,
		"expires": "2026-06-21T14:20:03.15135742-04:00",
		"keys": {
			"956fbe52f38738cb5c31b1cbff06842ceda8e429f55164a61349cd233a50813f": {
				"keytype": "ecdsa-x509",
				"keyval": {
					"private": null,
					"public": "..."
				}
			}
			// ...
		},
		"roles": {
			"root": {
				"keyids": [
					"956fbe52f38738cb5c31b1cbff06842ceda8e429f55164a61349cd233a50813f"
				],
				"threshold": 1
			}
			// ...
		},
		"version": 1
	},
	"signatures": [{
		"keyid": "956fbe52f38738cb5c31b1cbff06842ceda8e429f55164a61349cd233a50813f",
		"method": "ecdsa",
		"sig": "70BvEbnWA1bHoc2IjnIE8XFI1N4I1WQxyd2GDNX/Avfmy89Z5J1w0pL507LpLd1DbV6p6CtqPgCdb5oCMeXt5w=="
	}
	}]
}
```

Additionally, the underlying TUF apis are updated to allow multiple root keys and certs, but that is not currently exposed to the CLI.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -173,10 +173,14 @@ func rootCertKey(gun string, privKey data.PrivateKey) (data.PublicKey, error) {
 // timestamp key and possibly other serverManagedRoles), but the created repository
 // result is only stored on local disk, not published to the server. To do that,
 // use r.Publish() eventually.
-func (r *NotaryRepository) Initialize(rootKeyID string, serverManagedRoles ...string) error {
-	privKey, _, err := r.CryptoService.GetPrivateKey(rootKeyID)
-	if err != nil {
-		return err
+func (r *NotaryRepository) Initialize(rootKeyIDs []string, serverManagedRoles ...string) error {
+	privKeys := []data.PrivateKey{}
+	for _, keyID := range rootKeyIDs {
+		privKey, _, err := r.CryptoService.GetPrivateKey(keyID)
+		if err != nil {
+			return err
+		}
+		privKeys = append(privKeys, privKey)
 	}
 
 	// currently we only support server managing timestamps and snapshots, and
@@ -206,16 +210,20 @@ func (r *NotaryRepository) Initialize(rootKeyID string, serverManagedRoles ...st
 		}
 	}
 
-	rootKey, err := rootCertKey(r.gun, privKey)
-	if err != nil {
-		return err
+	rootKeys := []data.PublicKey{}
+	for _, privKey := range privKeys {
+		rootKey, err := rootCertKey(r.gun, privKey)
+		if err != nil {
+			return err
+		}
+		rootKeys = append(rootKeys, rootKey)
 	}
 
 	var (
 		rootRole = data.NewBaseRole(
 			data.CanonicalRootRole,
 			notary.MinThreshold,
-			rootKey,
+			rootKeys...,
 		)
 		timestampRole data.BaseRole
 		snapshotRole  data.BaseRole
@@ -271,7 +279,7 @@ func (r *NotaryRepository) Initialize(rootKeyID string, serverManagedRoles ...st
 
 	r.tufRepo = tuf.NewRepo(r.CryptoService)
 
-	err = r.tufRepo.InitRoot(
+	err := r.tufRepo.InitRoot(
 		rootRole,
 		timestampRole,
 		snapshotRole,
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -167,7 +167,7 @@ func initializeRepo(t *testing.T, rootType, gun, url string,
 
 	repo, rec, rootPubKeyID := createRepoAndKey(t, rootType, tempBaseDir, gun, url)
 
-	err = repo.Initialize(rootPubKeyID, serverManagedRoles...)
+	err = repo.Initialize([]string{rootPubKeyID}, serverManagedRoles...)
 	if err != nil {
 		os.RemoveAll(tempBaseDir)
 	}
@@ -241,7 +241,7 @@ func TestInitRepositoryManagedRolesIncludingRoot(t *testing.T) {
 
 	repo, rec, rootPubKeyID := createRepoAndKey(
 		t, data.ECDSAKey, tempBaseDir, "docker.com/notary", "http://localhost")
-	err = repo.Initialize(rootPubKeyID, data.CanonicalRootRole)
+	err = repo.Initialize([]string{rootPubKeyID}, data.CanonicalRootRole)
 	require.Error(t, err)
 	require.IsType(t, ErrInvalidRemoteRole{}, err)
 	// Just testing the error message here in this one case
@@ -261,7 +261,7 @@ func TestInitRepositoryManagedRolesInvalidRole(t *testing.T) {
 
 	repo, rec, rootPubKeyID := createRepoAndKey(
 		t, data.ECDSAKey, tempBaseDir, "docker.com/notary", "http://localhost")
-	err = repo.Initialize(rootPubKeyID, "randomrole")
+	err = repo.Initialize([]string{rootPubKeyID}, "randomrole")
 	require.Error(t, err)
 	require.IsType(t, ErrInvalidRemoteRole{}, err)
 	// no key creation happened
@@ -278,7 +278,7 @@ func TestInitRepositoryManagedRolesIncludingTargets(t *testing.T) {
 
 	repo, rec, rootPubKeyID := createRepoAndKey(
 		t, data.ECDSAKey, tempBaseDir, "docker.com/notary", "http://localhost")
-	err = repo.Initialize(rootPubKeyID, data.CanonicalTargetsRole)
+	err = repo.Initialize([]string{rootPubKeyID}, data.CanonicalTargetsRole)
 	require.Error(t, err)
 	require.IsType(t, ErrInvalidRemoteRole{}, err)
 	// no key creation happened
@@ -298,10 +298,35 @@ func TestInitRepositoryManagedRolesIncludingTimestamp(t *testing.T) {
 
 	repo, rec, rootPubKeyID := createRepoAndKey(
 		t, data.ECDSAKey, tempBaseDir, "docker.com/notary", ts.URL)
-	err = repo.Initialize(rootPubKeyID, data.CanonicalTimestampRole)
+	err = repo.Initialize([]string{rootPubKeyID}, data.CanonicalTimestampRole)
 	require.NoError(t, err)
+
 	// generates the target role, the snapshot role
 	rec.requireCreated(t, []string{data.CanonicalTargetsRole, data.CanonicalSnapshotRole})
+}
+
+func TestInitRepositoryMultipleRootKeys(t *testing.T) {
+	// Temporary directory where test files will be created
+	tempBaseDir, err := ioutil.TempDir("/tmp", "notary-test-")
+	require.NoError(t, err, "failed to create a temporary directory")
+	defer os.RemoveAll(tempBaseDir)
+
+	ts, _, _ := simpleTestServer(t)
+	defer ts.Close()
+
+	repo, rec, rootPubKeyID := createRepoAndKey(
+		t, data.ECDSAKey, tempBaseDir, "docker.com/notary", ts.URL)
+	rootPubKey2, err := repo.CryptoService.Create("root", repo.gun, data.ECDSAKey)
+	require.NoError(t, err, "error generating second root key: %s", err)
+
+	err = repo.Initialize([]string{rootPubKeyID, rootPubKey2.ID()}, data.CanonicalTimestampRole)
+	require.NoError(t, err)
+
+	// generates the target role, the snapshot role
+	rec.requireCreated(t, []string{data.CanonicalTargetsRole, data.CanonicalSnapshotRole})
+
+	// has two root keys
+	require.Len(t, repo.tufRepo.Root.Signed.Roles[data.CanonicalRootRole].KeyIDs, 2)
 }
 
 // Initializing a new repo fails if unable to get the timestamp key, even if
@@ -317,7 +342,7 @@ func TestInitRepositoryNeedsRemoteTimestampKey(t *testing.T) {
 
 	repo, rec, rootPubKeyID := createRepoAndKey(
 		t, data.ECDSAKey, tempBaseDir, "docker.com/notary", ts.URL)
-	err = repo.Initialize(rootPubKeyID, data.CanonicalTimestampRole)
+	err = repo.Initialize([]string{rootPubKeyID}, data.CanonicalTimestampRole)
 	require.Error(t, err)
 	require.IsType(t, store.ErrMetaNotFound{}, err)
 
@@ -339,7 +364,7 @@ func TestInitRepositoryNeedsRemoteSnapshotKey(t *testing.T) {
 
 	repo, rec, rootPubKeyID := createRepoAndKey(
 		t, data.ECDSAKey, tempBaseDir, "docker.com/notary", ts.URL)
-	err = repo.Initialize(rootPubKeyID, data.CanonicalSnapshotRole)
+	err = repo.Initialize([]string{rootPubKeyID}, data.CanonicalSnapshotRole)
 	require.Error(t, err)
 	require.IsType(t, store.ErrMetaNotFound{}, err)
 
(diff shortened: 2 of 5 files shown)


riyazdf:
jenkins, test this please

riyazdf:
thank you @ecordell for this awesome work!

I had some design considerations about how this might work with a `--rootcert` flag; I think it would be nice to optionally allow users to provide their own certificate for each root key they use.  This would address the other part of #731, which is critical for pinning to a CA

Maybe we could only allow for up to 1 key on init:
`notary init --rootkey root.key` - uses the provided private key and generates a self-signed cert for the root.json
`notary init --rootkey root.key --rootcert cert.pem` - uses the provided private key and certificate (provided that they form a valid keypair) for the root.json

And if we wanted to add more keys:
`notary key add <GUN> root root.key [--rootcert cert.pem]` - same behavior as above with certs, either using the one if provided (and valid), otherwise generating one

ping @endophage for his thoughts

ecordell:
Thanks for taking a look @riyazdf !

To limit the scope, I've simply rebased on master and changed `rootkeys` to `rootkey` to support only a single key being passed in. Adding additional root keys and supporting certs other than self-signed, generated certs can come in a future PR.

riyazdf:
jenkins, test this please

endophage:
I'll review this fully next week but in the meantime, thank you! This is a feature we've had quite a few people asking for. I largely agree with @riyazdf's suggestions, though it might be nice to find a way to cleanly combine the key and cert so that if we do decide to allow multiple keys/certs to be added on init there's a nice way to do it.

Also worth noting there might be a use case to add multiple certs but fewer keys (i.e. you want other keys to be able to sign root but you personally only hold one of them. This is a specific use case we discussed with one user at DockerCon).

cmd/notary/tuf.go
```diff
@@ -285,7 +320,7 @@ func (t *tufCommander) tufInit(cmd *cobra.Command, args []string) error {
 		cmd.Printf("Root key found, using: %s\n", rootKeyID)
 	}
 
-	if err = nRepo.Initialize(rootKeyID); err != nil {
+	if err = nRepo.Initialize([]string{rootKeyID}); err != nil {
```
riyazdf: If other root keys already exist on this host, this `rootKeyID` may not necessarily correspond to the `--rootkey` we pass in.  We should change the conditional logic above such that it prioritizes keys from the `--rootkey` flag if we passed any in

cmd/notary/tuf.go
```diff
@@ -86,17 +89,21 @@ type tufCommander struct {
...
 
 	input  string
 	output string
 	quiet  bool
 }
 
 func (t *tufCommander) AddToCommand(cmd *cobra.Command) {
-	cmd.AddCommand(cmdTUFInitTemplate.ToCommand(t.tufInit))
+	cmdTUFInit := cmdTUFInitTemplate.ToCommand(t.tufInit)
+	cmdTUFInit.Flags().StringVar(&t.rootKey, "rootkey", "", "Root key to initialize the repository with.")
```
riyazdf: super nit: can we drop the period at the end of this flag description for consistency?

ecordell:
@riyazdf Fixed the issues you noted and rebased.

riyazdf:
jenkins, test this please

client/client_test.go
```diff
@@ -298,7 +298,27 @@ func TestInitRepositoryManagedRolesIncludingTimestamp(t *testing.T) {
...
+
+	ts, _, _ := simpleTestServer(t)
+	defer ts.Close()
+
+	repo, rec, rootPubKeyID := createRepoAndKey(
+		t, data.ECDSAKey, tempBaseDir, "docker.com/notary", ts.URL)
+	rootPubKey2, err := repo.CryptoService.Create("root", repo.gun, data.ECDSAKey)
+	require.NoError(t, err, "error generating second root key: %s", err)
+
+	err = repo.Initialize([]string{rootPubKeyID, rootPubKey2.ID()}, data.CanonicalTimestampRole)
```
riyazdf: could we verify that both keys were included in the root metadata?

Something like `require.Len(t, repo.tufRepo.Root.Signed.Roles[data.CanonicalRootRole].KeyIDs, 2)`

cmd/notary/tuf.go
```diff
@@ -268,7 +275,38 @@ func (t *tufCommander) tufInit(cmd *cobra.Command, args []string) error {
...
+
+		pemBytes, err := ioutil.ReadAll(keyFile)
+		if err != nil {
+			return fmt.Errorf("Error reading input file: %v", err)
+		}
+		if err = cryptoservice.CheckRootKeyIsEncrypted(pemBytes); err != nil {
+			return err
+		}
+
+		privKey, err := trustmanager.ParsePEMPrivateKey(pemBytes, "")
```
riyazdf: I don't think we need the `ParsePEMPrivateKey` with empty passphrase call since we're only dealing with root keys which must always be encrypted, and we check the PEM block structure integrity above

cmd/notary/integration_test.go
```diff
@@ -81,6 +81,43 @@ func setupServer() *httptest.Server {
...
+			tempDir, "private", "root_keys", rID+".key"))
+	}
+
+	// init repo
+	_, err = runCommand(t, tempDir, "-s", server.URL, "init", "gun", "--rootkey", rootFiles[0])
+	require.NoError(t, err)
+
+	// publish repo
+	_, err = runCommand(t, tempDir, "-s", server.URL, "publish", "gun")
+	require.NoError(t, err)
```
riyazdf: It would be awesome to add a check to this test that this initialized repo used the specified key from the flag (especially if we had more rootkeys in the `private/root_keys` dir)

Unfortunately I can't think of a way to verify which root key was used without manually unmarshalling metadata (unless we wanted to bring in a `NotaryRepository` from the client library), so it might be a little messy :|

riyazdf:
thank you @ecordell for working on this!  I have a couple of small nits and test case additions, but I really like how this looks, cc @cyli @endophage for additional review.

I'll also link https://github.com/docker/notary/pull/813 for tracking, since it is somewhat coupled with the discussion above

cyli:
@riyazdf @endophage maybe `--rootkey path-to-privkey:path-to-rootcert`?

cmd/notary/integration_test.go
cyli: Apologies for the trouble - would it also be possible to add some or all of the following negative test cases?  Similar test cases may already exist for importing a root key.

1.  Passing a non-existent file with the `--rootkey` flag should cause the command to fail
1.  Passing a non-PEM file with the `--rootkey` flag should cause the command to fail
1.  Passing an unencrypted PEM file with the `--rootkey` flag should cause the command to fail
1.  Failure to enter the correct passphrase with a valid `--rootkey` file should cause the command to fail
1.  Passing a targets or snapshot key with the `--rootkey` flag should cause the command to fail, even if the correct password is entered

cyli:
@ecordell This is awesome - thanks so much for adding this!  Being able to import existing root keys and certs will become really important for our use cases in the future.  It looks fantastic aside from the test nitpicks above!

riyazdf:
@cyli: @endophage and I just discussed that it might be useful to allow for multiple `--rootcert` and multiple `--rootkey` flags on initialization, potentially allowing for initializing with extra certs that you may not have the keys for.

The use case would be to add other admins to be root-signers, but without having to share private key material directly.  Ex: `notary init --rootkey alice.key --rootcert alice.crt --rootcert bob.crt --rootcert eve.crt`

cyli:
That makes sense - and I guess certs and keys don't really need to be
associated?

client/client_test.go
ecordell: Definitely, will update

cmd/notary/tuf.go
ecordell: Good point, removing

cmd/notary/integration_test.go
ecordell: No problem; we already have the key id here so we can just `notary key list` and check that the id listed is a match. I'll update with that test.

ecordell:
Thanks @riyazdf and @cyli for taking the time to review!

I agree that

cmd/notary/integration_test.go
ecordell: Sorry @cyli GitHub caching caused me to miss your comment. 

The integration tests seem to only test the "happy path" - not saying that's all they should test, just noting that it's in keeping with the current style.

IIRC the cases you listed are tested at the api levels, because the addition in this PR uses all the same code paths for key loading. I will double check to make sure.

cyli: @ecordell  You are definitely right in that the utility functions you use for loading the key should all have be tested already.  I agree with you that it's pointless to re-test this functionality here; I  mainly wanted to test that particular loading checks (like ensuring that we only allow importing encrypted keys - we did not have to include this check, after all) are called and that their errors are propagated (for instance that we don't just fall back on generating a random root key if importing a key fails), and wanted some way to trigger said failures.  

For instance, for case 3, it's to test that the root key is required to be encrypted - you could for instance completely skip that check and allow importing an unencrypted root key (which may or may not be a valid choice).

Also, for case 4, cryptoservice's `AddKey` will definitely fail if the roles are mismatched, but I'm also wondering if we should do a separate check, in case you are importing a targets key from a different location (for instance, you might have a separate `~/.docker/trust` and `~/.notary` directory with different keys, and the `AddKey` check will only fail if you import a targets key from `~/.notary` into a repo in `~/.notary`, but won't fail if you import a targets key that exists only in `~/.docker/trust` into `~/.notary`).  But maybe this case doesn't matter as much.

ecordell: Thanks @cyli. I updated with integration tests for the cases you enumerated - deciding whether or not to change the behavior in those cases is, as you point out, up for discussion, but at least we can track those changes now with the tests.

ecordell:
@riyazdf @cyli I think I've addressed all of your comments at this point, please let me know if there's anything else I should do.

cyli:
jenkins, test this please

dnwake:
Sorry to make things more complicated, but I have created new pull request at https://github.com/docker/notary/pull/821 based on top of a (probably slightly out-of-date) version of this commit.  It adds the --rootcert flag to allow a root cert and root key to be imported together.

I'm very willing to work with everyone here to get all the functionality approved asap -- please let me know the best way to co-ordinate things.  Cheers.

cmd/notary/integration_test.go
```diff
@@ -81,6 +81,98 @@ func setupServer() *httptest.Server {
...
+	_, err = runCommand(t, tempDir, "-s", server.URL, "init", "gun2", "--rootkey", "bad_file")
+	require.Error(t, err, "Init with nonexistent key file should error")
+
+	// check error if file is invalid format
+	badKeyFilename := filepath.Join(tempDir, "bad_key.key")
+	nonPEMKey := []byte("thisisnotapemkey")
+	err = ioutil.WriteFile(badKeyFilename, nonPEMKey, 0644)
+	require.NoError(t, err)
+
+	_, err = runCommand(t, tempDir, "-s", server.URL, "init", "gun2", "--rootKey", badKeyFilename)
```
cyli: I think the flag for these next few test cases should have `--rootkey` as opposed to `--rootKey`?

cmd/notary/integration_test.go
cyli: 👍   Thanks so much!  :)

cyli:
Thanks so much for all your work and your responsiveness on this @ecordell!  Aside from one last nitpick about the `--rootkey` flag in tests, this LGTM!

riyazdf:
@ecordell thank you for updating!

LGTM after resolving @cyli's comments

@dnwake: thanks for checking in -- I think we'll probably want to merge this PR first before #821 but I'll take a look and review in the meantime.  We can rebase on top of this once merged :)

cmd/notary/tuf.go
```diff
@@ -268,7 +275,36 @@ func (t *tufCommander) tufInit(cmd *cobra.Command, args []string) error {
...
+
+		pemBytes, err := ioutil.ReadAll(keyFile)
+		if err != nil {
+			return fmt.Errorf("Error reading input file: %v", err)
+		}
+		if err = cryptoservice.CheckRootKeyIsEncrypted(pemBytes); err != nil {
+			return err
+		}
+
+		privKey, _, err := trustmanager.GetPasswdDecryptBytes(t.retriever, pemBytes, "", "imported root")
```
riyazdf: something I just noticed from #821 - it might make sense to just change the last arg here to `data.CanonicalRootRole`, so that `NOTARY_ROOT_PASSPHRASE` can be used to automate this

ecordell:
Updated and rebased!

cyli:
jenkins, test this please

riyazdf:
LGTM after changes, thank you again @ecordell for contributing this!

client/client.go
```diff
@@ -173,10 +173,14 @@ func rootCertKey(gun string, privKey data.PrivateKey) (data.PublicKey, error) {
 // timestamp key and possibly other serverManagedRoles), but the created repository
 // result is only stored on local disk, not published to the server. To do that,
 // use r.Publish() eventually.
-func (r *NotaryRepository) Initialize(rootKeyID string, serverManagedRoles ...string) error {
-	privKey, _, err := r.CryptoService.GetPrivateKey(rootKeyID)
-	if err != nil {
-		return err
+func (r *NotaryRepository) Initialize(rootKeyIDs []string, serverManagedRoles ...string) error {
+	privKeys := []data.PrivateKey{}
```
endophage: In all likelihood `privKeys` will have the same length as `rootKeyIDs` so we should initialize appropriately: `privKeys := make([]data.PrivateKey, 0, len(rootKeyIDs))`

client/client.go
```diff
@@ -206,16 +210,20 @@ func (r *NotaryRepository) Initialize(rootKeyID string, serverManagedRoles ...st
 		}
 	}
 
-	rootKey, err := rootCertKey(r.gun, privKey)
-	if err != nil {
-		return err
+	rootKeys := []data.PublicKey{}
```
endophage: Same as above comment for `privKeys`

cmd/notary/tuf.go
```diff
@@ -268,7 +275,36 @@ func (t *tufCommander) tufInit(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	rootKeyList := nRepo.CryptoService.ListKeys(data.CanonicalRootRole)
+	var rootKeyList []string
+
+	if t.rootKey != "" {
```
endophage: This addition is pushing this function right up to the cyclomatic complexity threshold before goreportcard marks it as a problem. The interior of this `if` blocks seems ideally suited to be broken out as its own function.

cmd/notary/tuf.go
```diff
@@ -268,7 +275,36 @@ func (t *tufCommander) tufInit(cmd *cobra.Command, args []string) error {
...
+		privKey, _, err := trustmanager.GetPasswdDecryptBytes(t.retriever, pemBytes, "", data.CanonicalRootRole)
+		if err != nil {
+			return err
+		}
+
+		err = nRepo.CryptoService.AddKey(data.CanonicalRootRole, "", privKey)
+		if err != nil {
+			return fmt.Errorf("Error importing key: %v", err)
+		}
+		rootKeyList = []string{data.PublicKeyFromPrivate(privKey).ID()}
```
endophage: Should just be able to do `privKey.ID()` and it handles the stripping of the private bytes for you.

endophage:
Left some comments on minor stuff but overall good to merge this now. LGTM and thank you!

client/client.go
ecordell: I'm not so sure - I think you should be able to initialize with one private key and multiple public keys, so that these two don't match up.

cmd/notary/tuf.go
ecordell: Definitely, and I have a note on my desk to make this exact change and just forgot. I can make a new PR if appropriate

client/client.go
endophage: That's not implemented in this PR though. Although @dnwake's PR for `--rootcert` may make my comment moot.

cmd/notary/tuf.go
cyli: I think @riyazdf got to it in #829 :)

Issue #734, opened 2016-05-10
Warnings when a certificate will expire soon


endophage:
Output warnings when any x509 type key will expire soon.

PR #802, opened 2016-06-24
Output warnings for cert within 6 months expiry, check intermediate cert expiry

### LINKED ISSUE #860: Warn if a delegation key certificate is expiring
Since we allow delegation public keys to be certs.

riyazdf:
Addresses #734 for root certificates, and cleans up some of the certificate validation logic.  Also ensures we use non-expired and non-SHA1 certificates for intermediate certs.  Though there is no way to include specific certs to the root at the moment, we will need this validation.

Last commit closes #860


trustmanager/x509utils.go
```diff
@@ -233,6 +233,10 @@ func ValidateCertificate(c *x509.Certificate) error {
 	if (tomorrow).Before(c.NotBefore) || now.After(c.NotAfter) {
 		return fmt.Errorf("certificate is expired")
 	}
+	// If this certificate is expiring within 6 months, put out a warning
+	if (c.NotAfter).Before(time.Now().AddDate(0, 6, 0)) {
+		logrus.Warn("certificate is expiring within 6 months")
```
endophage: what other info can we include to make it clear _which_ certificate is expiring?

trustpinning/certs.go
```diff
@@ -266,3 +284,24 @@ func parseAllCerts(signedRoot *data.SignedRoot) (map[string]*x509.Certificate, m
 
 	return leafCerts, intCerts
 }
+
+func checkCertExpiry(cert *x509.Certificate) error {
```
cyli: Hmm... We seem to have 2 separate sets of checks - these (which check the expiration with no leeway), and the expiration check in `ValidateCertificate`.  Likewise, we check that the RSA key must be at least 2048 in `ValidateCertificate`, but we don't check that for root leaf certificates or intermediate certificates.  For intermediate and root certs, we disallow certs with SHA1.

Is there a reason they should be different?  (might make sense if people are bringing their own CA to maybe be more permissive for the root CA?)

Can we unify the checks into `ValidateCertificate`?  Maybe let that function take a boolean as to whether the expiry check should be performed, etc.?

riyazdf: Yeah, BYO-CA might warrant having different checks but I definitely agree that we should consolidate these checks into one place -- I'll update this 👍

trustmanager/x509utils.go
```diff
@@ -222,16 +222,22 @@ func ParsePEMPublicKey(pubKeyBytes []byte) (data.PublicKey, error) {
...
+		now := time.Now()
+		tomorrow := now.AddDate(0, 0, 1)
+		// Give one day leeway on creation "before" time, check "after" against today
+		if (tomorrow).Before(c.NotBefore) || now.After(c.NotAfter) {
+			return fmt.Errorf("certificate with CN %s is expired", c.Subject.CommonName)
+		}
+		// If this certificate is expiring within 6 months, put out a warning
+		if (c.NotAfter).Before(time.Now().AddDate(0, 6, 0)) {
+			logrus.Warnf("certificate with CN %s is near expiry", c.Subject.CommonName)
+		}
```
cyli: Did we also want to add the SHA1 check in here?

riyazdf: yes!  oops, I forgot to include it in the unification 😄

cyli:
Thanks for checking the intermediate certs too!  If it isn't too much trouble, can we add a check for the SHA1 algorithms too for the `ValidateCertificate` tests?  If not, no worries.  

Other than that, this LGTM!

riyazdf:
@cyli of course, great idea!  Done.

trustpinning/golang_ugh.patch
```diff
@@ -0,0 +1,11 @@
+diff --git a/README.md b/README.md
```
endophage: Did you mean to include this file?

trustpinning/lol.out
```diff
@@ -0,0 +1,23 @@
+diff --git a/trustpinning/certs.go b/trustpinning/certs.go
```
endophage: did you mean to include this file?

tuf/utils/x509.go
```diff
@@ -258,16 +258,27 @@ func ParsePEMPublicKey(pubKeyBytes []byte) (data.PublicKey, error) {
...
-	tomorrow := now.AddDate(0, 0, 1)
-	// Give one day leeway on creation "before" time, check "after" against today
-	if (tomorrow).Before(c.NotBefore) || now.After(c.NotAfter) {
-		return fmt.Errorf("certificate is expired")
+	if checkExpiry {
+		now := time.Now()
+		tomorrow := now.AddDate(0, 0, 1)
+		// Give one day leeway on creation "before" time, check "after" against today
+		if (tomorrow).Before(c.NotBefore) || now.After(c.NotAfter) {
+			return fmt.Errorf("certificate with CN %s is expired", c.Subject.CommonName)
```
endophage: The whole `if checkExpiry` block should probably be moved to the end of the function. This return here will cause us to accept delegation certs in `tuf/tuf.go` that may have serious flaws but have expired and we returned from this validation prematurely, c.f. comment in tuf/tuf.go line 249 `Currently we do not reject expired delegation keys but warn if they might expire soon or have already`

riyazdf: good catch, updated!

tuf/tuf.go
```diff
@@ -245,6 +245,17 @@ func (tr *Repo) GetDelegationRole(name string) (data.DelegationRole, error) {
...
 					return err
 				}
+				// Check all public key certificates in the role for expiry
+				// Currently we do not reject expired delegation keys but warn if they might expire soon or have already
+				for keyID, pubKey := range delgRole.Keys {
+					certFromKey, err := utils.LoadCertFromPEM(pubKey.Public())
+					if err != nil {
+						continue
+					}
+					if err := utils.ValidateCertificate(certFromKey, true); err != nil {
```
endophage: This check probably needs to become more explicit as well. Right now it would proceed even if the problem was something like an RSA key being too short (which we should treat as a hard failure). Maybe create an `ErrCertExpired` (if it doesn't exist somewhere already) and only allow continuation if we get that specific error.

riyazdf: Added a `ErrCertExpired` and check for that specific error here.

endophage:
LGTM

PR #809, opened 2016-06-30
use tabwriter in favour of tablewriter

### LINKED ISSUE #805: Investigate if text/tabwriter is a sufficient substitute for tablewriter
It would reduce vendored dependencies and has a super simple interface to column align output.

endophage:
Closes #805 

The formatting isn't 100% identical but it's very close and it still looks good. I had to do some manual formatting for the role printing where tablewriter handled multi-line rows cleanly. Overall though I think this is a good tradeoff for the removal of the dependency.

No rush to merge this. Happy to leave it open for a while and see how people feel.


riyazdf:
this is awesome!  I'm still digging into the delegation prints for multiple keys/paths but here are some screenshots for the interested.  Main difference is the underlining at the top of the table is slightly different (which is fine with me):

`notary list docker.io/library/ubuntu`
screen shot 2016-06-30 at 11 44 25 am

`notary key list`
screen shot 2016-06-30 at 11 45 16 am

riyazdf:
LGTM!  Tested with multiple delegation keys and paths, looks good!

screen shot 2016-07-05 at 9 23 21 am

PR #814, opened 2016-07-01
WIP: Keynativestore

avaid96:
created functionality to have keydata storage in the system's native keystore


avaid96:
I accidentally `go fmt `ed the vendor directory. I'll undo it soon

HuKeping:
What about people use notary with out docker-credential-x

avaid96:
It won't change anything for anyone who chooses not to use the native store. It is an opt in feature which can be activated by following the instructions in advanced usage docs- setting "useNative" to "true" in the config.json

HuKeping:
It's true. But why should we call that a native?

avaid96:
[Docker credentials helper](https://github.com/docker/docker-credential-helpers/) uses this terminology to refer to the credentials stores of each OS and I wanted to follow the same convention. Also considered simply calling it keychain store since that might be more intuitive for many, but that makes it OS specific in a way. Open to suggestions if you have anything else in mind?

client/repo_pkcs11.go
```diff
@@ -4,19 +4,19 @@ package client
...
 import (
 	"fmt"
-	"net/http"
-
 	"github.com/docker/notary"
+	"github.com/docker/notary/passphrase"
 	"github.com/docker/notary/trustmanager"
 	"github.com/docker/notary/trustmanager/yubikey"
 	"github.com/docker/notary/trustpinning"
+	"net/http"
```
riyazdf: I don't think you need this import

client/repo.go
```diff
@@ -16,14 +17,22 @@ import (
 // (This is normally defaults to "~/.notary" or "~/.docker/trust" when enabling
 // docker content trust).
 func NewNotaryRepository(baseDir, gun, baseURL string, rt http.RoundTripper,
-	retriever notary.PassRetriever, trustPinning trustpinning.TrustPinConfig) (
+	retriever notary.PassRetriever, trustPinning trustpinning.TrustPinConfig, UseNative bool) (
```
riyazdf: to keep consistent with go convention, we should lower-case this to `useNative`

cmd/notary/tuf.go
```diff
@@ -692,6 +692,15 @@ func getRemoteTrustServer(config *viper.Viper) string {
 	return defaultServerURL
 }
 
+func getUseNative(config *viper.Viper) bool {
+	conf := config.Get("useNative")
```
riyazdf: I think this function can be reduced to `return config.GetBool("useNative")` -- we shouldn't need to parse strings for a boolean value

client/repo_pkcs11.go
```diff
@@ -4,19 +4,19 @@ package client
...
 )
 
 // NewNotaryRepository is a helper method that returns a new notary repository.
 // It takes the base directory under where all the trust files will be stored
 // (usually ~/.docker/trust/).
 func NewNotaryRepository(baseDir, gun, baseURL string, rt http.RoundTripper,
-	retriever notary.PassRetriever, trustPinning trustpinning.TrustPinConfig) (
+	retriever notary.PassRetriever, trustPinning trustpinning.TrustPinConfig, UseNative bool) (
```
riyazdf: can also lowercase to `useNative` here

trustmanager/keynativestore.go
```diff
@@ -0,0 +1,147 @@
...
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"github.com/docker/docker-credential-helpers/client"
+	"github.com/docker/docker-credential-helpers/credentials"
+	"github.com/docker/notary"
+	"github.com/docker/notary/tuf/data"
+	"strings"
+	//"os/exec"
+	//"regexp"
```
riyazdf: can we delete the unused imports?

client/repo_pkcs11.go
avaid96: thanks, removed it

client/repo.go
avaid96: makes sense, changed it. Should getUseNative change as well or should we keep that as is to keep with the config of the functions in tuf.go?

cmd/notary/tuf.go
avaid96: I thought of doing this just in case a user enters an invalid boolean, your call- I'm fine with either

client/repo_pkcs11.go
avaid96: yup, changed it

trustmanager/keynativestore.go
avaid96: yup, changed it

client/repo.go
riyazdf: the `getUseNative` function name follows go convention and is fine as-is 👍

client/repo.go
```diff
@@ -16,14 +17,22 @@ import (
...
 	*NotaryRepository, error) {
 
 	fileKeyStore, err := trustmanager.NewKeyFileStore(baseDir, retriever)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create private key store in directory: %s", baseDir)
 	}
-
+	keyStores := []trustmanager.KeyStore{fileKeyStore}
+	if UseNative {
+		nativeKeyStore, err := trustmanager.NewKeyNativeStore(passphrase.ConstantRetriever("password"))
```
riyazdf: Why do we have this hardcoded to `"password"`, but the other `NewKeyNativeStore` in `getKeyStores` take a prompt retriever?

client/repo_pkcs11.go
avaid96: Actually need this as per the function definition: 
```
func NewNotaryRepository(baseDir, gun, baseURL string, rt http.RoundTripper,
	retriever notary.PassRetriever, trustPinning trustpinning.TrustPinConfig, useNative bool) (
	*NotaryRepository, error)
```

client/repo.go
avaid96: changed it, thanks for catching this!

client/repo_pkcs11.go
riyazdf: oh my bad -- I got caught up with the reformatting.  Sounds good 👍

cmd/notary/main.go
```diff
@@ -98,7 +99,7 @@ func (n *notaryCommander) parseConfig() (*viper.Viper, error) {
 	// Setup the configuration details into viper
 	config.SetDefault("trust_dir", defaultTrustDir)
 	config.SetDefault("remote_server", map[string]string{"url": defaultServerURL})
-
+	config.SetDefault("useNative", dontusenative)
```
riyazdf: in go, `false` is the default value for booleans, so we don't actually need to set a default here.
If a user does not specify the `useNative` key in a provided config, it will automatically be false.

cmd/notary/tuf.go
avaid96: As per our discussion, now using config.GetBool("useNative")

docs/advanced_usage.md
```diff
@@ -113,6 +113,62 @@ Yubikey support requires
 (which are bundled with the PIV tools)</a> to be available in standard
 library locations.
 
+### Use the native store
```
riyazdf: Awesome docs! 👍 

I'm thinking it should be clear that this is for keys only, so it might be worth renaming the section to "Use the native key store"?

cmd/notary/main.go
avaid96: alright perfect, changed it!

docs/advanced_usage.md
```diff
@@ -113,6 +113,62 @@ Yubikey support requires
...
+
+We currently support native storage on OS X, Linux and Windows. The advantage of using the native store to manage keydata is 3-fold. It is more easily visible, it more secure and it is more convenient- since the native store is secured, there is no need for passphrases on keys. You can enable use of the native store by following the below instructions. The programs in this repository are written with the Go programming language. These instructions assume that you have set it up your machine.
```
riyazdf: some small nits on this line:
 - small typo: `it is more secure`
 - I think you should elaborate on "since the native store is secured" - you should mention how it already has OS-level access control (not sure how to best phrase this)
 - also, have we tested wincred?  If not, we should not release it

docs/advanced_usage.md
```diff
@@ -113,6 +113,62 @@ Yubikey support requires
...
+		"url": "https://notary-server:4443",
+		"root_ca": "root-ca.crt"
+	},
+	"useNative":"true"
+}
+``` 
+
+For Notary to detect this change you will have to re-make the notary client. You can do this easily by running ``` make client``` 
+
+Congratulations!  Notary will now use the Native Store for keydata management. You can open up keychain access or the equivalent to view keydata which working with Notary. **Please be careful not to accidentely delete Notary keys from keychain access. Especially the root key!**
```
riyazdf: typo nit: `accidentally`

client/repo_pkcs11.go
```diff
@@ -25,9 +25,19 @@ func NewNotaryRepository(baseDir, gun, baseURL string, rt http.RoundTripper,
 	}
 
 	keyStores := []trustmanager.KeyStore{fileKeyStore}
+	if UseNative {
+		nativeKeyStore, err := trustmanager.NewKeyNativeStore(passphrase.ConstantRetriever("password"))
```
riyazdf: same comment here as above with respect to using a prompt retriever instead of a constant `"password"`

avaid96:
everything attended to

cmd/notary/keys.go
```diff
@@ -454,12 +455,21 @@ func removeKeyInteractively(keyStores []trustmanager.KeyStore, keyID string,
 	var storesByIndex []trustmanager.KeyStore
 
 	for _, store := range keyStores {
-		for keypath, keyInfo := range store.ListKeys() {
-			if filepath.Base(keypath) == keyID {
+		if strings.Contains(store.Name(), "Native keychain store") {
```
riyazdf: I think we should be able to refactor this to not need to be split out -- in fact we probably don't need to `ListKeys()` at all here (unless I'm missing something)

riyazdf:
jenkins, test this please

avaid96:
The issue is because I'm not adding the vendor directory, will fix it once I get lists in credentials helper

cyli:
@avaid96 Sorry for the delay on this!  Would you mind giving us access to push to this branch so we can rebase?

avaid96:
@cyli Just allowed edits from maintainers. Looking forward to seeing this! I have finals for the next 2 weeks but would be happy to look at any comments that pop up after these 2 weeks 😄

cyli:
@avaid96 Awesome, thank you!

endophage:
There are a lot of changes in a lot of places (both here and in the credential helpers) that mean this would need a lot of updates. I'm going to close this for now, we'll reassess our use of the credential helpers when we get the chance. I think the first integration is likely to be just storing the user credentials for the notary login.

Issue #789, opened 2016-06-15
Update the auth permissions so that delete repo requires admin permission


cyli:
I think that means all permissions (`*`), not just push/pull.

Confirmation from @jlhawn that `scope=repository:library/foobar:*` is given only to namespace admins, as opposed to anyone with read-write access.

riyazdf:
I've been investigating this a little bit further -- this shouldn't be too much work on the delete side, but will require some changes on publishing.

Since we currently don't separate an initial publish vs. subsequent publishes, we'll have to check on the server if we have any trust data and send a new challenge for a token with admin (`*`) permission if this publish is initializing trust data for this repo.

riyazdf:
After digging around more, I'm currently of the opinion that the initial push should remain as is (requiring `"push"` and `"pull"`) rather than `*`.  Changing the required permissions would impose a hard break on old clients since they would not be able to form correct transports.

endophage:
Closed by #819

Issue #710, opened 2016-05-02
Command to delete a notary repository locally and remotely


cyli:
`client.Client.DeleteTrustData` removes the TUF metadata locally.  We should expose this via the command line interface, and also provide a flag to delete the repository remotely as well.

For remote deletion:
- we want to limit this to someone who owns the root key, as opposed to anyone with push access

PR #819, opened 2016-07-08
Delete GUN functionality, requires *

riyazdf:
Hooks up the handler for deleting a gun to the remote store and client library.  Also changes the handler to require the admin `*` permission in the auth token.

I did some reworking of the transport code to make getting an admin permission more explicit, which should help with making creating a notary repo require the `*` permission as well.  This PR as-is just defines the `getAdminTransport` function to be used if one was ever to want to put together a valid request with the `*` permission, but does not use it directly -- I figured it would be nice to include this, but I'm happy to expand this PR to include CLI commands for deletion, or remove that function for the time being if we'd prefer.

Partially addresses #789 and provides groundwork for #710 in the CLI


client/client_test.go
```diff
@@ -3274,6 +3274,85 @@ func TestDeleteRepo(t *testing.T) {
...
+	defer ts.Close()
+
+	// Create and publish a repo to delete
+	repo, rootKeyID := initializeRepo(t, data.ECDSAKey, gun, ts.URL, false)
+	defer os.RemoveAll(repo.baseDir)
+
+	require.NoError(t, repo.Publish())
+
+	// Create another repo to ensure it stays intact
+	livingGun := "stayingAlive"
```
cyli: 👍 Yay, thanks for ensuring that everything in the server isn't blown away.  Also awesome name.

client/client_test.go
```diff
@@ -3274,6 +3274,85 @@ func TestDeleteRepo(t *testing.T) {
...
+	requireRepoHasExpectedMetadata(t, repo, data.CanonicalTargetsRole, false)
+	requireRepoHasExpectedMetadata(t, repo, data.CanonicalSnapshotRole, false)
+	requireRepoHasExpectedMetadata(t, repo, data.CanonicalTimestampRole, false)
+
+	// Assert keys for this repo still exist locally
+	requireRepoHasExpectedKeys(t, repo, rootKeyID, true)
+
+	// Try connecting to the remote store directly and make sure that no metadata exists for this gun
+	remoteStore, err := getRemoteStore(repo.baseURL, repo.gun, repo.roundTrip)
+	require.NoError(t, err)
```
cyli: Non-blocking - wondering if we should check for `store.ErrMetaNotFound` specifically?

riyazdf: good idea!  Done.

cmd/notary/tuf_test.go
```diff
@@ -33,7 +44,21 @@ func TestTokenAuth200Status(t *testing.T) {
 	s := httptest.NewServer(http.HandlerFunc(NotAuthorizedTestHandler))
 	defer s.Close()
 
-	auth, err := tokenAuth(s.URL, baseTransport, gun, readOnly)
+	auth, err := tokenAuth(s.URL, baseTransport, gun, readOnly, false)
+	require.NoError(t, err)
+	require.NotNil(t, auth)
+}
+
+func TestAdminTokenAuth200Status(t *testing.T) {
```
cyli: Totally non-blocking: good to have these here, but I'm not sure our existing auth tests actually do anything either other than reflect the 401 unauthorized handler. We don't actually test which permissions are asked for - I'm wondering if it's worth writing a mock jwt authorizer, so we can ensure that notary is asking for the correct permissions.  However, a lot of that stuff is tested in the distribution library we depend on too.  

What do you think?  I can file an issue if that plumbing would be a useful thing for us to test.

cyli:
LGTM pending CI!  Thanks for adding this so quickly!

cmd/notary/tuf_test.go
riyazdf: Yeah, definitely - I was wondering how we could better test for this, and @endophage might have more context around the token auth code and tests in distribution.  I think a tracking issue would be good 👍

client/client_test.go
endophage: Image of turtle dancing to staying alive

cmd/notary/tuf.go
```diff
@@ -568,14 +568,37 @@ func (ps passwordStore) Basic(u *url.URL) (string, string) {
...
+	base, err := getBaseTransport(config)
+	if err != nil {
+		return nil, err
+	}
+	return tokenAuth(trustServerURL, base, gun, readOnly, false)
+}
+
+// getAdminTransport returns an admin http.RoundTripper to be used for all http requests,
+// and can be used for deleting repos
+func getAdminTransport(config *viper.Viper, gun string) (http.RoundTripper, error) {
```
endophage: Unless I'm missing something this is a dupe of `getTransport` immediately above with the only difference being `readOnly` is implicitly `false`. Seems unnecessary, especially as it's untested.

endophage: oh, sorry, saw the last arg for `tokenAuth` inverted too. Maybe we just update `getTransport` to take another arg?

endophage: On third thoughts, maybe define an enum and make readOnly of that type, then get rid of adminAccess in `tokenAuth`

endophage:
Love how this came out with the enum!

LGTM!

PR #822, opened 2016-07-08
Only use delegation passphrase env var for non-base roles

riyazdf:
Part of the issue noted in #820 

### DIFF
--- a/cmd/notary/main.go
+++ b/cmd/notary/main.go
@@ -10,6 +10,7 @@ import (
 	"github.com/Sirupsen/logrus"
 	"github.com/docker/notary"
 	"github.com/docker/notary/passphrase"
+	"github.com/docker/notary/tuf/data"
 	"github.com/docker/notary/version"
 	homedir "github.com/mitchellh/go-homedir"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


cmd/notary/main.go
```diff
@@ -234,7 +235,8 @@ func getPassphraseRetriever() notary.PassRetriever {
 		// For delegation roles, we can also try the "delegation" alias if it is specified
 		// Note that we don't check if the role name is for a delegation to allow for names like "user"
 		// since delegation keys can be shared across repositories
-		if v := env["delegation"]; v != "" {
+		// This cannot be a base role or imported key, though.
+		if v := env["delegation"]; !data.IsBaseRole(alias) && !strings.Contains(alias, "imported ") && v != "" {
```
cyli: 👍  Nice catch for if the base roles aren't provided but delegation roles are!

cyli: I'm wondering if we should should specify a constant or otherwise provide some kind of utility function for this, so we ensure that all imported keys have a similarly-formatted alias?

cyli:
Thanks for fixing this bug!  I have one caveat about the word "imported" being special, but this LGTM other than that!

endophage:
Want to circle back on this once the import refactor is merged

riyazdf:
@endophage: rebased, doesn't handle the `"imported"` keyword since that was removed

endophage:
LGTM pending CI

Issue #797, opened 2016-06-22
Refactor server/signer tests to be able to test multiple DBs (like rethink)


cyli:
It'd be nice to be able to run our storage tests against actual DBs - while we use an ORM for sql-like DBs and test against a sqlite database, our ORM may not support some of the functionality we use.

And there is no fake rethink DB right now to test, so we'd have to test our rethink code against a real DB.

riyazdf:
Closing after @cyli's awesome work in https://github.com/docker/notary/pull/824, https://github.com/docker/notary/pull/854, and https://github.com/docker/notary/pull/863

PR #824, opened 2016-07-11
Set rethink and mysql server DB storage tests to run on circle

cyli:
Part of #797

This adds some basic rethinkDB TUF CRUD tests, and fixes the following issues:

1.  `MemStorage`'s `UpdateMany` previously did a partial update even if there was a conflict. Now we check for conflicts before updating.
2.  `MemStorage`'s `UpdateMany` allowed you to insert the same version twice.
3.  `RethinkDB`'s `UpdateMany` did a partial update even if there was a conflict, because the    deletion criteria value (for timestamp_checksum) was passed as a list not as strings.
4.  `RethinkDB`'s `Delete` did not actually delete metadata, because the deletion criteria value
    (for gun) was passed as a list and not as strings.

This PR:
- [x] factor out server MetaStore tests to take a database/store so we can have tests for every type of DB
- [x] set up tests to run on CircleCI

One or more future PRs will factor out tests for `NewTUFMetaStorage` to be db-testable, and do the same for the signer tests.


Makefile
```diff
@@ -130,16 +130,13 @@ test:
 	@echo
 	go test -tags "${NOTARY_BUILDTAGS}" $(TESTOPTS) $(PKGS)
 
-test-full: TESTOPTS =
```
riyazdf: Any reason for removing this target?  Admittedly, I don't use it much because it doesn't include `fmt` and `misspell`

cyli: The only reason is because CI doesn't use it, and doesn't have the full linting.  I can just add those, though.

cyli: (also, someone could just do `make vet lint test`)

riyazdf: yeah I'm fine with either removing it or bringing it in parity with circle.  Agree that the current version isn't particularly useful

server/storage/rethink_realdb_test.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/docker/notary/storage/rethinkdb"
+	"github.com/docker/notary/tuf/data"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/dancannon/gorethink.v2"
+)
+
+var tlsOpts = tlsconfig.Options{InsecureSkipVerify: true}
+
+func rethinkSessionSetup(t *testing.T) (*gorethink.Session, string) {
+	// Get the MYSQL connection string from an environment variable
```
riyazdf: nit: `MYSQL` should be `RethinkDB`

Makefile
cyli: Combined all the lint functionality into one function, as per https://github.com/docker/notary/pull/830#issuecomment-232515700

Makefile
```diff
@@ -90,41 +90,26 @@ ${PREFIX}/bin/static/notary:
...
-
-# Requires that the following:
-# go get -u github.com/client9/misspell/cmd/misspell
-#
-# be run first
-
-# misspell target, don't include Godeps, binaries, python tests, or git files
-misspell:
-	@echo "+ $@"
+	# mispell - requires that the following be run first:
```
riyazdf: typo: `misspell`

Makefile
```diff
@@ -90,41 +90,26 @@ ${PREFIX}/bin/static/notary:
...
-
-# Requires that the following:
-# go get -u github.com/gordonklaus/ineffassign
-#
-# be run first
-
-# ineffassign target, don't include Godeps, binaries, python tests, or git files
-ineffassign:
-	@echo "+ $@"
+	# inefdesign - requires that the following be run first:
```
riyazdf: typo: `ineffassign`

cyli: Bah, brain broken.  It's right there on the next line.  :)  Thanks!

server/storage/memory_test.go
```diff
@@ -7,38 +9,58 @@ import (
...
 )
 
-func TestUpdateCurrent(t *testing.T) {
-	s := NewMemStorage()
-	s.UpdateCurrent("gun", MetaUpdate{"role", 1, []byte("test")})
+func assertExpectedMemoryTUFMeta(t *testing.T, expected []StoredTUFMeta, s *MemStorage) {
+	counter := make(map[string]int)
+	for _, tufObj := range expected {
+		k := entryKey(tufObj.Gun, tufObj.Role)
+		gun, ok := s.tufMeta[k]
```
riyazdf: can we rename this to `gunList` or `versionList` for clarity?

buildscripts/dbtests.sh
```diff
@@ -0,0 +1,52 @@
...
+  mysql*)
+    db="mysql"
+    dbContainerOpts="--name mysql_tests mysql mysqld --innodb_file_per_table"
+    DBURL="server@tcp(mysql_tests:3306)/notaryserver?parseTime=True"
+    ;;
+  rethink*)
+    db="rethink"
+    dbContainerOpts="--name rethinkdb_tests rdb-01 --bind all --driver-tls-key /tls/key.pem --driver-tls-cert /tls/cert.pem"
+    DBURL="rethinkdb_tests"
+    ;;
```
riyazdf: non-blocking nit: would be nice to have a default error case, something like:

```
       *)
            echo $"Usage: $0 dbName"
            exit 1
```

Currently I get the following if I don't provide any arg
```
...
+ docker-compose -p dbtests_ -f development..yml kill
ERROR: .IOError: [Errno 2] No such file or directory: u'./development..yml'
```

cyli: Good point, thanks!

cyli: Added

riyazdf:
LGTM!  Thank you for all of your work on this :)

server/storage/sqlite_test.go
```diff
@@ -0,0 +1,27 @@
+// +build !mysqldb
```
endophage: should this also include `!rethinkdb`?

endophage:
Seems a little odd/confusing that the sqlite tests will not run if the mysql tests are run, but will run with the rethinkdb tests. Seems like they should either always run, or have their own build flag (rather than being `!<something else>`

Other than that, LGTM.

server/storage/sqlite_test.go
cyli: Unfortunately `sqldb_tests` has some sample config generation that's needed by `tuf_store_test`, which didn't get refactored in this PR.  I plan on doing that in the next PR.

That's why `sqlite3` is still needed because without that refactor, the rethinkdb tests still run the `tuf_store_test`, which depends on stuff in `sqldb_test`.

cyli:
@endophage I am fixing that in the next PR - I needed to refactor some other tests in order to fix so me dependencies in `sqldb_tests`, so they're being run by rethink right now.

I'm planning on :  no flags, sqlite3 tests run.  mysqldb flag - mysql tests run.  rethinkdb - rethink tests run.  Does that sound right?

endophage:
@cyli sounds perfect! Merging now

Issue #806, opened 2016-06-28
Cleanup import-export and key architecture


endophage:
Thought there was an issue, apparently not. As discussed in the past, key storage should be cleaned up to have clear logical divisions. At the top level, a `CryptoService` should deal in operations: sign, create, etc... `trustmanager.KeyStore` should speak Public/PrivateKey objects, and the `trustmanager.Storage` interface should speak in `[]byte`.

Import/Export should then operate against a either the `KeyStore`, or `Storage` interface (TBD, based on password requirements).

We're going to go about cleaning this up in 3 steps as trying to do it all at once turned into a big mess that prevented me getting it commited last time. In order the steps are:

  - [x] Remove all existing import/export functionality - PR #807 
  - [x] Refactor CryptoService, KeyStore, and Storage to strictly abide by the rules above, acting on operations, Public/PrivateKey objects, and []byte, respectively - PR #808 
  - [x] Re-implement import/export on top of this new, cleaner stack - PR #825

HuKeping:
Maybe we should put on a label like `breaking-change` with red background color here to remind people that it could be not any guarantee of backwards compatibility with importing already exported keys.

But since it was not on a v1.0 yet, it's fine to do so.

endophage:
@HuKeping great idea! Will create and add the label now

endophage:
@HuKeping @andrewhsu I've been talking with the team here at Docker and we think an easy way to re-implement import/export is to use `.pem` files which already support containing multiple unique PEM blocks.

Export will pack all the keys being exported into a single `.pem` file, each delineated as a unique PEM block. Import can take any number of `.pem` files and will unpack each individual PEM block into a separate file.

I propose the following command formats:

```
notary export     # with no further arguments, exports all private keys on the host
notary export --gun my_gun [--gun my_gun]    # online operation that exports all keys currently valid within the gun(s). Flag can be provided multiple times 
notary export --key <keyID> [--key <keyID>]...   # exports only the provided key IDs. Flag can be provided multiple times.

notary import file1.pem file2.pem ...
```

All `export` commands will output to `stdout` by default and will take a `-o` flag to provide an output file. Current I do not intend to allow mixing the `--gun` and `--key` flags, but I may make it possible depending on the emergent ease/difficulty found during implementation. That flexibility could also be added at a later time, there's no inherent reason why it shouldn't be possible.

HuKeping:
Sounds good and I also think the scenario that wants to export all keys of one gun and one/two keys of another is not that common.

PR #825, opened 2016-07-11
#806 Part 3: Reimplement import export

endophage:
Depends on #808 

~~Still need to write some tests for the new `cmd/notary/keys.go` code. Not sure my parsing of CLI flags is correct.~~

~~N.B. import will no longer import to a yubikey. More discussion required but I think this needs to be more explicit around _when_ we import to a yubikey, maybe even a separate command, because it hits the key management APIs at a different level.~~ Functionality has been brought up to parity :-)

This PR contains parts 1 and 2 which have been closed in favour of merging this PR alone (means only this PR has to be rebased, not all 3)


cmd/notary/keys.go
```diff
@@ -597,14 +386,86 @@ func (k *keyCommander) keyPassphraseChange(cmd *cobra.Command, args []string) er
...
+	)
+	config, err := k.configGetter()
+	if err != nil {
+		return err
+	}
+
+	if k.outFile == "" {
+		out = cmd.Out()
+	} else {
+		f, err := os.OpenFile(k.outFile, os.O_TRUNC|os.O_CREATE|os.O_WRONLY, notary.PrivKeyPerms)
```
riyazdf: nit: might be worth explaining the bitwise or over these flags to explain that we're creating a new file in write mode, and possibly truncating

riyazdf: Actually I'm happy with the usage comment, which explains this at a higher level

cmd/notary/keys.go
```diff
@@ -597,14 +386,86 @@ func (k *keyCommander) keyPassphraseChange(cmd *cobra.Command, args []string) er
...
+		if err != nil {
+			return err
+		}
+		defer f.Close()
+		out = f
+	}
+
+	directory := config.GetString("trust_dir")
+	fileStore, err := store.NewPrivateSimpleFileStore(directory, notary.KeyExtension)
```
riyazdf: This should work fine, but is there a reason why we don't scope this down further to `<"trust_dir">/private`?

endophage: oh! good catch, I thought `NewPrivateSimpleFileStore` was adding that for some reason

endophage: Ah, we do it in NewKeyFileStore

cmd/notary/main_test.go
```diff
@@ -159,10 +159,6 @@ var exampleValidCommands = []string{
 	"key list",
 	"key rotate repo snapshot",
 	"key generate rsa",
-	"key backup tempfile.zip",
-	"key export e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 backup.pem",
```
riyazdf: I think it'd be good to bring back the `key import` and `key export` commands here.  This just tests for invalid number of arguments, if I remember correctly.

utils/keys.go
```diff
@@ -0,0 +1,115 @@
...
+type importStore interface {
+	Set(string, []byte) error
+}
+
+// ExportKeysByGUN exports all keys filtered to a GUN
+func ExportKeysByGUN(to io.Writer, s exportStore, gun string) error {
+	keys := s.ListFiles()
+	for _, k := range keys {
+		dir := filepath.Dir(k)
+		if dir == gun { // must be full GUN match
```
riyazdf: I think we need to chomp off any part of the filepath preceding `<some_path>/<"trust_dir">/private` in order to match against a GUN

riyazdf: Actually disregard that comment - I think we're ok because of the `NewPrivateSimpleFilestore`.  It'd be good to have a test that initializes multiple notary repos and then exports a select few of the GUNs' keys

utils/keys.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+	}
+	return nil
+}
+
+// ExportKeysByID exports all keys matching the given ID
+func ExportKeysByID(to io.Writer, s exportStore, ids []string) error {
+	want := make(map[string]struct{})
+	for _, id := range ids {
+		want[id] = struct{}{}
```
riyazdf: this is nifty, `unsafe.Sizeof(struct{}{})` is 0 👍

utils/keys.go
```diff
@@ -0,0 +1,115 @@
...
+		return err
+	}
+	var (
+		writeTo string
+		toWrite []byte
+	)
+	for block, rest := pem.Decode(data); block != nil; block, rest = pem.Decode(rest) {
+		loc, ok := block.Headers["path"]
+		if !ok || loc == "" {
+			continue // don't know where to copy this key. Skip it.
```
riyazdf: might be worth adding a debug log here in case users get confused why non-notary exported keys can't be imported due to the missing `"path"` header

utils/keys.go
endophage: I've seen it used elsewhere and while i don't have an official source, I've been told it's the idiomatic way to do sets (or at least set membership) in golang.

riyazdf:
this cleanup is awesome, thank you for tackling it!

I think it'd be good to add some of the following tests around notary repositories:
- [ ] Exporting keys from an initialized notary repository on disk
- [ ] Importing over existing notary repository keys when paths collide with pre-existing key files
- [ ] Exporting keys from a published notary repository, importing the same keys into a blank filesystem and successfully publishing to the same notary repository

storage/filestore.go
```diff
@@ -0,0 +1,222 @@
...
+// NewPrivateKeyFileStorage initializes a new filestore for private keys, appending
+// the notary.PrivDir to the baseDir.
+func NewPrivateKeyFileStorage(baseDir, fileExt string) (*FilesystemStore, error) {
+	baseDir = filepath.Join(baseDir, notary.PrivDir)
+	return NewFileStore(baseDir, fileExt, notary.PrivKeyPerms)
+}
+
+// NewPrivateSimpleFileStore is a wrapper to create an owner readable/writeable
+// _only_ filestore
+func NewPrivateSimpleFileStore(baseDir, fileExt string) (*FilesystemStore, error) {
```
cyli: With the modification of import/export to use `NewPrivateKeyFileStorage`, I think this constructor is no longer used anywhere else except `filestore_test.go`, where I don't think permissions are getting checked (we could use one of the other constructors?).

cmd/notary/keys.go
```diff
@@ -597,14 +386,91 @@ func (k *keyCommander) keyPassphraseChange(cmd *cobra.Command, args []string) er
...
+	directory := config.GetString("trust_dir")
+	fileStore, err := store.NewPrivateKeyFileStorage(directory, notary.KeyExtension)
+	if err != nil {
+		return err
+	}
+	if len(k.exportGUNs) > 0 {
+		if len(k.exportKeyIDs) > 0 {
+			return fmt.Errorf("Only the --gun or --key flag may be provided, not a mix of the two flags")
+		}
+		for _, gun := range k.exportGUNs {
```
cyli: Non-blocking: do we want to dedupe the GUNs, or do you just live with the output if you pass the same GUN multiple times?

utils/keys.go
```diff
@@ -0,0 +1,115 @@
...
+
+// ExportKeys copies a key from the store to the io.Writer
+func ExportKeys(to io.Writer, s exportStore, from string) error {
+	// get PEM block
+	k, err := s.Get(from)
+	if err != nil {
+		return err
+	}
+
+	// parse PEM blocks if there are more than one
```
cyli: nitpick:  could you elaborate on the use case for multiple PEM blocks per key in this comment?  e.g. x509 keys with intermediate certificates?

utils/keys.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	return nil
+}
+
+// ImportKeys expects an io.Reader containing one or more PEM blocks.
+// It reads PEM blocks one at a time until pem.Decode returns a nil
+// block.
+// Each block is written to the subpath indicated in the "path" PEM
+// header. If the file already exists, the file is truncated. Multiple
+// PEMs with the same "path" header are appended together.
```
cyli: nitpick: Maybe "Multiple PEMs with the same "path" header that are adjacent to each other will be appended together?

Since if they are separated by PEM with a different path, the second one will overwrite the first.

cmd/notary/keys.go
endophage: was just planning to let the user live with the output... it'll still import correctly, just the export and import will both take a little longer.

utils/keys.go
endophage: I didn't have a good use case up front but I wanted to future proof. It seems unlikely but maybe we'll use additional PEM blocks for metadata or something like that and this will export and import them all correctly.

HuKeping:
Tons of code changed, I think it's a good choice to just wait it to be merged for myself :smile:

cmd/notary/keys.go
```diff
@@ -80,55 +56,69 @@ var cmdKeyPasswdTemplate = usageTemplate{
 	Long:  "Changes the passphrase for the key with the given keyID.  Will require validation of the old passphrase.",
 }
 
+var cmdKeyImportTemplate = usageTemplate{
+	Use:   "import pemfile [ pemfile ... ]",
+	Short: "Imports all keys from all provided .pem files",
+	Long:  "Imports all keys from all provided .pem files by reading each PEM block from the file and writing that block to a unique object in the local keystore",
```
cyli: Do we want to mention that hardware storage is preferred in the long usage description?

cmd/notary/keys.go
```diff
@@ -80,55 +56,69 @@ var cmdKeyPasswdTemplate = usageTemplate{
...
+}
+
+var cmdKeyExportTemplate = usageTemplate{
+	Use:   "export",
+	Short: "Exports all keys from all local keystores. Can be filtered using the --key and --gun flags.",
+	Long:  "Exports all keys from all local keystores. Which keys are exported can be restricted by using the --key or --gun flags. By default the result is sent to stdout, it can be directed to a file with the -o flag.",
```
cyli: In the long usage description, should we mention that only on-disk keys can be exported (not hardware keys)?

utils/keys.go
```diff
@@ -0,0 +1,129 @@
...
+	// get PEM block
+	k, err := s.Get(from)
+	if err != nil {
+		return err
+	}
+
+	// parse PEM blocks if there are more than one
+	for block, rest := pem.Decode(k); block != nil; block, rest = pem.Decode(rest) {
+		// add from path in a header for later import
+		block.Headers["path"] = from
```
cyli: Non-blocking - since there seems to be no objection yet to #838, should we change this to GUN in this PR or are we waiting to do that in a future PR?

riyazdf:
I did a quick export - remove keys - re-import - publish test and it worked great 👍 

LGTM!!  Thank you for all of your hard work on this PR and the parts that preceded it :)

If you don't mind, I'd like to file a couple of follow-up issues: importing delegation keys without the `path` PEM header (`notary delegation import`?), and more end-to-end integration tests around exporting/importing.

cyli:
Thank you for all your work on this refactor!  LGTM!

endophage:
@riyazdf goal is to make the `path` header irrelevant through flattening of the key store

HuKeping:
Hi,  is there any plan to add new docs for this, I found I can no longer use the command like `notary key import delegation.key --role=targets/delegation_1`

Issue #815, opened 2016-07-02
Get goreportcard to 100% across the board


endophage:
[Report Card](https://goreportcard.com/report/github.com/docker/notary)

TODOs:

  - notary/cmd/notary/delegations.go
    - [ ] Line 111: warning: cyclomatic complexity 22 of function (*delegationCommander).delegationRemove() is high (> 15) (gocyclo)
    - [ ] Line 211: warning: cyclomatic complexity 21 of function (*delegationCommander).delegationAdd() is high (> 15) (gocyclo)
  - notary/cmd/notary/tuf.go
    - [ ] Line 130: warning: cyclomatic complexity 16 of function (*tufCommander).tufAddByHash() is high (> 15) (gocyclo)
  - notary/client/client.go
    - [ ] Line 575: warning: cyclomatic complexity 22 of function (*NotaryRepository).publish() is high (> 15) (gocyclo)
    - [ ] Line 176: warning: cyclomatic complexity 18 of function (*NotaryRepository).Initialize() is high (> 15) (gocyclo)
    - [ ] Line 880: warning: cyclomatic complexity 17 of function (*NotaryRepository).RotateKey() is high (> 15) (gocyclo)
  - notary/tuf/tuf.go
    - [ ] Line 134: warning: cyclomatic complexity 17 of function (*Repo).RemoveBaseKeys() is high (> 15) (gocyclo)
  - notary/server/handlers/default.go
    - [ ] Line 176: warning: cyclomatic complexity 16 of function getKeyHandler() is high (> 15) (gocyclo)
notary/storage/rethinkdb/bootstrap.go
    - [ ] Line 53: warning: cyclomatic complexity 16 of function (Table).create() is high (> 15) (gocyclo)
notary/cmd/notary/keys.go
    - [ ] Line 326: warning: cyclomatic complexity 20 of function (*keyCommander).keysImport() is high (> 15) (gocyclo)
  - notary/trustmanager/yubikey/yubikeystore.go
    - [ ] Line 448: warning: cyclomatic complexity 20 of function yubiListKeys() is high (> 15) (gocyclo)
notary/tuf/builder_test.go
    - [ ] Line 355: warning: cyclomatic complexity 18 of function TestGetConsistentInfo() is high (> 15) (gocyclo)
  - notary/server/handlers/validation_test.go
    - [x] Line 607: warning: timestamp assigned and not used (ineffassign)
    - [x] Line 607: warning: targets assigned and not used (ineffassign)

[...]

PR #830, opened 2016-07-11
Include ineffassign in CI, fix ineffassign in code

riyazdf:
Adds the `ineffassign` target to our Makefile and circle tests, indirectly part of #815 


cyli:
LGTM!  Thank you for adding this!

cyli:
Non-blocking, but wondering if it'd be worth adding a `full-lint` endpoint that just combines all of these?  That way we can run all the lints that are run on the CI with one command.

endophage:
Agree on adding a `full-lint` (could probably just be `lint` ??)

LGTM though. Merging

PR #831, opened 2016-07-12
Implement distribution/notifications into Notary

jchorl:


server/handlers/default.go
```diff
@@ -108,7 +110,19 @@ func atomicUpdateHandler(ctx context.Context, w http.ResponseWriter, r *http.Req
...
+	if !ok {
+		return errors.ErrUnknown.WithDetail("could not get notification broadcaster")
+	}
+	sourceRecordKey := ctx.Value("sourceRecord")
+	sourceRecord, ok := sourceRecordKey.(dNotifications.SourceRecord)
+	if !ok {
+		return errors.ErrUnknown.WithDetail("could not get notification sourceRecord")
+	}
+
+	return notifications.Notify("GUN update", sourceRecord, "Josh", r, broadcaster)
```
jchorl: Need to update the username that we pass to Notify, might try to parse it out of the JWT claims

server/handlers/default.go
```diff
@@ -108,7 +111,21 @@ func atomicUpdateHandler(ctx context.Context, w http.ResponseWriter, r *http.Req
...
+	}
+	sourceRecordKey := ctx.Value("sourceRecord")
+	sourceRecord, ok := sourceRecordKey.(dNotifications.SourceRecord)
+	if !ok {
+		return errors.ErrUnknown.WithDetail("could not get notification sourceRecord")
+	}
+
+	// try to get a username, which may be blank if the req is not authd
+	username, _ := ctx.Value(auth.UserNameKey).(string)
+	return notifications.Notify("GUN update", gun, sourceRecord, username, r, broadcaster)
```
riyazdf: since it's possible for this to error, should the notification be best effort (and we log appropriately if the notification failed)?

cmd/notary/tuf.go
```diff
@@ -604,6 +604,11 @@ func (ps passwordStore) Basic(u *url.URL) (string, string) {
 	return username, password
 }
 
+func (ps passwordStore) RefreshToken(*url.URL, string) string {
```
riyazdf: nit: it'd be nice to add simple unit tests for these two refresh functions

notifications/notifications.go
```diff
@@ -0,0 +1,50 @@
...
+
+type Endpoint struct {
+	Name      string        `json:"name"`      // identifies the endpoint in the registry instance.
+	Disabled  bool          `json:"disabled"`  // disables the endpoint
+	URL       string        `json:"url"`       // post url for the endpoint.
+	Headers   http.Header   `json:"headers"`   // static headers that should be added to all requests
+	Timeout   time.Duration `json:"timeout"`   // HTTP timeout
+	Threshold int           `json:"threshold"` // circuit breaker threshold before backing off on failure
+	Backoff   time.Duration `json:"backoff"`   // backoff duration
+	CA        string        `json:"ca"`        // acceptable CA to avoid x509 unknown authority error
```
riyazdf: clarifying question: do we want `omitempty` on any of these fields?

notifications/notifications_test.go
```diff
@@ -0,0 +1,47 @@
...
+		Method:     "PUT",
+		Header: http.Header{
+			"User-Agent": []string{"ua"},
+		},
+	}
+	rec := newRequestRecord(req)
+	require.Equal(t, context.RemoteAddr(req), rec.Addr)
+	require.Equal(t, req.Host, rec.Host)
+	require.Equal(t, req.Method, rec.Method)
+	require.Equal(t, rec.UserAgent, "ua")
```
riyazdf: nit: `"ua"` should go first since it's the `expected` arg

notifications/notifications_test.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func TestCreateEvent(t *testing.T) {
+	req, err := http.NewRequest("GET", "https://url.com", bytes.NewBufferString("hi"))
+	require.Nil(t, err)
+
+	evt := createEvent("NewThing", "GUN", dNotifications.SourceRecord{Addr: "localhost"}, "username", req)
+	require.NotEmpty(t, evt.ID)
+	require.NotEqual(t, evt.Timestamp, time.Time{})
```
riyazdf: it might be more useful to check if the time is within some window of `time.Now()` instead of checking against epoch

jchorl: I would rather not make tests time dependent. As long as the value is actually populated, it should be correct. It's a tradeoff, verifying times is not really possible to do correctly because we can't really mock out `time.Now()`.

cyli: Since the `createEvent` function is unexported, you could inject a time getting function into it.  E.g. `createEvent` could take a `func()` that returns a `time.Time{}`, and you can inject a fake `time.Now()` function that way.

jchorl: This is clever

notifications/notifications.go
```diff
@@ -0,0 +1,54 @@
...
+	Disabled  bool          // disables the endpoint
+	URL       string        // post url for the endpoint.
+	Headers   http.Header   // static headers that should be added to all requests
+	Timeout   time.Duration // HTTP timeout
+	Threshold int           // circuit breaker threshold before backing off on failure
+	Backoff   time.Duration // backoff duration
+	CA        string        // acceptable CA to avoid x509 unknown authority error
+}
+
+// Notify notifies registered endpoints that an event has occured
```
riyazdf: I think `misspell` is catching a typo for "occurred" on this line

NathanMcCauley:
We shouldn't implement this method because it is inherently lossy. We probably want to revive the change feed if folks need to get this kind of info out of notary.

CC @endophage

endophage:
Superseded by #901

PR #834, opened 2016-07-14
Wait for indexes in rethink table.wait

riyazdf:
We should call [indexWait](https://rethinkdb.com/api/javascript/index_wait/) in addition to just the table [wait](https://rethinkdb.com/api/javascript/wait/) to ensure that all indexes for the table are ready.

I can add unit-tests to the infrastructure in #824, though the rethink integration test checks out

### DIFF
--- a/storage/rethinkdb/bootstrap.go
+++ b/storage/rethinkdb/bootstrap.go
@@ -47,6 +47,16 @@ func (t Table) wait(session *gorethink.Session, dbName string) error {
 		resp.Close()
 	}
 
+	if err != nil {
+		return err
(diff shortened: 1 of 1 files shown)


cyli:
@riyazdf Thanks for noticing this!  And this LGTM!  And sure, unit tests would be great.  I'm not sure how to trigger indexes being slow to create, though.

endophage:
LGTM, good find!

I'm wondering if we can use a test wrapper for rethink that creates the indexes one by one with random delays between them? Work for a later time if it's even possible.

PR #842, opened 2016-07-15
Reissue certs

riyazdf:
Reissues certs for notary-server, notary-signer, secure.example.com (which were all expired).  These will expire again in a year, and use the same private key material.

Also reissued root-ca and intermediate-ca, and removed private key material upon creation.   These certs will expire in 10 years.

Closes #840 

Signed-off-by: Riyaz Faizullabhoy <riyaz.faizullabhoy@docker.com>

PR #843, opened 2016-07-18
Update certs

HuKeping:
This patch re-generate the root CA and the intermediate CA, the
trust chain is:
root-ca.crt -> intermediate-ca.crt -> notary-server.crt/notary-signer.crt.

This patch introduce the private key for root CA and the intermediate CA
which can be used to sign csr from user.

And also the pre-generated csr for user who would like to try the trust
chain.


HuKeping:
Oh, I missed #842 when I creat this patch, but still wonder if we could introduce the csr and the keys. With these file people can try to sign there own intermediate CA and then their own certs for server and signer.  It might be some help.

riyazdf:
Hi @HuKeping, in #842 we added a `regenerateTestingCerts.sh` script that you could modify to not remove the csr and keys -- in general that script is very prescriptive to our fixture certificate use-case but we'd be more than happy to accept pull requests to make it more generic :)

We would prefer not to keep any unnecessary private key material checked into source control wherever possible, so I'll close this PR for now.

PR #848, opened 2016-07-18
Refactor passphrase to cache passphrases individually

### LINKED ISSUE #524: Refactor passphrase retriever to separate targets and snapshot

riyazdf:
Caches each passphrase in the passphrase retriever separately, such that snapshot and targets passphrases are no longer necessarily shared and that we can also cache delegation role passphrases.

Note from previous discussions about how this affects the docker-client integration: the passphrase refactor will affect old repos since we'll ask for targets and snapshots separately, but it will not affect new repos created with Docker 1.11 since the server handles the snapshot key.  Also,  the environment variables for automation remain the same.

Closes #524, carries #541

### DIFF
--- a/passphrase/passphrase.go
+++ b/passphrase/passphrase.go
@@ -50,13 +50,10 @@ func PromptRetriever() notary.PassRetriever {
 }
 
 type boundRetriever struct {
-	in                             io.Reader
-	out                            io.Writer
-	aliasMap                       map[string]string
-	userEnteredTargetsSnapshotPass bool
-	targetsSnapshotPass            string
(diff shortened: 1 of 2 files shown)


endophage:
LGTM

cyli:
Tried this out locally - it successfully asked for my targets and snapshot passwords, and setting the environment variables still correctly works.  LGTM!  Thanks for cleaning this up!

PR #854, opened 2016-07-20
More server DB tests

cyli:
This finishes the refactor in #824 of the server storage tests refactor to take a DB type.

In addition, this fixes:

1.  Deleting a GUN in mysql - previously, we soft-deleted a gun, preventing us from being able to re-use the namespace.  We now hard-delete.
1.  The RethinkDB server health check needs to fail if the session user does not have the correct permissions to access the requisite tables.

### DIFF
--- a/server/storage/sql_models.go
+++ b/server/storage/sql_models.go
@@ -14,7 +14,6 @@ type TUFFile struct {
 
 // TableName sets a specific table name for TUFFile
 func (g TUFFile) TableName() string {
-	// NOTE: if this value changes, please also change it in SQLStorage.Delete
 	return "tuf_files"
 }
 
--- a/server/storage/sql_models.go
+++ b/server/storage/sql_models.go
@@ -14,6 +14,7 @@ type TUFFile struct {
 
 // TableName sets a specific table name for TUFFile
 func (g TUFFile) TableName() string {
+	// NOTE: if this value changes, please also change it in SQLStorage.Delete
 	return "tuf_files"
 }
 
--- a/server/storage/sqldb_test.go
+++ b/server/storage/sqldb_test.go
@@ -283,7 +283,7 @@ func TestSQLDBCheckHealthTableMissing(t *testing.T) {
 	require.Error(t, err, "Cannot access table:")
 }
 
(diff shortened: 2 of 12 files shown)


server/storage/sql_models.go
```diff
@@ -14,6 +14,7 @@ type TUFFile struct {
 
 // TableName sets a specific table name for TUFFile
 func (g TUFFile) TableName() string {
+	// NOTE: if this value changes, please also change it in SQLStorage.Delete
```
riyazdf: Is it worth making the table names constants in the `storage` package?

cyli: Unfortunately we didn't want to string-interpolate an SQL statement.  :|  Which  means hard-coding it in the DELETE string.  Probably the right way to fix this is to submit a patch to GORM upstream to enable hard-delete?  

Although there's this `scope.Search.Unscoped` check in the delete function that I don't really know how to use:  https://github.com/jinzhu/gorm/blob/master/callback_delete.go#L29 - maybe that does something interesting.

cyli: Although hmm... let me try:  https://github.com/jinzhu/gorm/issues/342

server/storage/sqldb_test.go
```diff
@@ -319,9 +283,9 @@ func TestDBCheckHealthTableMissing(t *testing.T) {
 	require.Error(t, err, "Cannot access table:")
 }
 
-// TestDBCheckHealthDBCOnnection asserts that if the DB is not connectable, the
+// TestSQLDBCheckHealthDBCOnnection asserts that if the DB is not connectable, the
```
riyazdf: nit: can we lowercase the "o" in `COnnection`?

server/storage/sql_models.go
cyli: Yes that seems to work, thanks!

riyazdf:
this is awesome, LGTM pending CI!

server/storage/sqldb_test.go
cyli: Oops!  Will fix

endophage:
LGTM!

endophage:
needs a rebase

PR #855, opened 2016-07-20
command to purge all keys from all delegations below a starting point

### LINKED ISSUE #696: Remove key from all delegations
We need a command that removes a key by ID from all delegations it appears in. If the key had permissions to sign any base roles, those (for now at least) should be handled via key rotations of the roles.

Use case: contributor leaves project, their key should be purged from any roles it could sign.

endophage:
~~Need to write tests.~~

Closes #696 

### DIFF
--- a/client/helpers.go
+++ b/client/helpers.go
@@ -40,7 +40,7 @@ func applyChangelist(repo *tuf.Repo, cl changelist.Changelist) error {
 		if err != nil {
 			return err
 		}
-		isDel := data.IsDelegation(c.Scope())
+		isDel := data.IsDelegation(c.Scope()) || data.IsWildDelegation(c.Scope())
 		switch {
 		case c.Scope() == changelist.ScopeTargets || isDel:
 			err = applyTargetsChange(repo, c)
--- a/client/helpers.go
+++ b/client/helpers.go
@@ -92,6 +92,10 @@ func changeTargetsDelegation(repo *tuf.Repo, c changelist.Change) error {
 		if err != nil {
 			return err
 		}
+		if data.IsWildDelegation(c.Scope()) {
+			return repo.PurgeDelegationKeys(c.Scope(), td.RemoveKeys)
+		}
+
 		delgRole, err := repo.GetDelegationRole(c.Scope())
 		if err != nil {
 			return err
--- a/client/helpers.go
+++ b/client/helpers.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"strings"
 	"time"
 
 	"github.com/Sirupsen/logrus"
@@ -116,10 +115,6 @@ func changeTargetsDelegation(repo *tuf.Repo, c changelist.Change) error {
 			removeTUFKeyIDs = append(removeTUFKeyIDs, canonicalToTUFID[canonID])
 		}
 
-		// If we specify the only keys left delete the role, else just delete specified keys
-		if strings.Join(delgRole.ListKeyIDs(), ";") == strings.Join(removeTUFKeyIDs, ";") && len(td.AddKeys) == 0 {
-			return repo.DeleteDelegation(c.Scope())
-		}
 		err = repo.UpdateDelegationKeys(c.Scope(), td.AddKeys, removeTUFKeyIDs, td.NewThreshold)
 		if err != nil {
 			return err
--- a/cmd/notary/delegations.go
+++ b/cmd/notary/delegations.go
@@ -32,7 +32,7 @@ var cmdDelegationRemoveTemplate = usageTemplate{
 	Long:  "Remove KeyID(s) from the specified Role delegation in a specific Global Unique Name.",
 }
 
-var cmdDelegationRemoveKeysTemplate = usageTemplate{
+var cmdDelegationPurgeKeysTemplate = usageTemplate{
 	Use:   "purge [ GUN ]",
 	Short: "Remove KeyID(s) from all delegations it is found in.",
 	Long:  "Remove KeyID(s) from all delegations it is found in, for which the signing keys are available. Warnings will be printed for delegations that cannot be updated.",
@@ -58,9 +58,9 @@ func (d *delegationCommander) GetCommand() *cobra.Command {
 	cmd := cmdDelegationTemplate.ToCommand(nil)
 	cmd.AddCommand(cmdDelegationListTemplate.ToCommand(d.delegationsList))
 
-	cmdRemDelgKeys := cmdDelegationRemoveKeysTemplate.ToCommand(d.delegationRemoveKeys)
-	cmdRemDelgKeys.Flags().StringSliceVar(&d.keyIDs, "key", nil, "Delegation keys to be removed from the GUN")
-	cmd.AddCommand(cmdRemDelgKeys)
+	cmdPurgeDelgKeys := cmdDelegationPurgeKeysTemplate.ToCommand(d.delegationPurgeKeys)
+	cmdPurgeDelgKeys.Flags().StringSliceVar(&d.keyIDs, "key", nil, "Delegation keys to be removed from the GUN")
+	cmd.AddCommand(cmdPurgeDelgKeys)
 
 	cmdRemDelg := cmdDelegationRemoveTemplate.ToCommand(d.delegationRemove)
 	cmdRemDelg.Flags().StringSliceVar(&d.paths, "paths", nil, "List of paths to remove")
@@ -75,10 +75,10 @@ func (d *delegationCommander) GetCommand() *cobra.Command {
 	return cmd
 }
 
-func (d *delegationCommander) delegationRemoveKeys(cmd *cobra.Command, args []string) error {
-	if len(args) > 1 {
+func (d *delegationCommander) delegationPurgeKeys(cmd *cobra.Command, args []string) error {
+	if len(args) != 1 {
 		cmd.Usage()
-		return fmt.Errorf("Please provide a Global Unique Name as an argument to remove")
+		return fmt.Errorf("Please provide a single Global Unique Name as an argument to remove")
 	}
 
 	if len(d.keyIDs) == 0 {
--- a/cmd/notary/delegations.go
+++ b/cmd/notary/delegations.go
@@ -32,6 +32,12 @@ var cmdDelegationRemoveTemplate = usageTemplate{
 	Long:  "Remove KeyID(s) from the specified Role delegation in a specific Global Unique Name.",
 }
 
+var cmdDelegationRemoveKeysTemplate = usageTemplate{
+	Use:   "purge [ GUN ]",
+	Short: "Remove KeyID(s) from all delegations it is found in.",
+	Long:  "Remove KeyID(s) from all delegations it is found in, for which the signing keys are available. Warnings will be printed for delegations that cannot be updated.",
+}
+
 var cmdDelegationAddTemplate = usageTemplate{
 	Use:   "add [ GUN ] [ Role ] <X509 file path 1> ...",
 	Short: "Add a keys to delegation using the provided public key X509 certificates.",
@@ -45,11 +51,16 @@ type delegationCommander struct {
 
 	paths                         []string
 	allPaths, removeAll, forceYes bool
+	keyIDs                        []string
 }
 
 func (d *delegationCommander) GetCommand() *cobra.Command {
 	cmd := cmdDelegationTemplate.ToCommand(nil)
 	cmd.AddCommand(cmdDelegationListTemplate.ToCommand(d.delegationsList))
+
+	cmdRemDelgKeys := cmdDelegationRemoveKeysTemplate.ToCommand(d.delegationRemoveKeys)
+	cmdRemDelgKeys.Flags().StringSliceVar(&d.keyIDs, "key", nil, "Delegation keys to be removed from the GUN")
+	cmd.AddCommand(cmdRemDelgKeys)
 
 	cmdRemDelg := cmdDelegationRemoveTemplate.ToCommand(d.delegationRemove)
 	cmdRemDelg.Flags().StringSliceVar(&d.paths, "paths", nil, "List of paths to remove")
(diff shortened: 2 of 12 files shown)


endophage:
@cyli @riyazdf ready for review

tuf/data/roles_test.go
```diff
@@ -105,44 +105,61 @@ func TestErrInvalidRole(t *testing.T) {
...
+		path.Join("  "+CanonicalTargetsRole, "level1"):                                      f,
+		path.Join(CanonicalTargetsRole, "level1"+"  "):                                      f,
+		path.Join(CanonicalTargetsRole, "white   space"+"level2"):                           f,
+		path.Join(CanonicalTargetsRole, strings.Repeat("x", 256-len(CanonicalTargetsRole))): f,
+
+		// true checks
+		path.Join(CanonicalTargetsRole, "level1"):                     tr,
+		path.Join(CanonicalTargetsRole, "level1", "level2", "level3"): tr,
+		path.Join(CanonicalTargetsRole, "under_score"):                tr,
+		path.Join(CanonicalTargetsRole, "hyphen-hyphen"):              tr,
```
riyazdf: thanks for this refactoring 👍

tuf/data/roles.go
```diff
@@ -96,6 +97,16 @@ func IsBaseRole(role string) bool {
 	return false
 }
 
+// IsWildDelegation determines if a role represents a valid wildcard delegation
```
riyazdf: Fun name, I had to include this
wild_delegation

tuf/data/roles.go
```diff
@@ -96,6 +97,16 @@ func IsBaseRole(role string) bool {
...
+// path, i.e. targets/*, targets/foo/*.
+// The wildcard may only appear as the final part of the delegation and must
+// be a whole segment, i.e. targets/foo* is not a valid wildcard delegation.
+func IsWildDelegation(role string) bool {
+	base := filepath.Dir(role)
```
riyazdf: I think we should be using `path` here to keep consistent with `IsDelegation` and other delegation checking logic (though on linux/OSX this shouldn't be an issue)

tuf/tuf.go
```diff
@@ -392,11 +392,81 @@ func (tr *Repo) UpdateDelegationKeys(roleName string, addKeys data.KeyList, remo
...
+
+	purgeKeys := func(tgt *data.SignedTargets, validRole data.DelegationRole) interface{} {
+		var (
+			deleteCandidates    []string
+			err                 error
+			canSign             bool
+			needsUpdateCantSign bool
+		)
+		err = tr.VerifyCanSign(validRole.Name)
+		canSign = err == nil
```
riyazdf: nit: can we collapse these above two lines into:`canSign = tr.VerifyCanSign(validRole.Name) == nil`?
I think it reads a little cleaner if we don't plan on doing anything else with the error

tuf/tuf.go
```diff
@@ -392,11 +392,81 @@ func (tr *Repo) UpdateDelegationKeys(roleName string, addKeys data.KeyList, remo
...
+		for id, key := range tgt.Signed.Delegations.Keys {
+			var (
+				canonID string
+				ok      bool
+			)
+			if canonID, ok = tufIDToCanon[id]; !ok {
+				canonID, err = utils.CanonicalKeyID(key)
+				if err != nil {
+					// TODO: should we return here or just log?
```
riyazdf: I think that either is acceptable in the context of purging keys, but I'd prefer to error in case something was somehow corrupted on that delegation

cmd/notary/delegations.go
```diff
@@ -32,6 +32,12 @@ var cmdDelegationRemoveTemplate = usageTemplate{
 	Long:  "Remove KeyID(s) from the specified Role delegation in a specific Global Unique Name.",
 }
 
+var cmdDelegationPurgeKeysTemplate = usageTemplate{
+	Use:   "purge [ GUN ]",
+	Short: "Remove KeyID(s) from all delegations it is found in.",
```
cyli: minor nit:  Maybe just "from all delegation roles"?

cmd/notary/delegations.go
```diff
@@ -45,12 +51,17 @@ type delegationCommander struct {
...
 	allPaths, removeAll, forceYes bool
+	keyIDs                        []string
 }
 
 func (d *delegationCommander) GetCommand() *cobra.Command {
 	cmd := cmdDelegationTemplate.ToCommand(nil)
 	cmd.AddCommand(cmdDelegationListTemplate.ToCommand(d.delegationsList))
 
+	cmdPurgeDelgKeys := cmdDelegationPurgeKeysTemplate.ToCommand(d.delegationPurgeKeys)
+	cmdPurgeDelgKeys.Flags().StringSliceVar(&d.keyIDs, "key", nil, "Delegation keys to be removed from the GUN")
```
cyli: minor nit:  "Delegation key IDs to be removed..."?

tuf/tuf.go
```diff
@@ -392,11 +392,81 @@ func (tr *Repo) UpdateDelegationKeys(roleName string, addKeys data.KeyList, remo
...
+				validRole.Name,
+				validRole.Name,
+			)
+			return nil
+		}
+		if len(deleteCandidates) == 0 {
+			// none of the interesting keys were present
+			return nil
+		}
+		// delete candidate keys from all roles. Keep only roles that still have keys
```
riyazdf: I think the second half of this comment needs to be updated since we're keeping all roles

cmd/notary/delegations_test.go
```diff
@@ -14,23 +14,36 @@ import (
...
 
-var testTrustDir = "trust_dir"
-
-func setup() *delegationCommander {
+func setup(trustDir string) *delegationCommander {
 	return &delegationCommander{
 		configGetter: func() (*viper.Viper, error) {
 			mainViper := viper.New()
-			mainViper.Set("trust_dir", testTrustDir)
+			mainViper.Set("trust_dir", trustDir)
```
cyli: \o/  Yay injecting all the things!

cmd/notary/delegations.go
```diff
@@ -64,6 +75,53 @@ func (d *delegationCommander) GetCommand() *cobra.Command {
...
+		d.retriever,
+		trustPin,
+	)
+	if err != nil {
+		return err
+	}
+
+	err = nRepo.RemoveDelegationKeys("targets/*", d.keyIDs)
+	if err != nil {
+		return fmt.Errorf("failed to remove delegation: %v", err)
```
cyli: maybe "failed to remove keys from delegations"?

tuf/tuf_test.go
```diff
@@ -166,6 +166,144 @@ func TestUpdateDelegations(t *testing.T) {
 	require.False(t, ok, "no empty targets file should be created for deepest delegation")
 }
 
+func TestPurgeDelegationsKeyFromTop(t *testing.T) {
```
riyazdf: in this test or a new one, can we test what happens if we:

 - try to purge a key ID that isn't signed into any delegation
 - remove the targets private key from our keystore, and try to purge a valid key ID from delegations

tuf/tuf.go
riyazdf: Also, we should update the docs in `advanced_usage.md` to reflect this change (there's currently this sentence: `If you specify all key IDs currently in the delegation role, you will delete the role entirely.`)

cmd/notary/delegations.go
endophage: ah yes!

tuf/data/roles.go
cyli: Do we also want to add a check that the last character of the role is `*`?  `path.Base(role)` will trim off any trailing slashes.

tuf/tuf_test.go
endophage: Can add first test here. Second one probably belongs at the cmd/notary package.

tuf/data/roles.go
endophage: @cyli wasn't aware of that with `path.Base`, will update to account for that. Good catch!

tuf/data/roles_test.go
endophage: Just wanted consistency within the file :-)

tuf/tuf.go
endophage: Can do!

tuf/data/roles.go
cyli: Oop, looks like the `path.Dir` will then count the `*` as part of the dir, if there are trailing slashes, so hopefully `IsDelegation` will fail that one.  But possibly slashes in the middle might be ignored?  https://play.golang.org/p/ChBOllvy42

cmd/notary/delegations_test.go
endophage: also using temp dirs rather than `./trust_dir`

tuf/tuf.go
```diff
@@ -326,16 +326,16 @@ func delegationUpdateVisitor(roleName string, addKeys data.KeyList, removeKeys,
 			}
 		}
 		// We didn't find the role earlier, so create it only if we have keys to add
+		// that are also sufficient to meet the threshold.
```
cyli: Is this comment still true?  It seems like we're creating the role even if there are no keys.

tuf/tuf.go
```diff
@@ -392,11 +392,81 @@ func (tr *Repo) UpdateDelegationKeys(roleName string, addKeys data.KeyList, remo
...
+	tufIDToCanon := make(map[string]string)
+
+	purgeKeys := func(tgt *data.SignedTargets, validRole data.DelegationRole) interface{} {
+		var (
+			deleteCandidates    []string
+			err                 error
+			canSign             bool
+			needsUpdateCantSign bool
+		)
+		err = tr.VerifyCanSign(validRole.Name)
```
cyli: IRL discussion - we could probably iterate through the keys first, find any deletable key IDs, and then if there are any, then check if this role is signable.  That might simplify some of the nested if statements later on.

Also a non-blocking thing:  `VerifyCanSign` currently walks the delegation tree to find the role before figuring out if we have signing keys.  Do we want to factor out a private function that can just take a role and find out if there are any signing keys?  That way we don't have to double-walk, since this walking function already gives us the delegation role to check.

cyli: Actually, a reason for modifying the `VerifyCanSign` - we probably want to give it a copy of the delegation role with the deletable keys removed, since if you attempt to remove your own signing key, it should skip removing from the role, right?

cyli: Oops NM about that last comment.  Sorry I think I got confused.  Since the tiles are mutated in this function before the walker appends the roles to the list of roles to walk, the key will have already been removed so the check is good.

tuf/tuf.go
endophage: I've updated the advanced_usage doc to be accurate per this PR, I'll need to update it again when I push the witness PR as that will provide a way to remedy the situation.

tuf/data/roles_test.go
```diff
@@ -105,44 +105,61 @@ func TestErrInvalidRole(t *testing.T) {
...
+		CanonicalTargetsRole:   f,
+		CanonicalSnapshotRole:  f,
+		CanonicalTimestampRole: f,
+		"foo":   f,
+		"foo/*": f,
+		path.Join(CanonicalRootRole, "*"):           f,
+		path.Join(CanonicalSnapshotRole, "*"):       f,
+		path.Join(CanonicalTimestampRole, "*"):      f,
+		path.Join(CanonicalTargetsRole, "*", "foo"): f,
+		path.Join(CanonicalTargetsRole, "*", "*"):   f,
```
cyli: Can also add failure cases for trailing slashes after the `*`, and extra slashes between the rest of the delegation and the `*`?

endophage: Done

riyazdf:
thank you for adding the tests, LGTM!  Documentation CI errors are unrelated

client/helpers.go
```diff
@@ -92,6 +91,10 @@ func changeTargetsDelegation(repo *tuf.Repo, c changelist.Change) error {
 		if err != nil {
 			return err
 		}
+		if data.IsWildDelegation(c.Scope()) {
```
cyli: This is never getting exercised, which is weird because `cmd/notary/integration_test.go`'s `TestPurge` should exercise this bit of code.  

In `applyChangelist`, where we switch based on whether the scope is a delegation (L43) I think we also have to  change that line to be `isDel := data.IsDelegation(c.Scope()) || data.IsWildDelegation(c.Scope())`

endophage: Hmmm, you're absolutely right, and that's making me wonder why the test in cmd/notary/delegations_test.go is even passing.

Going to add some more checks to that test to try and make it fail, then will update this and see if it passes again.

endophage: Oh, I see now. I'm calling the purge command but not publish so it's not actually applying the change. I must have got sidetracked half way through the test and forgotten to come back to it.

cmd/notary/integration_test.go
```diff
@@ -1291,3 +1292,51 @@ func TestMain(m *testing.M) {
...
+	require.NoError(t, err)
+
+	// remove targets key and ensure we fail to publish
+	os.RemoveAll(filepath.Join(tempDir, notary.PrivDir, notary.NonRootKeysSubdir))
+
+	_, err = runCommand(t, tempDir, "delegation", "purge", "gun", "--key", keyID)
+	require.NoError(t, err)
+
+	_, err = runCommand(t, tempDir, "-s", server.URL, "publish", "gun")
+	require.Error(t, err)
```
cyli: When I printed out the error here, I got:  
`Metadata hierarchy is incomplete: no snapshot was included in update and server does not hold current snapshot key for repository`

Do we always publish a new snapshot even if nothing's changed?  

Can we just try to remove the targets key?  I think that as the feature is currently implemented, this publish shouldn't fail if we remove the targets key - since we don't have the signing key, the changelist (the key purge) will not be applied, and there should be no changes to the repo?  I forget exactly how publish works if there are no changes.

cyli:
Awesome, thanks for tackling this!  LGTM!

PR #863, opened 2016-07-21
Per-DB signer tests

cyli:
This adds specific mysql DB and rethink DB tests for the signer.

It also does the following:

- factors the caching part of the `KeyStore` implementations into a wrapper `KeyStore`
- renames `KeyDBStore` to `SQLKeyDBStore` and add SQL types to the Gorm model (otherwise the schema is not created correctly in MySQL tests)
- changes `SQLKeyDBStore`'s `RemoveKey` to always succeed, even if the key doesn't exist
- store the `RethinkKeyDBStore`'s public and private bytes as `[]byte`, otherwise rethink won't know that it's binary data and will mangle the result - this doesn't matter for the private key bytes, since we encrypt it in such a way that it's base64 encoded, but I changed it anyway in case we want to change encryption method in the future
- fixes `RethinkKeyDBStore`'s key rotation function - it now uses the same get query as `GetKey`
- update the signer healthcheck to fail if we don't have permission to access the DB

### DIFF
--- a/signer/keydbstore/cachedkeystore.go
+++ b/signer/keydbstore/cachedkeystore.go
@@ -1,0 +1,75 @@
+package keydbstore
+
+import (
+	"sync"
+
+	"github.com/docker/notary/trustmanager"
+	"github.com/docker/notary/tuf/data"
+)
+
+// Note:  once trustmanager's file KeyStore has been flattened, this can be moved to trustmanager
+
+type cachedKeyStore struct {
+	trustmanager.KeyStore
+	lock       *sync.Mutex
+	cachedKeys map[string]*cachedKey
+}
+
+type cachedKey struct {
+	role string
+	key  data.PrivateKey
+}
+
+// NewCachedKeyStore returns a new trustmanager.KeyStore that includes caching
+func NewCachedKeyStore(baseStore trustmanager.KeyStore) trustmanager.KeyStore {
+	return &cachedKeyStore{
+		KeyStore:   baseStore,
+		lock:       &sync.Mutex{},
+		cachedKeys: make(map[string]*cachedKey),
+	}
+}
+
+// AddKey stores the contents of a private key. Both role and gun are ignored,
+// we always use Key IDs as name, and don't support aliases
+func (s *cachedKeyStore) AddKey(keyInfo trustmanager.KeyInfo, privKey data.PrivateKey) error {
+	if err := s.KeyStore.AddKey(keyInfo, privKey); err != nil {
+		return err
+	}
+
+	// Add the private key to our cache
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	s.cachedKeys[privKey.ID()] = &cachedKey{
+		role: keyInfo.Role,
+		key:  privKey,
+	}
+
+	return nil
+}
+
+// GetKey returns the PrivateKey given a KeyID
+func (s *cachedKeyStore) GetKey(keyID string) (data.PrivateKey, string, error) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	cachedKeyEntry, ok := s.cachedKeys[keyID]
+	if ok {
+		return cachedKeyEntry.key, cachedKeyEntry.role, nil
+	}
+
+	// retrieve the key from the underlying store and put it into the cache
+	privKey, role, err := s.KeyStore.GetKey(keyID)
+	if err == nil {
+		// Add the key to cache
+		s.cachedKeys[privKey.ID()] = &cachedKey{key: privKey, role: role}
+	}
+	return privKey, role, err
+}
+
+// RemoveKey removes the key from the keyfilestore
+func (s *cachedKeyStore) RemoveKey(keyID string) error {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	delete(s.cachedKeys, keyID)
+	return s.KeyStore.RemoveKey(keyID)
+}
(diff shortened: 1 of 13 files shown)


signer/keydbstore/cachedkeystore.go
```diff
@@ -0,0 +1,75 @@
...
+		return cachedKeyEntry.key, cachedKeyEntry.role, nil
+	}
+
+	// retrieve the key from the underlying store and put it into the cache
+	privKey, role, err := s.KeyStore.GetKey(keyID)
+	if err == nil {
+		// Add the key to cache
+		s.cachedKeys[privKey.ID()] = &cachedKey{key: privKey, role: role}
+	}
+	return privKey, role, err
```
riyazdf: non-blocking nit: while this shouldn't make any difference in actual values returned, it could be clearer if this returned `data.PrivateKey{}, "", err`

signer/keydbstore/rethink_keydbstore.go
```diff
@@ -77,7 +79,7 @@ func rdbPrivateKeyFromJSON(data []byte) (interface{}, error) {
 // PrivateKeysRethinkTable is the table definition for notary signer's key information
 var PrivateKeysRethinkTable = rethinkdb.Table{
 	Name:             RDBPrivateKey{}.TableName(),
-	PrimaryKey:       RDBPrivateKey{}.KeyID,
+	PrimaryKey:       "key_id",
```
riyazdf: do we need the hardcoded string instead of the constant here?

signer/keydbstore/rethink_realkeydbstore_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+	var rows []RDBPrivateKey
+	require.NoError(t, res.All(&rows))
+
+	return rows
+}
+
+func TestRethinkKeyCanOnlyBeAddedOnce(t *testing.T) {
+	dbStore, _ := rethinkDBSetup(t, "signerAddTests")
+	// defer cleanup()
```
riyazdf: should we uncomment this line?

signer/keydbstore/sql_keydbstore.go
```diff
@@ -17,58 +16,56 @@ const (
...
-	PassphraseAlias string `sql:"not null"`
-	Public          string `sql:"not null"`
-	Private         string `sql:"not null"`
+	KeyID           string `sql:"type:varchar(255);not null;unique;index:key_id_idx"`
+	EncryptionAlg   string `sql:"type:varchar(255);not null"`
+	KeywrapAlg      string `sql:"type:varchar(255);not null"`
+	Algorithm       string `sql:"type:varchar(50);not null"`
+	PassphraseAlias string `sql:"type:varchar(50);not null"`
+	Public          string `sql:"type:blob;not null"`
+	Private         string `sql:"type:blob;not null"`
```
riyazdf: thank you for specifying these 👍

riyazdf:
this is great, thank you for adding these fixes and tests!  LGTM pending CI

signer/keydbstore/rethink_realkeydbstore_test.go
cyli: Ah yes, thanks!  Missed that one.

signer/keydbstore/rethink_keydbstore.go
cyli: `RDBPrivateKey{}`.KeyID was just the empty string, because it initialized the empty struct and so that value is just the zero value.  If we wanted to declare a constant and use it to set the field value, and hence use the constant here, or if we wanted to pull out the field value, I think we're going to need some reflection to do so, or better yet:  https://github.com/oleiade/reflections.

If we wanted to add that though, maybe it can be a different PR?

riyazdf: Ah yes, sorry I missed that.  I'm fine with a hardcoded string for now and following up with reflection in a separate PR

signer/keydbstore/cachedkeystore.go
```diff
@@ -0,0 +1,76 @@
...
+	cachedKeys map[string]*cachedKey
+}
+
+type cachedKey struct {
+	role string
+	key  data.PrivateKey
+}
+
+// NewCachedKeyStore returns a new trustmanager.KeyStore that includes caching
+func NewCachedKeyStore(baseStore trustmanager.KeyStore) trustmanager.KeyStore {
```
endophage: I'm thinking this could also replace a chunk of `trustmanager.GenericKeyStore` down the road, particularly if we get around to separating the password stuff out. We'd end up with something like `NewCacheKeyStore( NewGenericKeyStore( NewPasswdStorage( NewFileStore(keyDirectory) )))`   :-D

signer/keydbstore/cachedkeystore.go
```diff
@@ -0,0 +1,76 @@
...
+		role: keyInfo.Role,
+		key:  privKey,
+	}
+
+	return nil
+}
+
+// GetKey returns the PrivateKey given a KeyID
+func (s *cachedKeyStore) GetKey(keyID string) (data.PrivateKey, string, error) {
+	s.lock.Lock()
```
endophage: we can move this lock down to line 62. Reading from maps is thread safe, we may just end up doing some duplicate `s.KeyStore.GetKey` calls. However, if that call is slow for some reason, it'll be more efficient to not lock this whole method

endophage:
LGTM! Thank you so much for adding the rethink tests!

Issue #852, opened 2016-07-19
delegation roles doesn't publish successfully


HuKeping:
There is a `targets` role which has three delegations roles `targets/distribution`, `targets/docker`, `targets/notary` under the GUN `huawei.com`:
```
$ notary delegation list huawei.com

          ROLE                    PATHS                                        KEY IDS                                THRESHOLD  
---------------------------------------------------------------------------------------------------------------------------------
  targets/distribution   huawei.com/distribution   bcfe06daf6db5916baa4f1389e32ebf282c17593f0b67e996fb46b4714fe99cd   1          
  targets/docker         huawei.com/docker         fff2024979a5282ca697b3b231f87499fbe8082cc2be7555329aa5d1bead5007   1          
  targets/notary         huawei.com/notary         6f5c271304f639c25df19b2da725d876ab04e62525cece6a01050c8cee15e4ff   1 
``` 

After I publish content(the README.MD of each repo) with delegation roles respectively , I get:
```
$ notary list huawei.com

               NAME                                             DIGEST                                SIZE (BYTES)           ROLE          
-------------------------------------------------------------------------------------------------------------------------------------------
  huawei.com/distribution/readme   d28f42179340667755eb8ae29101437ce5c60cec099d8d4820e4f4dd67b663f1   4984           targets/distribution  
  huawei.com/docker/readme         495cefcebf8b2ed4de71a4e54f2b11c2533114fe9e884dc1be4502ac2ba6d7de   13755          targets/docker        
  huawei.com/notary/readme         9d9e890af64dd0f44b8a1538ff5fa0511cc31bf1ab89f3a3522a9a581a70fad8   6563           targets/notary 
```

Then I publish the `huawei.com/docker/readme` with the `targets` role, which I get:
```
$ notary list huawei.com

               NAME                                             DIGEST                                SIZE (BYTES)           ROLE          
-------------------------------------------------------------------------------------------------------------------------------------------
  huawei.com/distribution/readme   d28f42179340667755eb8ae29101437ce5c60cec099d8d4820e4f4dd67b663f1   4984           targets/distribution  
  huawei.com/docker/readme         495cefcebf8b2ed4de71a4e54f2b11c2533114fe9e884dc1be4502ac2ba6d7de   13755          targets               
  huawei.com/notary/readme         9d9e890af64dd0f44b8a1538ff5fa0511cc31bf1ab89f3a3522a9a581a70fad8   6563           targets/notary 
```

[...]

PR #864, opened 2016-07-21
Ensure that we do not clobber in-memory key id data on delegation list

riyazdf:
Brought up by @HuKeping in https://github.com/docker/notary/issues/852#issuecomment-234137903, I noticed that `translateDelegationsToCanonicalIDs` was mistakenly editing in-memory metadata by replacing TUF key IDs with canonical key IDs, which would cause nested delegations to have issues when building into `data.DelegationRole` structs.

Changed pass-by-reference to pass-by-value and added a test that publishes a nested delegation and ensures that `delegation list` behaves as expected.

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -2325,6 +2325,25 @@ func TestPublishTargetsDelegationSuccessNeedsToDownloadRoles(t *testing.T) {
 	// delegation parents all get signed
 	ownerRec.requireAsked(t, []string{data.CanonicalTargetsRole, "targets/a"})
 
+	// assert both delegation roles appear to the other repo in a call to GetDelegationRoles
+	delgRoleList, err := delgRepo.GetDelegationRoles()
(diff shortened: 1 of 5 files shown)


HuKeping:
I personally tried this and it works as expected, thanks for the quick fixing!

riyazdf:
@HuKeping thank you for trying this out 👍

cyli:
Great catch!  Thanks for fixing this so fast, and for the test!  Non-blocking, but would it make sense to add a moe specific unit test for `GetDelegationRoles` to ensure that it does not alter any of the repo metadata? 

Otherwise LGTM!

HuKeping:
Pre-expansion patch live now :) @cyli

riyazdf:
@cyli: great idea, added on to a delegation publishing test

endophage:
LGTM

Issue #846, opened 2016-07-18
Server Key Rotation: Schema updates


endophage:
Signer should do all key management.

PR #869, opened 2016-07-23
Consolidate the signer tests

cyli:
The RPC api tests and the GRPC client tests were somewhat redundant, and I wanted to make sure they could definitely talk to each other easily.

Cherry-picked from the changes for #846, because I want to check how much coverage this change by itself has.


Godeps/Godeps.json
```diff
@@ -504,18 +504,6 @@
 			"Rev": "f368244301305f414206f889b1735a54cfc8bde8"
 		},
 		{
-			"ImportPath": "github.com/spf13/cobra/cobra",
```
cyli: `github.com/spf13/cobra/cobra` and `github.com/spf13/cobra/cobra/cmd` don't seem to be imported from anywhere even from our vendored libraries, and they're also a different version.  This might be causing Godep issues.

`github.com/spf13/cobra/doc` just doesn't seem to be used anywhere.

cyli:
There is some weirdness with the way we generate coverage - it's saying some lines aren't covered that are definitely being covered, but it looks like the client and rpc packages aren't specified as dependencies of the package the new tests are in, so no coverage output is generated for those packages.  Trying to figure out how to fix that.

Godeps/Godeps.json
riyazdf: Hm that's weird -- I'm fine with deleting those deps in this PR or a separate one.  It seems like the actual source files for these deps are still in our `vendor` directory though.

cyli: Ah ok.  I'll split this off into a separate PR and remove them, then.

proto/signer.proto
```diff
@@ -22,9 +22,6 @@ service KeyManagement {
 service Signer {
   // Sign calculates a cryptographic signature using the Key associated with a KeyID and returns the signature
   rpc Sign(SignatureRequest) returns (Signature) {}
-
-  // CheckHealth returns the HealthStatus with the service
-  rpc CheckHealth(Void) returns (HealthStatus) {}
```
riyazdf: Why are we removing the `CheckHealth` from this interface?  I realize that we have another health check with the `KeyManagement` service but wasn't sure if we still wanted both

cyli: It just seemed redundant because they were the same server, and seem like they always will be.  I don't mind having it in both, or splitting it off entirely into its own service (I think grpc now has its own health service too, but we might have to upgrade).

riyazdf: Makes sense, I agree.  It might be worth noting somewhere that the signer's healthcheck is hitting the `KeyManagement` interface level in case we expand on the signer in the future

riyazdf:
LGTM!  I'm fine merging this as-is, and diagnosing codecov in a separate PR

Godeps/Godeps.json
cyli: Thanks for fixing this in #874!

endophage:
LGTM. Looks like we have a lot more tests to write on the signer but this is great noise reduction on what we already have.

PR #872, opened 2016-07-25
Flattened the filekeystore: Post0.4

### LINKED ISSUE #838: Discussion: Can we fully flatten the keystore?
We've discussed it internally on a number of occasions, I'd like to have the discussion with a larger group. 

## Current State

Keys are stored in a directory called `private` within you configured `trust_dir`. Root keys are stored in `private/root_keys/*.key`, delegation keys are stored in the top level of `private/tuf_keys/*.key`, and all other keys (targets, snapshot), are stored in `private/tuf_keys/<gun>/*.key`.

## Proposal

I'd like to flatten the entire key storage. All keys would be stored at `private/<CanonicalKeyID>.key`. The only interesting piece of information, the GUN of targets and snapshot keys, would be moved to the PEM headers, where we already store the role.

## Results

This will greatly simplify a lot of key management code. I would implement migration code that would run when filestorage for keys is initialized to move any existing keys to the new location, allowing us to fully remove all other code that has to be aware of the current pathing.

However, it would mean the scope of some keys becomes more opaque to a user digging around in the `private` directory. I have some big PRs open right now to do a lot of cleanup around key storage and import/export so my desire would be to ensure we have sufficient CLI functionality that nobody ever has to dig into `private`. We should not become beholden to, and unable to change, notary internal data structure decisions.

cc (a bunch of possibly interested parties) @HuKeping @andrewhsu @dnwake @ecordell @diogomonica @NathanMcCauley @riyazdf @cyli @avaid96

avaid96:
- Stores all keys under the private directory

- Key data now contains the gun that was otherwise inferred by filepath

- Root keys delegations keys etc don't get a GUN

- Backwards compatibility to 0.1 (already there) and 0.3 (added tests and fixtures) has been tested for 

closes #838


riyazdf:
jenkins, test this please.

avaid96:
jenkins, test this please.

tuf/utils/x509.go
```diff
@@ -409,6 +409,17 @@ func KeyToPEM(privKey data.PrivateKey, role string) ([]byte, error) {
 			"role": role,
 		}
 	}
+	if gun != "" {
+		headers = map[string]string{
```
cyli: Totally non-blocking - can we just set the header value for the given key in two if blocks, rather than redeclaring it in three if blocks?

cmd/notary/integration_test.go
```diff
@@ -856,11 +858,11 @@ func TestClientDelegationsPublishing(t *testing.T) {
 	require.NotContains(t, output, "No delegations present in this repository.")
 
 	// remove the targets key to demonstrate that delegates don't need this key
-	keyDir := filepath.Join(tempDir, "private", "tuf_keys")
-	require.NoError(t, os.Remove(filepath.Join(keyDir, "gun", targetKeyID+".key")))
+	keyDir := filepath.Join(tempDir, "private")
+	require.NoError(t, os.Remove(filepath.Join(keyDir, targetKeyID+".key")))
 
 	// Note that we need to use the canonical key ID, followed by the base of the role here
```
riyazdf: we should update this comment to explain that delegation keys need not have any PEM headers to specify the role, given that the filename is the canonical key ID

fixtures/compatibility/README.md
```diff
@@ -2,6 +2,8 @@ This directory contains sample repositories from different versions of Notary cl
 
 Notary client makes no guarantees of future-compatibility though (that is, repositories produced by newer clients may not be able to be read by old clients.)
 
+Backwards compatibility has been tested in ../../client/backwards_compatibility_test.go
```
riyazdf: nit: use backticks (like `) to pretty print this.  Also I think just `client/backwards_compatibility_test.go` is enough detail

client/backwards_compatibility_test.go
```diff
@@ -131,6 +131,66 @@ func Test0Dot1RepoFormat(t *testing.T) {
 	require.NoError(t, err)
 }
 
+// We can read and publish from notary0.1 repos
```
riyazdf: nit: this comment should say 0.3

trustmanager/keystore.go
```diff
@@ -265,49 +266,54 @@ func copyKeyInfoMap(keyInfoMap map[string]KeyInfo) map[string]KeyInfo {
...
 }
 
 // getKeyRole finds the role for the given keyID. It attempts to look
 // both in the newer format PEM headers, and also in the legacy filename
-// format. It returns: the role, whether it was found in the legacy format
-// (true == legacy), and an error
-func getKeyRole(s Storage, keyID string) (string, bool, error) {
+// format. It returns: the role, whether it was found in the legacy(0.1) format
+// (true == legacy),  whether it was found in a notary0.3 formal (true == notary0.3) and an error
+func getKeyRole(s Storage, keyID string) (string, bool, bool, error) {
```
riyazdf: instead of returning two booleans for each legacy format (which could potentially grow), how about instead returning the path of the key relative to the base `private` directory as a string?  Then the `legacy` and `notary03` checks can go away, and all accesses to that path are handled generically?

tuf/utils/x509.go
```diff
@@ -443,6 +454,9 @@ func EncryptPrivateKey(key data.PrivateKey, role, passphrase string) ([]byte, er
 		return nil, fmt.Errorf("unable to encrypt key - invalid PEM file produced")
 	}
 	encryptedPEMBlock.Headers["role"] = role
+	if gun != "" {
```
riyazdf: nit: Is there a reason why we couldn't handle empty guns like we do for empty roles?  Since we're doing this special gun handling, should the `"role"` header have an if block as well in case delegation keys do not have roles?

tuf/utils/x509_test.go
```diff
@@ -124,15 +124,15 @@ func TestKeyOperations(t *testing.T) {
...
 	require.NoError(t, err)
 
 	// Encrypt our EC Key
-	encryptedECKey, err := EncryptPrivateKey(ecKey, "root", "ponies")
+	encryptedECKey, err := EncryptPrivateKey(ecKey, "root", "ilove", "ponies")
 	require.NoError(t, err)
 
 	// Encrypt our RSA Key
-	encryptedRSAKey, err := EncryptPrivateKey(rsaKey, "root", "ponies")
+	encryptedRSAKey, err := EncryptPrivateKey(rsaKey, "root", "ilove", "ponies")
```
riyazdf: love 'em

trustmanager/keystore.go
avaid96: So the reason I thought against it is that we're doing a GUN lookup from the keyInfoMap currently. Yes, we could infer the GUN from the PEM/filepath but I thought we should take a decision on that one so I have left it alone for now. I'm open to either way, just thought I'd follow the format we were using for now but I'm open to changing it to follow your suggestion

tuf/utils/x509.go
avaid96: This is where we set the PEM headers. As per conversation with @endophage, we decided to have no gun field instead of an empty gun in case we have a root/delegation key. So, I wrote the if block so that a `gun: ""` field doesn't get added to these keys where it would be irrelevant

avaid96:
made all recommended changes that I reacted to with a thumbs up and commented on those that I think we should further discuss. Not pushing the changes files yet because I don't want the comments to be squashed

tuf/utils/x509.go
endophage: @riyazdf as far as Go is concerned, it'll basically be the same, but if somebody for some reason manually looks inside the PEM, it's mildly confusing to have a header with an empty value so I suggested it would be better to no include it at all if it would otherwise be `""`

trustmanager/keystore.go
endophage: @avaid96 with a migration on initialization of a key filestore, there should only be one way to get the role, as any legacy version should have been translated already.

avaid96: As of now, this doesn't include a migration but includes checks that allow full backwards compatibility to older versions. I am open to writing migration on initialization as well but was wondering if that may cause any issues in docker content trust?

endophage: If it migrates the 2 known structures to the new one it will cover DCT and will mean we can remove the legacy checks everywhere else.

avaid96: sounds good, I'll get working on adding in that migration

avaid96:
TODO: with a migration on initialization of a key filestore, there should only be one way to get the role, as any legacy version should have been translated already

tuf/utils/x509.go
riyazdf: Agreed that we don't want PEM headers with empty values.  Should we do the same for the `"role"` value then?

I've traced through the code and it seems like this might be a condition on a passphrase change for imported delegation keys since that calls `AddKey` on a key that could potentially have no role, and it would be nice to keep this consistent between GUNs and roles.  I can envision delegation keys in particular not needing a role in the PEM headers, since a key could be used across multiple delegation roles.

avaid96: Agreed, I am for this change. Let's wait on @endophage and take a call

trustmanager/keystore_test.go
```diff
@@ -89,16 +88,18 @@ func TestKeyStoreInternalState(t *testing.T) {
...
+
 		require.NoError(t, err, "could not generate PEM")
 
 		// write the key file to the correct location
-		// Pretend our GUN is docker.com/notary
-		keyPath := filepath.Join(tempBaseDir, "private", getSubdir(role))
-		if role == data.CanonicalTargetsRole || role == data.CanonicalSnapshotRole {
-			keyPath = filepath.Join(keyPath, gun)
-		}
+		keyPath := filepath.Join(tempBaseDir, "private")
```
cyli: nit:  could we use the `notary.PrivDir` constant here?

cyli: (I think there's a couple more usages of "private" below too)

cyli:
Apologies! I'm sure this would work fine with the new code - could we add a delegation to the fixtures so we have a delegation key imported into the key store?  It should be fine, since it could have had any role anyway, but just wanted to make sure that something with that that particular key can still be used.

tuf/utils/x509.go
endophage: @riyazdf I thought for delegations we set the role to `"user"`. If that's not the case then yeah, we should make the header not appear.

cyli: @endophage IIRC previously we could import a delegation key as any role, so it can be "user" but it could by anything (our docs recommend user).  You couldn't really import a key such that it had no role header.

riyazdf: @cyli's right that we required a role when importing a key if one did not exist in the PEM headers, so in our docs we suggested `"user"`, but it wasn't hardcoded anywhere.

I'd like to make sure that the new `notary key import` can import delegation keys directly from openssl so maybe we can have a default role value or no header at all?

trustmanager/keystore_test.go
riyazdf: We can also use `notary.RootKeysSubdir` and `notary.NonRootKeysSubdir` instead of `"root_keys"` and `"tuf_keys"`, respectively

avaid96:
Thanks for this comment @cyli 

"Apologies! I'm sure this would work fine with the new code - could we add a delegation to the fixtures so we have a delegation key imported into the key store? It should be fine, since it could have had any role anyway, but just wanted to make sure that something with that that particular key can still be used.

This looks great! So clean!"

it picked up a case that I hadn't looked for. Made the necessary changes with a new fixture that uses delegations and tests that with the new keystore

cyli:
jenkins, ok to test

cmd/notary/integration_test.go
```diff
@@ -2269,7 +2293,7 @@ func TestClientKeyImport(t *testing.T) {
 	// if there is hardware available, root will only be on hardware, and not
 	// on disk
 	assertNumKeys(t, tempDir, 2, 4, !rootOnHardware())
-	_, err = os.Open(filepath.Join(tempDir, "private", notary.NonRootKeysSubdir, newKeyID+".key"))
+	_, err = os.Open(filepath.Join(tempDir, notary.PrivDir, notary.NonRootKeysSubdir, newKeyID+".key"))
```
cyli: Should the `notary.NonRootKeysSubdir` be removed?  Since now it should import, if imported successfully, to just the `notary.PrivDir` directory?

avaid96: thanks for catching that, since it was checking that the file didn't exist I missed that one 😄  will update it with the next batch of changes

trustmanager/keystore.go
```diff
@@ -40,7 +42,46 @@ func NewKeyFileStore(baseDir string, p notary.PassRetriever) (*GenericKeyStore,
...
+		} else if strings.HasPrefix(fileDir, notary.NonRootKeysSubdir) {
+			fileDir = strings.TrimPrefix(fileDir, notary.NonRootKeysSubdir)
+			if fileDir != "" {
+				block.Headers["gun"] = fileDir[1:]
+			}
+			if strings.Contains(keyID, "_") {
+				role := strings.Split(keyID, "_")[1]
+				keyID = strings.TrimSuffix(keyID, "_"+role)
+				block.Headers["role"] = role
+			}
```
cyli: Should we be removing the gun here if the role is not a base role?  (so that delegation keys have no guns?)

trustmanager/keystore.go
```diff
@@ -40,7 +42,46 @@ func NewKeyFileStore(baseDir string, p notary.PassRetriever) (*GenericKeyStore,
...
+		}
+		var keyPEM bytes.Buffer
+		_ = pem.Encode(&keyPEM, block)
```
cyli: Do we want to comment here that we are ignore the error return value because since block came from decoding the PEM bytes in the first place, and all we're doing is adding some headers, it shouldn't error in the first place?

trustmanager/keystore.go
cyli: IRL discussion with @avaid96:  as a counter argument, since it's a migration, shouldn't we be preserving any information that is already in the previous keystore? (e.g. if a delegation role has a GUN)

trustmanager/keystore.go
```diff
@@ -218,30 +217,14 @@ func (s *GenericKeyStore) ListKeys() map[string]KeyInfo {
...
-
 	// being in a subdirectory is for backwards compatibliity
-	err = s.store.Remove(filepath.Join(getSubdir(role), name))
+	err := s.store.Remove(keyID)
 	if err != nil {
 		return err
 	}
 
-	// Remove this key from our keyInfo map if we removed from our filesystem
 	delete(s.keyInfoMap, filepath.Base(keyID))
```
cyli: is the `filepath.Base` needed anymore here?  (shouldn't hurt anything, but the keyID shouldn't have any more path segments in it)

trustmanager/keystore.go
```diff
@@ -218,31 +218,15 @@ func (s *GenericKeyStore) ListKeys() map[string]KeyInfo {
...
-	}
-
 	delete(s.cachedKeys, keyID)
 
-	name := keyID
-	if legacy {
-		name = keyID + "_" + role
-	}
-
 	// being in a subdirectory is for backwards compatibliity
```
cyli: Non-blocking: I think this comment is no longer relevant, since it's no longer in a subdirectory

trustmanager/keystore.go
```diff
@@ -263,55 +246,43 @@ func copyKeyInfoMap(keyInfoMap map[string]KeyInfo) map[string]KeyInfo {
...
-	keyID, role, gun := inferKeyInfoFromKeyPath(filename)
-	if role == "" {
-		block, _ := pem.Decode(pemBytes)
-		if block == nil {
-			return "", KeyInfo{}, fmt.Errorf("could not decode PEM block for key %s", filename)
-		}
-		if keyRole, ok := block.Headers["role"]; ok {
-			role = keyRole
-		}
+	//keyID, role, gun := inferKeyInfoFromKeyPath(filename)
```
cyli: Should this commented out line be removed?

trustmanager/keystore.go
```diff
@@ -263,55 +246,43 @@ func copyKeyInfoMap(keyInfoMap map[string]KeyInfo) map[string]KeyInfo {
...
 	}
 	return keyID, KeyInfo{Gun: gun, Role: role}, nil
 }
 
 // getKeyRole finds the role for the given keyID. It attempts to look
 // both in the newer format PEM headers, and also in the legacy filename
-// format. It returns: the role, whether it was found in the legacy format
-// (true == legacy), and an error
-func getKeyRole(s Storage, keyID string) (string, bool, error) {
+// format. It returns: the role, whether it was found in the legacy(0.1) format
```
cyli: Should this comment be updated since the legacy booleans are no longer returned?

utils/keys.go
```diff
@@ -97,6 +92,8 @@ func ExportKeys(to io.Writer, s Exporter, from string) error {
 // header. If the file already exists, the file is truncated. Multiple
 // adjacent PEMs with the same "path" header are appended together.
 func ImportKeys(from io.Reader, to []Importer, fallbackRole string, fallbackGun string, passRet notary.PassRetriever) error {
+	// here is a small flowchart I made to clear up my understand while writing the cases in this function
+	// it is very rough, but it may help while reading this piece of code http://imgur.com/HQICWeO
```
cyli: 👍  This chart is so awesome!  Do we want to check it into the repo instead of keeping it in imgur?

avaid96: Sure, I'd be happy to but the only issue is that we would then need to maintain it if it's in the repo. Totally your call

utils/keys.go
```diff
@@ -117,53 +115,47 @@ func ImportKeys(from io.Reader, to []Importer, fallbackRole string, fallbackGun
...
 			if fallbackRole == "" {
 				block.Headers["role"] = notary.DefaultImportRole
 			} else {
 				block.Headers["role"] = fallbackRole
 			}
 		}
+
+		// A root key or a delegations key should not have a gun
+		// Note that a key that is not any of the canonical roles (except root) is a delegations key and should not have a gun
+		if block.Headers["role"] != tufdata.CanonicalSnapshotRole && block.Headers["role"] != tufdata.CanonicalTargetsRole && block.Headers["role"] != tufdata.CanonicalTimestampRole {
```
cyli: Totally non-blocking nitpick:  This might be a good place for a switch, so we don't have to repeat `block.Headers["role"] !=` over and over:

```
switch block.Headers["role"] {
case tufdata.CanonicalSnapshotRole, tufdata.CanonicalTargetsRole, tufdata.CanonicalTimestampRole:
    if block.Headers...
default:
    delete(block.Headers, "gun")
}
```

utils/keys.go
```diff
@@ -117,53 +115,47 @@ func ImportKeys(from io.Reader, to []Importer, fallbackRole string, fallbackGun
...
+		}
+
 		loc, ok := block.Headers["path"]
 		// only if the path isn't specified do we get into this parsing path logic
 		if !ok || loc == "" {
 			// if the path isn't specified, we will try to infer the path rel to trust dir from the role (and then gun)
 			// parse key for the keyID which we will save it by.
 			// if the key is encrypted at this point, we will generate an error and continue since we don't know the ID to save it by
+
+			// note that any key exported by an older client
```
cyli: I think this comment got truncated

utils/keys.go
cyli: If the logic changes the comment would have to change anyway to not refer to the chart :)

utils/keys.go
avaid96: switched it, thanks @cyli 😄

cmd/notary/integration_test.go
```diff
@@ -2185,8 +2196,15 @@ func TestClientKeyImport(t *testing.T) {
 	// if there is hardware available, root will only be on hardware, and not
 	// on disk
 	assertNumKeys(t, tempDir, 2, 3, !rootOnHardware())
-	_, err = os.Open(filepath.Join(tempDir, "private", notary.NonRootKeysSubdir, "somegun", newKeyID+".key"))
+	_, err = os.Open(filepath.Join(tempDir, notary.PrivDir, newKeyID+".key"))
 	require.NoError(t, err)
+	_, err = os.Open(filepath.Join(tempDir, notary.PrivDir, newKeyID+".key"))
```
riyazdf: nit: this check seems like a repeat of the above `os.Open` and `NoError` check

cmd/notary/integration_test.go
```diff
@@ -2240,8 +2258,14 @@ func TestClientKeyImport(t *testing.T) {
 	// if there is hardware available, root will only be on hardware, and not
 	// on disk
 	assertNumKeys(t, tempDir, 2, 4, !rootOnHardware())
-	_, err = os.Open(filepath.Join(tempDir, "private", notary.NonRootKeysSubdir, newKeyID+".key"))
+	_, err = os.Open(filepath.Join(tempDir, notary.PrivDir, newKeyID+".key"))
+	require.NoError(t, err)
+	_, err = os.Open(filepath.Join(tempDir, notary.PrivDir, newKeyID+".key"))
```
riyazdf: seems like a repeat here too

avaid96: ah my bad, totally slipped through- I'll remove the duplicates in the next push with addressing further comments

tuf/utils/x509.go
```diff
@@ -453,7 +454,12 @@ func EncryptPrivateKey(key data.PrivateKey, role, gun, passphrase string) ([]byt
...
 	}
-	encryptedPEMBlock.Headers["role"] = role
+	if role != "" {
+		encryptedPEMBlock.Headers["role"] = role
+	}
+	if gun != "" {
+		encryptedPEMBlock.Headers["gun"] = gun
+	}
 
 	if gun != "" {
```
riyazdf: this seems to be repeated code

trustmanager/keystore.go
```diff
@@ -40,7 +42,47 @@ func NewKeyFileStore(baseDir string, p notary.PassRetriever) (*GenericKeyStore,
...
 		return nil, err
 	}
-	return NewGenericKeyStore(fileStore, p), nil
+	migrateTo0Dot4(fileStore)
+	store := NewGenericKeyStore(fileStore, p)
+	return store, nil
+}
+
+func migrateTo0Dot4(s Storage) {
+	for _, file := range s.ListFiles() {
```
riyazdf: Instead of listing all files and checking on each one, can we instead check if the `notary.RootKeysSubdir` and `notary.NonRootKeysSubdir` exist and only migrate if keys live there?  And then we can delete those subdirectories once we've finished a successful migration

cmd/notary/keys_test.go
```diff
@@ -569,9 +574,9 @@ func TestExportKeys(t *testing.T) {
 	require.NoError(t, err)
 
 	fileStore, err := store.NewPrivateKeyFileStorage(tempBaseDir, notary.KeyExtension)
-	err = fileStore.Set(filepath.Join(notary.NonRootKeysSubdir, "discworld/ankh"), bBytes)
+	err = fileStore.Set("discworld/ankh", bBytes)
```
endophage: Should just be "ankh"

cmd/notary/keys_test.go
```diff
@@ -569,9 +574,9 @@ func TestExportKeys(t *testing.T) {
...
 	require.NoError(t, err)
-	err = fileStore.Set(filepath.Join(notary.NonRootKeysSubdir, "discworld/morpork"), cBytes)
+	err = fileStore.Set("discworld/morpork", cBytes)
```
endophage: Should just be "morpork"

const.go
```diff
@@ -33,8 +33,10 @@ const (
 	// PrivDir is the directory, under the notary repo base directory, where private keys are stored
 	PrivDir = "private"
 	// RootKeysSubdir is the subdirectory under PrivDir where root private keys are stored
+	// The only reason we need this constant is compatibility with older versions
```
endophage: Add `DEPRECATED` notice on front of this comment and the one for `NonRootKeysSubdir`

trustmanager/keystore.go
```diff
@@ -40,7 +42,47 @@ func NewKeyFileStore(baseDir string, p notary.PassRetriever) (*GenericKeyStore,
 	if err != nil {
 		return nil, err
 	}
-	return NewGenericKeyStore(fileStore, p), nil
+	migrateTo0Dot4(fileStore)
```
endophage: would be great to move this migration down a level so it happens in `store.NewPrivateKeyFileStorage` where it can be optimized to only walk the `root_keys` and `tuf_keys` directories.

trustmanager/keystore.go
```diff
@@ -40,7 +42,47 @@ func NewKeyFileStore(baseDir string, p notary.PassRetriever) (*GenericKeyStore,
...
+				keyID = strings.TrimSuffix(keyID, "_"+role)
+				block.Headers["role"] = role
+			}
+		}
+		var keyPEM bytes.Buffer
+		// since block came from decoding the PEM bytes in the first place, and all we're doing is adding some headers we ignore the possibility of an error while encoding the block
+		pem.Encode(&keyPEM, block)
+		s.Set(keyID, keyPEM.Bytes())
+	}
+	os.RemoveAll(filepath.Join(s.Location(), notary.RootKeysSubdir))
```
endophage: check the `s.Location()` and `notary.*KeysSubdir` both have safe values

trustmanager/keystore.go
```diff
@@ -173,40 +181,32 @@ func (s *GenericKeyStore) AddKey(keyInfo KeyInfo, privKey data.PrivateKey) error
 func (s *GenericKeyStore) GetKey(name string) (data.PrivateKey, string, error) {
```
endophage: Can we update `name` to `keyID`

trustmanager/keystore.go
```diff
@@ -263,55 +246,41 @@ func copyKeyInfoMap(keyInfoMap map[string]KeyInfo) map[string]KeyInfo {
...
-		if keyRole, ok := block.Headers["role"]; ok {
-			role = keyRole
-		}
+	var keyID, role, gun string
+	keyID = filepath.Base(filename)
+	block, _ := pem.Decode(pemBytes)
+	if block == nil {
+		return "", KeyInfo{}, fmt.Errorf("could not decode PEM block for key %s", filename)
+	}
+	if keyRole, ok := block.Headers["role"]; ok {
```
endophage: `role := block.Headers["role"]`

utils/keys.go
```diff
@@ -106,8 +103,10 @@ func ImportKeys(from io.Reader, to []Importer, fallbackRole string, fallbackGun
 		toWrite []byte
 	)
 	for block, rest := pem.Decode(data); block != nil; block, rest = pem.Decode(rest) {
-		// if there is a path then we set the gun header from this path
-		if rawPath := block.Headers["path"]; rawPath != "" {
+		// if there is a legacy path then we set the gun header from this path
+		// this is the case when a user attempts to import a key bundle generated by an older client
+		if rawPath := block.Headers["path"]; rawPath != filepath.Base(rawPath) {
+			// this is a legacy filepath and we should try to deduce the gun name from it
 			pathWOFileName := strings.TrimSuffix(rawPath, filepath.Base(rawPath))
```
endophage: pathWOFileName := filepath.Dir(rawPath)

utils/keys.go
```diff
@@ -117,53 +116,47 @@ func ImportKeys(from io.Reader, to []Importer, fallbackRole string, fallbackGun
...
 			}
 		}
+
+		// A root key or a delegations key should not have a gun
+		// Note that a key that is not any of the canonical roles (except root) is a delegations key and should not have a gun
+		switch block.Headers["role"] {
+		case tufdata.CanonicalSnapshotRole, tufdata.CanonicalTargetsRole, tufdata.CanonicalTimestampRole:
+			// check if the key is missing a gun header or has an empty gun and error out since we don't know what gun it belongs to
+			if block.Headers["gun"] == "" {
+				logrus.Info("failed to import key to store: Cannot have canonical role key without a gun, don't know what gun it belongs to")
```
endophage: include path (keyID) so use knows which key failed

utils/keys.go
```diff
@@ -117,53 +116,47 @@ func ImportKeys(from io.Reader, to []Importer, fallbackRole string, fallbackGun
 				}
 			}
```
endophage: Need to `block.Headers["path"] = filepath.Base(rawPath)` otherwise later code will incorrectly use the legacy style path.

avaid96: fixed, with test in TestBlockHeaderPrecedenceGunFromPath. Will move the migration down a level tomorrow morning and this should be good to go 😄

cmd/notary/integration_test.go
```diff
@@ -2125,7 +2131,7 @@ func TestClientKeyImport(t *testing.T) {
 	// if there is hardware available, root will only be on hardware, and not
```
avaid96: These 8 tests could go into 8 separate testing functions but the setup for each is common at the top so not sure if we would like to repeat that across testing functions. CC @david regarding our conversation IRL earlier today

storage/filestore.go
```diff
@@ -61,6 +67,71 @@ type FilesystemStore struct {
...
+func (f *FilesystemStore) migrateTo0Dot4() {
+	if _, err := os.Stat(filepath.Join(f.Location(), notary.RootKeysSubdir)); !os.IsNotExist(err) {
+		// root_keys exists, migrate things from it
+		listOnlyRootKeysDirStore, _ := NewFileStore(filepath.Join(f.baseDir, notary.RootKeysSubdir), f.ext, notary.PrivKeyPerms)
+		for _, file := range listOnlyRootKeysDirStore.ListFiles() {
+			f.moveKeyTo0Dot4Location(filepath.Join(notary.RootKeysSubdir, file))
+		}
+		// delete the old directory
+		rootKeysSubDir := filepath.Join(f.Location(), notary.RootKeysSubdir)
+		if rootKeysSubDir == "" || rootKeysSubDir == "/" {
```
riyazdf: maybe this should check if `rootKeysSubDir == f.Location()`?  I'm not sure if we need this check since `f.Location()` is already cleaned

Also, I don't think you need to introduce the `rootKeysSubDir` variable and can use `listOnlyRootKeysDirStore.Location()` instead  (same thing below for `listOnlyNonRootKeysDirStore`_

trustmanager/keystore.go
```diff
@@ -40,7 +40,8 @@ func NewKeyFileStore(baseDir string, p notary.PassRetriever) (*GenericKeyStore,
 	if err != nil {
 		return nil, err
 	}
-	return NewGenericKeyStore(fileStore, p), nil
```
riyazdf: nit: I think we can keep this one-line return

avaid96: @endophage I did it as well 😆

storage/filestore.go
avaid96: I think the intention of this check is to make sure that we don't accidentally delete a users root dir (very rare case but still)
I'll clean up the variables as you recommended

storage/filestore.go
```diff
@@ -61,6 +67,71 @@ type FilesystemStore struct {
 	perms   os.FileMode
 }
 
+func (f *FilesystemStore) moveKeyTo0Dot4Location(file string) {
+	keyID := filepath.Base(file)
+	fileDir := filepath.Dir(file)
+	d, _ := f.Get(file)
+	block, _ := pem.Decode(d)
+	if block == nil {
+		logrus.Warn("Key data for ", file, " may have been tampered with/ is invalid. The key has not been migrated and may not be available")
```
riyazdf: super nit : can we mention how we couldn't decode the PEM or filestructure?  Something like `Key data for <file> could not be decoded as a valid PEM block.  The key will not been migrated and may not be available`

riyazdf:
this has come together really nicely, thank you @avaid96 for your hard work on this PR!

LGTM after comments

cmd/notary/integration_test.go
cyli: That setup op top can be factored out as a helper function, and return the temp files instead  :)  Up to you though.

storage/filestore.go
```diff
@@ -61,6 +67,73 @@ type FilesystemStore struct {
...
+	rootKeysSubDir := filepath.Join(f.Location(), notary.RootKeysSubdir)
+	nonRootKeysSubDir := filepath.Join(f.Location(), notary.NonRootKeysSubdir)
+	if _, err := os.Stat(rootKeysSubDir); !os.IsNotExist(err) && f.Location() != rootKeysSubDir {
+		// root_keys exists, migrate things from it
+		listOnlyRootKeysDirStore, _ := NewFileStore(rootKeysSubDir, f.ext, notary.PrivKeyPerms)
+		for _, file := range listOnlyRootKeysDirStore.ListFiles() {
+			f.moveKeyTo0Dot4Location(filepath.Join(notary.RootKeysSubdir, file))
+		}
+		// delete the old directory
+		if rootKeysSubDir == "" || rootKeysSubDir == "/" {
```
cyli: Non-blocking:  Can this ever happen, since we control the constants `notary.RootKeysSubdir` and `notary.NonRootKeysSubdir` - they're not empty, and we do a `filepath.Join`, so won't it always contain at least some of the constnat values?

storage/filestore.go
```diff
@@ -61,6 +67,73 @@ type FilesystemStore struct {
...
+	var keyPEM bytes.Buffer
+	// since block came from decoding the PEM bytes in the first place, and all we're doing is adding some headers we ignore the possibility of an error while encoding the block
+	pem.Encode(&keyPEM, block)
+	f.Set(keyID, keyPEM.Bytes())
+}
+
+func (f *FilesystemStore) migrateTo0Dot4() {
+	rootKeysSubDir := filepath.Join(f.Location(), notary.RootKeysSubdir)
+	nonRootKeysSubDir := filepath.Join(f.Location(), notary.NonRootKeysSubdir)
+	if _, err := os.Stat(rootKeysSubDir); !os.IsNotExist(err) && f.Location() != rootKeysSubDir {
```
cyli: I don't think `f.Location() != rootKeysSubDir`, is ever false, since we do a `filepath.Join`.  Although it's possible that that once we clean it this could be the case.  So possibly `rootKeysSubDir` and `nonRootKeysSubDir` should be cleaned, as well as joined?  e.g.  `filepath.Clean(filepath.Join(...))`

storage/filestore.go
cyli: Update:  I'm not sure `filepath.Abs(filepath.Clean(filepath.Join(...)))` could result in `rootKeysSubDir` being either empty or "/".

https://play.golang.org/p/OCxAa4jtjL were the only cases I tried, so I might have missed some stuff.

client/backwards_compatibility_test.go
```diff
@@ -131,6 +131,70 @@ func Test0Dot1RepoFormat(t *testing.T) {
...
+	require.Len(t, newTargetsKeys, 1)
+	require.NotEqual(t, oldTargetsKeys[0], newTargetsKeys[0])
+
+	// rotate the snapshot key to the server and ensure that the server can re-generate the snapshot
+	// and we can download the snapshot
+	require.NoError(t, repo.RotateKey(data.CanonicalSnapshotRole, true))
+	require.NoError(t, repo.Publish())
+	err = repo.Update(false)
+	require.NoError(t, err)
+}
```
cyli: Since we're migrating these, could we add a test for the 0.3 and the 0.1 repos to ensure that once a notary repo is created, all the keys are migrated to the new format?

cyli:
So sorry I didn't comment sooner.  :(  I will be happy to address the comments above in a few days.

storage/filestore.go
endophage: @cyli is 100% correct. I asked @avaid96 to put these checks in as a sanity check in case anyone is making updates in the future and does something silly.

@cyli has also made the suggestion that we check `f.Location` and the `(Non)RootKeysSubdir` constants at the start instead of checking the joined values here and later. That would stop a possible walk over `/` which would take a long time.

cmd/notary/integration_test.go
avaid96: I think I may just leave it as is because they're pretty similar and re-setting up- even as a helper function will simply add time to the tests as discussed with @endophage

storage/filestore.go
avaid96: @cyli you're 100% correct about the fact that as of now we won't enter these statements but as correctly pointed out by @endophage these are just to make the migrations futureproof

@cyli the prior checks make sense. I have implemented something similar in the latest push (coming soon)- checking the concatenated value but checking this before the walk over the directory. Let me know if you like this or would prefer separate checks. I prefer this because it will be simpler to understand what is going on

storage/filestore.go
avaid96: f.location may be equal to rootkeyssubdir in case someone in the future does remove the deprecated constans notary.(Non)RootKeysSubDir. This was on suggestion from @riyazdf 

You're totally right though, we should anyways run a filepath.clean on the joined values

avaid96:
No issues at all @cyli. Thanks for the awesome comments. I think they're all mostly addressed besides the migration test which I can write some time this week :smile: Let me know if there's anything else!

avaid96:
added the texts for the migrations @cyli :smiley:

cyli:
@avaid96 So so sorry I hadn't finished going over this yet.  Thank you for adding those tests!  LGTM, and thank you so much for slogging through all of this and keeping it up to date!

avaid96:
bumping this for a final review and merge. Has one slightly out of date LGTM from @riyazdf (doesn't include last commit) and another up to date one from @cyli

cyli:
@avaid96 Many apologies, this seems to require some conflict resolution.  :|  I think github allows collaboration on forks now (https://github.com/blog/2247-improving-collaboration-with-forks), so if it's enabled we'd be happy to rebase :)

riyazdf:
Final commit LGTM!

And yeah, unfortunately we'll need to rebase -- let us know if you hit that checkbox and we can handle that :)

avaid96:
No worries, I rebased. Sorry I think Circle ran into something :disappointed: Could you rebuild?

PR #873, opened 2016-07-25
Make a single script to run go tests with coverage

cyli:
We needed to make the dependency calculation for coverage include more packages, because as discovered in #869, a test file may import and use code that is not a dependency of the non-test version of the package itself, and `go list -f {{.Deps}}` does not actually get the test dependencies.

So in #869, for example, the test touches both the signer client and the signer RPC code, but coverage wasn't generated for either because the `signer` package it was in would not have touched it at all.   Moving the test into the `signer/client` package wouldn't have generated coverage reports for `signer/api`, and vice versa.

So rather than a complex calculation, we're just going to include all notary packages.  It would be simpler if we could just do `--coverpkg=all`, but that includes everything the vendor directory.

So the test script:  
  - includes all notary packages only for coverage calculation purposes
  - we get a ton of warnings about a package not actually being covered by the test if we do that, so this suppresses those specific warnings
  - pretty-formats the output so that the "ok" status of each package is not followed by a list of all notary packages again, to make the results more easily readable at a glance

Also there are some `cmd/` package changes because otherwise, if we try to load all packages in notary, we get flag conflicts when it tries to load `cmd/notary-server` and `cmd/notary-signer` at once, so this PR changes those to variables scoped to an object, rather than global state.


cyli:
Meh I may have done a bunch of unnecessary work - sec...

cyli:
Ok, I think this fixes it.  Ready for review :)

codecov.yml
```diff
@@ -15,5 +15,7 @@ coverage:
     # changes tells us if there are unexpected code coverage changes in other files
     # which were not changed by the diff
     changes: false
+  ignore:  # ignore testutils for coverage
+    - "tuf/testutils/*"
```
riyazdf: thanks for adding this!  A super non-blocking nit while we're editing the codecov config: did we want to change the badge colors (something like `range: 50...100` might work)?

cyli: Sure!

codecov.yml
```diff
@@ -15,5 +15,7 @@ coverage:
     # changes tells us if there are unexpected code coverage changes in other files
     # which were not changed by the diff
     changes: false
```
riyazdf: nit: does this new way of generating coverage packages mean we still need this?  I think the answer is yes because of how codecov seems to rebase, but I'm not entirely sure.

riyazdf:
thank you for tackling this!  LGTM

codecov.yml
cyli: Yes, because you  might write tests that touch a package that you didn't change, and the coverage for that package would change, but there were no changes in that package itself.

I think if this were enabled, codecov would add a separate check that would fail.  The three checks that codecov can do are project (we want to gate on total project coverage), patch (that allows gating based on what % of the diff is covered by tests), and failing if there are unexpected changes. 

We're only enabling the first check.

endophage:
LGTM!

PR #875, opened 2016-07-26
Witness command

### LINKED ISSUE #562: Merge clients
Merge the Notary client with the TUF client.

Not parallelizable with #561

endophage:
Fixes #562.


tuf/builder.go
```diff
@@ -543,13 +548,20 @@ func (rb *repoBuilder) loadDelegation(roleName string, content []byte, minVersio
...
 		return err
 	}
+	// verify signature
+	if err := signed.VerifySignatures(signedObj, delegationRole.BaseRole); err != nil {
+		rb.invalidRoles.Targets[roleName] = signedTargets
+		return err
+	}
 
 	if err := signed.VerifyVersion(&(signedTargets.Signed.SignedCommon), minVersion); err != nil {
+		rb.invalidRoles.Targets[roleName] = signedTargets
```
cyli: Should this be a recoverable error in that it should be witness-able?  If the version is too low, doesn't that means there's a newer version already (and hence witness could be reverting existing changes)?

cmd/notary/tuf.go
```diff
@@ -83,6 +83,12 @@ var cmdTUFVerifyTemplate = usageTemplate{
 	Long:  "Verifies if the data passed in STDIN is included in the remote trusted collection identified by the Global Unique Name.",
 }
 
+var cmdWitnessTemplate = usageTemplate{
+	Use:   "witness [ GUN ] <role> ...",
+	Short: "Marks roles to be re-signed the next time they're published",
```
NathanMcCauley: Is this the default mode of `witness`?

I was hoping this command could be used for addition of signatures and that that would be the primary use case for this command.

endophage: The long form usage on the next line elaborates on the exact behaviour. Does that more closely match what you were thinking?

client/witness.go
```diff
@@ -0,0 +1,53 @@
...
+		}
+		successful = append(successful, role)
+	}
+	return successful, err
+}
+
+func witnessTargets(repo *tuf.Repo, invalid *tuf.Repo, role string) error {
+	if r, ok := repo.Targets[role]; ok {
+		// role is already valid, mark for re-signing/updating
+		r.Dirty = true
```
cyli: This bumps the version too - should we have a second flag for witness that does not bump the version and just signs?

client/witness.go
```diff
@@ -0,0 +1,53 @@
...
+		return nil
+	}
+	if invalid == nil {
+		// role isn't recognized, even as invalid
+		return ErrInvalidLocalRole{Role: role}
+	}
+	if r, ok := invalid.Targets[role]; ok {
+		// role is recognized but invalid, move to valid data and mark for re-signing
+		repo.Targets[role] = r
+		r.Dirty = true
```
cyli: Similarly here - do we always want to bump the version?

tuf/builder.go
```diff
@@ -545,11 +557,19 @@ func (rb *repoBuilder) loadDelegation(roleName string, content []byte, minVersio
 	}
 
 	if err := signed.VerifyVersion(&(signedTargets.Signed.SignedCommon), minVersion); err != nil {
+		// don't capture in invalidRoles because the role we received is rollback
```
cyli: grammar nitpick:  "is a rollback" or "has been rollbacked"?

endophage:
@cyli re bumping version, I was implementing this version just to solve the immediate recovery case. Until we have actual thresholding it's kind of a moot point how we recover.

My instinct though is that because it was, in the past, a valid role that has become invalid, we should not allow the invalid version to become valid again, we should require it's published as a new version. That seems safer than allowing a version to flip flop between states, which seems like it could be abused to serve 2 different versions of a role. In fact we may want to look at adding an additional check that says "if I've seen valid version X and know it to have checksum Y, I will not accept a version X from the server with a checksum that differs from Y". Either that or require the version strictly increases (I think the current logic is >=)

cyli:
@endophage Ok, that makes sense for recovering invalid roles.  Should we just ignore the witness command though if the role is already valid?  It seems pointless to re-sign.

endophage:
I feel like there's still a use case for preemptively resigning if a role will expire soon. Like if I know I'm going on vacation but my data will expire while I'm away. I should be able to make it good in advance. Because it's an explicit user command rather than us trying to be clever about resigning I think this is the right behaviour.

cyli:
In that case, would actually witnessing (e.g. for thresholds) require an extra flag to the witness command?  Would it make sense for bumping the expiry and version to take a flag instead?

The word "witness" makes more sense to me to indicate that you are sort of rubber-stamping something, or agreeing with something that is already there.  This behavior seems like it's just always re-generating new data, and not really "witnessing" something in the general sense of the word.

endophage:
Your understanding of what this code currently does is spot on.

Because the staging/state of partially signed role for thresholds is TBD I was designing it to meet the current use case only (resigning a role that has become invalid). A full witness feature for thresholds is probably going to necessarily be online (to determine the available variations that could be witnessed), and require interactivity. In that situation, I would guess we would add a flag, as you suggested, to shortcut all that and just do the behaviour implemented in this PR (which would also be offered as an interactive option).

I'm just trying to keep the current command simple because a required flag feels wrong. The obvious question will be "what does the command do without the flag?", to which the answer would be "nothing".

client/tufclient.go
```diff
@@ -11,17 +12,26 @@ import (
 	"github.com/docker/notary/tuf/signed"
 )
 
-// Client is a usability wrapper around a raw TUF repo
-type Client struct {
+// ErrCorruptedCache - local data is incorrect
```
riyazdf: Can we delete this error?  I think it's unused

client/witness.go
```diff
@@ -0,0 +1,53 @@
+package client
+
+import (
+	"github.com/docker/notary/client/changelist"
+	"github.com/docker/notary/tuf"
+	"path/filepath"
+)
+
+// Witness creates change objects to witness (i.e. resign) the given
```
riyazdf: nit: `re-sign` instead of `resign` 😄

client/witness.go
```diff
@@ -0,0 +1,53 @@
...
+		c := changelist.NewTUFChange(
+			changelist.ActionUpdate,
+			role,
+			changelist.TypeWitness,
+			"",
+			nil,
+		)
+		err = cl.Add(c)
+		if err != nil {
+			break
```
riyazdf: nit: can we add a debug or warning log to inform the user that the change wasn't successfully added to the changelist?

client/witness.go
```diff
@@ -0,0 +1,53 @@
...
+
+func witnessTargets(repo *tuf.Repo, invalid *tuf.Repo, role string) error {
+	if r, ok := repo.Targets[role]; ok {
+		// role is already valid, mark for re-signing/updating
+		r.Dirty = true
+		return nil
+	}
+	if invalid == nil {
+		// role isn't recognized, even as invalid
+		return ErrInvalidLocalRole{Role: role}
```
riyazdf: I'm not sure if this is the right error because it'll print `"notary does not permit the client managing the <ROLE> key"` - since this could technically be called on a delegation role that may not exist?

cmd/notary/integration_test.go
```diff
@@ -1366,3 +1366,129 @@ func TestPurge(t *testing.T) {
...
+// Initialize repo and test witnessing. The following steps are performed:
+//   1. init a repo
+//   2. add a delegation with a key and --all-paths
+//   3. add a target to the delegation
+//   4. list targets and ensure it really is in the delegation
+//   5. add a new (different) key to the delegation
+//   6. remove the key from the delegation
+//   7. list targets and ensure the target is no longer visible
+//   8. witness the delegation
+//   9.list targets and ensure target is visible again
```
riyazdf: super non-blocking nit: align spaces on comment

client/witness.go
cyli: Yes apologies, that one is just used for key rotation.  :|  Also, can we add tests for this case where we try to witness a role that doesn't exist?   Or even if it did exist once, but the user explicitly deleted it (i.e. it has been removed from the snapshot)?

Also would be great to test witnessing a valid role too.  :)

client/witness.go
riyazdf: I'm fine with bumping the version after reading through the main-thread discussion but we should keep this consistent with the command's usage description (which mentions that publishing a new version is only for invalid roles)

client/witness.go
endophage: That should already be happening when the error is returned up to the witness command code, where it also takes care of messaging which roles were successfully marked for re-signing.

client/witness.go
endophage: Error fixed and good suggestions on the tests, caught a silent failure!

cmd/notary/tuf.go
```diff
@@ -83,6 +83,12 @@ var cmdTUFVerifyTemplate = usageTemplate{
 	Long:  "Verifies if the data passed in STDIN is included in the remote trusted collection identified by the Globally Unique Name.",
 }
 
+var cmdWitnessTemplate = usageTemplate{
+	Use:   "witness [ GUN ] <role> ...",
+	Short: "Marks roles to be re-signed the next time they're published",
+	Long:  "Marks roles to be re-signed the next time they're published. If the role has no currently valid signatures, or is otherwise invalid, a new version is published. If a role has some valid signatures and is not otherwise invalid, new signatures are added without modifying the signed role data.",
```
cyli: Non-blocking - should we modify this comment since for now, if the role is valid, the version is bumped?

tuf/builder.go
```diff
@@ -545,11 +557,19 @@ func (rb *repoBuilder) loadDelegation(roleName string, content []byte, minVersio
 	}
 
 	if err := signed.VerifyVersion(&(signedTargets.Signed.SignedCommon), minVersion); err != nil {
+		// don't capture in invalidRoles because the role we received is a rollback
```
cyli: Can we add a test for this case where if the role is too low, `invalidRoles` will not contain the version that was attempted?

cyli:
Thanks for adding this!  Can we add `witness gun targets/releases` to the list of example valid commands here:  https://github.com/docker/notary/blob/master/cmd/notary/main_test.go#L151?  That way we can test the number of command line arguments.

Other than that and the version test, this LGTM!

cmd/notary/integration_test.go
```diff
@@ -1366,3 +1366,150 @@ func TestPurge(t *testing.T) {
...
+	_, err = runCommand(t, tempDir, "-s", server.URL, "publish", "gun")
+	require.NoError(t, err)
+
+	// 10. list targets and ensure target is visible again
+	output, err = runCommand(t, tempDir, "-s", server.URL, "list", "gun")
+	require.NoError(t, err)
+	require.Contains(t, output, targetName)
+	require.Contains(t, output, targetHash)
+
+	// 11. witness an invalid role and check for error on publish
```
riyazdf: this entire test flow is awesome! 👍

cmd/notary/integration_test.go
```diff
@@ -1366,3 +1366,150 @@ func TestPurge(t *testing.T) {
...
+	_, err = runCommand(t, tempDir, "witness", "gun", "targets/made/up")
```
riyazdf: Can we also add a test to try to witness non-targets or delegation roles?  I was thinking `root`, `snapshot`, `timestamp` - this should be already covered by changelist helper logic, but would be nice to have.

client/witness.go
endophage: Updated long form usage description.

endophage:
@cyli @riyazdf I think the other cases you suggested should be covered now. I also made some additional updates in the application of changelists. We were inconsistent about when we returned an error or just logged an continued if we didn't recognize a given change. With the new functionality to help manage the changelist, we now more consistently error and the user will have to remove the offending change from their changelist.

cyli:
LGTM!  Thank you for addressing all the testing so quickly!

riyazdf:
LGTM

PR #880, opened 2016-07-27
Signer key management refactor

cyli:
This addresses parts of #846.

This currently doesn't do cleanup, but does record the last used time for keys so that a DB admin can clean up old, not-used keys.

Other stuff that should be changed in the signer as part of making server key rotations work, but maybe should be addressed in a different PR:

- Signer should take rate limiting configuration so that the RPC API will not create a new key if one exists that was created within the last X time.  This is tricky though because keydbstore implements KeyStore, as opposed to cryptoservice, and also doesn't provide a way to list keys based on gun and role.

   We've discussed eliminating the `KeyStore` layer and just having a DB-based `cryptoservice`, which gives us some more fine-grained control over controlling key creation, or just also having our cryptoservice do extra db-work.

- Signer should provide a way to delete all keys for a given GUN so that when a repo is deleted from the server, the server can tell the signer to remove all the keys as well.


HuKeping:
> Signer should provide a way to delete all keys for a given GUN so that when a repo is deleted from the server, the server can tell the signer to remove all the keys as well.

+1 for this feature , can I supposed that with this clean-up on signer side, if we delete a repo, the metadata of it will all be wiped off from notary **service**

cyli:
@HuKeping We would like that to be the case, yes.  The delete endpoint al ready removes all the TUF metadata from the DB.

endophage:
Wondering if marking as active should be implicit in signing (and therefore retrieval from the database). Then we wouldn't have to bubble up the `markActive` concept, which currently seems incompatible with potentially having multiple storage backends.

cyli:
@endophage Yep - definitely agree it can be moved to `GetPrivateKey` now - at the time I wasn't sure how `Get` was going to work for the server trying to get the current key, but https://github.com/docker/notary/pull/889 implements a different function to do that so.  If that method looks good to you, I'll merge `MarkActive` into `GetPrivateKey` here.

> which currently seems incompatible with potentially having multiple storage backends.
Just verifying - ultimately, didn't we want change the backends to implement `CryptoService` instead of `KeyStore`?  Since it's unlikely you'd want multiple keystores for a signer?

cyli:
@endophage Moved the marking as active to the the private key's `Sign` function, as per your suggestion.  Thanks this is much cleaner :)

cyli:
Closing in favor of #889

PR #720, opened 2016-05-05
Expand testclient to test delegations and root rotations

cyli:
Add more extensive delegation tests and root rotation tests to our test client.

Also, this can be used against Docker Hub.

PR #882, opened 2016-07-28
adding --role flag to key import 

### LINKED ISSUE #881: Add --role flag for key import
Currently a user can't import a key with a role flag. This becomes an issue becuase notary is not able to work with keys generated by openssl `openssl genrsa -out delegation.key 2048` which don't have PEM headers for role

avaid96:
closes #881 


avaid96:
This is still WIP
Discussion: 
1. should we have a --gun flag for compatibility with the flat key store? as of now you can specify the path if you want a key associated with a gun but this won't be possible in a flat keystore. followup: should we even have a path if we have a flat keystore
2. if there is a role flag should we limit to importing only one key or add that to all specified keys?

HuKeping:
Thanks for the quick fixing!

HuKeping:
Do you get this error when import keys:
```
failed to import key to store: PEM headers did not contain import path
```

cyli:
@avaid96 I think taking a GUN makes sense, and a GUN header makes sense more than a path header in the flattened keystore.  I think #825 included both headers - can import guess the path based on the GUN for now until flatten keystore is merged?

utils/keys.go
```diff
@@ -103,6 +103,9 @@ func ImportKeys(from io.Reader, to []Importer) error {
 		toWrite []byte
 	)
 	for block, rest := pem.Decode(data); block != nil; block, rest = pem.Decode(rest) {
+		if role != "" {
```
riyazdf: I think this conditional logic should be slightly different to match the usage comment - we should check if `block.Headers["role"]` exists, and set it to `role` if it's empty.

Also, I think this logic should combine with the pathing logic below since to-be-imported delegation keys may not have a `path` PEM header.  We can check if we have a provided role and if it's a delegation we can write it to `private/tuf_keys/<KEY_ID>.key`.  If we follow @cyli's suggestion for GUN headers we can include that in the path we construct for non-delegation keys as well.

cyli:
@HuKeping Yep, that's what we are trying to fix.  :)

utils/keys.go
```diff
@@ -93,7 +94,7 @@ func ExportKeys(to io.Writer, s Exporter, from string) error {
 // Each block is written to the subpath indicated in the "path" PEM
 // header. If the file already exists, the file is truncated. Multiple
 // adjacent PEMs with the same "path" header are appended together.
-func ImportKeys(from io.Reader, to []Importer) error {
+func ImportKeys(from io.Reader, to []Importer, role string, gun string, passRet notary.PassRetriever) error {
```
cyli: nit:  can we call `role` and `gun` -> `fallbackRole` and `fallbackGun` or some such, to make it clear they're only used if there is no provided role or gun/path?

utils/keys.go
```diff
@@ -103,11 +104,45 @@ func ImportKeys(from io.Reader, to []Importer) error {
 		toWrite []byte
 	)
 	for block, rest := pem.Decode(data); block != nil; block, rest = pem.Decode(rest) {
+		if block.Headers["role"] == "" {
+			// no worries about if check as for GUN here because empty roles will get a role:notary.DefaultImportRole
+			block.Headers["role"] = role
+		}
+		if block.Headers["gun"] == "" {
```
cyli: Non-blocking:  this has probably already been discussed and resolved, and I just completely forgot all the cases, but is it ok for the gun and path to mismatch?  e.g. gun to be `docker.io/hello/world` and path to be `/tuf_keys/docker.io/nothello/notworld/DEADBEEF0A...key`?  I know we don't really use the header GUN for anything right now, so it doesn't much matter - just wondering if it's worth trying to grab the GUN from the path first before falling back on no gun?

utils/keys.go
```diff
@@ -26,7 +27,7 @@ type Importer interface {
 // ExportKeysByGUN exports all keys filtered to a GUN
 func ExportKeysByGUN(to io.Writer, s Exporter, gun string) error {
 	keys := s.ListFiles()
-	sort.Strings(keys) // ensure consistenct. ListFiles has no order guarantee
+	sort.Strings(keys) // ensure consistent. ListFiles has no order guarantee
```
riyazdf: nit: I think this was originally a typo for `consistency`

utils/keys.go
```diff
@@ -103,11 +104,52 @@ func ImportKeys(from io.Reader, to []Importer) error {
...
+			if block.Headers["role"] == "" {
+				// now we have no clue where to copy this key so we skip it since we have no path or role
+				logrus.Info("failed to import key to store: PEM headers did not contain import path")
+				continue
+			}
+			// if the path isn't specified, we will try to infer the path rel to trust dir from the role (and then gun)
+
+			decodedKey, err := utils.ParsePEMPrivateKey(pem.EncodeToMemory(block), "")
+			if err != nil {
+				logrus.Info("failed to import key to store: Invalid key generated, key may be encrypted and not contains path header")
```
riyazdf: typo: `...and does not contain path header`

utils/keys.go
```diff
@@ -103,11 +104,52 @@ func ImportKeys(from io.Reader, to []Importer) error {
...
+				continue
+			}
+			keyID := decodedKey.ID()
+
+			if block.Headers["role"] == "root" {
```
riyazdf: nit: we should use `data.CanonicalRootRole` instead of hardcoding `"root"`

trustmanager/keystore.go
```diff
@@ -152,7 +152,7 @@ func (s *GenericKeyStore) AddKey(keyInfo KeyInfo, privKey data.PrivateKey) error
 	}
 
 	if chosenPassphrase != "" {
-		pemPrivKey, err = utils.EncryptPrivateKey(privKey, keyInfo.Role, chosenPassphrase)
+		pemPrivKey, err = utils.EncryptPrivateKey(privKey, keyInfo.Role, "", chosenPassphrase)
```
riyazdf: I think it's ok to pass on the `keyInfo.Gun` here because we set it to empty string for delegations and root roles above

utils/keys.go
```diff
@@ -103,11 +104,52 @@ func ImportKeys(from io.Reader, to []Importer) error {
 		toWrite []byte
 	)
 	for block, rest := pem.Decode(data); block != nil; block, rest = pem.Decode(rest) {
+		if block.Headers["role"] == "" {
+			// no worries about if check as for GUN here because empty roles will get a role:notary.DefaultImportRole
+			block.Headers["role"] = fallbackRole
+		}
+		if rawPath := block.Headers["path"]; rawPath != "" {
+			pathWOFileName := strings.TrimSuffix(rawPath, filepath.Base(rawPath))
+			if strings.Contains(pathWOFileName, notary.NonRootKeysSubdir) {
```
riyazdf: can we use `strings.HasPrefix` here instead of `strings.Contains`?

Also, we can include slashes in the `TrimSuffix` and `TrimPrefix` instead of having to slice when assigning to the PEM header

avaid96: First one sounds good
Second one: http://weknowmemes.com/wp-content/uploads/2013/08/windows-logic-meme.jpg

utils/test.key
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
riyazdf: nit: can we put these keys in the `fixtures` directory?  Also, you may not need this key without the role header because we other keys (ex: `secure.example.com.key`) without headers that we can test with

utils/keys_test.go
```diff
@@ -298,20 +362,74 @@ func TestImportKeys2InOneFile(t *testing.T) {
...
+
+	in := bytes.NewBuffer(byt)
+
+	err := ImportKeys(in, []Importer{s}, "", "", passphraseRetriever)
 	require.NoError(t, err)
 
 	bFinal, bRest := pem.Decode(s.data["ankh"])
 	require.Equal(t, b.Bytes, bFinal.Bytes)
-	require.Len(t, bFinal.Headers, 0) // path header is stripped during import
+	require.Equal(t, "", bFinal.Headers["path"]) // path header is stripped during import
```
riyazdf: can we change these checks to something like `require.NotContains(t, bFinal.Headers, "path")`?  It's possible that this equality check could pass if we had a `"path"` header without any value, but instead we should test the absence of that header entirely

cmd/notary/integration_test.go
```diff
@@ -1366,3 +1372,183 @@ func TestPurge(t *testing.T) {
...
+	require.NoError(t, err)
+
+	nBytes, err = tempFile4.Write(pemBytes)
+	require.NoError(t, err)
+	tempFile4.Close()
+	require.Equal(t, len(pemBytes), nBytes)
+	newKeyID := privKey.ID()
+
+	// import the key
+	_, err = runCommand(t, tempDir, "key", "import", tempFile4.Name(), "-r", "somerole")
```
riyazdf: Can we add an integration test that imports a delegation key with `notary key import -r targets/releases` and then successfully publishes targets to that role?  We can piggyback off of existing delegation publishing tests in `TestClientDelegationsPublishing`

utils/keys_test.go
avaid96: changed this to something like this: 
```
if _, ok := bFinal.Headers["path"]; !ok {
		require.False(t, ok, "expected no path header, should have been removed at import")
	}
```
I don't think NotContains will work since `bFinal.Headers` is a map

cmd/notary/integration_test.go
avaid96: Just tried the first one on the client made with the edited code and it seems to work fine, will write a test soon. 

For the next two, just to clarify my understanding with a couple of cases, where my.key does not have a path header: 

- `notary key import my.key`: key is imported to tuf_keys/ with role: notary.DefaultImportRole

- `notary key import my.key -g randomgun`: key is imported to tuf_keys/ with role: notary.DefaultImportRole

- `notary key import my.key -r targets/releases -g randomgun`: key is imported to tuf_keys/ with role: notary.DefaultImportRole

- `notary key import my.key -r snapshot -g randomgun`: key is imported to tuf_keys/randomgun/ with role: notary.DefaultImportRole

cyli: I think the first two look correct.  For the second two, assuming that the key does not have a role specified:

> `notary key import my.key -r targets/releases -g randomgun: key` is imported to tuf_keys/ with role: notary.DefaultImportRole

Should this be imported with the provided role (`targets/releases`)?  Or are we overriding all delegation roles to just be "delegation"?

> `notary key import my.key -r snapshot -g randomgun`: key is imported to tuf_keys/randomgun/ with role: notary.DefaultImportRole

Should this be imported with the role "snapshot" since that was provided as a flag?

avaid96: Agree on the last one, that was a typo by me, I've edited it accordingly. 

On the third one, I thought we were overriding the roles but I think what you're saying makes more sense because it will help when the client looks up keys for publishing. 

Thanks a ton for the quick review @cyli 😄

riyazdf:
jenkins, ok to test.

utils/keys.go
```diff
@@ -103,11 +105,73 @@ func ImportKeys(from io.Reader, to []Importer) error {
...
+			// if the key is encrypted at this point, we will generate an error and continue since we don't know the ID to save it by
+			decodedKey, err := utils.ParsePEMPrivateKey(pem.EncodeToMemory(block), "")
+			if err != nil {
+				logrus.Info("failed to import key to store: Invalid key generated, key may be encrypted and does not contain path header")
+				continue
+			}
+			keyID := decodedKey.ID()
+			if block.Headers["role"] == tufdata.CanonicalRootRole {
+				// this is a root key so import it to trustDir/root_keys/
+				delete(block.Headers, "gun")
```
cyli: Non-blocking:  Should we move this logic out of the if block?  E.g. after setting the path, we can delete the "gun" header if the role is root or a delegation?  Since currently it's only deleted if the path is not specified.

utils/keys.go
```diff
@@ -103,11 +105,73 @@ func ImportKeys(from io.Reader, to []Importer) error {
...
+			// if the path isn't specified, we will try to infer the path rel to trust dir from the role (and then gun)
+			// parse key for the keyID which we will save it by.
+			// if the key is encrypted at this point, we will generate an error and continue since we don't know the ID to save it by
+			decodedKey, err := utils.ParsePEMPrivateKey(pem.EncodeToMemory(block), "")
+			if err != nil {
+				logrus.Info("failed to import key to store: Invalid key generated, key may be encrypted and does not contain path header")
+				continue
+			}
+			keyID := decodedKey.ID()
+			if block.Headers["role"] == tufdata.CanonicalRootRole {
```
cyli: Totally non-blocking nitpick: this might be a good place to use switch, since it shortens everything:

```
switch block.Headers["role"] {
case tufdata.CanonicalRootRole:
    ...
case tufdata.CanonicalSnapshotRole, tufdata.CanonicalTargetsRole, tufdata.CanonicalTimestampRole:
    ...
default:  // delegation
    ...
}
```

utils/keys_test.go
```diff
@@ -232,17 +249,22 @@ func TestImportKeys(t *testing.T) {
...
-	err := ImportKeys(in, []Importer{s})
+	err := ImportKeys(in, []Importer{s}, "", "", passphraseRetriever)
 	require.NoError(t, err)
 
 	bFinal, bRest := pem.Decode(s.data["ankh"])
 	require.Equal(t, b.Bytes, bFinal.Bytes)
-	require.Len(t, bFinal.Headers, 0) // path header is stripped during import
+	_, ok := bFinal.Headers["path"]
+	require.False(t, ok, "expected no path header, should have been removed at import")
+	require.Equal(t, notary.DefaultImportRole, bFinal.Headers["role"]) // if no role is specified we assume it is a delegation key
```
cyli: Should we add an assertion that no gun is included in the header as well?

utils/keys_test.go
```diff
@@ -252,17 +274,70 @@ func TestImportNoPath(t *testing.T) {
...
-	b.Bytes = make([]byte, 1000)
-	rand.Read(b.Bytes)
+	from, _ := os.OpenFile("../fixtures/secure.example.com.key", os.O_RDONLY, notary.PrivKeyPerms)
+	defer from.Close()
+	b.Bytes, _ = ioutil.ReadAll(from)
 
-	bBytes := pem.EncodeToMemory(b)
+	in := bytes.NewBuffer(b.Bytes)
+
+	err := ImportKeys(in, []Importer{s}, "root", "", passphraseRetriever)
```
cyli: Non-blocking: can we use the constant `data.CanonicalRootRole` here instead of "root"?

utils/keys_test.go
```diff
@@ -252,17 +274,70 @@ func TestImportNoPath(t *testing.T) {
...
+	require.NoError(t, err)
+
+	for key := range s.data {
+		// no path but role included should work
+		require.Equal(t, key, filepath.Join(notary.RootKeysSubdir, "12ba0e0a8e05e177bc2c3489bdb6d28836879469f078e68a4812fc8a2d521497"))
+	}
+
+	err = ImportKeys(in, []Importer{s}, "", "", passphraseRetriever)
```
cyli: Can we reset `s =NewTestImportStore()` here before importing again?  That way we can assert that there is no data in `s`?  You are totally correct, and that the new import should fail, but but because it's importing the same key again just without a role and gun, we can't distinguish whether it actually failed to import or imported it again to the same name (which shouldn't happen, because it cannot have formulated a name, but just in case :))

cmd/notary/integration_test.go
```diff
@@ -1526,3 +1532,374 @@ func TestWitness(t *testing.T) {
...
+	tempFile2.Close()
+	require.Equal(t, len(pemBytes), nBytes)
+
+	// import the key
+	_, err = runCommand(t, tempDir, "key", "import", tempFile3.Name())
+	require.NoError(t, err)
+
+	// if there is hardware available, root will only be on hardware, and not
+	// on disk
+	_, _ = assertNumKeys(t, tempDir, 2, 1, !rootOnHardware())
```
riyazdf: can we test that we imported this key under the default `delegation` role, and that it's directly under the `tuf_keys` directory?

utils/keys_test.go
```diff
@@ -252,17 +274,70 @@ func TestImportNoPath(t *testing.T) {
 	b := &pem.Block{
```
cyli: Totally non-blocking - I think `b` may not be needed anymore?  We just have to read the file into a byte slice, and pass it to the buffer.  The import itself constructs a new PEM block itself - is this pem block used other than for temporary storage?

cmd/notary/integration_test.go
```diff
@@ -1526,3 +1532,374 @@ func TestWitness(t *testing.T) {
...
+	_, err = runCommand(t, tempDir, "-s", server.URL, "publish", "gun")
+	require.NoError(t, err)
+
+	// list delegations - none yet
+	output, err = runCommand(t, tempDir, "-s", server.URL, "delegation", "list", "gun")
+	require.NoError(t, err)
+	require.Contains(t, output, "No delegations present in this repository.")
+
+	// publish repo
+	_, err = runCommand(t, tempDir, "-s", server.URL, "publish", "gun")
```
riyazdf: nit: this publish is unnecessary

cmd/notary/integration_test.go
```diff
@@ -1526,3 +1532,374 @@ func TestWitness(t *testing.T) {
...
+
+	// the changelist still exists so no need to add the target again
+	// just import the key and publish
+	output, err = runCommand(t, tempDir, "-s", server.URL, "status", "gun")
+	require.NoError(t, err)
+	require.Contains(t, output, "targets/releases")
+	require.Contains(t, output, "sdgkadga")
+
+	// import the key
+	_, err = runCommand(t, tempDir, "key", "import", keyFile.Name(), "-r", "targets/releases", "-g", "gun")
```
riyazdf: nit: we don't need to include the gun here right?

cmd/notary/integration_test.go
```diff
@@ -1526,3 +1532,374 @@ func TestWitness(t *testing.T) {
...
+	// check that changelist is applied
+	output, err = runCommand(t, tempDir, "-s", server.URL, "status", "gun")
+	require.NoError(t, err)
+	require.NotContains(t, output, "targets/releases")
+
+	// list targets for targets/releases - we should see our target!
+	output, err = runCommand(t, tempDir, "-s", server.URL, "list", "gun", "--roles", "targets/releases")
+	require.NoError(t, err)
+	require.NotContains(t, output, "No targets present")
+	require.Contains(t, output, "sdgkadga")
```
riyazdf: nit: can we also check that `targets/releases` is in the output?  Just to confirm that we published under that role and not another one

utils/keys_test.go
```diff
@@ -252,17 +274,70 @@ func TestImportNoPath(t *testing.T) {
...
+	}
+}
+
+func TestBlockHeaderPrecedence(t *testing.T) {
+	s := NewTestImportStore()
+
+	b := &pem.Block{
+		Headers: make(map[string]string),
+	}
+	from, _ := os.OpenFile("../fixtures/precedence.example.com.key", os.O_RDONLY, notary.PrivKeyPerms)
```
cyli: Totally non-blocking nit - I was wondering if it'd be clearer to just synthesize the PEM block with the right headers (similar to the test code in `TestImportKeys2InOneFile`) rather than create a new fixture, just because it'd be easier in the test to see what the headers in the key file are, or to modify them at need.

cmd/notary/keys.go
```diff
@@ -97,7 +99,12 @@ func (k *keyCommander) GetCommand() *cobra.Command {
 			"Required for timestamp role, optional for snapshot role")
 	cmd.AddCommand(cmdRotateKey)
 
-	cmd.AddCommand(cmdKeyImportTemplate.ToCommand(k.importKeys))
+	cmdKeysImport := cmdKeyImportTemplate.ToCommand(k.importKeys)
+	cmdKeysImport.Flags().StringVarP(
+		&k.keysImportRole, "role", "r", "", "Role to import key with - Notary can use this to infer the path to store the key. A specified path will take precedence")
+	cmdKeysImport.Flags().StringVarP(
```
riyazdf: nit: can we rephrase both of these flag descriptions to something shorter like:

`Role to import key with, if a role is not already given in a PEM header`

I think the pathing logic can be explained from informative error messages rather than adding complexity to the CLI flags, since the error should depend on how the user imports the keys (straight from openssl vs. from `notary key export`)

utils/keys_test.go
```diff
@@ -252,17 +274,70 @@ func TestImportNoPath(t *testing.T) {
...
-	in := bytes.NewBuffer(bBytes)
+	in := bytes.NewBuffer(b.Bytes)
 
-	err := ImportKeys(in, []Importer{s})
+	err := ImportKeys(in, []Importer{s}, "somerole", "somegun", passphraseRetriever)
 	require.NoError(t, err)
 
-	require.Len(t, s.data, 0)
+	for key := range s.data {
+		// block header role= root should take precedence over command line flag
```
cyli: 👍  Awesome test, thank you! 

Could we also add one for falling back on the delegation role if no role is given and no role is provided?  (also maybe one for overriding the gun if the path specifies a different gun?)

tuf/utils/x509_test.go
```diff
@@ -165,6 +168,19 @@ func TestKeyOperations(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, rsaKey.Private(), decryptedRSAKey.Private())
 
+	// quick test that gun headers are being added appropriately
+	// Encrypt our RSA Key, one type of key should be enough since headers are treated the same
+	testGunKey, err := EncryptPrivateKey(rsaKey, "root", "ilove", "ponies")
```
riyazdf: non-blocking nit: while you're in here, can we change `"root"` to `data.CanonicalRootRole` in this file?

utils/keys_test.go
cyli: Update:  actually apologies, falling back on the default role can just be tested in the next test `TestImportKeys2InOneFile`, since that one has no role specified and no fallback role provided.

utils/keys.go
```diff
@@ -103,11 +105,73 @@ func ImportKeys(from io.Reader, to []Importer) error {
 		toWrite []byte
 	)
 	for block, rest := pem.Decode(data); block != nil; block, rest = pem.Decode(rest) {
+		// if there is a path then we set the gun header from this path
+		if rawPath := block.Headers["path"]; rawPath != "" {
+			pathWOFileName := strings.TrimSuffix(rawPath, filepath.Base(rawPath))
+			if strings.HasPrefix(pathWOFileName, notary.NonRootKeysSubdir) {
+				gunName := strings.TrimPrefix(pathWOFileName, notary.NonRootKeysSubdir)
```
riyazdf: Can we add a test-case that if we initialize a repo, export keys, then import keys to a fresh file-based keystore the `gun` PEM headers are set?

Also, for this logic - it seems like delegation keys may end up with a `gun` header with an empty value?  Should we add a check for `gunName` being an empty string?

utils/keys_test.go
```diff
@@ -298,20 +373,80 @@ func TestImportKeys2InOneFile(t *testing.T) {
...
+func TestImportKeys2InOneFileNoPath(t *testing.T) {
+	s := NewTestImportStore()
+
+	b := &pem.Block{
+		Headers: make(map[string]string),
+	}
+	from, _ := os.OpenFile("../fixtures/secure.example.com.key", os.O_RDONLY, notary.PrivKeyPerms)
+	b.Bytes, _ = ioutil.ReadAll(from)
+	rand.Read(b.Bytes)
+	b.Headers["path"] = "ankh"
```
cyli: Should one of these be missing a path to test what happens if one of the keys does not have a path specified?  Otherwise this test seems functionally the same as `TestImportKeys2InOneFile`  (importing multiple keys in one file that all have paths specified)

cmd/notary/keys.go
```diff
@@ -408,8 +415,8 @@ func (k *keyCommander) importKeys(cmd *cobra.Command, args []string) error {
 	for _, file := range args {
 		from, err := os.OpenFile(file, os.O_RDONLY, notary.PrivKeyPerms)
 		defer from.Close()
-
-		if err = utils.ImportKeys(from, importers); err != nil {
+		passRetreiver := k.getRetriever()
```
cyli: Non-blocking nit - can we change this param name to "passRetriever"  ("retriever" is misspelled :))

utils/keys.go
```diff
@@ -26,7 +28,7 @@ type Importer interface {
 // ExportKeysByGUN exports all keys filtered to a GUN
 func ExportKeysByGUN(to io.Writer, s Exporter, gun string) error {
 	keys := s.ListFiles()
-	sort.Strings(keys) // ensure consistenct. ListFiles has no order guarantee
+	sort.Strings(keys) // ensure consistency. ListFiles has no order guarantee
```
cyli: Thank you for fixing this typo!

cmd/notary/integration_test.go
```diff
@@ -1526,3 +1532,374 @@ func TestWitness(t *testing.T) {
...
+
+	tempFile6, err := ioutil.TempFile("", "pemfile")
+	require.NoError(t, err)
+	// close later, because we might need to write to it
+	defer os.Remove(tempFile6.Name())
+
+	privKey, err = utils.GenerateECDSAKey(rand.Reader)
+	require.NoError(t, err)
+
+	pemBytes, err = utils.EncryptPrivateKey(privKey, "root", "", testPassphrase)
```
riyazdf: non-blocking nit: can we change "root" to data.CanonicalRootRole in this file?

cmd/notary/integration_test.go
avaid96: Do you'll think that I should move the constant notary.DefaultImportRole to where the canonical roles are? @cyli @riyazdf

cmd/notary/integration_test.go
```diff
@@ -1526,3 +1532,374 @@ func TestWitness(t *testing.T) {
...
+	require.NoError(t, err)
+	// close later, because we might need to write to it
+	defer os.Remove(tempFile.Name())
+
+	// -- tests --
+	// test 1, no path but role=root included with non-encrypted key
+	privKey, err := utils.GenerateECDSAKey(rand.Reader)
+	require.NoError(t, err)
+
+	pemBytes, err := utils.EncryptPrivateKey(privKey, "root", "", "")
```
cyli: Non-blocking:  can we use `data.CanonicalRootRole` instead of "root" in this test here and a little further down?  Similarly with `data.CanonicalSnapshotRole` and "snapshot"?

cmd/notary/integration_test.go
cyli: Hmm... it might make sense to leave it where it is, since `tuf/data` contains the constants that are specific to TUF, and `notary.DefaultImportRole` is our own default value for on a non-TUF-spec operation.

utils/keys_test.go
```diff
@@ -232,17 +249,22 @@ func TestImportKeys(t *testing.T) {
...
+	require.Equal(t, notary.DefaultImportRole, bFinal.Headers["role"]) // if no role is specified we assume it is a delegation key
 	require.Len(t, bRest, 0)
 
 	cFinal, cRest := pem.Decode(s.data["morpork"])
 	require.Equal(t, c.Bytes, cFinal.Bytes)
-	require.Len(t, cFinal.Headers, 0)
+	_, ok = cFinal.Headers["path"]
+	require.False(t, ok, "expected no path header, should have been removed at import")
+	require.Equal(t, data.CanonicalSnapshotRole, cFinal.Headers["role"])
+	require.Equal(t, "somegun", cFinal.Headers["gun"])
```
cyli: Could we also assert that the imported bytes are encrypted?

cmd/notary/integration_test.go
avaid96: You're correct, was just following a similar structure as TestClientDelegationsPublishing. I'll remove the extra publish from there as well! 💃

cyli:
Also, this totally fixes https://github.com/docker/notary/pull/720 :)  And those tests all pass with this PR.

utils/keys_test.go
avaid96: done, thanks 😄

utils/keys.go
avaid96: of course 😄

utils/keys.go
avaid96: done, thanks 😄

cmd/notary/integration_test.go
avaid96: done, thanks 😄

utils/keys_test.go
avaid96: done, thanks 😄

cmd/notary/integration_test.go
avaid96: okay perfect, thanks 😄 

above comment is in reply to my earlier comment which I deleted: "Should I move notary.DefaultImportRole to tuf/data"

utils/keys_test.go
avaid96: added it, thanks 😄

utils/keys.go
avaid96: done, thanks 😄

utils/keys_test.go
```diff
@@ -315,21 +314,50 @@ func TestNonRootPathInference(t *testing.T) {
...
+		require.Equal(t, key, filepath.Join(notary.NonRootKeysSubdir, "anothergun", "12ba0e0a8e05e177bc2c3489bdb6d28836879469f078e68a4812fc8a2d521497"))
+		final, rest := pem.Decode(s.data[key])
+		require.Len(t, rest, 0)
+		require.Equal(t, final.Headers["role"], "snapshot")
+		require.Equal(t, final.Headers["gun"], "anothergun")
 	}
-	from, _ := os.OpenFile("../fixtures/precedence.example.com.key", os.O_RDONLY, notary.PrivKeyPerms)
+}
+
+func TestBlockHeaderPrecedencePathGun(t *testing.T) {
```
cyli: Non-blocking nitpick:  should this be "TestBlockHeaderPrecedencePathRole", since the header has both the path and the role, but not the gun?

utils/keys_test.go
```diff
@@ -397,30 +427,27 @@ func TestImportKeys2InOneFile(t *testing.T) {
...
+	b.Headers["role"] = data.CanonicalSnapshotRole
+	bBytes := pem.EncodeToMemory(b)
 
-	b2 := &pem.Block{
-		Headers: make(map[string]string),
-	}
-	b2.Bytes, _ = ioutil.ReadAll(from)
-	rand.Read(b2.Bytes)
-	b2.Headers["path"] = "ankh"
+	from2, _ := os.OpenFile("../fixtures/secure.example.com.key", os.O_RDONLY, notary.PrivKeyPerms)
```
cyli: Non-blocking: since we've already read all the bytes from this file once, we just use `fromBytes` for both `b2` and `c`, rather than having to re-read the file.

utils/keys_test.go
```diff
@@ -431,22 +458,99 @@ func TestImportKeys2InOneFileNoPath(t *testing.T) {
...
+	bFinal, bRest := pem.Decode(keyBytes)
+	require.Equal(t, "somegun", bFinal.Headers["gun"])
+	require.Len(t, bRest, 0)
+
+	// we should fail to parse it without the passphrase
+	privKey, err = utils.ParsePEMPrivateKey(keyBytes, "")
+	require.Equal(t, err, errors.New("could not decrypt private key"))
+	require.Nil(t, privKey)
+
+	// we should sucdeed to parse it with the passphrase
```
cyli: spelling nitpick:  "succeed"

utils/keys_test.go
avaid96: Done, thanks for catching that! 😃

utils/keys_test.go
```diff
@@ -298,20 +401,153 @@ func TestImportKeys2InOneFile(t *testing.T) {
...
+	cBytes := pem.EncodeToMemory(c)
+
+	byt := append(bBytes, cBytes...)
+
+	in := bytes.NewBuffer(byt)
+
+	err := ImportKeys(in, []Importer{s}, "", "", passphraseRetriever)
+	require.NoError(t, err)
+
+	bFinal, bRest := pem.Decode(s.data[filepath.Join(notary.NonRootKeysSubdir, "12ba0e0a8e05e177bc2c3489bdb6d28836879469f078e68a4812fc8a2d521497")])
```
cyli: Non-blocking question, just because I don't know what we decided for this:  it seems like we are ok with one of the base non-root roles not having a gun also (we don't delete it, but we don't error if we can't infer one) - we infer a path being just the non-root subdirectory + the key ID, as you've included in your test.

I'm ok with that, since the user messed this up, but this would mean that earlier versions of docker probably wouldn't be able to use this key?  cc @riyazdf @endophage

avaid96: I'm okay with either method, but I would be tempted to lean towards not importing it since we don't import encrypted keys without a path etc (cases where the user messed up as well). Would make sense to have some consistency there. This would mean changing the code and this test etc since as @cyli mentioned correctly we now just import the key. Happy to do it once we decide @riyazdf @endophage

utils/keys.go
endophage: Looks like the code is duplicated between the `switch` and `if ... else if ...`

endophage:
LGTM when CI goes green!

cmd/notary/integration_test.go
```diff
@@ -2028,3 +2030,499 @@ func TestClientTUFAddByHashWithAutoPublish(t *testing.T) {
...
+
+	privKey, err = utils.GenerateECDSAKey(rand.Reader)
+	require.NoError(t, err)
+
+	pemBytes, err = utils.EncryptPrivateKey(privKey, "", "", "")
+	require.NoError(t, err)
+
+	nBytes, err = tempFile3.Write(pemBytes)
+	require.NoError(t, err)
+	tempFile2.Close()
```
riyazdf: nit: `tempFile3.Close()`

cmd/notary/keys.go
```diff
@@ -408,8 +415,8 @@ func (k *keyCommander) importKeys(cmd *cobra.Command, args []string) error {
 	for _, file := range args {
 		from, err := os.OpenFile(file, os.O_RDONLY, notary.PrivKeyPerms)
 		defer from.Close()
-
-		if err = utils.ImportKeys(from, importers); err != nil {
+		passRetriever := k.getRetriever()
```
riyazdf: super non-blocking nit: we can remove this line and just directly use `k.getRetriever` in the next line since we don't use the `passRetriever` variable elsewhere

utils/keys.go
```diff
@@ -103,11 +105,89 @@ func ImportKeys(from io.Reader, to []Importer) error {
...
+			// Key is not encrypted- ask for a passphrase and encrypt this key
+			var chosenPassphrase string
+			for attempts := 0; ; attempts++ {
+				var giveup bool
+				chosenPassphrase, giveup, err = passRet(loc, block.Headers["role"], true, attempts)
+				if err != nil {
+					continue
+				}
+				if giveup || attempts > 10 {
+					fmt.Println("exceeded attempts")
```
riyazdf: I think this should be erroring out instead of printing

utils/keys.go
```diff
@@ -103,11 +105,89 @@ func ImportKeys(from io.Reader, to []Importer) error {
...
+		// we are now all set for import but let's first encrypt the key
+		blockBytes := pem.EncodeToMemory(block)
+		// check if key is encrypted, note: if it is encrypted at this point, it will have had a path header
+		if privKey, err := utils.ParsePEMPrivateKey(blockBytes, ""); err == nil {
+			// Key is not encrypted- ask for a passphrase and encrypt this key
+			var chosenPassphrase string
+			for attempts := 0; ; attempts++ {
+				var giveup bool
+				chosenPassphrase, giveup, err = passRet(loc, block.Headers["role"], true, attempts)
+				if err != nil {
```
riyazdf: discussed IRL but this should be:

```
if err == nil {
    break
 }
```

And the `break` below can be removed

utils/keys.go
```diff
@@ -103,11 +105,89 @@ func ImportKeys(from io.Reader, to []Importer) error {
...
+					continue
+				}
+				if giveup || attempts > 10 {
+					fmt.Println("exceeded attempts")
+				}
+				break
+			}
+			blockBytes, _ = utils.EncryptPrivateKey(privKey, block.Headers["role"], block.Headers["gun"], chosenPassphrase)
```
riyazdf: we should handle the error from `EncryptPrivateKey` and error out if it's not nil

utils/keys.go
avaid96: thanks for catching that, changed it 😄

riyazdf:
LGTM pending CI.  Thank you for all of your hard work on this feature!

avaid96:
Of course, thanks for the detailed review on this PR! 😄

cyli:
Thank you for all your work on this @avaid96!  :D

PR #886, opened 2016-07-29
Introduce a flag for auto publishing

### LINKED ISSUE #885: -p flag for auto-publish on staged commands
It'd be nice to have a `-p` flag (or similar) to automatically attempt to publish after staging a change in a notary CLI command.  For example: `notary init -p`, `notary add -p`, `notary remove -p`, `notary delegation add -p`, `notary delegation remove -p`, and `notary addhash -p`

HuKeping:
Closes #885 

Add flag "-p", "--publish" to "notary add", with this flag being set,
notary will attemp to publish after adding the change.

### DIFF
--- a/cmd/notary/integration_test.go
+++ b/cmd/notary/integration_test.go
@@ -1716,6 +1716,15 @@ func TestClientTUFAddWithAutoPublish(t *testing.T) {
 	// the change should still exists.
 	output, err = runCommand(t, tempDir, "status", gun)
 	require.NoError(t, err)
+	require.Contains(t, output, target)
+
+	// add a target with auto publish being enabled but with a malformed URL.
+	_, err = runCommand(t, tempDir, "-s", "For the Horde!", "add", "-p", gun, target, tempFile.Name())
+	require.Error(t, err, "Trust server url has to be in the form of http(s)://URL:PORT.")
+	// add a target with auto publish being enabled but with an unaccessible URL.
+	_, err = runCommand(t, tempDir, "-s", "https://notary-server-on-the-moon:12306", "add", "-p", gun, target, tempFile.Name())
+	require.NotNil(t, err)
+	require.Equal(t, err, nstorage.ErrOffline{})
 
 	// add a target with auto publish being enabled, and with the server URL
 	_, err = runCommand(t, tempDir, "-s", server.URL, "add", "-p", gun, target2, tempFile.Name())
@@ -1978,8 +1987,6 @@ func TestClientTUFAddByHashWithAutoPublish(t *testing.T) {
 	targetData := []byte{'a', 'b', 'c'}
 	target256Bytes := sha256.Sum256(targetData)
 	targetSha256Hex := hex.EncodeToString(target256Bytes[:])
-	target512Bytes := sha512.Sum512(targetData)
-	targetSha512Hex := hex.EncodeToString(target512Bytes[:])
 
 	err := ioutil.WriteFile(filepath.Join(tempDir, "tempfile"), targetData, 0644)
 	require.NoError(t, err)
@@ -1987,8 +1994,6 @@ func TestClientTUFAddByHashWithAutoPublish(t *testing.T) {
 	var (
 		output  string
 		target1 = "sdgkadga"
-		target2 = "asdfasdf"
-		target3 = "qwerty"
 	)
 	// -- tests --
 
@@ -2014,60 +2019,4 @@ func TestClientTUFAddByHashWithAutoPublish(t *testing.T) {
 	output, err = runCommand(t, tempDir, "-s", server.URL, "lookup", "gun", target1)
 	require.NoError(t, err)
 	require.Contains(t, output, target1)
-
-	// remove target
-	_, err = runCommand(t, tempDir, "-s", server.URL, "remove", "-p", "gun", target1)
-	require.NoError(t, err)
-
-	// list repo - don't see target
-	output, err = runCommand(t, tempDir, "-s", server.URL, "list", "gun")
-	require.NoError(t, err)
-	require.False(t, strings.Contains(string(output), target1))
-
-	// add a target just by sha512
-	_, err = runCommand(t, tempDir, "-s", server.URL, "addhash", "-p", "gun", target2, "3", "--sha512", targetSha512Hex)
-	require.NoError(t, err)
-
-	// check status - no targets
-	output, err = runCommand(t, tempDir, "status", "gun")
-	require.NoError(t, err)
-	require.False(t, strings.Contains(string(output), target2))
-
-	// list repo - see target
-	output, err = runCommand(t, tempDir, "-s", server.URL, "list", "gun")
-	require.NoError(t, err)
-	require.Contains(t, output, target2)
-
-	// lookup target and repo - see target
-	output, err = runCommand(t, tempDir, "-s", server.URL, "lookup", "gun", target2)
-	require.NoError(t, err)
-	require.Contains(t, output, target2)
-
-	// remove target
-	_, err = runCommand(t, tempDir, "-s", server.URL, "remove", "-p", "gun", target2)
-	require.NoError(t, err)
-
-	// add a target by sha256 and sha512
-	_, err = runCommand(t, tempDir, "-s", server.URL, "addhash", "-p", "gun",
-		target3, "3", "--sha256", targetSha256Hex, "--sha512", targetSha512Hex)
-	require.NoError(t, err)
-
-	// check status - no targets
-	output, err = runCommand(t, tempDir, "status", "gun")
-	require.NoError(t, err)
-	require.False(t, strings.Contains(string(output), target3))
-
-	// list repo - see target
-	output, err = runCommand(t, tempDir, "-s", server.URL, "list", "gun")
-	require.NoError(t, err)
-	require.Contains(t, output, target3)
-
-	// lookup target and repo - see target
-	output, err = runCommand(t, tempDir, "-s", server.URL, "lookup", "gun", target3)
-	require.NoError(t, err)
-	require.Contains(t, output, target3)
-
-	// remove target
-	_, err = runCommand(t, tempDir, "-s", server.URL, "remove", "-p", "gun", target3)
-	require.NoError(t, err)
 }
--- a/cmd/notary/integration_test.go
+++ b/cmd/notary/integration_test.go
@@ -1667,7 +1667,7 @@ func TestClientTUFInitWithAutoPublish(t *testing.T) {
 	_, err = runCommand(t, tempDir, "-s", server.URL, "init", gunNoPublish)
 	require.NoError(t, err)
 	// list repo - exepect error
-	output, err = runCommand(t, tempDir, "-s", server.URL, "list", gunNoPublish)
+	_, err = runCommand(t, tempDir, "-s", server.URL, "list", gunNoPublish)
 	require.NotNil(t, err)
 	require.Equal(t, err, nstorage.ErrMetaNotFound{Resource: "root"})
 }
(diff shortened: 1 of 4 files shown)


cyli:
jenkins, test this please

endophage:
Looks like a great start. Would like to see this option take effect on every command that currently generates a changelist item.

cyli:
jenkins, test this please

cmd/notary/tuf.go
```diff
@@ -241,9 +244,17 @@ func (t *tufCommander) tufAdd(cmd *cobra.Command, args []string) error {
...
 		return err
 	}
-	cmd.Printf(
-		"Addition of target \"%s\" to repository \"%s\" staged for next publish.\n",
-		targetName, gun)
+
+	cmd.Printf("Addition of target \"%s\" to repository \"%s\" staged for next publish.\n", targetName, gun)
+
+	// Auto publish the local trusted collection to remote.
+	if t.autoPublish == true {
```
riyazdf: nit: can we refactor this line to `if t.autoPublish {`?

cmd/notary/tuf.go
```diff
@@ -115,6 +117,7 @@ func (t *tufCommander) AddToCommand(cmd *cobra.Command) {
 
 	cmdTUFAdd := cmdTUFAddTemplate.ToCommand(t.tufAdd)
 	cmdTUFAdd.Flags().StringSliceVarP(&t.roles, "roles", "r", nil, "Delegation roles to add this target to")
+	cmdTUFAdd.Flags().BoolVarP(&t.autoPublish, "publish", "p", false, "Automatically attempt to publish after adding the change. Will also publish those pre-generated changelists.")
```
riyazdf: nit on the phrasing: `"Automatically attempt to publish after staging the change. Will also publish existing staged changes."`

cmd/notary/tuf.go
```diff
@@ -241,9 +244,17 @@ func (t *tufCommander) tufAdd(cmd *cobra.Command, args []string) error {
...
 	}
-	cmd.Printf(
-		"Addition of target \"%s\" to repository \"%s\" staged for next publish.\n",
-		targetName, gun)
+
+	cmd.Printf("Addition of target \"%s\" to repository \"%s\" staged for next publish.\n", targetName, gun)
+
+	// Auto publish the local trusted collection to remote.
+	if t.autoPublish == true {
+		c := cobra.Command{
```
riyazdf: instead of setting up another command, we can call `nRepo.Publish` directly

riyazdf:
I agree with @endophage's comment - it would be great if this PR could add the flag to all commands listed in #885 and also if we could add some simple tests

HuKeping:
Please take a look if this change is OK, I'll add the others if so.

cmd/notary/tuf.go
```diff
@@ -249,9 +252,27 @@ func (t *tufCommander) tufAdd(cmd *cobra.Command, args []string) error {
...
+		if err != nil {
+			return err
+		}
+
+		nRepo, err := notaryclient.NewNotaryRepository(
+			config.GetString("trust_dir"), gun, getRemoteTrustServer(config), rt, t.retriever, trustPin)
+		if err != nil {
+			return err
+		}
+
```
riyazdf: this looks good, the only thing I would add is to have a print similar to how we have one in `tufPublish`, something like:

```
...
cmd.Println("Auto-publishing changes to", gun)
return nRepo.Publish()
```

endophage: this is also quite a few lines of code that I think will be identical everywhere we want to do this so I'd wrap the whole thing, including the `if t.autoPublish` in its own function and then anywhere we want to end with an auto publish we can just do something like `return maybeAutoPublish(t.autoPublish, nRepo, <other args>...)`

endophage: Also, some of the commands we'd want to autopublish are in other places (i.e. `notary delegation add/remove`) so it would be good to work out a more accessible place for the flag.

cyli:
jenkins, test this please

cmd/notary/tuf.go
```diff
@@ -123,6 +125,7 @@ func (t *tufCommander) AddToCommand(cmd *cobra.Command) {
 
 	cmdTUFAdd := cmdTUFAddTemplate.ToCommand(t.tufAdd)
 	cmdTUFAdd.Flags().StringSliceVarP(&t.roles, "roles", "r", nil, "Delegation roles to add this target to")
+	cmdTUFAdd.Flags().BoolVarP(&t.autoPublish, "publish", "p", false, "Automatically attempt to publish after staging the change. Will also publish existing staged changes.")
```
cyli: Since this is a flag that is going to be added to multiple commands, I'm wondering if we should make the help text be a constant?

cmd/notary/tuf.go
```diff
@@ -834,3 +839,30 @@ func (a *authRoundTripper) RoundTrip(req *http.Request) (*http.Response, error)
 	// Return the last response
 	return resp, nil
 }
+
+func publish(cmd *cobra.Command, doPublish bool, gun string, config *viper.Viper, passRetriever notary.PassRetriever) error {
```
cyli: Non-blocking - it's unfortunate that we can't inject a `RoundTripper` into an existing `NotaryRepository`.  

I'm wondering whether we can just have `publish` take a `NotaryRepository`, and refactor this function into a general `NotaryRepository` builder that creates a `NotaryRepository` with either a `nil` transport or a real transport, depending on whether it needs to be online. 

Also, as a future refactor we can do for `NotaryRepository`, I'm wondering if we should have the online functions `Publish`, `List`, etc. take a `RoundTripper` as an argument as opposed to having one stored on the `NotaryRepository`.

cyli:
Thank you for your work on this @HuKeping!  This looks much cleaner, and am looking forward to having this flag on the offline commands!  :)

cmd/notary/tuf.go
HuKeping: We can benefit from the constant especially when we want to modify the help text, we may missing some if we don't use that. But also I'm wondering if it was a good idea to use a constant for a command help text, I haven't get an example.

HuKeping:
Tests ready.

cmd/notary/tuf.go
riyazdf: also, I think it'd be nice to rename this to `maybePublish` or `maybeAutoPublish` so it's clear that we may not actually publish

cmd/notary/integration_test.go
```diff
@@ -1366,3 +1367,159 @@ func TestPurge(t *testing.T) {
...
+	require.NoError(t, err)
+	tempFile.Close()
+	defer os.Remove(tempFile.Name())
+
+	var (
+		output       string = ""
+		gun          string = "MistsOfPandaria"
+		gunNoPublish string = "Legion"
+
+		// This might be changed via the implmentation, please be care.
```
riyazdf: typo: `implementation`

cmd/notary/tuf.go
riyazdf: I agree with @cyli - also for when we implement the `delegation add` and `delegation remove` auto-publish flags, I think the text will be the same, and we've used constants for similar important text like `tufRootKeyGenerationWarning`

riyazdf:
@HuKeping would you be alright with also including `notary delegation add -p` and `notary delegation remove -p` in this PR so we get all of the auto-publish flags in simultaneously?  It would iron out the UX for all of our printing and flag information, and would also ensure the design of the auto-publish helper functions are solid.

If you would prefer a separate PR that's fine too, but those are the only two auto-publish commands left and you've made some really awesome progress in this PR already :)

HuKeping:
@riyazdf  It should be in one PR, will add the `delegation` case

HuKeping:
Covers all the commands listed in #885

riyazdf:
jenkins, test this please

cmd/notary/integration_test.go
```diff
@@ -1526,3 +1527,317 @@ func TestWitness(t *testing.T) {
...
+
+		// This might be changed via the implementation, please be careful.
+		emptyList = "\nNo targets present in this repository.\n\n"
+	)
+	// -- tests --
+
+	// init repo with auto publish being enabled
+	_, err = runCommand(t, tempDir, "-s", server.URL, "init", "-p", gun)
+	require.NoError(t, err)
+	// list repo - exepect empty list
```
riyazdf: typo: `expect`

cmd/notary/integration_test.go
```diff
@@ -1526,3 +1527,317 @@ func TestWitness(t *testing.T) {
...
+	output, err = runCommand(t, tempDir, "-s", server.URL, "list", gun)
+	require.NoError(t, err)
+	require.Equal(t, output, emptyList)
+
+	// init repo without auto publish being enabled
+	//
+	// Use this test to guarantee that we won't break the normal init process.
+	_, err = runCommand(t, tempDir, "-s", server.URL, "init", gunNoPublish)
+	require.NoError(t, err)
+	// list repo - exepect error
```
riyazdf: typo: `expect`

cmd/notary/integration_test.go
```diff
@@ -1526,3 +1527,317 @@ func TestWitness(t *testing.T) {
...
+	_, err = runCommand(t, tempDir, "-s", server.URL, "list", gunNoPublish)
+	require.NotNil(t, err)
+	require.Equal(t, err, nstorage.ErrMetaNotFound{Resource: "root"})
```
riyazdf: these tests are great!

cmd/notary/integration_test.go
```diff
@@ -1526,3 +1527,317 @@ func TestWitness(t *testing.T) {
...
+	)
+	// -- tests --
+
+	// init repo with auto publish being enabled
+	_, err = runCommand(t, tempDir, "-s", server.URL, "init", "-p", gun)
+	require.NoError(t, err)
+	// add a target with auto publish being enabled
+	_, err = runCommand(t, tempDir, "add", "-s", server.URL, "-p", gun, target, tempFile.Name())
+	require.NoError(t, err)
+	_, err = runCommand(t, tempDir, "add", "-s", server.URL, "-p", gun, targetWillBeRemoved, tempFile.Name())
```
riyazdf: nit: can we add a test that checks that the autopublish flag will publish multiple staged changes?  For example, requiring that if we ran `notary add -p` here but didn't autopublish above, both targets are published?

riyazdf: Another test idea: try an autopublish command without giving a server URL, and check that we failed to publish.  And ensure that the change is still staged

cyli:
Should we also add a `-p` flag for the adding by hash command?

cmd/notary/integration_test.go
HuKeping: Added those test case with commit `Enhance test for auto publish`

HuKeping:
Add the `notary addhash` case

riyazdf:
jenkins, test this please

HuKeping:
Finally all green lol

cmd/notary/integration_test.go
```diff
@@ -1526,3 +1527,447 @@ func TestWitness(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// add a target with auto publish being enabled, and with the server URL
+	_, err = runCommand(t, tempDir, "-s", server.URL, "add", "-p", gun, target2, tempFile.Name())
+	require.NoError(t, err)
+	// list repo, since the auto publish flag will try to publish all the staged changes,
+	// so the target and target2 should be in the list.
+	output, err = runCommand(t, tempDir, "-s", server.URL, "list", gun)
+	require.NoError(t, err)
+	require.True(t, strings.Contains(output, target))
```
riyazdf: super non-blocking nit: these string contains can be simplified to `require.Contains(t, output, target)`

cmd/notary/integration_test.go
```diff
@@ -1526,3 +1527,447 @@ func TestWitness(t *testing.T) {
...
+
+	// add a target with auto publish being enabled, but without the server URL
+	_, err = runCommand(t, tempDir, "add", "-p", gun, target, tempFile.Name())
+	require.NotNil(t, err)
+	require.Equal(t, err, nstorage.ErrOffline{})
+	// check status, since we only fail the auto publishment in the previous step,
+	// the change should still exists.
+	output, err = runCommand(t, tempDir, "status", gun)
+	require.NoError(t, err)
+
```
cyli: Would it be worth checking that the output actually contains the first target here?  `notary status` returns no error if the changelists can be read, so an empty changelist would also not error.

cmd/notary/integration_test.go
```diff
@@ -1526,3 +1527,447 @@ func TestWitness(t *testing.T) {
...
+		// This might be changed via the implementation, please be careful.
+		emptyList = "\nNo targets present in this repository.\n\n"
+	)
+	// -- tests --
+
+	// init repo with auto publish being enabled but with a malformed URL.
+	_, err = runCommand(t, tempDir, "-s", "For the Horde!", "init", "-p", gun)
+	require.Error(t, err, "Trust server url has to be in the form of http(s)://URL:PORT.")
+	// init repo with auto publish being enabled but with an unaccessible URL.
+	_, err = runCommand(t, tempDir, "-s", "https://notary-server-on-the-moon:12306", "init", "-p", gun)
```
riyazdf: this is an awesome test!  Can we add a similar malformed URL and inaccessible URL test for an originally offline command, like `add`?  For `init` it's less applicable because the command would already talk to notary server without publishing to setup the timestamp.

cmd/notary/integration_test.go
```diff
@@ -1526,3 +1527,447 @@ func TestWitness(t *testing.T) {
...
+	output, err = runCommand(t, tempDir, "-s", server.URL, "list", "gun")
+	require.NoError(t, err)
+	require.True(t, strings.Contains(string(output), target1))
+
+	// lookup target and repo - see target
+	output, err = runCommand(t, tempDir, "-s", server.URL, "lookup", "gun", target1)
+	require.NoError(t, err)
+	require.True(t, strings.Contains(string(output), target1))
+
+	// remove target
```
cyli: Totally non-blocking - since this test case is just testing the autopublish functionality, the removal and testing different types of hashes may not be as necessary.

cyli:
Thanks for adding these @HuKeping!  Aside from @riyazdf's test comment, this LGTM!

cmd/notary/integration_test.go
HuKeping: fixed in commit `Use require.Contains other than require.True when neccessary` , also fix the others test in this file.

cmd/notary/integration_test.go
HuKeping: Fixed in commit `Use require.Contains other than require.True when neccessary`, also fix the others test in this file

cmd/notary/integration_test.go
HuKeping: addressed in commit `Refactor and clean up test`

cmd/notary/integration_test.go
HuKeping: addressed in commit `Refactor and clean up test`

riyazdf:
jenkins, test this please.

riyazdf:
LGTM.  Thank you @HuKeping for your continued work on this feature!

cmd/notary/integration_test.go
```diff
@@ -119,7 +120,7 @@ func TestInitWithRootKey(t *testing.T) {
 	// check that the root key used for init is the one listed as root key
 	output, err := runCommand(t, tempDir, "key", "list")
 	require.NoError(t, err)
-	require.True(t, strings.Contains(output, data.PublicKeyFromPrivate(privKey).ID()))
+	require.Contains(t, output, data.PublicKeyFromPrivate(privKey).ID())
```
endophage: Thanks for cleaning these up!

riyazdf:
@HuKeping - sorry for the inconvenience but could you please rebase this branch?

HuKeping:
rebased

cyli:
jenkins test this please

HuKeping:
Great to see this has been merged! Thanks for all your continued  review on this! :tada:

endophage:
@HuKeping thanks for picking it up! This is a really useful addition we're excited to see added.

Issue #752, opened 2016-05-19
Investigate behaviour of timestamp failure cases.


endophage:
It appears we may always default to the cached timestamp, even if we retrieve the remote one successfully but it subsequently fails to verify (for signatures in the case we've seen). In this particular case, we should instead fetch a new root.

We should only fall back to the cached timestamp if there was a network problem that prevented retrieval of the remote one.

PR #889, opened 2016-07-29
Rotate remote key endpoint

### LINKED ISSUE #846: Server Key Rotation: Schema updates

riyazdf:
Stacked on #880 to make use of the key marking functionality

Implements remote key rotation for snapshot and timestamp keys via `notary key rotate <GUN> <ROLE> -r`.

In case this operation is interrupted, we will attempt to sign in any "pending keys" in the database on the next remote rotation -- each role/gun pair can only have one pending key.

This PR also removes the `timestamp_keys` table in the notary server database, and instead parses out public keys from the TUF metadata in the `tuf_files` table.  If TUF metadata does not yet exist for a repo, we attempt to create new remote keys (using the same "pending" logic described above to prevent key explosion).

Closes #846, #847, #752
Supersedes #553, #880


migrations/server/mysql/0004_drop_timestamp_key.up.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE `timestamp_keys`;
```
cyli: \o/

server/handlers/default.go
```diff
@@ -245,6 +245,75 @@ func getKeyHandler(ctx context.Context, w http.ResponseWriter, r *http.Request,
...
 }
 
+// RotateKeyHandler rotates the remote key for the specified role, returning the public key
+func RotateKeyHandler(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+	defer r.Body.Close()
+	vars := mux.Vars(r)
+	return rotateKeyHandler(ctx, w, r, vars)
+}
+
+func rotateKeyHandler(ctx context.Context, w http.ResponseWriter, r *http.Request, vars map[string]string) error {
```
cyli: Non-blocking - since this code is basically the same as that for `getKeyHandler`, except for logging and which function `GetTimestampKey` vs `RotateTimestampKey` and `GetSnapshotKey` vs `RotateSnapshotKey`, I was wondering if we could refactor at least the logging part to be the same except for plugging in the method?

server/handlers/validation_test.go
```diff
@@ -675,13 +675,6 @@ func TestValidateSnapshotGenerateNoPrev(t *testing.T) {
 	repo, cs, err := testutils.EmptyRepo(gun)
 	require.NoError(t, err)
 	store := storage.NewMemStorage()
-	snapRole, err := repo.GetBaseRole(data.CanonicalSnapshotRole)
```
cyli: 👍  Thanks for cleaning these up!

server/snapshot/snapshot.go
```diff
@@ -16,35 +16,53 @@ import (
...
+	}
 
-		if _, ok := err.(*storage.ErrKeyExists); ok {
-			keyAlgorithm, public, err = store.GetKey(gun, data.CanonicalSnapshotRole)
-			if err != nil {
-				return nil, err
-			}
-			return data.NewPublicKey(keyAlgorithm, public), nil
-		}
+	// If we have a current root, parse out the public key ID for the snapshot role and get it from the underlying cryptoservice
```
cyli: nit:  "parse out the public key ID for the snapshot role, and return it" ?

server/storage/errors.go
```diff
@@ -4,10 +4,10 @@ import (
 	"fmt"
 )
 
-// ErrOldVersion is returned when a newer version of TUF metadada is already available
```
cyli: Thanks for fixing this typo!

server/snapshot/snapshot.go
riyazdf: thank you for catching that, I'll update this 👍

server/handlers/default.go
riyazdf: Definitely agree.  I can try to unify as much of this as possible 👍

server/timestamp/timestamp.go
```diff
@@ -22,34 +22,52 @@ import (
...
+	}
 
-		if _, ok := err.(*storage.ErrKeyExists); ok {
-			keyAlgorithm, public, err = store.GetKey(gun, data.CanonicalTimestampRole)
-			if err != nil {
-				return nil, err
-			}
-			return data.NewPublicKey(keyAlgorithm, public), nil
-		}
+	// If we have a current root, parse out the public key ID for the snapshot role and get it from the underlying cryptoservice
```
cyli: Similar comment nitpick here

migrations/server/mysql/0004_drop_timestamp_key.up.sql
cyli: Non-blocking - this is probably a bad idea from a historical point of view, but wondering if it's better to also remove the table from 0001 and add a `IF EXISTS` to this drop table command, that way it doesn't have to be created and then immediately dropped.

https://dev.mysql.com/doc/refman/5.7/en/drop-table.html to this if there is no error, and modify

riyazdf: I'd be more than happy to add a `IF EXISTS` to this drop table but I'm not sure if we should edit existing migrations?

cyli: Yeah I'm not sure either.  I guess you really only run migrations once, so creating and then destroying a table isn't an issue.

client/client_test.go
```diff
@@ -2732,6 +2790,14 @@ func TestRotateKeyAfterPublishServerManagementChange(t *testing.T) {
 		data.CanonicalTargetsRole:  false,
 		data.CanonicalRootRole:     false,
 	})
+	// check that the snapshot remote rotation creates new keys
```
cyli: non-blocking: I think this snapshot case may already be tested in the previous batch?  Although it rotates all those keys at once, as opposed to one at a time.  Possibly we can just add the timestamp role to the list of keys to rotate in these tests?

docs/advanced_usage.md
```diff
@@ -98,7 +98,8 @@ snapshot key to push their updates to the collection.
 Note that new collections created by a Docker 1.11 Engine client will have the server manage the snapshot key by default.
 To reclaim control of the snapshot key on the client, use the `notary key rotate` command without the `-r` flag.
 
-The targets key must be locally managed - to rotate the targets key, for instance in case of compromise, use the `notary key rotate targets` command without the `-r` flag.
+The root and targets key must be locally managed - to rotate either the root or targets key, for instance in case of compromise, use the `notary key rotate` command without the `-r` flag.
```
cyli: 👍  Thanks for updating the docs too!

server/handlers/default.go
cyli: I put this on the wrong line, so github hasn't collapsed this comment.  It has been addressed.  :)

server/handlers/default_test.go
```diff
@@ -172,6 +177,52 @@ func TestGetKeyHandlerCreatesOnce(t *testing.T) {
 	}
 }
 
+// Rotating a key for a non-supported role results in a 400.
+func TestRotateKeyHandlerInvalidRole(t *testing.T) {
+	state := defaultState()
+	for _, role := range []string{data.CanonicalRootRole, data.CanonicalTargetsRole, "targets/a", "invalidrole"} {
```
cyli: These are awesome!  Non-blocking:  wondering if we could unify this with the `TestGetKeyHandlerInvalidRole` the same way you did with `TestKeyHandlersNoRoleOrRepo`?   And apologies there was no existing test for getting a key with an invalid algorithm, but was wondering if that test could also be adapted to test both handlers?

server/server_test.go
```diff
@@ -304,3 +304,32 @@ func verifyGetResponse(t *testing.T, r *http.Response, expectedBytes []byte) {
...
+		context.Background(), "metaStore", storage.NewMemStorage())
+	ctx = context.WithValue(ctx, "keyAlgorithm", data.ED25519Key)
+
+	ccc := utils.NewCacheControlConfig(10, false)
+	handler := RootHandler(nil, ctx, signed.NewEd25519(), ccc, ccc, nil)
+	ts := httptest.NewServer(handler)
+	defer ts.Close()
+
+	rolesToStatus := map[string]int{
+		data.CanonicalTimestampRole: http.StatusOK, // just returning same as GetKey endpoint right now
```
cyli: Non-blocking - I'm not sure I understand the comment about the `GetKey` endpoint?

client/client_test.go
riyazdf: the snapshot case is definitely tested in the previous case but I wanted to do an additional rotation from a pre-existing server-managed snapshot key.

I can add the check that those keys exist on disk to `testRotateKeySuccess`

cyli: Ah ok - each invocation of `testRotateKeySuccess` wipes the repo and starts a new server - so if we wanted to test the additional rotation you'd probably want to do :

`testRotateKeySuccess(t, true, map[string]bool{data.CanonialSnapshotRole: true})`

That first boolean says whether to start the the repo off with a server-managed snapshot key.

server/handlers/default_test.go
riyazdf: Great idea, done!

server/server_test.go
riyazdf: yeah I think that's an out-of-date comment; removed it 👍

client/client_test.go
riyazdf: ah ok.  Fixed the test to do that

cyli:
I was taking this for a spin, and the remote snapshot rotation works great.  But for some reason when I rotate timestamp: 

1.  The first time I rotate it, it succeeds, but then when I list and look at the root.json, the timestamp is unchanged.
2. The second time I rotate it, it errors with an `ErrOldVersion` (also encountered https://github.com/docker/notary/pull/900).

I'm not sure why, since as far as I can tell the versions are all bumped.  Still trying to work it out.

riyazdf:
@cyli I've been digging into this and it seems like we're seeing the effect of this issue bubbling up: https://github.com/docker/notary/issues/752  It was never an issue before because we weren't making any changes to the timestamp role, but now that we can change keys and fall back to the cached timestamp, updating on the client had to be updated.  I've added changes to `downloadTimestamp` to handle this

client/tufclient.go
```diff
@@ -107,27 +107,36 @@ func (c *TUFClient) downloadTimestamp() error {
...
-	case cachedErr == nil:
-		logrus.Debug(remoteErr.Error())
-		logrus.Warn("Error while downloading remote metadata, using cached timestamp - this might not be the latest version available remotely")
-
-		err := c.newBuilder.Load(role, cachedTS, 1, false)
-		if err == nil {
-			logrus.Debug("successfully verified cached timestamp")
-		}
-		return err
+	case store.ErrMetaNotFound:
```
cyli: Non-blocking nitpick:  If both result in a break, I think we can use comma separated values here, so `case store.ErrMetaNotFound, store.ErrServerUnavailable:`

cyli: Also non-blocking question:  I think we don't translate it as any type atm, but should a 401 result in using the cache too?

riyazdf: I think 401s are translated to `ErrServerUnavailable{code: 401}` in the default case of `translateStatusToError`

cyli: Oh yes, that's the catchall, thanks!  (although that may be a misleading error name  :))

client/client_test.go
```diff
@@ -2768,6 +2834,11 @@ func testRotateKeySuccess(t *testing.T, serverManagesSnapshotInit bool,
 			keysToExpectCreated = append(keysToExpectCreated, role)
 		}
 	}
+
+	for _, role := range keysToExpectCreated {
```
cyli: Actually I'm really really sorry to do this after you've made this change, but apparently I may have just forgotten to do something with this because `requireRotationSuccessful` checks that the new key, so possibly we can skip this check entirely?  And also just not collect the expected to be created keys.  :(  Sorry sorry.

cyli:
Thanks so much for picking up this feature and fixing #752 too!  LGTM!

client/client_test.go
riyazdf: no problem, always happy to not include unneeded code 😄   Done.

server/snapshot/snapshot.go
```diff
@@ -16,35 +16,53 @@ import (
...
-	if _, ok := err.(*storage.ErrNoKey); ok {
+func GetOrCreateSnapshotKey(gun string, store storage.MetaStore, crypto signed.CryptoService, createAlgorithm string) (data.PublicKey, error) {
+	_, rootJSON, err := store.GetCurrent(gun, data.CanonicalRootRole)
+	if err != nil {
+		// If the error indicates we couldn't find the root, create a new key
+		if _, ok := err.(storage.ErrNotFound); !ok {
+			logrus.Errorf("Error when retrieving root role for GUN %s: %v", gun, err)
+			return nil, err
+		}
 		key, err := crypto.Create(data.CanonicalSnapshotRole, gun, createAlgorithm)
```
cyli: group review:  we can just `return crypto.Create(data.CanonicalSnapshotRole, gun, createAlgorithm)` here - similarly for `GetOrCreateTimestampKey`

server/snapshot/snapshot.go
```diff
@@ -16,35 +16,53 @@ import (
...
+
+	snapshotRole, err := repoSignedRoot.BuildBaseRole(data.CanonicalSnapshotRole)
+	if err != nil {
+		logrus.Errorf("Failed to extract snapshot role from root for GUN %s", gun)
 		return nil, err
 	}
+
+	// We currently only support single keys for snapshot and timestamp, so we can return the first and only key in the map
+	for _, pubKey := range snapshotRole.Keys {
+		return pubKey, nil
```
cyli: group review:  we probably need to get the key by ID from the signer to make sure we have the key, otherwise notary 0.3's rotate snapshot key to server will fail because this might just return the locally-owned snapshot key (doesn't check if the signer actually has a key).

If we make it through this loop without finding a key that the signer has, we should probably create a new key.

Similarly with `GetOrCreateTimestampKey`

signer/keydbstore/rethink_keydbstore.go
```diff
@@ -241,6 +266,45 @@ func (rdb RethinkDBKeyStore) RotateKeyPassphrase(keyID, newPassphraseAlias strin
...
+// Create will attempt to first re-use an inactive key for the same role, gun, and algorithm.
+// If one isn't found, it will create a private key and add it to the DB as an inactive key
+func (rdb RethinkDBKeyStore) Create(role, gun, algorithm string) (data.PublicKey, error) {
+	dbPrivateKey := RDBPrivateKey{}
+	res, err := gorethink.DB(rdb.dbName).Table(dbPrivateKey.TableName()).
+		Filter(gorethink.Row.Field("gun").Eq(gun)).
+		Filter(gorethink.Row.Field("role").Eq(role)).
+		Filter(gorethink.Row.Field("algorithm").Eq(algorithm)).
+		Filter(gorethink.Row.Field("last_used").Eq(time.Time{})).
+		Run(rdb.sess)
```
cyli: group review: we might want to sort by rethink ID just so if there are multiple unused keys, calling create multiple times returns the same value rather than switching back enough.

riyazdf: we're using the keyID as the primary key, so I've changed this to sort by that in case we get in a weird state with multiple key IDs

server/snapshot/snapshot.go
riyazdf: done!

server/snapshot/snapshot.go
riyazdf: 👍  done

server/snapshot/snapshot_test.go
```diff
@@ -46,60 +47,68 @@ func TestGetSnapshotKeyCreate(t *testing.T) {
 
 	require.Nil(t, err, "Expected nil error")
 
-	// trying to get the same key again should return the same value
-	require.Equal(t, k, k2, "Did not receive same key when attempting to recreate.")
+	// trying to get the key for the same gun and role will create a new key unless we have existing TUF metadata
```
cyli: Could we also add a test for if there is existing TUF metadata, but the key in the metadata is not held by the signer?  (similarly for timestamp?)

riyazdf: good idea, added to `TestGetSnapshotKeyExistingMetadata` and `TestGetTimestampKeyExistingMetadata`

server/timestamp/timestamp_test.go
```diff
@@ -230,3 +226,82 @@ func TestCreateTimestampNoKeyInCrypto(t *testing.T) {
...
+	require.Nil(t, err, "Expected nil error")
+
+	require.Equal(t, k, k2, "Did not receive same key when attempting to recreate.")
+	require.NotNil(t, k2, "Key should not be nil")
+
+	// try wiping out the cryptoservice data, and ensure we create a new key because the signer doesn't hold the key specified by TUF
+	crypto = signed.NewEd25519()
+	k3, err := GetOrCreateTimestampKey("gun", store, crypto, data.ED25519Key)
+	require.Nil(t, err, "Expected nil error")
+	require.NotEqual(t, k, k3, "Did not receive same key when attempting to recreate.")
```
cyli: Non-blocking nitpick: should this be "Received same key when..."?

server/timestamp/timestamp_test.go
```diff
@@ -230,3 +226,82 @@ func TestCreateTimestampNoKeyInCrypto(t *testing.T) {
...
+
+func TestGetTimestampKeyCreateWithCorruptedStore(t *testing.T) {
+	store := CorruptedStore{storage.NewMemStorage()}
+	crypto := signed.NewEd25519()
+	k, err := GetOrCreateTimestampKey("gun", store, crypto, data.ED25519Key)
+	require.Error(t, err, "Expected error")
+	require.Nil(t, k, "Key should be nil")
+}
+
+func TestGetSnapshotKeyCreateWithInvalidAlgo(t *testing.T) {
```
cyli: Should this be "...GetTimestampKey..."?

cyli:
Thanks for all your work on this!  This LGTM again aside from some minor nits on naming!

server/timestamp/timestamp_test.go
riyazdf: yes - fixed both typos!

endophage:
LGTM. Excited to merge this! @riyazdf can you rebase just so we can get a final test run to make sure everything is good, then we can merge if it's all green :-D

riyazdf:
@endophage sounds good, just rebased!

cyli:
Yay all green!

Issue #226, opened 2015-10-16
Look into just using grpc/health


cyli:
If the notary-signer key-management and signer services are always meant to be run on the same machine, we can use https://godoc.org/google.golang.org/grpc/health instead of having a health check on each interface.

cyli:
Closing in favor of #894

PR #890, opened 2016-08-01
Update protobuf

### LINKED ISSUE #888: Update to protobuf 3.0.0
Protobuf 3.x is now GA. We should update to it rather than continuing to use a beta version.

HuKeping:
Fixes #888 

Didn't have an experience to update the Godpes, please feel free to close this if I did that wrong.


signer/client/signer_trust.go
```diff
@@ -23,7 +23,6 @@ import (
 
 // The only thing needed from grpc.ClientConn is it's state.
 type checkableConnectionState interface {
-	State() grpc.ConnectivityState
```
HuKeping: State() is an Experimental API and has been removed long time ago.

signer/client/signer_trust.go
```diff
@@ -180,16 +179,7 @@ func (trust *NotarySigner) ListAllKeys() map[string]string {
...
-
-	// Do not bother starting checking at all if the connection is broken.
-	if trust.clientConn.State() != grpc.Idle &&
-		trust.clientConn.State() != grpc.Ready {
-		return fmt.Errorf("Not currently connected to trust server.")
-	}
-
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
-	// the kmClient should point at the same server as the sClient, so a single
-	// healthcheck should suffice
```
HuKeping: Since we have already removed the `CheckHealth` interface from `Signer`, I think it might be good to remove this comment as well.

cyli:
jenkins test this please

HuKeping:
rebased

cyli:
jenkins, test this please

signer/client/signer_trust.go
cyli: This one might make sense to leave because folks might assume `kmClient` and `sClient` should both have health checks, if they were different services.

signer/client/signer_trust.go
cyli: Thanks - we should probably remove that at some point.  :)

signer/client/signer_trust.go
```diff
@@ -180,16 +179,7 @@ func (trust *NotarySigner) ListAllKeys() map[string]string {
 // CheckHealth checks the health of one of the clients, since both clients run
 // from the same GRPC server.
 func (trust *NotarySigner) CheckHealth(timeout time.Duration) error {
-
-	// Do not bother starting checking at all if the connection is broken.
-	if trust.clientConn.State() != grpc.Idle &&
```
riyazdf: thank you for removing the `State()` calls but I think we should still check the client connection status, it seems that the `grpc.ConnectivityState` is still a typed int corresponding to the same constants: https://godoc.org/google.golang.org/grpc#ConnectivityState

signer/client/signer_trust.go
endophage: @cyli been wondering about that since we removed it. Would it be possible to have the endpoint defined in both APIs but have our signer satisfy both with a single function? I don't know if there's a use case for putting key management and signing behind different services with different permissions. I could imagine somebody wanting to lock down key management more strongly than signing, especially if it's possible to cause keys to be deleted.

signer/client/signer_trust.go
cyli: Actually nm, see @riyazdf 's comment https://github.com/docker/notary/pull/890/files#r73199398 :)

signer/client/signer_trust.go
cyli: @endophage That is how it worked previously - we can always put it back.  It just seems like if there are two health checks and hence potentially two services, we also should check both of them in the client health check?

HuKeping:
Do you think it make sense to update the health check mechanism in a separated PR and make this one only update the vendor.

I'd like to add a new health check grpc service. Per [the official documentation](https://github.com/grpc/grpc/blob/master/doc/health-checking.md) . An empty name triggers a check for generic health of the whole Notary-Signer service and ofcourse register both of the `kmClient` and `sClient` when they are ready.

cyli:
@HuKeping Yes, I think any health check changes should probably go into a separate PR.  And yes, using the GRPC service would be great!  You could finally close https://github.com/docker/notary/issues/226 :D

signer/client/signer_trust.go
HuKeping: You are right @riyazdf , ConnectivityState is still there but it can not directly accessed by `ClientConn` any more. It was now hold by another structure `addrConn` which I would like to use when doing the refactor of the health check mechanism as I said below.

HuKeping:
OK @cyli , then  rebased

cyli:
jenkins, test this please

signer/client/signer_trust.go
cyli: It looks like `addrConn` is unexported too.  :|  We might not be able to check the connection status here before actually making the request (which will fail if there's no connection, it just might be slower).

cyli:
LGTM

signer/client/signer_trust.go
HuKeping: That can be done with the [health check hook](https://github.com/grpc/grpc-go/blob/master/health/health.go). I plan to add it to the Notary-Signer service.

HuKeping:
@cyli `clientConn` has been removed with the new commit.

cyli:
jenkins, test this please

riyazdf:
thank you @HuKeping for updating this!  LGTM

PR #891, opened 2016-08-01
Change debug phrasing for pkcs11 initialization failure

riyazdf:
Small update to our debug statement for when we fail to initialize a PKCS11 environment, since this is a common source of confusion.


cyli:
LGTM!  Thank you - this message is less scary looking :)

endophage:
Given we know what it means, would it make sense to make it even more user friendly and just say something like "No Yubikey found, using alternative key storage."

riyazdf:
@endophage I like that messaging better, updated 👍

endophage:
Cool, LGTM!

endophage:
Before merging, as you'd already LGTM'd the initial message, @cyli: any thoughts on the updated message?

cyli:
That's definitely even less scary than the previous message.  👍

PR #895, opened 2016-08-03
notary delete CLI command

riyazdf:
Wires up `DeleteTrustData` from the client library to a `notary delete` command, which will delete all local TUF metadata by default and additionally the remote metadata if the `--remote` flag is given.

Closes https://github.com/docker/notary/issues/710

### DIFF
--- a/cmd/notary/integration_test.go
+++ b/cmd/notary/integration_test.go
@@ -24,10 +24,12 @@ import (
 	"github.com/Sirupsen/logrus"
 	ctxu "github.com/docker/distribution/context"
 	"github.com/docker/notary"
+	"github.com/docker/notary/client"
 	"github.com/docker/notary/cryptoservice"
 	"github.com/docker/notary/passphrase"
 	"github.com/docker/notary/server"
 	"github.com/docker/notary/server/storage"
+	notaryStorage "github.com/docker/notary/storage"
 	"github.com/docker/notary/trustmanager"
 	"github.com/docker/notary/tuf/data"
 	"github.com/docker/notary/tuf/utils"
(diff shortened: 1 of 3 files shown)


cmd/notary/integration_test.go
```diff
@@ -245,6 +247,153 @@ func TestClientTUFInteraction(t *testing.T) {
...
+
+	server := setupServer()
+	defer server.Close()
+
+	tempFile, err := ioutil.TempFile("", "targetfile")
+	require.NoError(t, err)
+	tempFile.Close()
+	defer os.Remove(tempFile.Name())
+
+	// Setup certificate
```
cyli: Non-blocking nitpick - setting up the cert and adding a delegation role + target to the delegation seems to also be done by `TestPurge` and `TestClientDelegationsInteraction` and `TestClientDelegationsPublishing`, although this one doesn't need the key ID of the cert.  I was wondering if we could factor this logic out to a helper function?

cmd/notary/integration_test.go
```diff
@@ -245,6 +247,153 @@ func TestClientTUFInteraction(t *testing.T) {
...
+	// list delegations - see role because remote data still exists
+	output, err = runCommand(t, tempDir, "-s", server.URL, "delegation", "list", "gun")
+	require.NoError(t, err)
+	require.True(t, strings.Contains(string(output), "targets/delegation"))
+
+	// Trying to delete the repo with the remote flag fails if it can't connect to the server
+	output, err = runCommand(t, tempDir, "-s", "https://invalidurl:4443", "delete", "gun", "--remote")
+	require.Error(t, err)
+	require.IsType(t, notaryStorage.ErrOffline{}, err)
+
```
cyli: Do we want to assert that after this command the local metadata is still there (since getTransport fails before we make the call to `DeleteTrustData`)?

cmd/notary/integration_test.go
```diff
@@ -245,6 +247,153 @@ func TestClientTUFInteraction(t *testing.T) {
...
+	// Delete the repo remotely and locally, pointing to the correct server
+	output, err = runCommand(t, tempDir, "-s", server.URL, "delete", "gun", "--remote")
+	require.NoError(t, err)
+	assertNoMetadataForGun(t, tempDir, "gun")
+	_, err = runCommand(t, tempDir, "-s", server.URL, "list", "gun")
+	require.Error(t, err)
+	require.IsType(t, client.ErrRepositoryNotExist{}, err)
+
+	// Silent success on extraneous deletes
```
cyli: Awesome detail on the tests!

cmd/notary/tuf.go
```diff
@@ -83,6 +83,12 @@ var cmdTUFVerifyTemplate = usageTemplate{
 	Long:  "Verifies if the data passed in STDIN is included in the remote trusted collection identified by the Globally Unique Name.",
 }
 
+var cmdTUFDeleteTemplate = usageTemplate{
+	Use:   "delete [ GUN ]",
+	Short: "Deletes all content for a trusted collection",
+	Long:  "Deletes all content for a trusted collection identified by the Globally Unique Name, can delete both local and remote data based on flags.",
```
cyli: nitpick:  "Deletes all local content for a trusted collection identified by the Globally Unique Name.  Remote data can also be deleted based on flags." ?

cyli: (I don't feel strongly at all on the wording, since the flag specifies additional detail - mainly was thinking there should be a semicolon or period between the two parts of the sentence :))

cyli:
Thanks for adding this so quickly @riyazdf!  Couple of minor nits on tests, but otherwise LGTM!

cmd/notary/integration_test.go
riyazdf: good idea, I'll add this

cyli: Ah right, sorry, I got confused -`getTransport` I guess only tries to make a connection if auth is set up.

riyazdf: Yeah, I've differentiated the cases for an invalid URL vs. a well-formed URL that doesn't actually point to notary in the tests so we have this tracked

cyli: 👍 thank you!

cmd/notary/integration_test.go
riyazdf: Good idea - added a helper to get a private key, cert, and keyID

cyli:
LGTM!  Thanks for fixing this so fast!

endophage:
LGTM! Tests look great too 👍

PR #896, opened 2016-08-03
changed GetAllTargetMetadataByName to return a TargetSignedStruct, also verifies messages against signatures

avaid96:
changed GetAllTargetMetadataByName to return a TargetSignedStruct which now contains signature data


cyli:
jenkins, ok to test

cyli:
jenkins, retest this please

client/client.go
```diff
@@ -478,14 +478,20 @@ func (r *NotaryRepository) GetTargetByName(name string, roles ...string) (*Targe
...
 }
 
+// TargetSignedStruct is a struct that contains information about the Target and the list of signatures
+type TargetSignedStruct struct {
+	Target     Target
+	Signatures []data.Signature
+}
+
 // GetAllTargetMetadataByName searches the entire delegation role tree to find the specified target by name for all
 // roles, and returns a map of role strings to Target structs for each time it finds the specified target.
```
riyazdf: nit: we should update this comment to reflect the new behavior

cyli:
jenkins, retest this please

client/client_test.go
```diff
@@ -3617,38 +3617,45 @@ func TestGetAllTargetInfo(t *testing.T) {
...
-	require.Equal(t, targetData["targets/level1"], *level1CurrentTarget)
-	require.Equal(t, targetData["targets/level2"], *level2CurrentTarget)
+	require.NotNil(t, targetSignatureData)
+	require.Len(t, targetSignatureData, 3)
+	require.Len(t, targetSignatureData[data.CanonicalTargetsRole].Signatures, 1)
+	require.Equal(t, (targetSignatureData[data.CanonicalTargetsRole]).Target, *targetsCurrentTarget)
+	require.Equal(t, (targetSignatureData["targets/level1"]).Target, *level1CurrentTarget)
+	require.Len(t, targetSignatureData["targets/level1"].Signatures, 1)
+	require.Equal(t, (targetSignatureData["targets/level2"]).Target, *level2CurrentTarget)
+	require.Len(t, targetSignatureData["targets/level2"].Signatures, 1)
```
riyazdf: In addition to checking the length of the signature, can we check that the signature `KeyID` corresponds to the role's key, and that the signature verifies correctly?

client/client.go
```diff
@@ -478,14 +478,20 @@ func (r *NotaryRepository) GetTargetByName(name string, roles ...string) (*Targe
 
 }
 
+// TargetSignedStruct is a struct that contains information about the Target and the list of signatures
+type TargetSignedStruct struct {
```
cyli: Non-blocking - this is just my misunderstanding of what we wanted:  did we want to also include the role in here so that we can look up any x509 certificates too?

avaid96: You're right, I misunderstood @endophage, working on a fix right now. It should be up soon :)

client/client.go
```diff
@@ -495,7 +502,12 @@ func (r *NotaryRepository) GetAllTargetMetadataByName(name string) (map[string]T
 		// We found the target and validated path compatibility in our walk,
 		// so add it to our list
 		if resultMeta, foundTarget := tgt.Signed.Targets[name]; foundTarget {
-			targetInfoMap[validRole.Name] = Target{Name: name, Hashes: resultMeta.Hashes, Length: resultMeta.Length}
+			targetInfo := TargetSignedStruct{
+				validRole,
```
endophage: Should use named fields, i.e.
```
targetInfo := TargetSignedStruct{
    Role: validRole,
    Target: Target{Name: name, Hashes: resultMeta.Hashes, Length: resultMeta.Length},
    Signatures: tgt.Signatures,
}
```

endophage:
Looks great. What other tests are you planning to add? It seems like the tests you've already added are probably sufficient or is there some edge case I'm missing?

avaid96:
@endophage as per Riyaz' comment "In addition to checking the length of the signature, can we check that the signature KeyID corresponds to the role's key, and that the signature verifies correctly?" I was thinking of adding a test that checks that the signature is as expected in the target

client/client.go
```diff
@@ -478,14 +478,21 @@ func (r *NotaryRepository) GetTargetByName(name string, roles ...string) (*Targe
...
+type TargetSignedStruct struct {
+	Role       data.DelegationRole
+	Target     Target
+	Signatures []data.Signature
+}
+
 // GetAllTargetMetadataByName searches the entire delegation role tree to find the specified target by name for all
-// roles, and returns a map of role strings to Target structs for each time it finds the specified target.
-func (r *NotaryRepository) GetAllTargetMetadataByName(name string) (map[string]Target, error) {
+// roles, and returns a map of role strings to TargetSignedStructs for each time it finds the specified target.
```
cyli: nitpick: maybe "returns a list of TargetSignedStructs..."?

client/client_test.go
```diff
@@ -3617,38 +3617,67 @@ func TestGetAllTargetInfo(t *testing.T) {
...
+			require.Len(t, tarSigStr.Signatures, 1)
+			require.Equal(t, tarSigStr.Target, *level1CurrentTarget)
+			makeSureWeHitEachCase = makeSureWeHitEachCase + 2
+		case "targets/level2":
+			require.Len(t, tarSigStr.Signatures, 1)
+			require.Equal(t, tarSigStr.Target, *level2CurrentTarget)
+			makeSureWeHitEachCase = makeSureWeHitEachCase + 3
+		}
+	}
+	require.Equal(t, makeSureWeHitEachCase, 6)
```
cyli: 👍 nice!

client/client_test.go
```diff
@@ -3617,38 +3617,67 @@ func TestGetAllTargetInfo(t *testing.T) {
...
+
+	targetSignatureData, err = repo.GetAllTargetMetadataByName("other")
+	require.NoError(t, err)
+	require.NotNil(t, targetSignatureData)
+	require.Len(t, targetSignatureData, 1)
+	require.Equal(t, (targetSignatureData[0]).Target, *level1OtherTarget)
```
cyli: Can we also add a role name check for the data for "other" and "latest"?

client/client.go
```diff
@@ -478,14 +478,21 @@ func (r *NotaryRepository) GetTargetByName(name string, roles ...string) (*Targe
 
 }
 
+// TargetSignedStruct is a struct that contains information about the Target and the list of signatures
```
riyazdf: nit: can we change this description to

` TargetSignedStruct is a struct that contains a Target, the role it was found in, and the list of signatures for that role`

client/client_test.go
```diff
@@ -3617,38 +3620,76 @@ func TestGetAllTargetInfo(t *testing.T) {
...
+		case "targets/level1":
+			require.Len(t, tarSigStr.Signatures, 1)
+			require.Equal(t, key1.ID(), tarSigStr.Signatures[0].KeyID)
+			require.Equal(t, tarSigStr.Target, *level1CurrentTarget)
+			makeSureWeHitEachCase = makeSureWeHitEachCase + 2
+		case "targets/level2":
+			require.Len(t, tarSigStr.Signatures, 1)
+			require.Equal(t, key2.ID(), tarSigStr.Signatures[0].KeyID)
+			require.Equal(t, tarSigStr.Target, *level2CurrentTarget)
+			makeSureWeHitEachCase = makeSureWeHitEachCase + 4
```
riyazdf: nit: not sure if it's worth keeping this as a loop instead of just indexing by the role name

tuf/data/types.go
```diff
@@ -111,6 +111,7 @@ type Signature struct {
 	KeyID     string       `json:"keyid"`
 	Method    SigAlgorithm `json:"method"`
 	Signature []byte       `json:"sig"`
+	IsValid   bool         `json:"-isvalid"`
```
riyazdf: should the json tag for the key just be `-` so it's ignored?  Based on https://golang.org/pkg/encoding/json/#Marshal

tuf/signed/verify.go
```diff
@@ -92,7 +93,7 @@ func VerifySignatures(s *data.Signed, roleData data.BaseRole) error {
 }
 
 // VerifySignature checks a single signature and public key against a payload
```
riyazdf: nit: we should mention that this function will actually mutate the signature to mark it as valid if it verifies against the public key and message in the description

tuf/data/types.go
cyli: I think `json:"-"` is the syntax for skipping importing/exporting the field?  It might be also be good, just for our own sanity since `VerifySignatures` is now setting this value, to add a test that marshals a `data.Signed` object using canonical JSON before calling `VerifySignatures`, and then marsahalling it again after and ensuring that the bytes haven't changed.  This will make sure the `IsValid` bool is not being exported.

tuf/signed/verify_test.go
```diff
@@ -152,6 +163,10 @@ func TestUnknownKeyBelowThreshold(t *testing.T) {
 	s.Signatures = append(s.Signatures)
 	err = VerifySignatures(s, roleWithKeys)
 	require.IsType(t, ErrRoleThreshold{}, err)
+	require.Len(t, s.Signatures, 2)
+	// order of the list is sometimes flaky, so using contains instead
+	require.Contains(t, []bool{s.Signatures[0].IsValid, s.Signatures[1].IsValid}, true)
+	require.Contains(t, []bool{s.Signatures[0].IsValid, s.Signatures[1].IsValid}, false)
```
riyazdf: nit: instead of doing it like this, could we have individual checks against each key as specified by key ID?  It'll be more verbose, but I think it'll be clearer to understand (and debug if the tests breaks)

tuf/data/types.go
cyli: Oops, sorry, @riyazdf beat me to it :D

tuf/signed/verify.go
```diff
@@ -77,10 +77,11 @@ func VerifySignatures(s *data.Signed, roleData data.BaseRole) error {
 		if key.ID() != sig.KeyID {
 			return ErrInvalidKeyID{}
 		}
-		if err := VerifySignature(msg, sig, key); err != nil {
+		if err := VerifySignature(msg, &sig, key); err != nil {
 			logrus.Debugf("continuing b/c %s", err.Error())
 			continue
 		}
+		s.Signatures[i].IsValid = sig.IsValid
```
riyazdf: The double-assign between `VerifySignature` and `VerifySignatures` we have to do here for `IsValid` on the same signature feels like we should be using pointers to refer to `Signatures` from `Signed`, and only having `VerifySignature` do any mutation.

That said, it's a significant change for all of our data structures given that we aren't making heavy use of `IsValid`.  cc @endophage @cyli for more discussion - maybe instead we could have a separate utility function to filter valid signatures?

cyli: Maybe we can just set `IsValid` to true, since `VerifySignature` will error if the signature is not valid, and the line should never be reached if it's invalid?  That way `VerifySignature` mutates nothing, and only `VerifySignatures` actually does the mutation.

riyazdf: @cyli that sounds good.  The only other place `VerifySignature` is called outside of testing is in `Sign`, and I'm fine with signatures there not having to care about `IsValid` if we aren't going to export the field 👍

avaid96: sounds good, I'll edit the functions accordingly. Thanks for the awesome and quick review 🐳

cyli: We could also have `Sign` set the `IsValid` field once it's finished verifying it, too.  Another way to do this, if we want `VerifySignature` to actually do the mutation (if we cared about `Sign` signatures having the correct boolean set):

```
	for i := range s.Signatures {
		sig := &(s.Signatures[i])
                ...
		if err := VerifySignature(msg, sig, key); err != nil {
                    ...
```

So we can skip the additional assignment.

avaid96: Using this method suggested by @cyli in the last comment if that's fine with all? @riyazdf @endophage

riyazdf: sounds good 👍

tuf/data/types.go
avaid96: I've added the test, thanks😃

tuf/signed/verify.go
avaid96: done, thanks! :)

trustpinning/certs_test.go
```diff
@@ -183,7 +183,8 @@ func TestValidateRootWithPinnedCert(t *testing.T) {
 	// This call to trustpinning.ValidateRoot should also succeed with the correct Cert ID (same as root public key ID), even though we passed an extra bad one
 	validatedSignedRoot, err = trustpinning.ValidateRoot(nil, &testSignedRoot, "docker.com/notary", trustpinning.TrustPinConfig{Certs: map[string][]string{"docker.com/notary": {rootPubKeyID, "invalidID"}}, DisableTOFU: true})
 	require.NoError(t, err)
-	require.Equal(t, validatedSignedRoot, typedSignedRoot)
+	typedSignedRoot.Signatures[0].IsValid = true
```
riyazdf: non-blocking nit: can we add a comment here and to the other `IsValid` assignment below explaining that this extra assignment is necessary because `ValidateRoot` calls through to a successful `VerifySignatures` which marks `IsValid`?

tuf/signed/verify_test.go
```diff
@@ -77,6 +81,30 @@ func TestExactlyEnoughSigs(t *testing.T) {
...
+	before, err := json.MarshalCanonical(s.Signatures[0])
+	require.NoError(t, err)
+	require.False(t, s.Signatures[0].IsValid)
+	require.NoError(t, VerifySignature(b, &(s.Signatures[0]), k))
+	// the IsValid field changed
+	require.True(t, s.Signatures[0].IsValid)
+	after, err := json.MarshalCanonical(s.Signatures[0])
+	require.NoError(t, err)
+	// but the marshalled byte strings stay the same since IsValid is not exported
+	require.Equal(t, before, after)
```
riyazdf: awesome test!

riyazdf:
LGTM!

endophage:
LGTM. The `Signature.IsValid` stuff looks great! I think it's also going to be really useful when we get to more verbose displays of Notary state.

PR #900, opened 2016-08-06
We are returning a pointer to an error rather than an error itself.

### LINKED ISSUE #898: Correctly translate ErrOldVersion errors
Both the `storage` and the `server` packages have `ErrOldVersion`, and for one we return 400 (https://github.com/docker/notary/blob/master/server/handlers/default.go#L108) and the other one results in a 500.  Both should return 400, and really we should just eliminate one of them.

cyli:
Fixes #898 

Here I'm just changing this to return the concrete type everywhere.  We could instead just change the type we compare against in https://github.com/docker/notary/blob/master/server/handlers/default.go#L108 to be a pointer instead of a concrete type.


riyazdf:
thank you for catching and fixing this!  LGTM.

endophage:
Nice cleanup. LGTM!

PR #901, opened 2016-08-08
Implement GetAll for TUF data

jchorl:
GetAll TUF data between a specific date range

Signed-off-by: Josh Chorlton <josh.chorlton@docker.com>

### DIFF
--- a/server/errors/errors.go
+++ b/server/errors/errors.go
@@ -38,6 +38,12 @@ var (
 		Description:    "The client sent malformed JSON.",
 		HTTPStatusCode: http.StatusBadRequest,
 	})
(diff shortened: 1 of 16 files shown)


riyazdf:
Thanks for adding this!  Can we add the two following test cases for `QueryTUFFilesHandler`?
- Call `QueryTUFFilesHandler` with a valid `createdAfter` that gets parsed (maybe in `TestQueryTUFFilesHandlerNoStorage`?)
- Call `QueryTUFFilesHandler` with a store that errors on `GetAll`

endophage:
Closing for now as use case isn't materializing and further changes will likely be required by the time it does.

PR #904, opened 2016-08-09
added commands document (notary cheatsheet)

avaid96:

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -1,0 +1,100 @@
+<!--[metadata]>
++++
+title = "Understand the typical commands"
+description = "Description of key notary commands"
+keywords = ["docker, notary, notary-client"]
++++
+<![end-metadata]-->
+
+Notary is a tool for publishing and managing trusted collections of content. Publishers can digitally sign collections and consumers can verify integrity and origin of content. This ability is built on a straightforward key management and signing interface to create signed collections and configure trusted publishers.
+
+With Notary anyone can provide trust over arbitrary collections of data. Using The Update Framework (TUF) as the underlying security framework, Notary takes care of the operations necessary to create, manage and distribute the metadata necessary to ensure the integrity and freshness of your content.
+
+# Notary terminology 
+
+1. Collection: A collection is a conceptual source of target files of interest to the application.
+2. GUN: Notary uses Globally Unique Names (GUNs) to identify trust collections
+3. Target: we will refer to the files the framework is used to distribute as "target files".  Target files are opaque to the framework. Whether target files are packages containing multiple files, single text files, or executable binaries is irrelevant to Notary 
+4. Key roles: 
+    - Root = The root role delegates trust to specific keys trusted for all other top-level roles used in the system.
+    - Targets = The targets role's signature indicates which target files are trusted by clients. 
+        - Delegation roles = the targets role can delegate full or partial trust to other roles.  Delegating trust means that the targets role indicates another role (that is, another set of keys and the threshold required for trust) is trusted to sign target file metadata.
+    - Snapshot =  The snapshot role signs a metadata file that provides information about the latest version of all of the other metadata on the repository (excluding the timestamp file, discussed below).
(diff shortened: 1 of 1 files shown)


docs/commands.md
```diff
@@ -0,0 +1,86 @@
...
+### Admin actions
+`notary delegation add example.com/image targets/qa_team <newkey>`
+
+`notary delegation remove example.com/image targets/qa_team <oldkey>`
+
+`notary publish example.com/image`
+### Owner of old key after admin actions (Not always necessary depending on state)
+`notary witness example.com/image targets/qa_team`
+
+Witness marks roles to be re-signed the next time they're published. It recovers a role that has been made invalid by all keys being removed. It's only necessary if, for example, key4 was the current signer of targets/qa_team and its removal made the role's targets file invalid
```
NathanMcCauley: This comment makes no sense. I still maintain that `witness` shouldn't be described or thought of as a way to recover from key revocations.

If that is a secondary use case for `witness`, that's fine. But it can't be the primary, documented use for the command.

docs/commands.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Initialize a collection
+`notary init example.com/image`
+
+`notary publish example.com/image`
+
+## Add delegations
+`notary delegation add example.com/image targets/qa_team --all-paths <key>`
+
+Note: This will be staged as a changelist, you will need to publish for it to take effect
```
NathanMcCauley: Not sure it makes sense to call this out after every command. Maybe there should be a general doc somewhere that describes that at a high level how notary makes local changes that eventually need to be published?

docs/commands.md
```diff
@@ -0,0 +1,86 @@
...
+title = "Understand the typical commands"
+description = "Description of key notary commands"
+keywords = ["docker, notary, notary-client"]
++++
+<![end-metadata]-->
+
+# Understand the typical notary commands
+
+## Initialize a collection
+`notary init example.com/image`
```
NathanMcCauley: I think the use of `example.com` makes these a bit confusing. Could we pick a GUN that doesn't have a domain in it?

docs/commands.md
```diff
@@ -0,0 +1,86 @@
+<!--[metadata]>
```
NathanMcCauley: Do we need a mechanism for folks to add a specific hash?

docs/commands.md
avaid96: My original thought process when I made this comment was in context of the delegation key rotation. Happy to add a separate section for witness as well

docs/commands.md
avaid96: of course, we could go with `examplegun`?

endophage: That works. We should also define what a "GUN" is at the top. It's also defined elsewhere but a cheatsheet should be as self contained as possible.

docs/commands.md
avaid96: done 😃

docs/commands.md
```diff
@@ -0,0 +1,96 @@
...
+description = "Description of key notary commands"
+keywords = ["docker, notary, notary-client"]
++++
+<![end-metadata]-->
+
+# Understand the typical notary commands
+
+## Terminology
+
+1. GUN: Notary uses Globally Unique Names (GUNs) to identify trust collections
```
avaid96: I was thinking we could add explanations of what the canonical roles are, what a target is, what a delegation is here? Thoughts on those + any additions? @cyli @endophage @riyazdf @NathanMcCauley

cyli: That would probably be good to have a brief overview of the roles, and then maybe link to the TUF spec for more info?

riyazdf:
jenkins, ok to test

HuKeping:
@avaid96  would you mind if closing this in favour of #946

avaid96:
sure, closing in favour of #946. Thanks @riyazdf for picking this up :smiley:

PR #906, opened 2016-08-10
Passphrase fixups

### LINKED ISSUE #812: Passphrase attempt and errorring logic for keys could be improved
I was doing some local testing when I realized that it's possible to fail `> attempts` times on passphrases that are too short, but then notary will fail with `ErrAttemptsExceeded` after inputting a valid passphrase.  While most users may not run into this exact failure case, I think we could revisit and improve the passphrase attempt logic UX at the KeyStore level in general

```
🐳 $ bin/notary init repo
Root key found, using: 4327a8deba2cae19a4b393d2c6cf3a2f349652c84bc41bc1e09ade182d86b19a
Enter passphrase for root key with ID 4327a8d:
Enter passphrase for new targets key with ID f511f86 (repo):
Passphrase is too short. Please use a password manager to generate and store a good random passphrase.
Enter passphrase for new targets key with ID f511f86 (repo):
Passphrase is too short. Please use a password manager to generate and store a good random 
...
[truncated]
...
Enter passphrase for new targets key with ID f511f86 (repo):
Passphrase is too short. Please use a password manager to generate and store a good random passphrase.
Enter passphrase for new targets key with ID f511f86 (repo):
Repeat passphrase for new targets key with ID f511f86 (repo):

* fatal: failed to add key to filestore: maximum number of passphrase attempts exceeded
```

### LINKED ISSUE #899: When running in CI system (no TTY) give up on passphrase
When running in a CI system where no TTY is attached, if something goes wrong, notary winds up asking for a root pass phrase over and over without end.  We recently had our CI system run out of disk space due to gig's of logs with the repeating pattern:

```
Enter passphrase for new root key with ID fdd4003: 
Enter passphrase for new root key with ID fdd4003: 
Enter passphrase for new root key with ID fdd4003: 
```

Notary should be smarter and give up asking for a passphrase when there's no tty instead of looping.

riyazdf:
Fixes some logic in the keystore when counting number of passphrase attempts, especially when short passphrases are provided.

```
🐳 $ bin/notary init testGun
Root key found, using: 4327a8deba2cae19a4b393d2c6cf3a2f349652c84bc41bc1e09ade182d86b19a
Enter passphrase for root key with ID 4327a8d:
Enter passphrase for new targets key with ID 454c204 (testGun):
Passphrase is too short. Please use a password manager to generate and store a good random passphrase.
Enter passphrase for new targets key with ID 454c204 (testGun):
Passphrase is too short. Please use a password manager to generate and store a good random passphrase.
...
Enter passphrase for new targets key with ID 454c204 (testGun):
Passphrase is too short. Please use a password manager to generate and store a good random passphrase.

* fatal: failed to add key to filestore: maximum number of passphrase attempts exceeded
```

Also, ensures that `PromptRetriever()` has an actual terminal when requesting passphrases and will error out with `ErrNoInput` if that is not the case instead of trying to read input over and over.

Closes #899 
Closes #812

### DIFF
--- a/passphrase/passphrase.go
+++ b/passphrase/passphrase.go
@@ -47,7 +47,7 @@ var (
 
 // PromptRetriever returns a new Retriever which will provide a prompt on stdin
 // and stdout to retrieve a passphrase. stdin will be checked if it is a terminal,
-// else the PromptRetriever will error when attempting to retreive passphrases.
+// else the PromptRetriever will error when attempting to retrieve a passphrase.
 // Upon successful passphrase retrievals, the passphrase will be cached such that
 // subsequent prompts will produce the same passphrase.
 func PromptRetriever() notary.PassRetriever {
--- a/passphrase/passphrase.go
+++ b/passphrase/passphrase.go
@@ -40,12 +40,22 @@ var (
 	// ErrTooManyAttempts is returned if the maximum number of passphrase
 	// entry attempts is reached.
 	ErrTooManyAttempts = errors.New("Too many attempts")
+
+	// ErrNoInput is returned if we do not have a valid input method for passphrases
+	ErrNoInput = errors.New("Please either use environment variables or STDIN with a terminal to provide key passphrases")
 )
 
 // PromptRetriever returns a new Retriever which will provide a prompt on stdin
-// and stdout to retrieve a passphrase. The passphrase will be cached such that
+// and stdout to retrieve a passphrase. stdin will be checked if it is a terminal,
+// else the PromptRetriever will error when attempting to retreive passphrases.
+// Upon successful passphrase retrievals, the passphrase will be cached such that
 // subsequent prompts will produce the same passphrase.
 func PromptRetriever() notary.PassRetriever {
+	if !term.IsTerminal(os.Stdin.Fd()) {
+		return func(string, string, bool, int) (string, bool, error) {
+			return "", false, ErrNoInput
+		}
+	}
 	return PromptRetrieverWithInOut(os.Stdin, os.Stdout, nil)
 }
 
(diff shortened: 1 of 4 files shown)


trustmanager/keystore.go
```diff
@@ -142,13 +142,12 @@ func (s *GenericKeyStore) AddKey(keyInfo KeyInfo, privKey data.PrivateKey) error
 	name := filepath.Join(keyInfo.Gun, privKey.ID())
 	for attempts := 0; ; attempts++ {
 		chosenPassphrase, giveup, err = s.PassRetriever(name, keyInfo.Role, true, attempts)
-		if err != nil {
-			continue
+		if err == nil {
+			break
 		}
 		if giveup || attempts > 10 {
```
cyli: Non-blocking:  wondering if we should decrease this number to 3 or some such?  That would give the user 5 tries.  (currently it fails out after 12 tries - I was wondering whether that's too many)

cyli:
Thank you for fixing this!  LGTM - took it for a spin and it correctly terminates, and in a script without stdin it immediately fails.

passphrase/passphrase.go
```diff
@@ -85,13 +95,13 @@ func (br *boundRetriever) requestPassphrase(keyName, alias string, createNew boo
...
-	if term.IsTerminal(0) {
-		state, err := term.SaveState(0)
+	if term.IsTerminal(os.Stdin.Fd()) {
+		state, err := term.SaveState(os.Stdin.Fd())
 		if err != nil {
 			return "", false, err
 		}
-		term.DisableEcho(0, state)
-		defer term.RestoreTerminal(0, state)
+		term.DisableEcho(os.Stdin.Fd(), state)
```
endophage: Yeah, this is a good reason for us to switch to crypto/terminal in the future. Better API for retrieving passwords from the terminal.

trustmanager/keystore.go
endophage: I agree. 10 is feeling like too many. Also would be good to move the exact number to a const.

endophage:
LGTM. We can modify the number of retries separately

PR #913, opened 2016-08-12
Use crypto/terminal instead of docker/term

### LINKED ISSUE #907: Look at replacing the docker term package with crypto/terminal
Go includes the crypto/terminal package in the core libraries and docker's term package pulls in other dependencies. I think crypto/terminal supports all the functionality we need (read a password without local echo and check if descriptor is a terminal) and would unwind one of our more painful dependencies.

HuKeping:
Fixes #907 

The crypto/terminal package already supports all the functionality
we need:
- read a password without local echo
- check if the descriptor is a terminal


riyazdf:
jenkins, test this please.

passphrase/passphrase.go
```diff
@@ -134,8 +123,19 @@ func (br *boundRetriever) requestPassphrase(keyName, alias string, createNew boo
...
 
+	var (
+		passphrase []byte
+		err        error
+	)
+
 	stdin := bufio.NewReader(br.in)
-	passphrase, err := stdin.ReadBytes('\n')
+
+	if terminal.IsTerminal(int(os.Stdin.Fd())) {
```
riyazdf: thanks for adding this!  Since we repeat this same code a few times, could we factor it out into a helper function?

riyazdf:
I tested locally and this is great, thank you @HuKeping for fixing this!

Can we remove `github.com/docker/docker/pkg/term` (and maybe `github.com/docker/docker/pkg/system`?) from our `Godeps.json` and `vendor/` dependencies?

endophage:
There's also some `Azure/go-ansiterm` packages in Godeps that may drop off too with this change 👍

HuKeping:
Does it make sense to cleanup the Godep in a separated PR for it would be always tons of code lines change, but I'm fine if you think it's better to make it in this PR.

riyazdf:
@HuKeping if you don't mind, I think it'd be better to also update the dependencies in this same PR.  That way we have a complete switchover without any unused code lying around

HuKeping:
OK @riyazdf , just find it a little bit harder to review

riyazdf:
jenkins, test this please.

riyazdf:
LGTM pending CI.  Thank you @HuKeping for also removing those dependencies!  The amount of deleted code is awesome.

cyli:
LGTM!  Thank you for tackling this @HuKeping!

riyazdf:
LGTM!

PR #915, opened 2016-08-13
made fix as per open issue 310

### LINKED ISSUE #310: Suspect logic in trustmanager.GetPasswdDecryptBytes
Just a low-priority drive-by note while working on something else:

* After the cycle, `retErr` will always be nil, so it is currently useless.
* The `attempts` check should happen either before prompting for a passphrase, or after trying to use it (in fact, probably in the loop condition). The current way, prompting for a passphrase, ignoring it, and failing, does not make sense.

avaid96:
closes #310 

### DIFF
--- a/trustmanager/keystore.go
+++ b/trustmanager/keystore.go
@@ -319,7 +319,6 @@ func getSubdir(alias string) string {
 func GetPasswdDecryptBytes(passphraseRetriever notary.PassRetriever, pemBytes []byte, name, alias string) (data.PrivateKey, string, error) {
 	var (
 		passwd  string
-		retErr  error
 		privKey data.PrivateKey
 	)
 	for attempts := 0; ; attempts++ {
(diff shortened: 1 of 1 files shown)


trustmanager/keystore.go
```diff
@@ -328,26 +327,22 @@ func GetPasswdDecryptBytes(passphraseRetriever notary.PassRetriever, pemBytes []
 			err    error
 		)
 		passwd, giveup, err = passphraseRetriever(name, alias, false, attempts)
+		if attempts > 10 {
```
riyazdf: Should this check go before we call `passphraseRetriever`, since we've determined that we have tried enough times and don't actually use `passwd`?

trustmanager/keystore.go
```diff
@@ -328,26 +327,22 @@ func GetPasswdDecryptBytes(passphraseRetriever notary.PassRetriever, pemBytes []
...
 		}
-		if attempts > 10 {
-			return nil, "", ErrAttemptsExceeded{}
-		}
 
 		// Try to convert PEM encoded bytes back to a PrivateKey using the passphrase
 		privKey, err = utils.ParsePEMPrivateKey(pemBytes, passwd)
 		if err != nil {
-			retErr = ErrPasswordInvalid{}
+			logrus.Info(ErrPasswordInvalid{})
```
riyazdf: I'm fine with removing this log and just having these lines simplified to:

```
if err == nil {
    // We managed to parse the PrivateKey. We've succeeded!
    break
}
```

trustmanager/keystore.go
```diff
@@ -319,35 +319,28 @@ func getSubdir(alias string) string {
...
-		if err != nil {
-			retErr = ErrPasswordInvalid{}
-		} else {
+		if err == nil {
 			// We managed to parse the PrivateKey. We've succeeded!
-			retErr = nil
 			break
 		}
 	}
-	if retErr != nil {
```
cyli: We might want to check if privKey is null here, and if so, return some kind of error.  Alternately, in line 339, do `privKey, retErr = ...` and leave this check here.

It's possible for us to never have decrypted the private key because `utils.ParsePEMPrivateKey` could have failed over and over.

avaid96: since we only exit the loop to reach this part if the error is nil and if it isn't then we will end up returning with ErrAttemptsExceeded or ErrPasswordInvalid, I thought this check was unnecessary and not hitting this part. So the retErr should always be nil over here. Am I missing a case?

cyli: Ah my bad yes, sorry, we are not checking the attempts as part of the for loop condition, just inside the for loop itself, so I got confused.  Thanks!

avaid96: perfect, thanks for the super quick review 😃

cyli:
Thanks for fixing this @avaid96!  LGTM pending CI!

riyazdf:
jenkins, test this please.

riyazdf:
LGTM pending CI!

riyazdf:
jenkins, ok to test.

PR #800, opened 2016-06-23
Validate root rotations against trust pinning where appropriate

riyazdf:
Check new roots against the trust pinning when performing validation, but do not reject the new root if TOFUs is disabled because we've already trusted a previous root.
 
Closes #735.

Signed-off-by: Riyaz Faizullabhoy <riyaz.faizullabhoy@docker.com>

PR #921, opened 2016-08-16
marking isValid for signed objects in builder and making sure that they're being marked with a test

avaid96:


riyazdf:
jenkins, ok to test

trustpinning/certs.go
```diff
@@ -161,6 +161,7 @@ func ValidateRoot(prevRoot *data.SignedRoot, root *data.Signed, gun string, trus
 		logrus.Debugf("failed to verify TUF data for: %s, %v", gun, err)
 		return nil, &ErrValidationFail{Reason: "failed to validate integrity of roots"}
 	}
+	signedRoot.Signatures = root.Signatures
```
riyazdf: this case is a little more tricky, since we've also validated against the old root and probably don't want to include the `isValid` markings from that check.

I had taken a stab at this in this commit (in the trust-pinning root rotation PR): https://github.com/docker/notary/pull/800/commits/a05717f80b6ceb243a19afa9b74e7ca35acab8c6

endophage:
@avaid96 can we merge this to release/0.4.0 rather than master?

trustpinning/certs.go
endophage: Good thought. Given the current use I'd be OK with `IsValid` only working correctly for delegations if it's not straightforward to fix for other roles too. We can file an issue to track implementing it for other roles.

trustpinning/certs_test.go
```diff
@@ -340,6 +341,7 @@ func TestValidateRootWithPinnerCertAndIntermediates(t *testing.T) {
 	err = signed.Sign(cs, signedRoot, []data.PublicKey{ecdsax509Key}, 1, nil)
 	require.NoError(t, err)
 
+	signedRoot.Signatures[0].IsValid = true
```
endophage: Is it necessary to set this (in any of the cases in this file) as we don't seem to check it anywhere?

tuf/builder_test.go
```diff
@@ -98,6 +98,52 @@ func TestBuilderOnlyAcceptsDelegationsAfterParent(t *testing.T) {
 	require.NoError(t, builder.Load("targets/a/b", meta["targets/a/b"], 1, false))
 }
 
+func TestMarkingIsValid(t *testing.T) {
```
endophage: 👍 great test

avaid96:
as per discussion IRL with @riyazdf. All changes regarding files certs.go and certs_test.go will come in with his earlier PR #800 as it covers all changes that need to be made to those files

trustpinning/certs.go
riyazdf: I think root is the only role that has this weird case, but @avaid96 can confirm

trustpinning/certs_test.go
riyazdf: I think this is due to the equality checks `require.Equal(t, typedSignedRoot, validatedRoot)`

trustpinning/certs.go
avaid96: I think it should be fine for all roles except root. It was already fine for snapshot and timestamp and per this PR is fixed for targets / delegations. #800 should fix it for root as well

avaid96:
seem to have picked up some other commits when I switched the base to release/0.4- my bad, I'll try to fix them 😄

avaid96:
done, should be ready for review pending CI

endophage:
LGTM!

riyazdf:
LGTM

PR #938, opened 2016-09-01
Fix ineffective assignments, local constification cleanup

riyazdf:
updates to `ineffassign` found some assignments that needed to be cleaned up, as well as a couple of errors that needed to be returned.  I also fixed up some of the hash constification (`notary.SHA256`/`notary.SHA512`) while in one of the test files


utils/http_test.go
```diff
@@ -51,6 +51,7 @@ func TestRootHandlerError(t *testing.T) {
 	defer ts.Close()
 
 	res, err := http.Get(ts.URL)
+	require.NoError(t, err)
 	if res.StatusCode != http.StatusInternalServerError {
 		t.Fatalf("Expected 500, received %d", res.StatusCode)
```
HuKeping: would you mind also replace the `500` to `http.StatusInternalServerError`

riyazdf: I can if you'd like, though it isn't as useful to use the constant here because this is for a test error message instead of an actual data key or comparison value.

I'm leaning to leave it as is because I think it's more readable, otherwise it'll look like:
`t.Fatalf("Expected %d, received %d", http.StatusInternalServerError, res.StatusCode)`

cyli:
Thank you for the massive cleanup as well as for adding the extra tests!  LGTM!

utils/http_test.go
endophage: should be `require.NotEqual(t, http.StatusInternalServerError, res.StatusCode)` rather than the whole `if ...`

endophage: probably my old code for what it's worth :-P

riyazdf: @endophage: good point, just amended the latest commit to clean that up and the test immediately above which had similar checks 👍

endophage:
LGTM!

Issue #841, opened 2016-07-15
Implement Threshold Signing mechanisms on the client/server


cyli:
Although we validate thresholds right now, currently there is no way to set a threshold higher than 1, because we had no good user story for how this would work.

This has been discussed over email with @gtank and @ecordell, and @ecordell has kindly opened https://github.com/theupdateframework/tuf/issues/339 as a result.

Posting the relevant parts of this discussion here for everyone to review.

(from @endophage:)
There are a few UX questions we need to answer with any implementation of thresholds, and are largely what have encouraged us to implement an alternate mechanism:

  - how do I as a signer (or "witness"), know which versions are available and which versions I can/should/need to sign? 
  - how do I distinguish pending versions to sign, as I may not want to witness all pending versions?
  - how do we prevent things from being deleted by accident: i.e. assuming threshold > 1, targets v1 is empty and signed by necessary threshold. User A generates v2, adds target "a", and pushes that partially signed targets file to server. User B wants to add target "b". Does he pull the untrusted (lacking threshold signatures) v2, does he pull v1 and create a conflicting v2, or does he pull v1 in the knowledge there is a v2 and he should create v3 but without the as yet untrusted "a" target? Alternatively do we block User B from pushing until v2 is finalized? 
  - how do we handle a threshold change that makes a partially signed thing become valid (i.e. it had threshold 2 but 1 sig, and the threshold is lowered to 1)?

cyli:
(from @ecordell:)

The approach we've been pondering is allowing partially signed, and thus "invalid" metadata to be pushed to the TUF server. I should add that we have some patches to Notary playing with these ideas (haven't fleshed out that test suite yet - sorry!) With that in mind I think we have decent answers to the UX questions you posed.

> how do I as a signer (or "witness"), know which versions are available and which versions I can/should/need to sign? 

If you request the metadata, it's either valid (meets threshold/keys) or invalid (doesn't meet threshold). If it's not valid and you're a "normal" user or a deployment system, you stop - the TUF repo is currently in an update process. There could be a mechanism to pull the latest valid metadata so that users don't have to be blocked (though some may wish to be in this case) or we could have explicitly staged metadata.

If the metadata is invalid and you're a "witness," you would pull the image and, if it meets whatever requirements you're verifying, you would `notary witness` with your key, and push that metadata back up to the server.

[...]

Issue #445, opened 2016-01-12
History API


vikstrous:
Provide a way to query old versions of a repo's tags/state

endophage:
Cleaning out old issues. I don't think we're going to support this beyond the changefeed which can be queried for activity, then the versions from the changefeed can be used to query for specific files.

Any ability to visualize changes to specific tags will need to be done by a different service based on processing data held in notary repos.

Issue #410, opened 2016-01-05
Allow downloading older versions of Notary-hosted files


mtrmac:
For investigation into past behavior (both debugging and forensic investigation where an image came from) it would be useful to allow downloading older versions of the TUF metadata. The server is already storing them, so this is a matter of:
* Exposing the already stored previous versions through the server’s REST API:
 * At least a download-by-version
 * Probably also download-by-SHA256-hash to allow going from `snapshot.json` directly to the specific contents of that snapshot, though this can be emulated client-side (Actually the client should always be using this to avoid races with concurrent repo updates.)
 * Perhaps also listing available versions and relevant time frames, taking advantage of the  `created_at` timestamps.
* Implementing a client for the API in `notary`, at least for downloading the raw data.
* Perhaps even providing `notary verify --as-of-version` or `notary-verify --as-of-time`?

Issue #408, opened 2016-01-05
Make repository updates atomic and generally safe to use in scripts


mtrmac:
It should always be *safe* to run `notary` to modify the local state or the on-server repo from a script, and if possible it should be *reliable*. This is necessary both for general build automation and for automation of key management (#406 and related). E.g.

* `NotaryRepository.Publish` is not handling the on-disk changelist atomically: it loads pending changes from disk, applies them, and then deletes all on-disk changes, even newly added ones which it has not applied. (To be fair, this matters only if concurrent pushes happen for the same repository from the same user account on the same machine, which is probably a corner case. Still, an unprivileged user setting a cron job under their own user account is not entirely implausible.)

* Handling of `notary publish` on the server (`POST /v2/$imageName/_trust/tuf/`) is atomic, and the version number checking does ensure that no updates should be lost, so it is safe—but it is not reliable, the operation just fails on a version conflict. Because we are maintaining the changelist of to-be-applied operations, we should be able to do better in common cases: simply retry the update on the client (how many times?), or perhaps obtain some kind of on-server writer lock before pulling latest data and sending an update.

PR #942, opened 2016-09-07
Support deprecating old root keys after rotation

### LINKED ISSUE #835: Old public keys should be stored in versioned metadata, not roles
Currently, rotating root keys versions the role and keeps the old keys around in the metadata indefinitely. Here's a quick example after adding two root keys:

root.json

``` json
{
    "signed": {
        "_type": "Root",
        "consistent_snapshot": false,
        "expires": "2026-07-10T15:54:46.480499158-04:00",
        "keys": {
            "072c38acffb2309e45bb942cfbd44d15a644b303713dda5dd0f2af25c71a6a7a": {
                "keytype": "ed25519",
                "keyval": {
                    "private": null,
                    "public": "uuyVXZHUGX3Sau6Yk3rC6mK2fq2fXODgKwXj8jkOo6I="
                }
            },
            "117e3e9ccce8a85a236c2156b1729e57ebd427b544a65c9dfa2d2145c64534e9": {
                "keytype": "ecdsa",
                "keyval": {
                    "private": null,
                    "public": "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEcl5LAZXA5VLrKq5t+EzcWyskVOyTIANtqh4r0znfE+Qv867Lbx29vlU6A1BsoWtQGORIotzJ78aSbu5DWPCVgQ=="
                }
            },
            "5ccfcc693384568483aa960453b9e13fbb5d10d5251f1ab5ed5fc76c58d7372e": {
                "keytype": "ecdsa-x509",
                "keyval": {
                    "private": null,
                    "public": "LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUJaekNDQVF5Z0F3SUJBZ0lSQUttbTF5R0VJZlIxa3ZZcldjWmJNYXd3Q2dZSUtvWkl6ajBFQXdJd0dURVgKTUJVR0ExVUVBeE1PY1hWaGVTOTBaWE4wTVRVNU16a3dIaGNOTVRZd056RXlNVGsxTkRRMFdoY05Nall3TnpFdwpNVGsxTkRRMFdqQVpNUmN3RlFZRFZRUURFdzV4ZFdGNUwzUmxjM1F4TlRrek9UQlpNQk1HQnlxR1NNNDlBZ0VHCkNDcUdTTTQ5QXdFSEEwSUFCSjRodWx6VmxETnZSS1ZYN1RZVytCeUNNdEFWNnlDZkI0Ymp1MXVZWlkwcUhRcU4KYkxORmRhblRLN2VlMmx2bVowalRuS3ZoMGVTZE5reFh2VThjbWhLak5UQXpNQTRHQTFVZER3RUIvd1FFQXdJRgpvREFUQmdOVkhTVUVEREFLQmdnckJnRUZCUWNEQXpBTUJnTlZIUk1CQWY4RUFqQUFNQW9HQ0NxR1NNNDlCQU1DCkEwa0FNRVlDSVFEbzlnTkhkSU5JZzFhN1BMcGQ1ZE1pcU1WMlRubzNVcEg1NTdWaVVvYTF5Z0loQU1JemcrM1YKcG5HcGRYTURFWkx5cnhGWHYzOURaUVdsTGlZbjd4K0x2d2FtCi0tLS0tRU5EIENFUlRJRklDQVRFLS0tLS0K"
                }
            },
            "86958cab1fb3cbaeae393fe7b7af3109adf627effc8e0ba39e23b185d266c61e": {
                "keytype": "ecdsa",
                "keyval": {
                    "private": null,
                    "public": "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEIowYCNsTPsmTkA4WzO5VdwOtVs3AZJnkW2pdPIKa30uB/oURJGx6gMGfuv3yZEu87hlVUOXT2DpIIRJIcyru5A=="
                }
            },
            "997c16ec6386ae57914a69983c712a05fa6382e4826251e9283c2cda21186045": {
                "keytype": "ecdsa",
                "keyval": {
                    "private": null,
                    "public": "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEp5vDm97fT4AnrqS7qmcTrE7Mu4aeFIVAGXU9k3kigWALgRQ9c+g+FLeX4hxd3vJwOd7JjaSsalAyZC/wDduDGw=="
                }
            },
            "ca35263785be4f26f01d216c6f6238810c9ea997f35dcd2bbfc812a94e492278": {
                "keytype": "ecdsa",
                "keyval": {
                    "private": null,
                    "public": "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEZMTJCaBeOdWis/ZtnX30W+TTuYBRz7dGIhTH5zGAlVcK1gY6qOKDJWtI6Nvuo73xsvKIpPXP2CkxhNCq16L66Q=="
                }
            }
        },
        "roles": {
            "root": {
                "keyids": [
                    "5ccfcc693384568483aa960453b9e13fbb5d10d5251f1ab5ed5fc76c58d7372e",
                    "117e3e9ccce8a85a236c2156b1729e57ebd427b544a65c9dfa2d2145c64534e9",
                    "997c16ec6386ae57914a69983c712a05fa6382e4826251e9283c2cda21186045"
                ],
                "threshold": 1
            },
            "root.1": {
                "keyids": [
                    "5ccfcc693384568483aa960453b9e13fbb5d10d5251f1ab5ed5fc76c58d7372e"
                ],
                "threshold": 1
            },
            "root.2": {
                "keyids": [
                    "5ccfcc693384568483aa960453b9e13fbb5d10d5251f1ab5ed5fc76c58d7372e",
                    "117e3e9ccce8a85a236c2156b1729e57ebd427b544a65c9dfa2d2145c64534e9"
                ],
                "threshold": 1
            },
            "snapshot": {
                "keyids": [
                    "ca35263785be4f26f01d216c6f6238810c9ea997f35dcd2bbfc812a94e492278"
                ],
                "threshold": 1
            },
            "targets": {
                "keyids": [
                    "86958cab1fb3cbaeae393fe7b7af3109adf627effc8e0ba39e23b185d266c61e"
                ],
                "threshold": 1
            },
            "timestamp": {
                "keyids": [
                    "072c38acffb2309e45bb942cfbd44d15a644b303713dda5dd0f2af25c71a6a7a"
                ],
                "threshold": 1
            }
        },
        "version": 3
    },
    "signatures": [{
        "keyid": "5ccfcc693384568483aa960453b9e13fbb5d10d5251f1ab5ed5fc76c58d7372e",
        "method": "ecdsa",
        "sig": "GEe8atrsRO5fy5dhcYQlsV0K2sr+Uwum1utBBViLLqTEdsYXGLPN6dWgpMQU8y6U4J/WvqV43xnMRLxaBmXwvw=="
    }]
}
```

I think the older keys should be available in `root.2.json` and `root.1.json`, not as additional roles on a single `root.json` file. This way clients that have older metadata can request intermediate metadata to become current. 

I saw the discussion on #267 and #648, but I don't think this implementation follows TUF. I wanted to bring this up for discussion before working on it, since it would be a breaking change.

### LINKED ISSUE #1014: Fix  formatted I/O
https://github.com/docker/notary/blob/master/trustpinning/certs.go#L143

```
logrus.Debugf("checking root against trust_pinning config", gun)
```

We can either use `logrus.Debug` or add the `%s` to it.

ecordell:
This adds support for root key rotation as described in #835. 

Previously, old root keys needed to sign for all subsequent `root.json` files in perpetuity. This changes root rotation so that new keys are trusted if signed by the previous root key/threshold. To support this behavior, root files can now be requested by version number from the server.

This also adds an additional command line flag for `rotate`, `--key`, which allows specifying keys to rotate the root to. This supports specifying multiple keys, but all must be available to sign locally (i.e. this doesn't support threshold/witness behavior at the moment).

Example:

``` bash
# test/collection root.json signed by A, B
$ notary key rotate test/collection root --key=C --key=D
# test/collection root.json now signed by A, B, C, D
$ notary key rotate test/collection root --key=E
# test/collection root.json now signed by C, D, E
```

Closes #1014 
Closes #835

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -818,7 +818,7 @@ func (r *NotaryRepository) Update(forWrite bool) error {
 		// notFound.Resource may include a checksum so when the role is root,
 		// it will be root or root.<checksum>. Therefore best we can
 		// do it match a "root." prefix
-		if notFound, ok := err.(store.ErrMetaNotFound); ok && strings.HasPrefix(notFound.Resource, data.CanonicalRootRole+".") {
+		if notFound, ok := err.(store.ErrMetaNotFound); ok && strings.Contains(notFound.Resource, data.CanonicalRootRole) {
 			return r.errRepositoryNotExist()
 		}
 		return err
@@ -837,8 +837,11 @@ func (r *NotaryRepository) Update(forWrite bool) error {
 // is initialized or not. If set to true, we will always attempt to download
 // and return an error if the remote repository errors.
 //
-// Populates a tuf.RepoBuilder with this root metadata (only use
-// TUFClient.Update to load the rest).
+// Populates a tuf.RepoBuilder with this root metadata. If the root metadata
+// downloaded is a newer version than what is on disk, then intermediate
+// versions will be downloaded and verified in order to rotate trusted keys
+// properly. Newer root metadata must always be signed with the previous
+// threshold and keys.
 //
 // Fails if the remote server is reachable and does not know the repo
 // (i.e. before the first r.Publish()), in which case the error is
@@ -918,8 +921,10 @@ func (r *NotaryRepository) bootstrapClient(checkInitialized bool) (*TUFClient, e
 	return NewTUFClient(oldBuilder, newBuilder, remote, r.cache), nil
 }
 
-// RotateKey removes all existing keys associated with the role, and either
-// creates and adds one new key or delegates managing the key to the server.
+// RotateKey removes all existing keys associated with the role. If no keys are
+// specified in keyList, then this creates and adds one new key or delegates
+// managing the key to the server. If key(s) are specified by keyList, then they are
+// used for signing the role. Currently keys may only be specified for root keys (encrypted)
 // These changes are staged in a changelist until publish is called.
 func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool, keyList []string) error {
 	if err := checkRotationInput(role, serverManagesKey); err != nil {
@@ -942,6 +947,11 @@ func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool, keyList
 		pubKeyList = make(data.KeyList, 0, len(keyList))
 		for _, keyID := range keyList {
 			pubKey = r.CryptoService.GetKey(keyID)
+			if pubKey == nil {
+				errFmtMsg = "unable to find key: %s"
+				err = fmt.Errorf("no key with id %s could be found", keyID)
+				break
+			}
 			pubKeyList = append(pubKeyList, pubKey)
 		}
 	case !serverManagesKey && len(keyList) == 0:
--- a/client/client.go
+++ b/client/client.go
@@ -9,7 +9,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
-	"strings"
+	"regexp"
 	"time"
 
 	"github.com/Sirupsen/logrus"
@@ -146,7 +146,7 @@ func rootCertKey(gun string, privKey data.PrivateKey) (data.PublicKey, error) {
 	x509PublicKey := utils.CertToKey(cert)
 	if x509PublicKey == nil {
 		return nil, fmt.Errorf(
-			"cannot use regenerated certificate: format %v", cert.PublicKeyAlgorithm)
+			"cannot use regenerated certificate: format %s", cert.PublicKeyAlgorithm)
 	}
 
 	return x509PublicKey, nil
@@ -815,10 +815,11 @@ func (r *NotaryRepository) Update(forWrite bool) error {
 	}
 	repo, invalid, err := c.Update()
 	if err != nil {
-		// notFound.Resource may include a checksum so when the role is root,
-		// it will be root or root.<checksum>. Therefore best we can
-		// do it match a "root." prefix
-		if notFound, ok := err.(store.ErrMetaNotFound); ok && strings.Contains(notFound.Resource, data.CanonicalRootRole) {
+		// notFound.Resource may include a version or checksum so when the role is root,
+		// it will be root, <version>.root or root.<checksum>.
+		notFound, ok := err.(store.ErrMetaNotFound)
+		isRoot, _ := regexp.MatchString(`\.?`+data.CanonicalRootRole+`\.?`, notFound.Resource)
+		if ok && isRoot {
 			return r.errRepositoryNotExist()
 		}
 		return err
@@ -924,7 +925,7 @@ func (r *NotaryRepository) bootstrapClient(checkInitialized bool) (*TUFClient, e
 // RotateKey removes all existing keys associated with the role. If no keys are
 // specified in keyList, then this creates and adds one new key or delegates
 // managing the key to the server. If key(s) are specified by keyList, then they are
-// used for signing the role. Currently keys may only be specified for root keys (encrypted)
+// used for signing the role.
 // These changes are staged in a changelist until publish is called.
 func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool, keyList []string) error {
 	if err := checkRotationInput(role, serverManagesKey); err != nil {
@@ -942,10 +943,26 @@ func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool, keyList
 		pubKey, err = rotateRemoteKey(r.baseURL, r.gun, role, r.roundTrip)
 		pubKeyList = make(data.KeyList, 0, 1)
 		pubKeyList = append(pubKeyList, pubKey)
-		errFmtMsg = "unable to rotate remote key: %s"
-	case !serverManagesKey && len(keyList) > 0:
-		pubKeyList = make(data.KeyList, 0, len(keyList))
-		for _, keyID := range keyList {
+		if err != nil {
+			return nil, fmt.Errorf("unable to rotate remote key: %s", err)
+		}
+		return pubKeyList, nil
+	}
+
+	// If no new keys are passed in, we generate one
+	if len(newKeys) == 0 {
+		pubKeyList = make(data.KeyList, 0, 1)
+		pubKey, err = r.CryptoService.Create(role, r.gun, data.ECDSAKey)
+		pubKeyList = append(pubKeyList, pubKey)
+	}
+	if err != nil {
+		return nil, fmt.Errorf("unable to generate key: %s", err)
+	}
+
+	// If a list of keys to rotate to are provided, we add those
+	if len(newKeys) > 0 {
+		pubKeyList = make(data.KeyList, 0, len(newKeys))
+		for _, keyID := range newKeys {
 			pubKey = r.CryptoService.GetKey(keyID)
 			if pubKey == nil {
 				errFmtMsg = "unable to find key: %s"
@@ -968,6 +985,13 @@ func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool, keyList
 	// if this is a root role, generate a root cert for the public key
 	if role == data.CanonicalRootRole {
 		for i, pubKey := range pubKeyList {
+			keyInfo, err := r.CryptoService.GetKeyInfo(pubKey.ID())
+			if err != nil {
+				return err
+			}
+			if keyInfo.Role != role {
+				return fmt.Errorf("attempted rotating root key but given %s key instead", keyInfo.Role)
+			}
 			privKey, _, err := r.CryptoService.GetPrivateKey(pubKey.ID())
 			if err != nil {
 				return err
--- a/client/client.go
+++ b/client/client.go
@@ -146,7 +146,7 @@ func rootCertKey(gun string, privKey data.PrivateKey) (data.PublicKey, error) {
 	x509PublicKey := utils.CertToKey(cert)
 	if x509PublicKey == nil {
 		return nil, fmt.Errorf(
-			"cannot use regenerated certificate: format %s", cert.PublicKeyAlgorithm)
+			"cannot use regenerated certificate: format %v", cert.PublicKeyAlgorithm)
 	}
 
 	return x509PublicKey, nil
@@ -921,21 +921,33 @@ func (r *NotaryRepository) bootstrapClient(checkInitialized bool) (*TUFClient, e
 // RotateKey removes all existing keys associated with the role, and either
 // creates and adds one new key or delegates managing the key to the server.
 // These changes are staged in a changelist until publish is called.
-func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool) error {
+func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool, keyList []string) error {
 	if err := checkRotationInput(role, serverManagesKey); err != nil {
 		return err
 	}
 	var (
-		pubKey    data.PublicKey
-		err       error
-		errFmtMsg string
+		pubKey     data.PublicKey
+		pubKeyList data.KeyList
+		err        error
+		errFmtMsg  string
 	)
-	switch serverManagesKey {
-	case true:
+
+	switch {
+	case serverManagesKey:
 		pubKey, err = rotateRemoteKey(r.baseURL, r.gun, role, r.roundTrip)
+		pubKeyList = make(data.KeyList, 0, 1)
+		pubKeyList = append(pubKeyList, pubKey)
 		errFmtMsg = "unable to rotate remote key: %s"
-	default:
+	case !serverManagesKey && len(keyList) > 0:
+		pubKeyList = make(data.KeyList, 0, len(keyList))
+		for _, keyID := range keyList {
+			pubKey = r.CryptoService.GetKey(keyID)
+			pubKeyList = append(pubKeyList, pubKey)
+		}
+	case !serverManagesKey && len(keyList) == 0:
+		pubKeyList = make(data.KeyList, 0, 1)
 		pubKey, err = r.CryptoService.Create(role, r.gun, data.ECDSAKey)
+		pubKeyList = append(pubKeyList, pubKey)
 		errFmtMsg = "unable to generate key: %s"
 	}
 
@@ -945,18 +957,21 @@ func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool) error {
 
 	// if this is a root role, generate a root cert for the public key
 	if role == data.CanonicalRootRole {
-		privKey, _, err := r.CryptoService.GetPrivateKey(pubKey.ID())
-		if err != nil {
-			return err
-		}
-		pubKey, err = rootCertKey(r.gun, privKey)
-		if err != nil {
-			return err
+		for i, pubKey := range pubKeyList {
+			privKey, _, err := r.CryptoService.GetPrivateKey(pubKey.ID())
+			if err != nil {
+				return err
+			}
+			pubKey, err = rootCertKey(r.gun, privKey)
+			if err != nil {
+				return err
+			}
+			pubKeyList[i] = pubKey
 		}
 	}
 
 	cl := changelist.NewMemChangelist()
-	if err := r.rootFileKeyChange(cl, role, changelist.ActionCreate, pubKey); err != nil {
+	if err := r.rootFileKeyChange(cl, role, changelist.ActionCreate, pubKeyList); err != nil {
 		return err
 	}
 	return r.publish(cl)
@@ -980,12 +995,10 @@ func checkRotationInput(role string, serverManaged bool) error {
 	return nil
 }
 
-func (r *NotaryRepository) rootFileKeyChange(cl changelist.Changelist, role, action string, key data.PublicKey) error {
-	kl := make(data.KeyList, 0, 1)
-	kl = append(kl, key)
+func (r *NotaryRepository) rootFileKeyChange(cl changelist.Changelist, role, action string, keyList []data.PublicKey) error {
 	meta := changelist.TUFRootData{
 		RoleName: role,
-		Keys:     kl,
+		Keys:     keyList,
 	}
 	metaJSON, err := json.Marshal(meta)
 	if err != nil {
--- a/client/client.go
+++ b/client/client.go
@@ -10,6 +10,8 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/Sirupsen/logrus"
@@ -27,6 +29,9 @@ import (
 
 const (
 	tufDir = "tuf"
+
+	// SignWithAllOldVersions is a sentinel constant for LegacyVersions flag
+	SignWithAllOldVersions = -1
 )
 
 func init() {
@@ -36,16 +41,17 @@ func init() {
 // NotaryRepository stores all the information needed to operate on a notary
 // repository.
 type NotaryRepository struct {
-	baseDir       string
-	gun           string
-	baseURL       string
-	tufRepoPath   string
-	cache         store.MetadataStore
-	CryptoService signed.CryptoService
-	tufRepo       *tuf.Repo
-	invalid       *tuf.Repo // known data that was parsable but deemed invalid
-	roundTrip     http.RoundTripper
-	trustPinning  trustpinning.TrustPinConfig
+	baseDir        string
+	gun            string
+	baseURL        string
+	tufRepoPath    string
+	cache          store.MetadataStore
+	CryptoService  signed.CryptoService
+	tufRepo        *tuf.Repo
+	invalid        *tuf.Repo // known data that was parsable but deemed invalid
+	roundTrip      http.RoundTripper
+	trustPinning   trustpinning.TrustPinConfig
+	LegacyVersions int // number of versions back to fetch roots to sign with
 }
 
 // NewFileCachedNotaryRepository is a wrapper for NewNotaryRepository that initializes
@@ -90,13 +96,14 @@ func repositoryFromKeystores(baseDir, gun, baseURL string, rt http.RoundTripper,
 	cryptoService := cryptoservice.NewCryptoService(keyStores...)
 
 	nRepo := &NotaryRepository{
-		gun:           gun,
-		baseDir:       baseDir,
-		baseURL:       baseURL,
-		tufRepoPath:   filepath.Join(baseDir, tufDir, filepath.FromSlash(gun)),
-		CryptoService: cryptoService,
-		roundTrip:     rt,
-		trustPinning:  trustPin,
+		gun:            gun,
+		baseDir:        baseDir,
+		baseURL:        baseURL,
+		tufRepoPath:    filepath.Join(baseDir, tufDir, filepath.FromSlash(gun)),
+		CryptoService:  cryptoService,
+		roundTrip:      rt,
+		trustPinning:   trustPin,
+		LegacyVersions: 0, // By default, don't sign with legacy roles
 	}
 
 	nRepo.cache = cache
@@ -640,10 +647,16 @@ func (r *NotaryRepository) publish(cl changelist.Changelist) error {
 	// we send anything to remote
 	updatedFiles := make(map[string][]byte)
 
+	// Fetch old keys to support old clients
+	legacyKeys, err := r.oldKeysForLegacyClientSupport(r.LegacyVersions, initialPublish)
+	if err != nil {
+		return err
+	}
+
 	// check if our root file is nearing expiry or dirty. Resign if it is.  If
 	// root is not dirty but we are publishing for the first time, then just
 	// publish the existing root we have.
-	if err := signRootIfNecessary(updatedFiles, r.tufRepo, initialPublish); err != nil {
+	if err := signRootIfNecessary(updatedFiles, r.tufRepo, legacyKeys, initialPublish); err != nil {
 		return err
 	}
 
@@ -661,7 +674,7 @@ func (r *NotaryRepository) publish(cl changelist.Changelist) error {
 	}
 
 	if snapshotJSON, err := serializeCanonicalRole(
-		r.tufRepo, data.CanonicalSnapshotRole); err == nil {
+		r.tufRepo, data.CanonicalSnapshotRole, nil); err == nil {
 		// Only update the snapshot if we've successfully signed it.
 		updatedFiles[data.CanonicalSnapshotRole] = snapshotJSON
 	} else if signErr, ok := err.(signed.ErrInsufficientSignatures); ok && signErr.FoundKeys == 0 {
@@ -683,9 +696,12 @@ func (r *NotaryRepository) publish(cl changelist.Changelist) error {
 	return remote.SetMulti(updatedFiles)
 }
 
-func signRootIfNecessary(updates map[string][]byte, repo *tuf.Repo, initialPublish bool) error {
+func signRootIfNecessary(updates map[string][]byte, repo *tuf.Repo, extraSigningKeys data.KeyList, initialPublish bool) error {
+	if len(extraSigningKeys) > 0 {
+		repo.Root.Dirty = true
+	}
 	if nearExpiry(repo.Root.Signed.SignedCommon) || repo.Root.Dirty {
-		rootJSON, err := serializeCanonicalRole(repo, data.CanonicalRootRole)
+		rootJSON, err := serializeCanonicalRole(repo, data.CanonicalRootRole, extraSigningKeys)
 		if err != nil {
 			return err
 		}
@@ -700,11 +716,113 @@ func signRootIfNecessary(updates map[string][]byte, repo *tuf.Repo, initialPubli
 	return nil
 }
 
+// Fetch back a `legacyVersions` number of roots files, collect the root public keys
+// This includes old `root` roles as well as legacy versioned root roles, e.g. `1.root`
+func (r *NotaryRepository) oldKeysForLegacyClientSupport(legacyVersions int, initialPublish bool) (data.KeyList, error) {
+	if initialPublish {
+		return nil, nil
+	}
+
+	var oldestVersion int
+	prevVersion := r.tufRepo.Root.Signed.Version
+
+	if legacyVersions == SignWithAllOldVersions {
+		oldestVersion = 1
+	} else {
+		oldestVersion = r.tufRepo.Root.Signed.Version - legacyVersions
+	}
+
+	if oldestVersion < 1 {
+		oldestVersion = 1
+	}
+
+	if prevVersion <= 1 || oldestVersion == prevVersion {
+		return nil, nil
+	}
+	oldKeys := make(map[string]data.PublicKey)
+
+	c, err := r.bootstrapClient(true)
+	// require a server connection to fetch old roots
+	if err != nil {
+		return nil, err
+	}
+
+	for v := prevVersion; v >= oldestVersion; v-- {
+		logrus.Debugf("fetching old keys from version %d", v)
+		// fetch old root version
+		versionedRole := fmt.Sprintf("%d.%s", v, data.CanonicalRootRole)
+
+		raw, err := c.remote.GetSized(versionedRole, -1)
+		if err != nil {
+			logrus.Debugf("error downloading %s: %s", versionedRole, err)
+			continue
+		}
+
+		signedOldRoot := &data.Signed{}
+		if err := json.Unmarshal(raw, signedOldRoot); err != nil {
+			return nil, err
+		}
+		oldRootVersion, err := data.RootFromSigned(signedOldRoot)
+		if err != nil {
+			return nil, err
+		}
+
+		// extract legacy versioned root keys
+		oldRootVersionKeys := getOldRootPublicKeys(oldRootVersion)
+		for _, oldKey := range oldRootVersionKeys {
+			oldKeys[oldKey.ID()] = oldKey
+		}
+	}
+	oldKeyList := make(data.KeyList, 0, len(oldKeys))
+	for _, key := range oldKeys {
+		oldKeyList = append(oldKeyList, key)
+	}
+	return oldKeyList, nil
+}
+
+// get all the saved previous roles keys < the current root version
+func getOldRootPublicKeys(root *data.SignedRoot) data.KeyList {
+	oldRootKeys := make(data.KeyList, 0, len(root.Signed.Roles))
+	// now go through the old roles
+	for roleName := range root.Signed.Roles {
+		// must be either `root` or `x.root`
+		isVersionedRoot := false
+
+		nameTokens := strings.Split(roleName, ".")
+		if len(nameTokens) == 2 && nameTokens[0] == data.CanonicalRootRole {
+			isVersionedRoot = true
+		}
+
+		if isVersionedRoot {
+			version, err := strconv.Atoi(nameTokens[1])
+			if err != nil || version >= root.Signed.Version {
+				continue
+			}
+		} else {
+			if roleName != data.CanonicalRootRole {
+				continue
+			}
+		}
+
+		// ignore invalid roles, which shouldn't happen
+		oldRole, err := root.BuildBaseRole(roleName)
+		if err != nil {
+			continue
+		}
+
+		for _, key := range oldRole.ListKeys() {
+			oldRootKeys = append(oldRootKeys, key)
+			logrus.Debugf("found old key %s for supporting legacy clients", key.ID())
+		}
+	}
+	return oldRootKeys
+}
+
 func signTargets(updates map[string][]byte, repo *tuf.Repo, initialPublish bool) error {
 	// iterate through all the targets files - if they are dirty, sign and update
 	for roleName, roleObj := range repo.Targets {
 		if roleObj.Dirty || (roleName == data.CanonicalTargetsRole && initialPublish) {
-			targetsJSON, err := serializeCanonicalRole(repo, roleName)
+			targetsJSON, err := serializeCanonicalRole(repo, roleName, nil)
 			if err != nil {
 				return err
 			}
@@ -753,7 +871,7 @@ func (r *NotaryRepository) bootstrapRepo() error {
 func (r *NotaryRepository) saveMetadata(ignoreSnapshot bool) error {
 	logrus.Debugf("Saving changes to Trusted Collection.")
 
-	rootJSON, err := serializeCanonicalRole(r.tufRepo, data.CanonicalRootRole)
+	rootJSON, err := serializeCanonicalRole(r.tufRepo, data.CanonicalRootRole, nil)
 	if err != nil {
 		return err
 	}
@@ -784,7 +902,7 @@ func (r *NotaryRepository) saveMetadata(ignoreSnapshot bool) error {
 		return nil
 	}
 
-	snapshotJSON, err := serializeCanonicalRole(r.tufRepo, data.CanonicalSnapshotRole)
+	snapshotJSON, err := serializeCanonicalRole(r.tufRepo, data.CanonicalSnapshotRole, nil)
 	if err != nil {
 		return err
 	}
@@ -931,15 +1049,25 @@ func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool, keyList
 	if err := checkRotationInput(role, serverManagesKey); err != nil {
 		return err
 	}
-	var (
-		pubKey     data.PublicKey
-		pubKeyList data.KeyList
-		err        error
-		errFmtMsg  string
-	)
 
-	switch {
-	case serverManagesKey:
+	pubKeyList, err := r.pubKeyListForRotation(role, serverManagesKey, keyList)
+	if err != nil {
+		return err
+	}
+
+	cl := changelist.NewMemChangelist()
+	if err := r.rootFileKeyChange(cl, role, changelist.ActionCreate, pubKeyList); err != nil {
+		return err
+	}
+	return r.publish(cl)
+}
+
+// Given a set of new keys to rotate to and a set of keys to drop, returns the list of current keys to use
+func (r *NotaryRepository) pubKeyListForRotation(role string, serverManaged bool, newKeys []string) (pubKeyList data.KeyList, err error) {
+	var pubKey data.PublicKey
+
+	// If server manages the key being rotated, request a rotation and return the new key
+	if serverManaged {
 		pubKey, err = rotateRemoteKey(r.baseURL, r.gun, role, r.roundTrip)
 		pubKeyList = make(data.KeyList, 0, 1)
 		pubKeyList = append(pubKeyList, pubKey)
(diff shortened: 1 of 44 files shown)


cmd/notary/keys.go
```diff
@@ -13,11 +13,12 @@ import (
 	"github.com/docker/notary/trustmanager"
 	"github.com/docker/notary/utils"
 
+	"os"
```
HuKeping: would you mind if moving this package up to those canonical packages

client/client.go
```diff
@@ -923,16 +926,28 @@ func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool) error {
...
-		errFmtMsg string
+		pubKey     data.PublicKey
+		pubKeyList data.KeyList
+		err        error
+		errFmtMsg  string
 	)
-	switch serverManagesKey {
-	case true:
+
+	switch {
```
HuKeping: This `switch` seems a littler wired to me, can we use the original format, or just the `if-else`

client/client.go
```diff
@@ -923,16 +926,28 @@ func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool) error {
...
+	case serverManagesKey:
 		pubKey, err = rotateRemoteKey(r.baseURL, r.gun, role, r.roundTrip)
+		pubKeyList = make(data.KeyList, 0, 1)
+		pubKeyList = append(pubKeyList, pubKey)
 		errFmtMsg = "unable to rotate remote key: %s"
```
HuKeping: It would be good enough to use `return fmt.Errorf("unable to rotate remote key: %s", err)`  here and below, what do you think?

client/client.go
```diff
@@ -923,16 +926,28 @@ func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool) error {
...
-	default:
+	case !serverManagesKey && len(keyList) > 0:
+		pubKeyList = make(data.KeyList, 0, len(keyList))
+		for _, keyID := range keyList {
+			pubKey = r.CryptoService.GetKey(keyID)
+			pubKeyList = append(pubKeyList, pubKey)
+		}
```
HuKeping: This case lack of a verbose error message like the other two.

client/tufclient.go
```diff
@@ -78,25 +79,79 @@ func (c *TUFClient) update() error {
...
-	_, err := c.tryLoadCacheThenRemote(consistentInfo)
-	return err
+	}
+	return nil
+}
+
+// updateRootVersions updates the root from it's current version to a target, rotating keys
+// as they are found
+func (c *TUFClient) updateRootVersions(currentRoot []byte, fromVersion, toVersion int) error {
+	if fromVersion == toVersion {
```
HuKeping: Since this is the base case to break this recursive function , does it make sense to add a log here?

client/tufclient.go
```diff
@@ -198,6 +253,27 @@ func (c TUFClient) getTargetsFile(role data.DelegationRole, ci tuf.ConsistentInf
...
+
+	// We can't read an exact size for the root metadata without risking getting stuck in the TUF update cycle
+	// since it's possible that downloading timestamp/snapshot metadata may fail due to a signature mismatch
+	if !consistentInfo.ChecksumKnown() {
+		logrus.Debugf("Loading root with no expected checksum")
+
+		// get the cached root, if it exists, just for version checking
+		cachedRoot, _ := c.cache.GetSized(role, -1)
+		// prefer to download a new root
+		_, remoteErr := c.tryLoadRemote(consistentInfo, cachedRoot)
```
HuKeping: any reason why not just use `err` here?

cmd/notary/keys.go
ecordell: Will do - could've sworn gofmt does this for you.

client/client.go
ecordell: Thanks, I'll add one

client/tufclient.go
ecordell: This is not new code that I wrote, the git diff just shows it as new since it was shuffled a bit further down the file.

client/client.go
ecordell: Keeping things how they are helps keep lots of nested ifs out of the code, which I assume is why it was originally written this way.

client/client.go
ecordell: There are three distinct cases: the server manages the key, the server doesn't manage the key and keys to rotate to have been provided, or the server doesn't manage the key and no keys were provided.

I don't follow how if/else cleans this up. Is it just a style preference?

client/client.go
ecordell: Spoke too soon - there's no error case to handle here.

client/tufclient.go
ecordell: Will add. There are logs for every step though, so it wasn't especially unclear before, but I'll add an "update finished" log.

client/tufclient.go
HuKeping: ah I see ;)

client/client.go
HuKeping: ooops, sorry for that,  you are right

client/client.go
HuKeping: Hmm, you're right, it would definitely bring many more ifs to check the err with nil

client/client.go
HuKeping: Hi @ecordell it's just a style preference, personally :) 
Do you thinks this works for you,
```
        if serverManagesKey {
                pubKey, err = rotateRemoteKey(r.baseURL, r.gun, role, r.roundTrip)
                pubKeyList = make(data.KeyList, 0, 1)
                pubKeyList = append(pubKeyList, pubKey)
                errFmtMsg = "unable to rotate remote key: %s"
        } else {
                if len(keyList) == 0 { 
                        pubKeyList = make(data.KeyList, 0, 1)
                        pubKey, err = r.CryptoService.Create(role, r.gun, data.ECDSAKey)
                        pubKeyList = append(pubKeyList, pubKey)
                        errFmtMsg = "unable to generate key: %s"
                } else {
                        pubKeyList = make(data.KeyList, 0, len(keyList))
                        for _, keyID := range keyList {
                                pubKey = r.CryptoService.GetKey(keyID)
                                pubKeyList = append(pubKeyList, pubKey)
                        }   
                }   
        }
```
 but feel free to keep your code style.

client/client.go
```diff
@@ -906,7 +909,7 @@ func (r *NotaryRepository) bootstrapClient(checkInitialized bool) (*TUFClient, e
 // RotateKey removes all existing keys associated with the role, and either
 // creates and adds one new key or delegates managing the key to the server.
 // These changes are staged in a changelist until publish is called.
-func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool) error {
+func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool, keyList []string) error {
```
riyazdf: can we document what `keyList` is in the function description?

client/client.go
riyazdf: I'm fine with the switch statement, but I'm wondering if we want to allow providing a list of keys to for non-root roles as it seems to currently do?

And if so, can we test this?

cmd/notary/keys.go
```diff
@@ -226,12 +233,23 @@ func (k *keyCommander) keysRotate(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	var keyList []string
+
+	for _, keyFile := range k.rotateKeyFiles {
```
riyazdf: can we add an integration test that makes use of this flag?

cmd/notary/keys.go
```diff
@@ -33,7 +33,7 @@ var cmdKeyListTemplate = usageTemplate{
 }
 
 var cmdRotateKeyTemplate = usageTemplate{
-	Use:   "rotate [ GUN ] [ key role ]",
+	Use:   "rotate [ GUN ] [ key role ] [ key ]",
```
riyazdf: Since `key` is a flag and not a positional argument, we shouldn't need to include it on this part of the usage.  It will print with a `[flags]` appended to it

server/storage/rethinkdb_models.go
```diff
@@ -6,9 +6,10 @@ import (
 
 // These consts are the index names we've defined for RethinkDB
 const (
-	rdbSha256Idx        = "sha256"
-	rdbGunRoleIdx       = "gun_role"
-	rdbGunRoleSha256Idx = "gun_role_sha256"
+	rdbSha256Idx         = "sha256"
+	rdbGunRoleIdx        = "gun_role"
+	rdbGunRoleSha256Idx  = "gun_role_sha256"
+	rdbGunRoleVersionIdx = "gun_role_version"
```
riyazdf: Rethink can't reuse the same name for secondary indexes as it does for primary key (failure message here: https://circleci.com/gh/docker/notary/3840#tests/containers/3) -- can we just use the primary key?

riyazdf:
thanks @ecordell for your work on this!  I have a couple of things:

- I want to echo [my comment](https://github.com/docker/notary/pull/942#discussion_r78256815) about whether we should allow non-root roles to rotate to multiple keys -- it's fine with me but if so we should make sure it's adequately tested in unit and integration tests.  I'm also fine with leaving the scope of this PR to only root-roles and expanding on this later.

- nit: can we rename the PR to something like `Support specifying specific keys on root rotation`?  It's a little confusing because root key rotation is technically already supported :)

client/client.go
cyli: `r.CryptoService.GetKey(keyID)` can return `nil` for a key, though.  Should it be checked before adding to the `pubKeyList`?  Otherwise https://github.com/docker/notary/pull/942/files#diff-1cd41a2611da49ce63630d5a51bd82c5R960 might panic.

Not sure whether the key ID not existing should warrant an error, but maybe an empty `pubKeyList` should?

HuKeping: @cyli it seems like the `for range` will just abort if the keylist is `nil`

https://play.golang.org/p/QzFKDS5gaa

client/client.go
ecordell: Currently keys are checked to be "root keys" by checking that they're encrypted. If we want to support specifying keys for all rotations, I can amend this PR to include that, but it's not the current behavior.

server/storage/rethinkdb_models.go
ecordell: Fixed - just didn't notice that was already the primary key and didn't run the rethink tests locally.

ecordell:
@riyazdf thanks for taking a look!

1. Responded inline, but currently this will fail for non-root roles because the non-root keys are not generally encrypted. I dislike that this is an implicit restriction so I think we should choose one way or the other and I'll update.

2. I'll update the title

client/client.go
cyli: @HuKeping That's true - it won't cause any panics.  But then an empty list will be passed to `rootFileChangeList`, which might eventually result in either a failure in signing or a rejected update because you can't have an empty list of keys.  It's probably a more friendly error if we just say right away that there were invalid keys, as @ecordell has done here.  :)

ecordell:
@riyazdf 
 
 - Fixed the rethink queries
 - Supports rotating non-root keys as well
 - Added integration tests for rotating keys

Please take a look and let me know if there's anything I've forgotten. I think this is ready to go despite the codecov failure - take a look and see if you agree.

riyazdf:
jenkins, test this please

cyli:
jenkins, test this please

client/tufclient.go
```diff
@@ -78,25 +79,81 @@ func (c *TUFClient) update() error {
...
+// downloads all intermediate root files to allow proper key rotation.
+func (c *TUFClient) updateRoot() error {
 	role := data.CanonicalRootRole
-	consistentInfo := c.newBuilder.GetConsistentInfo(role)
 
-	// We can't read an exact size for the root metadata without risking getting stuck in the TUF update cycle
-	// since it's possible that downloading timestamp/snapshot metadata may fail due to a signature mismatch
-	if !consistentInfo.ChecksumKnown() {
-		logrus.Debugf("Loading root with no expected checksum")
+	// Get Current Root Version
```
cyli: The old builder should already have the root role loaded (from `NotaryRepository.bootstrapClient`), if the one on disk was valid.  If a new one has been downloaded since, it's only because it was validated against the previous one, so no updates should be necessary.  So I think we can just call `c.oldBuilder.GetLoadedVersion(role)` here?

client/tufclient.go
```diff
@@ -78,25 +79,81 @@ func (c *TUFClient) update() error {
...
+	currentRootConsistentInfo := c.oldBuilder.GetConsistentInfo(role)
+	currentRoot, _ := c.cache.GetSized(role, -1)
+	c.oldBuilder.Load(currentRootConsistentInfo.RoleName, currentRoot, 1, true)
+	currentVersion := c.oldBuilder.GetLoadedVersion(currentRootConsistentInfo.RoleName)
+
+	// Get New Root Version
+	raw, err := c.remote.GetSized(data.CanonicalRootRole, -1)
```
cyli: `c.tryLoadRemote` and  `c.c.tryLoadCacheThenRemote` both return the failed root bytes, if it fails to load.  If we can make `c.downloadRoot` also do so, then you can probably just call that here, and if it succeeds, we can skip the rest of the logic to check versions, since we have to do `downloadRoot` if it's a version behind anyway. 

If it fails, and it's not due to a `trustpinning.ErrRootRotationFail` (for instance if it's expired, invalid, or otherwise fails trust pinning), we can also bail without downloading intermediate versions.

client/tufclient.go
```diff
@@ -78,25 +79,81 @@ func (c *TUFClient) update() error {
...
-		return remoteErr
 	}
+	return nil
+}
 
-	_, err := c.tryLoadCacheThenRemote(consistentInfo)
-	return err
+// updateRootVersions updates the root from it's current version to a target, rotating keys
+// as they are found
+func (c *TUFClient) updateRootVersions(currentRoot []byte, fromVersion, toVersion int) error {
```
cyli: `currentRoot` seems unused in this function?  Although it could be.  (see comment about atomic updates a little further below)

client/tufclient.go
```diff
@@ -78,25 +79,81 @@ func (c *TUFClient) update() error {
...
+	if fromVersion == toVersion {
+		logrus.Debugf("finished updating root files")
+		return nil
+	}
+	nextVersion := fromVersion + 1
+	skipCheckExpiryAndChecksum := !(nextVersion == toVersion)
```
cyli: Totally non-blocking nitpick: would `nextVersion != toVersion` be easier to read?

client/tufclient.go
```diff
@@ -78,25 +79,81 @@ func (c *TUFClient) update() error {
...
+		logrus.Debugf("error downloading %s: %s", versionedRole, err)
+		return err
+	}
+	// If loading the most recent version, check expiry
+	if err := c.newBuilder.LoadOptions(data.CanonicalRootRole, raw, nextVersion, skipCheckExpiryAndChecksum, skipCheckExpiryAndChecksum, true); err != nil {
+		logrus.Debugf("downloaded %s is invalid: %s", versionedRole, err)
+		return err
+	}
+	logrus.Debugf("successfully verified downloaded %s", versionedRole)
+	if err := c.cache.Set(data.CanonicalRootRole, raw); err != nil {
```
cyli: Totally non-blocking:  we have an issue which has not been resolved yet about making updates atomic:  https://github.com/docker/notary/issues/408.  Eventually we probably want to update this whole update logic to download to a download cache before moving everything back to the on-disk cache.

In the meantime though, since we're only downloading roots to support root rotation, that means that if the update fails, the cache on disk will be inconsistent (there will be a root that doesn't match any of the other metadata).  We already pollute the cache if we fail to download other metadata, so this wouldn't be a deviation from existing behavior, but wondering if we should only try to write to cache if we successfully updated the root to the latest version?

This would be a good use of the `currentRoot` param in this function.

client/tufclient.go
cyli: There was no original reason to, other than it was factored down from something bigger where presumably this mattered.  :)

cmd/notary/integration_test.go
```diff
@@ -1289,6 +1289,177 @@ func TestClientKeyGenerationRotation(t *testing.T) {
 	}
 }
 
+// Tests key rotation
+func TestKeyRotation(t *testing.T) {
```
cyli: 👍 Thanks for adding these! 

Would you mind also adding one assertion that rotating to a key that doesn't exists fails?  :)

client/client.go
```diff
@@ -815,7 +815,7 @@ func (r *NotaryRepository) Update(forWrite bool) error {
 		// notFound.Resource may include a checksum so when the role is root,
 		// it will be root or root.<checksum>. Therefore best we can
 		// do it match a "root." prefix
-		if notFound, ok := err.(store.ErrMetaNotFound); ok && strings.HasPrefix(notFound.Resource, data.CanonicalRootRole+".") {
+		if notFound, ok := err.(store.ErrMetaNotFound); ok && strings.Contains(notFound.Resource, data.CanonicalRootRole) {
```
endophage: We should probably just improve the `ErrMetaNotFound` to have a field for the role.

ecordell: Updated, good idea

client/client.go
```diff
@@ -954,29 +976,30 @@ func (r *NotaryRepository) RotateKey(role string, serverManagesKey bool) error {
...
 	if role == data.CanonicalRootRole {
-		privKey, _, err := r.CryptoService.GetPrivateKey(pubKey.ID())
-		if err != nil {
-			return err
-		}
-		pubKey, err = rootCertKey(r.gun, privKey)
-		if err != nil {
-			return err
+		for i, pubKey := range pubKeyList {
+			privKey, _, err := r.CryptoService.GetPrivateKey(pubKey.ID())
```
ecordell: I updated this but it required adding some methods to the storage layer to get that information

client/tufclient.go
ecordell: Yup!

client/tufclient.go
ecordell: That cleans things up a lot, updated

client/tufclient.go
ecordell: I removed it for now - I agree atomic updates are really important, but I could see that potentially involving a different solution than passing currentRoot around. It will be easy to add back if that's appropriate.

cmd/notary/integration_test.go
ecordell: Added

server/server.go
```diff
@@ -151,7 +151,15 @@ func RootHandler(ac auth.AccessController, ctx context.Context, trust signed.Cry
...
 	r.Methods("GET").Path("/v2/{imageName:.*}/_trust/tuf/{tufRole:root|targets(?:/[^/\\s]+)*|snapshot|timestamp}.{checksum:[a-fA-F0-9]{64}|[a-fA-F0-9]{96}|[a-fA-F0-9]{128}}.json").Handler(createHandler(_serverEndpoint{
-		OperationName:       "GetRoleByHash",
+		OperationName:       "",
+		ErrorIfGUNInvalid:   notFoundError,
+		IncludeCacheHeaders: true,
+		CacheControlConfig:  consistent,
+		ServerHandler:       handlers.GetHandler,
+		PermissionsRequired: []string{"pull"},
+	}))
+	r.Methods("GET").Path("/v2/{imageName:.*}/_trust/tuf/{version:[0-9]+}.{tufRole:root|targets(?:/[^/\\s]+)*|snapshot|timestamp}.json").Handler(createHandler(_serverEndpoint{
```
ecordell: seems reasonable to me, I updated.

server/storage/interface.go
```diff
@@ -36,6 +36,11 @@ type MetaStore interface {
 	// not found, it returns storage.ErrNotFound
 	GetChecksum(gun, tufRole, checksum string) (created *time.Time, data []byte, err error)
 
+	// GetVersion returns the given TUF role file and creation date for the
+	// GUN with the provided version. If the given (gun, role, version) are
+	// not found, it returns storage.ErrNotFound
+	GetVersion(gun, tufRole, version string) (created *time.Time, data []byte, err error)
```
ecordell: I missed this comment; I don't feel strongly either way...

server/storage/rethinkdb.go
```diff
@@ -231,6 +232,28 @@ func (rdb RethinkDB) GetChecksum(gun, role, checksum string) (created *time.Time
...
 }
 
+// GetVersion gets a specific TUF record by its version
+func (rdb RethinkDB) GetVersion(gun, role, version string) (*time.Time, []byte, error) {
+	var file RDBTUFFile
+	v, err := strconv.Atoi(version)
+	if err != nil {
+		return nil, nil, ErrNotFound{}
+	}
+	res, err := gorethink.DB(rdb.dbName).Table(file.TableName(), gorethink.TableOpts{ReadMode: "majority"}).Get([]interface{}{gun, role, v}).Run(rdb.sess)
```
ecordell: That's how I originally had it but it doesn't work because that index is actually on the primary key, so you just use `Get`.

server/storage/storage_test.go
```diff
@@ -126,6 +126,18 @@ func testUpdateCurrentVersionCheck(t *testing.T, s MetaStore, oldVersionExists b
...
+// GetVersion should successfully retrieve a version of an existing TUF file,
+// but will return an error if the requested version does not exist.
+func testGetVersion(t *testing.T, s MetaStore) {
+	_, _, err := s.GetVersion("gun", "role", "2")
+	require.IsType(t, ErrNotFound{}, err, "Expected error to be ErrNotFound")
+
+	s.UpdateCurrent("gun", MetaUpdate{"role", 2, []byte("test")})
+	_, d, err := s.GetVersion("gun", "role", "2")
+	require.Nil(t, err, "Expected error to be nil")
+	require.Equal(t, []byte("test"), d, "Data was incorrect")
```
ecordell: Done

tuf/builder.go
```diff
@@ -242,11 +246,16 @@ func (rb *repoBuilder) GetConsistentInfo(roleName string) ConsistentInfo {
 }
 
 func (rb *repoBuilder) Load(roleName string, content []byte, minVersion int, allowExpired bool) error {
+	return rb.LoadOptions(roleName, content, minVersion, allowExpired, false, false)
+}
+
+// LoadOptions adds additional flags that should only be used for updating the root.json
+func (rb *repoBuilder) LoadOptions(roleName string, content []byte, minVersion int, allowExpired, skipChecksum, allowLoaded bool) error {
```
ecordell: That makes sense, I did the simple refactor for now and I think a larger refactor of the builders should happen in a future PR.

client/tufclient.go
```diff
@@ -78,25 +79,81 @@ func (c *TUFClient) update() error {
...
+		logrus.Debugf("Unable to write %s to cache: %s", versionedRole, err)
+	}
+	return c.updateRootVersions(raw, nextVersion, toVersion)
```
ecordell: I personally find recursive forms to be more understandable, but I've updated it. It probably makes more sense anyway, since go doesn't do any TCO.

client/client.go
cyli: I think `r.CryptoService.GetPrivateKey` returns a key, the role string, and the error?  So I think we can just use the second argument.  :)

client/tufclient.go
```diff
@@ -78,25 +80,68 @@ func (c *TUFClient) update() error {
...
-	_, err := c.tryLoadCacheThenRemote(consistentInfo)
-	return err
+// updateRootVersions updates the root from it's current version to a target, rotating keys
+// as they are found
+func (c *TUFClient) updateRootVersions(fromVersion, toVersion, newestVersion int) error {
+	skipCheckExpiryAndChecksum := toVersion != newestVersion
+	logrus.Debugf("updating root from version %d to version %d, up to %d", fromVersion, toVersion, newestVersion)
+
+	var versionedRole string
+	if toVersion == newestVersion {
```
endophage: This still shouldn't be necessary. Just download the `newestVersion` as `<version>.<role>`. That also mitigates any race conditions where you've pulled and are walking the root rotations and somebody else pushes.

client/tufclient.go
```diff
@@ -78,25 +80,68 @@ func (c *TUFClient) update() error {
...
+	signedRoot := &data.Signed{}
+	if err := json.Unmarshal(raw, signedRoot); err != nil {
+		return err
+	}
+	newestRoot, err := data.RootFromSigned(signedRoot)
+	if err != nil {
+		return err
+	}
+	newestVersion := newestRoot.Signed.Version
+	for i := currentVersion; i < newestVersion; i++ {
```
endophage: Was thinking this loop would be inside `updateRootVersions`, then you don't need to do the funky `i, i+1`. You'd just have `if err := updateRootVersions(from, to int); err != nil {return err}`

server/storage/rethinkdb.go
cyli: Oh, ok that makes sense.  Thanks for explaining!  :)

server/storage/rethinkdb.go
```diff
@@ -231,6 +232,28 @@ func (rdb RethinkDB) GetChecksum(gun, role, checksum string) (created *time.Time
...
+	if err != nil {
+		return nil, nil, err
+	}
+	defer res.Close()
+	if res.IsNil() {
+		return nil, nil, ErrNotFound{}
+	}
+	err = res.One(&file)
+	if err == gorethink.ErrEmptyResult {
```
cyli: Non-blocking:  just out of curiosity, can this happen, since you checked for `nil` before?

server/storage/interface.go
cyli: I wouldn't block the PR on this, as it's pretty minor and and can easily be fixed later, but my general preference for interfaces to be specific as to the types of arguments expected;  the version is defined as an int in the TUF data, etc., and this would let the compiler verify the type instead of getting an input error when it fails to convert.

server/storage/rethinkdb.go
ecordell: I'm not very familiar with rethink or it's go api, I just checked here since that's what GetChecksum did

ecordell:
@endophage @cyli I believe I've addressed all comments in the current PR, thanks!

Before merging, I think I need to implement the `--drop` flag from #841, unless consensus has changed?

I'd prefer to get the server and the client in together, but if we need to split them up that's fine. I'll need to rewrite the commits since I did some refactoring across both. Let me know which is preferred!

cyli:
jenkins, test this please

ecordell:
I believe this also at least partially addresses https://github.com/docker/notary/issues/410 and https://github.com/docker/notary/issues/445 , as it makes all metadata files addressable by version (assuming the db hasn't been pruned, etc etc)

ecordell:
Recap: The plan is to have a flag allowing users to sign with old keys. If provided, we pull old versions of the root file from the server, correlate the keys specified in the root role with those the signer has on disk, and sign with them.

We may choose to go back a certain amount of time (e.g. two months) or versions (e.g. two versions).

Food for thought: what about an automated `--legacy` flag that does some heuristics, like if you have >75% of the keys for the past X versions, sign without interaction?

ecordell:
I've implemented a potential solution for supporting legacy clients; I've pushed it up early for review (it still needs lots of testing).

Currently this hardcodes fetching the root keys from 5 previous versions. It's not clear to me that the number of versions back to fetch should necessarily be user-defined (if it looks like no one has rotated more than once, for example, we could just hardcode the number here to 20 and be done with it). If it's necessary to control how far back to sign for, should it be set by number of versions or date?

endophage:
I like the idea of doing the backwards fetch based on time. It feels logical and in the spirit of TUF to say "if a client hasn't updated such that they have a version no older than X, they're so far out of date they need to do a more manual recovery."

Would could have a default of say, 2 months (completely flexible on the specific period), while allowing a user to supply a flag to change the period. I don't think the time needs to be particularly long.

endophage:
I'd say if the immediately previous key isn't found (based on the old rotation logic for now), it should be a hard fail. Now that we have delete released, in that instance the user has to recover by deleting all their existing trust data and starting fresh, because no up to date client would accept the rotation.

If older keys aren't found, it should probably warn but publish. With this legacy signing mode, the user can  re-sign the root once they have recovered the older keys (if they consider that to even be necessary). That does mean we'll also need a way to re-sign roles, but that doesn't need to be in this PR, just needs to be in by the next release.

ecordell:
My only hangup with a time-based solution is that it's unpredictable. How many network calls is "4 months back"?

endophage:
I agree it has the potential to be a lot of network calls. Could the endpoint be optimised to return multiple versions in a single response?

Time just seems like a more natural scale to think on. It's easier for a human to remember "I changed my key in February, it's now October, I want to go back 8 months" (and yes, I acknowledge this could be a _lot_ of versions), vs thinking "I changed my key in February, I think I've published 20 new root.json versions since then"

I don't necessarily expect root rotations to be fast, my concerns are more about lowering the complexity to help humans use it. I'm sure we can find ways to get the speed elsewhere.

cyli:
jenkins, test this please

ecordell:
Thanks for the review @cyli! I'll work on this and update soon.

Before merging, I also need to double check that this handles cases where intermediate versions do not exist, or enforce that version numbers always increment by 1.

client/client_test.go
```diff
@@ -2900,6 +2977,170 @@ func TestRotateRootKey(t *testing.T) {
...
+	require.NoError(t, authorRepo.Update(false))
+	newRootRole, err := authorRepo.tufRepo.GetBaseRole(data.CanonicalRootRole)
+	require.False(t, newRootRole.Equals(oldRootRole))
+	require.NoError(t, err)
+	// not only is the root cert different, but the private key is too
+	newRootCertID := rootRoleCertID(t, authorRepo)
+	require.NotEqual(t, oldRootCertID, newRootCertID)
+	newCanonicalKeyID, err := utils.CanonicalKeyID(newRootRole.Keys[newRootCertID])
+	require.NoError(t, err)
+	require.NotEqual(t, oldCanonicalKeyID, newCanonicalKeyID)
```
ecordell: Definitely! Will fix.

client/tufclient.go
```diff
@@ -78,25 +80,71 @@ func (c *TUFClient) update() error {
...
+		return err
+	}
+	logrus.Debugf("finished updating root files")
+	return nil
+}
+
+// updateRootVersions updates the root from it's current version to a target, rotating keys
+// as they are found
+func (c *TUFClient) updateRootVersions(fromVersion, toVersion int) error {
+	for v := fromVersion; v <= toVersion; v++ {
```
ecordell: Because the initial and final version cases are unique, I've moved those out of this method and changed the range this method operates over

cmd/notary/tuf.go
```diff
@@ -464,11 +466,19 @@ func readRootKey(rootKeyFile string, retriever notary.PassRetriever) (data.Priva
...
 	if err = cryptoservice.CheckRootKeyIsEncrypted(pemBytes); err != nil {
-		return nil, err
+		if role == data.CanonicalRootRole {
+			return nil, err
+		}
+		isEncrypted = false
+	}
+	var privKey data.PrivateKey
+	if isEncrypted {
+		privKey, _, err = trustmanager.GetPasswdDecryptBytes(retriever, pemBytes, "", data.CanonicalRootRole)
```
ecordell: I based this off the current behavior in notary, which is that encrypted keys are for root keys only. Is it reasonable to allow all keys to be encrypted?

signer/keydbstore/rethink_keydbstore.go
```diff
@@ -215,6 +215,16 @@ func (rdb *RethinkDBKeyStore) GetKey(keyID string) data.PublicKey {
 	return data.NewPublicKey(dbPrivateKey.Algorithm, dbPrivateKey.Public)
 }
 
+// GetKeyInfo returns the gun and role given a KeyID, and does not activate the key
+func (rdb *RethinkDBKeyStore) GetKeyInfo(keyID string) (trustmanager.KeyInfo, error) {
+	dbPrivateKey, _, err := rdb.getKey(keyID)
+	if err != nil {
+		return trustmanager.KeyInfo{}, nil
```
ecordell: nice catch!

tuf/tuf.go
```diff
@@ -167,19 +165,12 @@ func (tr *Repo) RemoveBaseKeys(role string, keyIDs ...string) error {
...
-	// util.CanonicalKeyID conversion.
-	if role != data.CanonicalRootRole {
-		for k := range toDelete {
-			delete(tr.Root.Signed.Keys, k)
-			tr.cryptoService.RemoveKey(k)
-		}
+	// Remove keys no longer in use by any roles
+	for k := range toDelete {
+		delete(tr.Root.Signed.Keys, k)
+		tr.cryptoService.RemoveKey(k)
```
ecordell: Let's err on the safe side and not delete the key material here :+1:

ecordell:
@cyli @endophage Sorry for the delay in updating this - pretty sure I've addressed everything at this point.

Added a restriction that root updates must increment by 1 or 0 to the server. 

For legacy support, is the current solution acceptable? Should the default be to sign with all old keys instead? (I prefer the default to sign with only the immediately previous, but we could start with signing with all by default and then at some future point flip that).

cyli:
jenkins, test this please

cmd/notary/tuf.go
cyli: Oops, sorry I missed this question before for some reason!  Hm... I thought we encrypted all the keys on disk (hence it asks for your passphrase when a targets key is generated).  For instance, in `cmd/notary/keys.go` above, `readKey` is called when rotating, which could technically be for any role.

cyli:
Thanks @ecordell!  And no worries - we have also been super unresponsive :)  Thanks for your patience.  I'll have a look at this probably tomorrow or over the weekend!

endophage:
Going to test this out this afternoon/tomorrow. Excited to see this wrapped up!

client/client.go
```diff
@@ -700,11 +716,113 @@ func signRootIfNecessary(updates map[string][]byte, repo *tuf.Repo, initialPubli
...
+		oldKeyList = append(oldKeyList, key)
+	}
+	return oldKeyList, nil
+}
+
+// get all the saved previous roles keys < the current root version
+func getOldRootPublicKeys(root *data.SignedRoot) data.KeyList {
+	oldRootKeys := make(data.KeyList, 0, len(root.Signed.Roles))
+	// now go through the old roles
+	for roleName := range root.Signed.Roles {
```
cyli: Non-blocking, just a cleanup: `root.BuildBaseRole` will index into the map, so we don't have to iterate over it.  Rather than the loop, this can probably just be:

```
rootRole, err := root.BuildBaseRole(roleName)
if err != nil {
    return nil
}

return rootRole.ListKeys()
```

If there are other cleanups, this would be awesome to include, if not, it can be done later :)

client/client.go
```diff
@@ -918,48 +1023,98 @@ func (r *NotaryRepository) bootstrapClient(checkInitialized bool) (*TUFClient, e
...
+}
+
+func (r *NotaryRepository) pubKeysToCerts(role string, pubKeyList data.KeyList) (data.KeyList, error) {
+	// only generate certs for root keys
+	if role != data.CanonicalRootRole {
+		return pubKeyList, nil
+	}
+
+	for i, pubKey := range pubKeyList {
+		keyInfo, err := r.CryptoService.GetKeyInfo(pubKey.ID())
```
cyli: Non-blocking optimization:  `r.CryptoService.GetPrivateKey(pubKey.ID())` returns the private key, the role, and an error, so we wouldn't need the extra call to `r.CryptoService.GetKeyInfo` to check that the role is root.

client/client_test.go
```diff
@@ -2900,6 +2977,170 @@ func TestRotateRootKey(t *testing.T) {
...
+	err = authorRepo.Publish()
+	require.NoError(t, err)
+	logRepoTrustRoot(t, "post-publish", authorRepo)
+
+	// Verify the user can use the rotated repo, and see the added target.
+	err = userRepo.Update(false)
+	require.NoError(t, err)
 	_, err = userRepo.GetTargetByName("current")
 	require.NoError(t, err)
 	logRepoTrustRoot(t, "client", userRepo)
```
cyli: Thanks for adding the check!  :D  (commenting because github displays this in the wrong place in the diff now)

client/client_test.go
cyli: Thanks for adding that!  (commenting because github displays this in the wrong place in the diff now)

server/handlers/validation.go
```diff
@@ -43,8 +43,13 @@ func validateUpdate(cs signed.CryptoService, gun string, updates []storage.MetaU
 	}
 
 	if rootUpdate, ok := roles[data.CanonicalRootRole]; ok {
+		currentRootVersion := builder.GetLoadedVersion(data.CanonicalRootRole)
+		if rootUpdate.Version != currentRootVersion && rootUpdate.Version != currentRootVersion+1 {
```
cyli: `GetLoadedVersion` returns `1` if there is no root role yet.  Do we want to skip this check if there is no root on the server?  (For instance, if the user is pushing the latest metadata, but not the full history, from one notary server to another).

Non-blocking question:  I was wondering if there was an issue with allowing gaps in versions (handling the 404's on the client side?)  I'm totally ok with enforcing monotonically increasing versions for now, and then allowing for gaps in versions at a later point if we wanted, I am just checking to see if I'm missing an edge case or some other difficulty.

cyli: Also can we add a test for rejecting updates that skip ahead by more than 1, if we're going to go with this route?

cyli:
jenkins, test this please

cyli:
@ecordell As an update, we are probably going to do a group review of this to try to get this merged as soon as we get back from the holiday break.  Apologies for the scheduling issues, and thanks again for your patience!

Hope you have a happy holiday!

ecordell:
Thanks for the review @cyli! 

> (As a separate issue, I just wanted to note it here for now - we may also want to allow witness of a root to also be able to add all the legacy signatures, in case someone rotated a root but then later decided they want to support legacy clients).

Agreed this sounds like a great idea.

Should I work on fixing the codecov or is this good to go?

server/storage/rethinkdb_models.go
```diff
@@ -6,9 +6,10 @@ import (
 
 // These consts are the index names we've defined for RethinkDB
 const (
-	rdbSHA256Idx        = "sha256"
-	rdbGunRoleIdx       = "gun_role"
-	rdbGunRoleSHA256Idx = "gun_role_sha256"
+	rdbSha256Idx         = "sha256"
+	rdbGunRoleIdx        = "gun_role"
+	rdbGunRoleSha256Idx  = "gun_role_sha256"
```
cyli: Group review:  SHA in a variable name should probably remain all caps, since it's an acronym as per go naming conventions.

tuf/signed/interface.go
```diff
@@ -18,6 +19,9 @@ type KeyService interface {
 	// GetKey retrieves the public key if present, otherwise it returns nil
 	GetKey(keyID string) data.PublicKey
 
+	// GetKeyInfo retrieves the KeyInfo for a keyID
+	GetKeyInfo(keyID string) (trustmanager.KeyInfo, error)
```
riyazdf: Group review: seems like we don't need this, can it be deleted from the interface and implementations?

tuf/builder.go
```diff
@@ -407,11 +427,15 @@ func (rb *repoBuilder) GenerateTimestamp(prev *data.SignedTimestamp) ([]byte, in
 	return sgndJSON, rb.repo.Timestamp.Signed.Version, nil
 }
 
+func (rb *repoBuilder) setPrevRoot(root *data.SignedRoot) {
```
cyli: Group review:  This doesn't seem to be used, since in `LoadRootForUpdate` the `prevRoot` is just being set directly.  And additionally, both this function and the variable it's setting are unexported.

client/tufclient.go
```diff
@@ -78,25 +80,92 @@ func (c *TUFClient) update() error {
...
+	raw, err := c.downloadRoot()
 
-		// get the cached root, if it exists, just for version checking
-		cachedRoot, _ := c.cache.GetSized(role, -1)
-		// prefer to download a new root
-		_, remoteErr := c.tryLoadRemote(consistentInfo, cachedRoot)
-		return remoteErr
+	// Return any non-rotation error. Rotation errors are okay since we haven't yet downloaded
+	// all intermediate root files
+	if _, ok := err.(*trustpinning.ErrRootRotationFail); err != nil && !ok {
```
cyli: Group review:  For readability, can we turn this into a switch statement?  
```
switch err.(type) {
    case *trustpinning.ErrRootRotationFail:
         continue
    case nil:
         return nil
    default:
         return error
}
```

client/tufclient.go
```diff
@@ -78,25 +80,92 @@ func (c *TUFClient) update() error {
...
+		if err != nil {
+			logrus.Debugf("error downloading %s: %s", versionedRole, err)
+			return err
+		}
+		if err := c.newBuilder.LoadRootForUpdate(raw, v, false); err != nil {
+			logrus.Debugf("downloaded %s is invalid: %s", versionedRole, err)
+			return err
+		}
+		logrus.Debugf("successfully verified downloaded %s", versionedRole)
+		if err := c.cache.Set(data.CanonicalRootRole, raw); err != nil {
```
cyli: Group review:  since this is transient data, could we not cache this for now?  If we wanted to improve our caching situation later (e.g. with atomic caching, or checksum-based caching, or version-based caching) it can be in a later PR.  

We haven't actually verified that this root belongs in the repo (that it matches with the timestamp and snapshot, for instance).

Writing a root to cache is an indication that we're pretty confident that this root is valid (we might use it in the future), and that is not necessarily true of these transient roots.

cmd/notary/tuf.go
cyli: Group review:  as a note, this whole function (and the subsequent `AddKey` calls) can probably be replaced with `utils.ImportKeys` at a later point (not necessarily in this PR - just wanted to note it here).

server/storage/rethinkdb_models.go
ecordell: :+1: I think this is a rebasing regression that was my fault

tuf/signed/interface.go
ecordell: :+1: Yup, this was rendered unnecessary by refactoring.

tuf/builder.go
ecordell: :+1: I will double check and remove

cyli:
Group review:  Fixing the code coverage would be awesome, but not super high priority.

There are a couple of specific test cases we'd like though:

1.  Testing the notary server validation prevents users from skipping a root version (or pushing a root version to a new repo that starts at > 1).
2.  Testing that if one of the intermediate roots does not verify (not properly root rotated), updating fails. 
3.  Testing that updating to the newest root still works if the intermediate roots are expired.

ecordell:
Thanks for the comments! I've addressed everything inline in a new commit, and added a test for (1). 

If I understand correctly, the other requested tests exist in (2) `TestRootRoleInvariant`, and (3) `TestUpdateFailsIfServerRootKeyChangedWithoutMultiSign`

cyli:
jenkins, test this please

tuf/signed/sign_test.go
```diff
@@ -60,6 +60,10 @@ func (mts *FailingCryptoService) GetKey(keyID string) data.PublicKey {
 	return nil
 }
 
+func (mts *FailingCryptoService) GetKeyInfo(keyID string) (trustmanager.KeyInfo, error) {
```
cyli: This function can probably be removed too.

tuf/signed/sign_test.go
```diff
@@ -88,6 +92,13 @@ func (mts *MockCryptoService) GetKey(keyID string) data.PublicKey {
 	return nil
 }
 
+func (mts *MockCryptoService) GetKeyInfo(keyID string) (trustmanager.KeyInfo, error) {
```
cyli: This function can probably be removed too.

signer/client/signer_trust.go
```diff
@@ -203,6 +204,11 @@ func (trust *NotarySigner) GetKey(keyid string) data.PublicKey {
 	return pubKey
 }
 
+// GetKeyInfo returns the gun and role given a keyID
```
cyli: This function can probably be removed too.

signer/keydbstore/rethink_keydbstore.go
```diff
@@ -215,6 +215,16 @@ func (rdb *RethinkDBKeyStore) GetKey(keyID string) data.PublicKey {
 	return data.NewPublicKey(dbPrivateKey.Algorithm, dbPrivateKey.Public)
 }
 
+// GetKeyInfo returns the gun and role given a KeyID, and does not activate the key
```
cyli: This function can probably be removed too.

signer/keydbstore/sql_keydbstore.go
```diff
@@ -228,6 +228,15 @@ func (s *SQLKeyDBStore) GetKey(keyID string) data.PublicKey {
 	return data.NewPublicKey(privKey.Algorithm, []byte(privKey.Public))
 }
 
+// GetKeyInfo returns role and GUN info of a key by ID
```
cyli: This function can probably be removed too.

cyli:
@ecordell 
Thanks for adding the version skip test!  (2) `TestRootRoleInvariant` tests that the latest root does not validate, but not intermediate downloaded roots, and (3) `TestUpdateFailsIfServerRootKeyChangedWithoutMultiSign` tests that if the cached root is expired, then an update succeeds so long as the new root is valid.

We were thinking for (3) something like having the local cache at v1 (either expired or not), and downloading an expired v2 and a valid v3?  It should successfully update to v3.

Similarly for (2), except v2 is not properly rotated.  The only reason being because some checks are skipped for intermediate roots (like checksum and expiry), and this would validate that the root signing is validated for the intermediate roots.

ecordell:
@cyli I gotcha - I added those extra tests to be safe, thanks for clarifying.

I also removed those other `GetKeyInfo` methods - somehow I missed them, sorry about that.

cyli:
@ecordell Thanks for adding them!    

> I also removed those other GetKeyInfo methods - somehow I missed them, sorry about that.

Not at all, easy to lose track :)

cyli:
jenkins, test this please

cyli:
LGTM!  Thank you so much for all your work and patience on this @ecordell!

ecordell:
Thanks! I've rebased on master, so hopefully everything will be good shortly.

cyli:
Jenkins, test this please

tuf/tuf.go
```diff
@@ -965,62 +927,6 @@ func (tr *Repo) SignRoot(expires time.Time) (*data.Signed, error) {
...
-
-	oldKeys := make([]data.PublicKey, 0, len(oldKeysMap))
-	for _, key := range oldKeysMap {
-		oldKeys = append(oldKeys, key)
-	}
-
-	return oldKeys
-}
-
 func oldRootVersionName(version int) string {
```
endophage: Followup PR: I think we can remove this function. I don't think it's used anywhere with the new root rotation logic.

client/client.go
```diff
@@ -700,11 +714,84 @@ func signRootIfNecessary(updates map[string][]byte, repo *tuf.Repo, initialPubli
 	return nil
 }
 
+// Fetch back a `legacyVersions` number of roots files, collect the root public keys
+// This includes old `root` roles as well as legacy versioned root roles, e.g. `1.root`
+func (r *NotaryRepository) oldKeysForLegacyClientSupport(legacyVersions int, initialPublish bool) (data.KeyList, error) {
```
riyazdf: followup PR: I'm pretty happy with this function but would like to add some unit tests around overly large or negative legacyVersions and the skipping logic

HuKeping:
Thanks for your patience and great work on this @ecordell !

ecordell:
Thanks for all of the reviews! I can't merge myself if someone with write access would like to :) Glad to finally have this wrapped up.

PR #943, opened 2016-09-08
Bump go to version 1.7.1

cyli:

Brew doesn't have the 1.7.1 formula yet, so can't update the yubikey tester just yet.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.7.0
+FROM golang:1.7.1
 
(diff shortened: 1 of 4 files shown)


HuKeping:
It seems like we have just bumped to Go1.7 yesterday :last_quarter_moon:

riyazdf:
thanks for the bump, LGTM!

Not sure why the docs status isn't being properly updated to green but all is good 👍

HuKeping:
LGTM! Thanks for tracking this!

PR #946, opened 2016-09-08
Add command reference docs

riyazdf:
This adds a command reference doc, so we can give our users a quicker way to find how to do things like rotate keys with notary CLI.

Inspired by and superseding #904 (thank you again @avaid96 for your work there!)

### DIFF
--- a/docs/command_reference.md
+++ b/docs/command_reference.md
@@ -12,9 +12,9 @@ weight=99
 # Notary Command Reference
 
 ## Terminology Reference
-1. **GUN**: Notary uses Globally Unique Names (GUNs) to identify trusted collections.
-2. **Target**: Notary refers to the files the framework is used to distribute as "target files".  Target files are opaque to the framework. Whether target files are packages containing multiple files, single text files, or executable binaries is irrelevant to Notary.
-3. **Trusted Collection**: A trusted collection is a set of target files of interest to the application.
+1. **GUN**: Notary uses Globally Unique Names (GUNs) to identify trusted collections.  Note that `<GUN>` can take on arbitrary values, but when working with Docker Content Trust they are structured like `<domain>/<account>/<repository>`.  For example `docker.io/library/alpine` for the [Docker Library Alpine image](https://hub.docker.com/r/library/alpine).
+2. **Target**: A target refers to a file in a to be distributed as part of a trusted collection.  Target files are opaque to the framework. Whether target files are packages containing multiple files, single text files, or executable binaries is irrelevant to Notary.
+3. **Trusted Collection**: A trusted collection is a set of target files of interest.
 4. **Notary Repository**: A notary repository refers to the set of signed metadata files that describe a trusted collection.
 5. **Key roles**:
     - **Root**: The root role delegates trust to specific keys trusted for all other top-level roles used in the system.
@@ -59,7 +59,7 @@ If you'd like to initialize your trusted collection with a specific root key, th
 $ notary init <GUN> --rootkey <key_file>
 ```
 
-Note that you will have to run a publish after this command for it to take effect, as there are staged changes to initialize the trusted collection that have not been pushed to a notary server.
+Note that you will have to run a publish after this command for it to take effect, because the Notary CLI client will create staged changes to initialize the trusted collection that have not yet been pushed to a notary server.
 ```bash
 $ notary publish <GUN>
 ```
@@ -81,8 +81,6 @@ $ notary status <GUN> --unstage 0
 # Alternatively, unstage all changes
 $ notary status <GUN> --reset
 ```
-
-Note that `<GUN>` can take on arbitrary values, but when working with Docker Content Trust they are structured like `<domain>/<account>/<repository>`.  For example `docker.io/library/alpine` for the [Docker Library Alpine image](https://hub.docker.com/r/library/alpine).
 
 When you're ready to publish your changes to the Notary server, run:
 
@@ -108,7 +106,7 @@ Users can sign content into a notary trusted collection by running:
 $ notary add -p <GUN> <target_name> <target_file>
 ```
 
-In the above command, the `<target_name>` corresponds to the name we want to associate the `<target_file>`'s signature with. Notary will sign the hash of the `<target_file>` into its trusted collection.
+In the above command, the `<target_name>` corresponds to the name we want to associate the `<target_file>` with in the trusted collection. Notary will sign the hash of the `<target_file>` into its trusted collection.
 Instead of adding a target by file, you can specify a hash and byte size directly:
 ```bash
 $ notary addhash -p <GUN> <target_name> <byte_size> --sha256 <sha256Hash>
@@ -184,7 +182,11 @@ Notary can import keys that are already in a PEM format:
 $ notary key import <pemfile> --role <key_role> --gun <key_gun>
 ```
 
-If no `--role` or `--gun` is given, notary will assume that the key is to be used for a delegation role.
+The `--role` and `--gun` flags can be provided to specify a role and GUN to import the key to if that information is not already included in PEM headers.
+Note that for root and delegation keys, the `--gun` flag is ignored because these keys can be shared across GUNs.
+If no `--role` or `--gun` is given, notary will assume that the key is to be used for a delegation role, which will appear as a `delegation` key in commands such as `notary key list`.
+
+
 Moreover, it's possible for notary to import multiple keys contained in one PEM file, each separated into separate PEM blocks.
 
 For each key it attempts to import, notary will prompt for a passphrase so that the key can be encrypted at rest.
@@ -235,13 +237,13 @@ You can also remove keys from a delegation role, such that those keys can no lon
 # Remove the given keys from a delegation role
 $ notary delegation remove -p <GUN> targets/<role> <keyID1> <keyID2>
 
-# Alternatively, you can remove keys from all delegation roles
+# Alternatively, you can remove keys from all delegation roles, in case of delegation key compromise
 $ notary delegation purge <GUN> --key <keyID1> --key <keyID2>
 ```
 
 ## Managing targets in delegation roles
 
-We can specify to sign content into specific delegation roles by specifying the `--roles` flag.  This also applies to `notary addhash` and `notary remove`.
+We can specify which delegation roles to sign content into by using the `--roles` flag.  This also applies to `notary addhash` and `notary remove`.
 Without the `--roles` flag, notary will attempt to operate on the base `targets` role:
 ```bash
 # Add content from a target file to a specific delegation role
@@ -251,7 +253,7 @@ $ notary add -p <GUN> <target_name> <target_file> --roles targets/<role>
 $ notary addhash -p <GUN> <target_name> <byte_size> --sha256 <sha256Hash> --roles targets/<role>
 
 # Remove content from a specific delegation role
-$ notary add -p <GUN> <target_name> <target_file> --roles targets/<role>
+$ notary remove -p <GUN> <target_name> <target_file> --roles targets/<role>
 ```
 
 Similarly, we can list all targets and prefer to show certain delegation roles' targets first with the `--roles` flag.
@@ -292,4 +294,4 @@ If you are receiving this error:
 * fatal: Get <URL>/v2/: x509: certificate signed by unknown authority
 ```
 The Notary CLI must be configured to trust the root certificate authority of the server it is communicating with.
-This can be configured by specifying the root certificate with the `--tlscacert` flag or by editing the Notary client configuration file.
+This can be configured by specifying the root certificate with the `--tlscacert` flag or by editing the Notary client configuration file.  Additionally, you can add the root certificate to your system CAs.
--- a/docs/command_reference.md
+++ b/docs/command_reference.md
@@ -14,8 +14,9 @@ weight=99
 ## Terminology Reference
 1. **GUN**: Notary uses Globally Unique Names (GUNs) to identify trusted collections.
 2. **Target**: Notary refers to the files the framework is used to distribute as "target files".  Target files are opaque to the framework. Whether target files are packages containing multiple files, single text files, or executable binaries is irrelevant to Notary.
-3. **Trusted Collection**: A trusted collection is a conceptual source of target files of interest to the application.
-4. **Key roles**:
+3. **Trusted Collection**: A trusted collection is a set of target files of interest to the application.
+4. **Notary Repository**: A notary repository refers to the set of signed metadata files that describe a trusted collection.
+5. **Key roles**:
     - **Root**: The root role delegates trust to specific keys trusted for all other top-level roles used in the system.
     - **Targets**: The targets role's signature indicates which target files are trusted by clients.
     - **Delegations**: the targets role can delegate full or partial trust to other roles.  Delegating trust means that the targets role indicates another role (that is, another set of keys and the threshold required for trust) is trusted to sign target file metadata.
@@ -32,8 +33,8 @@ Once you install the Notary CLI client, you can use it to manage your signing
 keys, authorize other team members to sign content, and rotate the keys if
 a private key has been compromised.
 
-When using the Notary CLI client you need to specify where is Notary server
-you want to communicate with with the `-s` flag, and where to store the private keys and cache for
+When using the Notary CLI client you need to specify where the URL of the Notary server
+you want to communicate is with the `-s` flag, and where to store the private keys and cache for
 the CLI client with the `-d` flag.  There are also fields in the [client configuration](/reference/client-config.md) to set these variables.
 
 ```bash
@@ -41,7 +42,7 @@ the CLI client with the `-d` flag.  There are also fields in the [client configu
 $ alias notary="notary -s <notary_server_url> -d <notary_cache_directory>
 ```
 
-When working Docker Content Trust, it is important to connect to Docker Hub's Notary server located at `https://notary.docker.io`, and specify notary's metadata cache as `~/.docker/trust`.
+When working Docker Content Trust, it is important to specify notary's client cache as `~/.docker/trust`.  Also, Docker Hub provides its own Notary server located at `https://notary.docker.io`, which contains trust data for many images including official images, though you are welcome to use your own notary server.
 
 ## Initializing a trusted collection
 
@@ -50,31 +51,24 @@ Notary can initialize a trusted collection with the `notary init` command:
 $ notary init <GUN>
 ```
 
-This command will generate targets and snapshot keys locally for the trusted collection, and try to locate a root key to use in the specified metadata cache (from `-d` or the config).
+This command will generate targets and snapshot keys locally for the trusted collection, and try to locate a root key to use in the specified notary client cache (from `-d` or the config).
 If notary cannot find a root key, it will generate one.  For all keys, notary will also prompt for a passphrase to encrypt the private key material at rest.
 
-If you'd like to initialize your trusted collection with a specific root key, there is a flag:
+If you'd like to initialize your trusted collection with a specific root key, there is a flag to provide it.  Notary will require that the key is encrypted:
 ```bash
 $ notary init <GUN> --rootkey <key_file>
 ```
 
-Note that you will have to run a publish after this command, as there are staged changes to initialize the trusted collection that have not been pushed to a ntoary server:
+Note that you will have to run a publish after this command for it to take effect, as there are staged changes to initialize the trusted collection that have not been pushed to a notary server.
 ```bash
 $ notary publish <GUN>
-```
-
-## Auto-publish changes
-
-Instead of manually running `notary publish` after each command, you can use the `-p` flag to auto-publish the changes from that command.
-For example:
-
-```bash
-$ notary init -p <GUN>
 ```
 
 ## Manage staged changes
 
 The Notary CLI client stages changes before publishing them to the server.
+These changes are staged locally in files such that the client can request the latest updates from the notary server, attempt to apply the staged changes on top of the new updates (like a `git rebase`), and then finally publish if the changes are still valid.
+
 You can manage the changes that are staged by running:
 
 ```bash
@@ -88,7 +82,7 @@ $ notary status <GUN> --unstage 0
 $ notary status <GUN> --reset
 ```
 
-Note that `<GUN>` can take on arbitrary values, but when working with Docker Content Trust they are structured like `<url>/<account>/<repository>`.  For example `docker.io/library/alpine` for the [Docker Library Alpine image](https://hub.docker.com/r/library/alpine).
+Note that `<GUN>` can take on arbitrary values, but when working with Docker Content Trust they are structured like `<domain>/<account>/<repository>`.  For example `docker.io/library/alpine` for the [Docker Library Alpine image](https://hub.docker.com/r/library/alpine).
 
 When you're ready to publish your changes to the Notary server, run:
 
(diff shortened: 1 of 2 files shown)


docs/command_reference.md
```diff
@@ -0,0 +1,235 @@
...
+weight=99
++++
+<![end-metadata]-->
+
+# Notary Command Reference
+
+## Terminology Reference
+1. **GUN**: Notary uses Globally Unique Names (GUNs) to identify trusted collections.
+2. **Target**: Notary refers to the files the framework is used to distribute as "target files".  Target files are opaque to the framework. Whether target files are packages containing multiple files, single text files, or executable binaries is irrelevant to Notary.
+3. **Trusted Collection**: A trusted collection is a conceptual source of target files of interest to the application.
```
cyli: I think I see what you mean by "source", but it may not be very clear since since notary doesn't necessarily identify the source of anything, and technically those files can be in disparate locations.  Maybe just "a list of target files"?  :|  I don't have a good suggestion.

cyli: Actually maybe I don't really know what a trusted collection means.  I interpreted it to be a collection of the targets in a notary repo, and maybe that's not what it actually means.  If it is, would it make sense to add `notary repository` here as a term and say how it's different than the trusted collection it represents?

docs/command_reference.md
```diff
@@ -0,0 +1,235 @@
...
+
+This command will generate targets and snapshot keys locally for the trusted collection, and try to locate a root key to use in the specified metadata cache (from `-d` or the config).
+If notary cannot find a root key, it will generate one.  For all keys, notary will also prompt for a passphrase to encrypt the private key material at rest.
+
+If you'd like to initialize your trusted collection with a specific root key, there is a flag:
+```bash
+$ notary init <GUN> --rootkey <key_file>
+```
+
+Note that you will have to run a publish after this command, as there are staged changes to initialize the trusted collection that have not been pushed to a ntoary server:
```
cyli: nitpick: "notary"

docs/command_reference.md
```diff
@@ -0,0 +1,235 @@
...
+content that was signed with those keys stop being trusted.
+
+For keys that are kept offline and managed by the Notary CLI client, such the
+keys with the root, targets, and snapshot roles, you can rotate them with:
+
+```bash
+$ notary key rotate <GUN> <key_role>
+```
+
+The Notary CLI client generates a new key for the role you specified, and
```
cyli: Would it make sense in this section to mention that `key rotate` replaces all existing keys on that role?

docs/command_reference.md
riyazdf: I think `a list of target files` makes more sense, I'll add that and `notary repository` as well 👍

docs/command_reference.md
riyazdf: Definitely 👍

cyli:
Would it make sense to add the add target/remove target references in this document too?  It's not very useful for docker users, but would be useful for manually signing things.  

Also, it might be useful to discuss the concept of staged changes and why notary does it and which types of commands produce staged changes.  I'm not sure if this should go into the  "manage staged changes" section or in other sections though, which is why I am adding an overall comment instead of a comment in a particular location.  :)

riyazdf:
@cyli agreed to both -- I initially left them out since they seemed basic enough from the getting started docs, but I think it'd be good to make this reference comprehensive.  I've also switched around the order of the first few sections and added a note about the "rebasing" notary does with local changelist files 👍

docs/command_reference.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Command Reference
+
+### Set up Notary CLI
+
+Once you install the Notary CLI client, you can use it to manage your signing
+keys, authorize other team members to sign content, and rotate the keys if
+a private key has been compromised.
+
+When using the Notary CLI client you need to specify where is Notary server
```
cyli: grammar nit:  maybe "the URL of the Notary server..."?

docs/command_reference.md
```diff
@@ -0,0 +1,279 @@
...
+When working Docker Content Trust, it is important to connect to Docker Hub's Notary server located at `https://notary.docker.io`, and specify notary's metadata cache as `~/.docker/trust`.
+
+## Initializing a trusted collection
+
+Notary can initialize a trusted collection with the `notary init` command:
+```bash
+$ notary init <GUN>
+```
+
+This command will generate targets and snapshot keys locally for the trusted collection, and try to locate a root key to use in the specified metadata cache (from `-d` or the config).
```
cyli: nit:  "notary client cache", since the keys are separate from the metadata cache specifically.

docs/command_reference.md
```diff
@@ -0,0 +1,279 @@
...
+
+Note that you will have to run a publish after this command, as there are staged changes to initialize the trusted collection that have not been pushed to a notary server.
+```bash
+$ notary publish <GUN>
+```
+
+## Manage staged changes
+
+The Notary CLI client stages changes before publishing them to the server.
+These chagnes are staged locally in files such that the client can request the latest updates from the notary server, attempt to rebase the staged changes against new updates, and then finally publish if the changes are still valid.
```
cyli: spelling nit: "changes"

docs/command_reference.md
```diff
@@ -0,0 +1,279 @@
...
+$ notary status <GUN>
+
+# Unstage a specific change
+$ notary status <GUN> --unstage 0
+
+# Alternatively, unstage all changes
+$ notary status <GUN> --reset
+```
+
+Note that `<GUN>` can take on arbitrary values, but when working with Docker Content Trust they are structured like `<url>/<account>/<repository>`.  For example `docker.io/library/alpine` for the [Docker Library Alpine image](https://hub.docker.com/r/library/alpine).
```
cyli: nitpick:  I'm not sure, but I think URLs generally include the protocol or scheme (e.g. "https").  Maybe for the GUN, `<url>` should be `domain` or `<registry>`?

docs/command_reference.md
```diff
@@ -0,0 +1,279 @@
...
+
+Users can sign content into a notary trusted collection by running:
+```bash
+$ notary add -p <GUN> <target_name> <target_file>
+```
+
+In the above command, the `<target_name>` corresponds to the name we want to associate the `<target_file>`'s signature with. Notary will sign the hash of the `<target_file>` into its trusted collection.
+Instead of adding a target by file, you can specify a hash directly:
+```bash
+$ notary addhash -p <GUN> <target_name> -sha256 <sha256Hash>
```
cyli: I think this also needs the size flag.

docs/command_reference.md
```diff
@@ -0,0 +1,279 @@
...
+```
+
+To check that your trust data was published successfully to the notary server, you can run:
+```bash
+$ notary list <GUN>
+```
+
+To remove targets from a trusted collection, you can run:
+```bash
+$ notary remove -p <GUN> <target_name>
```
cyli: You can also specify multiple roles to remove a target from - otherwise it defaults to only attempting to remove it from the `targets` role?

docs/command_reference.md
```diff
@@ -0,0 +1,279 @@
...
+keys with the root, targets, and snapshot roles, you can rotate them with:
+
+```bash
+$ notary key rotate <GUN> <key_role>
+```
+
+The Notary CLI client generates a new key for the role you specified, and
+prompts you for a passphrase to encrypt it.
+Then you're prompted for the passphrase for the key you're rotating, and if it
+is correct, the Notary CLI client contacts the Notary server to publish the
```
cyli: Non-blocking:  wondering if we should call out that Notary CLI auto-publishes immediately, no flag needed?

docs/command_reference.md
cyli: Oh oops, the multi-role stuff is below.  :)  Sorry - it might be good to include `notary remove -p <GUN> <target_name> --roles ...` down there too

docs/command_reference.md
```diff
@@ -0,0 +1,279 @@
...
+
+# Notary Command Reference
+
+## Terminology Reference
+1. **GUN**: Notary uses Globally Unique Names (GUNs) to identify trusted collections.
+2. **Target**: Notary refers to the files the framework is used to distribute as "target files".  Target files are opaque to the framework. Whether target files are packages containing multiple files, single text files, or executable binaries is irrelevant to Notary.
+3. **Trusted Collection**: A trusted collection is a set of target files of interest to the application.
+4. **Notary Repository**: A notary repository refers to the set of signed metadata files that describe a trusted collection.
+5. **Key roles**:
+    - **Root**: The root role delegates trust to specific keys trusted for all other top-level roles used in the system.
```
HuKeping: Does it make sense to link the top-level to the [relevant part of TUF spec](https://github.com/theupdateframework/tuf/blob/develop/docs/tuf-spec.txt#L246)

docs/command_reference.md
```diff
@@ -0,0 +1,279 @@
...
+When using the Notary CLI client you need to specify where is Notary server
+you want to communicate with with the `-s` flag, and where to store the private keys and cache for
+the CLI client with the `-d` flag.  There are also fields in the [client configuration](/reference/client-config.md) to set these variables.
+
+```bash
+# Create an alias to always have the notary client talking to the right notary server
+$ alias notary="notary -s <notary_server_url> -d <notary_cache_directory>
+```
+
+When working Docker Content Trust, it is important to connect to Docker Hub's Notary server located at `https://notary.docker.io`, and specify notary's metadata cache as `~/.docker/trust`.
```
HuKeping: The `https:/notary.docker.io` here is not very accurate, if one enabled the DCT and set up his own Notary service, what he need to specify is the URL of his notary server(and it might be the most instance).

docs/command_reference.md
```diff
@@ -0,0 +1,279 @@
...
+
+Notary can initialize a trusted collection with the `notary init` command:
+```bash
+$ notary init <GUN>
+```
+
+This command will generate targets and snapshot keys locally for the trusted collection, and try to locate a root key to use in the specified metadata cache (from `-d` or the config).
+If notary cannot find a root key, it will generate one.  For all keys, notary will also prompt for a passphrase to encrypt the private key material at rest.
+
+If you'd like to initialize your trusted collection with a specific root key, there is a flag:
```
HuKeping: I think there should be more detail about the customized root key, for example if people use `openssl` to generate a key, that key might not be acceptable by this command.

docs/command_reference.md
```diff
@@ -0,0 +1,279 @@
...
+```bash
+$ notary init <GUN> --rootkey <key_file>
+```
+
+Note that you will have to run a publish after this command, as there are staged changes to initialize the trusted collection that have not been pushed to a notary server.
```
HuKeping: maybe "Note that you will have to run a publish after this command to make it efficient"

docs/command_reference.md
HuKeping: Does it make sense to add a note for people that it was like a `git rebase`?

docs/command_reference.md
HuKeping: Aha, it was `<account>`, we always considered it as `<namespace>` :)

docs/command_reference.md
```diff
@@ -0,0 +1,279 @@
...
+```bash
+$ notary init -p <GUN>
+```
+
+
+## Adding and removing trust data
+
+Users can sign content into a notary trusted collection by running:
+```bash
+$ notary add -p <GUN> <target_name> <target_file>
```
HuKeping: Since `-p` is not required, would you mind if we just use `notary add <GUN> <target_name> <target_file>` here?

docs/command_reference.md
```diff
@@ -0,0 +1,279 @@
...
+
+The Notary CLI client generates a new key for the role you specified, and
+prompts you for a passphrase to encrypt it.
+Then you're prompted for the passphrase for the key you're rotating, and if it
+is correct, the Notary CLI client contacts the Notary server to publish the
+change.
+
+After a rotation, all previously existing keys for the specified role are replaced with the new key.
+
+You can also rotate keys that are stored in the Notary server, such as the keys
```
HuKeping: Non-blocking: "You can .... Notary server with flag `-r`" ?

docs/command_reference.md
```diff
@@ -0,0 +1,279 @@
...
+$ notary key export --gun <GUN1> --gun <GUN2> -o exported_keys.pem
+```
+When exporting multiple keys, all keys are outputted to a single PEM file in individual blocks. If the output flag `-o` is omitted, the PEM blocks are outputted to STDOUT.
+
+## Manage keys for delegation roles
+
+To delegate content signing to other users without sharing the targets key, retrieve a x509 certificate for that user and run:
+
+```bash
+$ notary delegation add -p <GUN> targets/<role> --all-paths user.pem user2.pem
```
HuKeping: I suppose the `--all-paths` should be an option?

docs/command_reference.md
```diff
@@ -0,0 +1,279 @@
...
+```
+
+The user can then import the private key for that certificate keypair (using `notary key import`) and use it for signing.
```
HuKeping: Maybe "The delegated user" will be more clear

docs/command_reference.md
```diff
@@ -0,0 +1,279 @@
...
+
+It's possible to add multiple certificates at once for a role:
+```bash
+$ notary delegation add -p <GUN> targets/<role> --all-paths user1.pem user2.pem user3.pem
+```
+
+You can also remove keys from a delegation role:
```
HuKeping: Should we also describe the consequence it brings?

The delegation role will not allowed to sign the content any more if I understand right.

docs/command_reference.md
```diff
@@ -0,0 +1,279 @@
...
+
+# Alternatively, you can remove keys from all delegation roles
+$ notary delegation purge <GUN> --key <keyID1> --key <keyID2>
+```
+
+## Managing targets in delegation roles
+
+We can specify to sign content into specific delegation roles by specifying the `--roles` flag.  This also applies to `notary addhash` and `notary remove`:
+```bash
+$ notary add -p <GUN> <target_name> <target_file> --roles targets/<role>
```
HuKeping: The flag `-p` is not required.

docs/command_reference.md
riyazdf: Yeah I put the multi-role stuff below because I hadn't introduced delegations yet, I'll add a `notary remove --roles` example too

docs/command_reference.md
riyazdf: I was trying to capture the general case, but I'll make sure to specify this is for images specifically wanting to use the notary server deployed alongside hub.docker.io

docs/command_reference.md
riyazdf: I'd rather not link to specific parts of the TUF spec in too many places in the command reference, since I feel like that's more relevant to the other documentation that explains more of the TUF concepts.  Do you think just the general TUF spec link I have below is enough?

docs/command_reference.md
riyazdf: I wanted to keep `-p` in commands after introducing it in the auto-publish section so that all of the commands will complete the full publish cycle.  I'll make a note in the previous section about how it'll be included going forward for commands that allow it (and that it is optional)

docs/command_reference.md
riyazdf: I think it's the common use case so I used that, but I'll make a note about it and `--paths`

docs/command_reference.md
riyazdf: @HuKeping great idea!

docs/command_reference.md
HuKeping: That make sense :+1:

docs/command_reference.md
HuKeping: agree, two many link doesn't make sense

docs/command_reference.md
```diff
@@ -0,0 +1,299 @@
...
+parent="mn_notary"
+weight=99
++++
+<![end-metadata]-->
+
+# Notary Command Reference
+
+## Terminology Reference
+1. **GUN**: Notary uses Globally Unique Names (GUNs) to identify trusted collections.
+2. **Target**: Notary refers to the files the framework is used to distribute as "target files".  Target files are opaque to the framework. Whether target files are packages containing multiple files, single text files, or executable binaries is irrelevant to Notary.
```
cyli: nitpick: "refers to the files the framework uses"

docs/command_reference.md
```diff
@@ -0,0 +1,299 @@
...
+3. **Trusted Collection**: A trusted collection is a set of target files of interest to the application.
```
cyli: nitpick:  By application do you mean application other than notary, or notary itself?  (it might be good to clarify this)

docs/command_reference.md
```diff
@@ -0,0 +1,299 @@
...
+
+This command will generate targets and snapshot keys locally for the trusted collection, and try to locate a root key to use in the specified notary client cache (from `-d` or the config).
+If notary cannot find a root key, it will generate one.  For all keys, notary will also prompt for a passphrase to encrypt the private key material at rest.
+
+If you'd like to initialize your trusted collection with a specific root key, there is a flag to provide it.  Notary will require that the key is encrypted:
+```bash
+$ notary init <GUN> --rootkey <key_file>
+```
+
+Note that you will have to run a publish after this command for it to take effect, as there are staged changes to initialize the trusted collection that have not been pushed to a notary server.
```
cyli: nitpick:  "as it creates staged changes..." might be clearer about where those staged changes came from.

docs/command_reference.md
```diff
@@ -0,0 +1,299 @@
...
+$ notary status <GUN>
+
+# Unstage a specific change
+$ notary status <GUN> --unstage 0
+
+# Alternatively, unstage all changes
+$ notary status <GUN> --reset
+```
+
+Note that `<GUN>` can take on arbitrary values, but when working with Docker Content Trust they are structured like `<domain>/<account>/<repository>`.  For example `docker.io/library/alpine` for the [Docker Library Alpine image](https://hub.docker.com/r/library/alpine).
```
cyli: Non-blocking:  should the description of the GUN be moved to the "terminology section", or alternately the `notary init` section where it is next mentioned?

docs/command_reference.md
```diff
@@ -0,0 +1,299 @@
...
+The remainder of this reference will include the `-p` auto-publish flag where it can be used, though it is optional and can be replaced with following each command with a `notary publish`.
+
+## Adding and removing trust data
+
+Users can sign content into a notary trusted collection by running:
+```bash
+$ notary add -p <GUN> <target_name> <target_file>
+```
+
+In the above command, the `<target_name>` corresponds to the name we want to associate the `<target_file>`'s signature with. Notary will sign the hash of the `<target_file>` into its trusted collection.
```
cyli: Non-blocking:  should we use the word "signature" here, since the rest of notary is about signing things?  Since we're not specifically signing this `<target_file>` itself, maybe something like "`<target_name>` corresponds to the name we want `<target_file>` to be known as in the trusted collection"?

docs/command_reference.md
```diff
@@ -0,0 +1,299 @@
...
+# Remove the given keys from a delegation role
+$ notary delegation remove -p <GUN> targets/<role> <keyID1> <keyID2>
+
+# Alternatively, you can remove keys from all delegation roles
+$ notary delegation purge <GUN> --key <keyID1> --key <keyID2>
+```
+
+## Managing targets in delegation roles
+
+We can specify to sign content into specific delegation roles by specifying the `--roles` flag.  This also applies to `notary addhash` and `notary remove`.
```
cyli: nitpick: maybe "we can specify which delegation roles to sign content into by using the `--roles` flag"?

docs/command_reference.md
```diff
@@ -0,0 +1,299 @@
...
+Without the `--roles` flag, notary will attempt to operate on the base `targets` role:
+```bash
+# Add content from a target file to a specific delegation role
+$ notary add -p <GUN> <target_name> <target_file> --roles targets/<role>
+
+# Add content by hash to a specific delegation role
+$ notary addhash -p <GUN> <target_name> <byte_size> --sha256 <sha256Hash> --roles targets/<role>
+
+# Remove content from a specific delegation role
+$ notary add -p <GUN> <target_name> <target_file> --roles targets/<role>
```
cyli: `notary remove`

cyli:
Thanks for working on this @riyazdf!  This will be a super helpful reference  :)  I have a couple of nitpicks above, but other than that, LGTM!

docs/command_reference.md
riyazdf: Good idea, I'll move it to terminology

docs/command_reference.md
riyazdf: Good point, I'll update this

docs/command_reference.md
```diff
@@ -0,0 +1,297 @@
...
+The remainder of this reference will include the `-p` auto-publish flag where it can be used, though it is optional and can be replaced with following each command with a `notary publish`.
+
+## Adding and removing trust data
+
+Users can sign content into a notary trusted collection by running:
+```bash
+$ notary add -p <GUN> <target_name> <target_file>
+```
+
+In the above command, the `<target_name>` corresponds to the name we want to associate the `<target_file>` known as in the trusted collection. Notary will sign the hash of the `<target_file>` into its trusted collection.
```
cyli: nitpick: "the name we want `<target_file>` to be known as in the trusted collection", or alternately "the name we want to associate the `<target file>` with in the trusted collection"  :)

docs/command_reference.md
```diff
@@ -0,0 +1,297 @@
...
+parent="mn_notary"
+weight=99
++++
+<![end-metadata]-->
+
+# Notary Command Reference
+
+## Terminology Reference
+1. **GUN**: Notary uses Globally Unique Names (GUNs) to identify trusted collections.  Note that `<GUN>` can take on arbitrary values, but when working with Docker Content Trust they are structured like `<domain>/<account>/<repository>`.  For example `docker.io/library/alpine` for the [Docker Library Alpine image](https://hub.docker.com/r/library/alpine).
+2. **Target**: Notary refers to the files the framework uses to distribute as "target files".  Target files are opaque to the framework. Whether target files are packages containing multiple files, single text files, or executable binaries is irrelevant to Notary.
```
endophage: First sentence reads poorly.

docs/command_reference.md
```diff
@@ -0,0 +1,297 @@
...
+```
+
+## Importing and exporting keys
+
+Notary can import keys that are already in a PEM format:
+```bash
+$ notary key import <pemfile> --role <key_role> --gun <key_gun>
+```
+
+The `--role` and `--gun` flags can be provided to specify a role and GUN to import the key to, which overrides any information contained in PEM headers.
```
endophage: I don't think we implemented it as override. They fill in blank data but don't override existing data afaik.

docs/command_reference.md
```diff
@@ -0,0 +1,297 @@
...
+Note that for root and delegation keys, the `--gun` flag is ignored because these keys can be shared across GUNs.
+If no `--role` or `--gun` is given, notary will assume that the key is to be used for a delegation role.
```
endophage: Might want to note that it will be available for signing regardless of role, but will be shown as a delegation key in commands such as `notary key list`

docs/command_reference.md
```diff
@@ -0,0 +1,297 @@
...
+$ notary delegation add -p <GUN> targets/<role> --all-paths user1.pem user2.pem user3.pem
+```
+
+You can also remove keys from a delegation role, such that those keys can no longer sign targets into the delegation role:
+
+```bash
+# Remove the given keys from a delegation role
+$ notary delegation remove -p <GUN> targets/<role> <keyID1> <keyID2>
+
+# Alternatively, you can remove keys from all delegation roles
```
endophage: Include the use case in which you might want to use this option: delegation key compromise.

docs/command_reference.md
```diff
@@ -0,0 +1,297 @@
...
+Usually most problems are fixed by ensuring you're communicating with the
+correct Notary server, using the `-s` flag, and that you're using the correct
+directory where your private keys are stored, with the `-d` flag.
+
+If you are receiving this error:
+```bash
+* fatal: Get <URL>/v2/: x509: certificate signed by unknown authority
+```
+The Notary CLI must be configured to trust the root certificate authority of the server it is communicating with.
+This can be configured by specifying the root certificate with the `--tlscacert` flag or by editing the Notary client configuration file.
```
endophage: or by adding the CA to your system CAs

docs/command_reference.md
riyazdf: oops you're right, I must've been thinking of something else 👍

endophage:
LGTM. Merge on green!

PR #948, opened 2016-09-09
Statically compile linux binaries, and try to fix OS X binaries

### LINKED ISSUE #826: Signing issues on OSX
Due to https://github.com/golang/go/issues/11887 the notary binary, when signed, will not run on OSX. We currently workaround the issue by rebuilding the application with striped dwarf symbols:

https://github.com/djs55/notary/commit/31e4b19260d57f0f59b596bb38aa4a6d5075be1a

Would be great to have a proper solution integrated upstream.

/cc @justincormack and @djs55

cyli:
Modify the cross-compile script to statically compile the linux binary (which includes libtool).

This hopefully compiles the OS X binary in such a way as to address #826.

### DIFF
--- a/buildscripts/cross.sh
+++ b/buildscripts/cross.sh
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
 
+# This script cross-compiles static (when possible) binaries for supported OS's
+# architectures.  The Linux binary is completely static, whereas Mac OS binary
+# has libtool statically linked in. but is otherwise not static because you
+# cannot statically link to system libraries in Mac OS.
+
 GOARCH="amd64"
 
 if [[ "${NOTARY_BUILDTAGS}" == *pkcs11* ]]; then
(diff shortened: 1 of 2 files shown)


buildscripts/cross.sh
```diff
@@ -15,19 +20,23 @@ for os in "$@"; do
 	if [[ "${GOOS}" == "darwin" ]]; then
 		export CC="o64-clang"
 		export CXX="o64-clang++"
-		# -ldflags=-s:  see https://github.com/golang/go/issues/11994
-		export LDFLAGS="${GO_LDFLAGS} -ldflags=-s"
+		# -ldflags=-s:  see https://github.com/golang/go/issues/11994 - TODO: this has been fixed in go 1.7.1
```
riyazdf: Not sure if I understand if this is fixed by #943?  Do we still need `LDFLAGS="-s"`?

cyli: It could be - I'd need to verify with @justincormack.

cyli:
Ok, checked with @justincormack, and the mac binary generated with `make cross` seems to be signable, so this should be able to produce release binaries that would work.

riyazdf:
Gave this a spin and worked great!  LGTM

buildscripts/cross.sh
justincormack: Seems to be fine without with 1.7.1

endophage:
I'm as uncertain on this as everyone else but it looks right based on everything else I've seen. LGTM

endophage:
One question before merging, would it obsolete the need for this change if we updated to Go 1.7.1? (and is that easy to do?)

cyli:
@endophage not all of this change, because this change involves passing the extra static and CTIME flags which we didn't correctly do the last time.  But it would obsolete the `-s` part, and yes that is easy to do

PR #951, opened 2016-09-12
Update expiry of fixture certs to at least 2 years in the future

cyli:


riyazdf:
LGTM on green!

cyli:
Updated this to do another 3 things:

1. use a sha256 signing algorithm instead of whatever is default (which may be SHA1 on older versions of openssl, which we reject)
1. use a different CN for the intermediate cert so we can more easily tell what signed what
1. also generate 2 other certs that are in the fixtures library

riyazdf:
thanks for the updates!  LGTM pending CI :)

endophage:
LGTM

PR #952, opened 2016-09-13
Try to fix flakey integration tests, also update the definition of notary server health

cyli:
Integration tests have been failing lately - it seems it's because the server comes up before the signer, and the integration tests start before the server manages to connect to the signer.

This fix should hopefully cause the tests to wait until the server is healthy for sure, so they might take longer to start.

This also changes the definition of notary server health - previously it was healthy but degraded if the signer is down - not it's just unhealthy, because we cannot do any writes.  

An alternate approach is to just have a separate endpoint to see whether it runs in degraded mode, or use a different health package that supports "warning" or degraded mode without necessarily failing the health check entirely.


riyazdf:
I'm good with this new "healthy" definition, thanks for updating this :)

LGTM

endophage:
This all LGTM!

PR #963, opened 2016-09-19
Fix confusing naming

HuKeping:

### DIFF
--- a/client/changelist/change.go
+++ b/client/changelist/change.go
@@ -20,7 +20,7 @@ const (
 // all changes in Snapshot and Timestamp are programmatically
 // generated base on Root and Targets changes.
 const (
-	TypeRootRole          = "role"
+	TypeBaseRole          = "role"
 	TypeTargetsTarget     = "target"
 	TypeTargetsDelegation = "delegation"
 	TypeWitness           = "witness"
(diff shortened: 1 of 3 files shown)


client/changelist/change.go
```diff
@@ -20,7 +20,7 @@ const (
 // all changes in Snapshot and Timestamp are programmatically
 // generated base on Root and Targets changes.
 const (
-	TypeRootRole          = "role"
+	TypeRootRole          = "root"
```
ecordell: I agree that this is confusing naming, but if you look where this constant is used it looks like this is referring to "root role" as in "not-delegated" e.g. `root`, `snapshot`, `target`, and `timestamp`.

E.g. see https://github.com/docker/notary/blob/b4d4ab6c96e19c9f679f59270130966eb42fbec9/client/helpers.go#L176 , where it refers to any "base" role change. 

I think it makes more sense to change the constant to `TypeRole` or maybe `TypeBaseRole`

endophage: @ecordell is correct on the use case. Some more context, it's used specifically to describe a change type that modifies a base role, as opposed to a delegation role. Probably `TypeBaseRole` makes the most sense, to make it clear it's not to be used for delegation roles.

HuKeping: Thanks @ecordell @endophage! I'll check it later!

HuKeping: Updated :)

PR #965, opened 2016-09-21
Allow non-cert public keys

ecordell:
This allows non-cert PEM encoded public keys to be loaded as delegation keys

### DIFF
--- a/trustpinning/certs_test.go
+++ b/trustpinning/certs_test.go
@@ -193,11 +193,7 @@ func TestValidateRootWithPinnedCert(t *testing.T) {
 func TestValidateRootWithPinnedCertAndIntermediates(t *testing.T) {
 	now := time.Now()
 	serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), 128)
-
-	pass := func(keyName, alias string, createNew bool, attempts int) (passphrase string, giveup bool, err error) {
-		return "password", false, nil
-	}
-	memStore := trustmanager.NewKeyMemoryStore(pass)
+	memStore := trustmanager.NewKeyMemoryStore(passphraseRetriever)
 	cs := cryptoservice.NewCryptoService(memStore)
 
 	// generate CA cert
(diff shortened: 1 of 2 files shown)


tuf/utils/x509.go
```diff
@@ -252,9 +252,31 @@ func ParsePEMPublicKey(pubKeyBytes []byte) (data.PublicKey, error) {
...
+		return nil, fmt.Errorf("unsupported PEM block type %q, expected CERTIFICATE or PUBLIC KEY", pemBlock.Type)
+	}
+}
+
+func keyTypeForPublicKey(pubKeyBytes []byte) (string, error) {
+	pub, err := x509.ParsePKIXPublicKey(pubKeyBytes)
+	if err != nil {
+		return "", fmt.Errorf("unable to parse pem encoded public key: %v", err)
+	}
+	_, ok := pub.(*ecdsa.PublicKey)
```
endophage: Use a type switch for clarity and ease of adding other key types in the future:
```
switch pub.(type) {
case *ecdsa.PublicKey:
    return data.ECDSAKey, nil
case *rsa.PublicKey:
    return data.RSAKey, nil
}
```

tuf/utils/x509.go
```diff
@@ -252,9 +252,31 @@ func ParsePEMPublicKey(pubKeyBytes []byte) (data.PublicKey, error) {
...
+	if ok {
+		return data.ECDSAKey, nil
+	}
+	_, ok = pub.(*rsa.PublicKey)
+	if ok {
+		return data.RSAKey, nil
 	}
+	return "", fmt.Errorf("unkown public key format")
```
endophage: I think we probably have a relevant error type for this somewhere.

tuf/utils/x509.go
```diff
@@ -252,9 +252,31 @@ func ParsePEMPublicKey(pubKeyBytes []byte) (data.PublicKey, error) {
 			return nil, fmt.Errorf("invalid certificate: %v", err)
 		}
 		return CertToKey(cert), nil
+	case "PUBLIC KEY":
+		keyType, err := keyTypeForPublicKey(pemBlock.Bytes)
```
endophage: Would it make sense to move this entire case block into a new constructor for `data.PublicKey` objects that takes a `[]byte` (in this case `pemBlock.Bytes`) and returns `(PublicKey, error)`?

trustpinning/certs_test.go
```diff
@@ -814,6 +814,52 @@ func generateRootKeyIDs(r *data.SignedRoot) {
 	}
 }
 
+func TestParsePEMPublicKey(t *testing.T) {
+	gun := "notary"
+	pass := func(keyName, alias string, createNew bool, attempts int) (passphrase string, giveup bool, err error) {
```
cyli: Non-blocking:  I think you can use `passphrase.ConstantRetriever("password")` for this, unless you don't want the extra import.

tuf/utils/x509.go
cyli: Non-blocking:  We seem to support ED25519 for `ParsePEMPrivateKey`, although I guess we concatenate the private and public parts.  I don't think it violates any existing spec to PEM-encode the public key, although I know it's not not supported by openssl or other software.  Would it make sense to either support that here, or remove support in `ParsePEMPrivateKey`?

tuf/utils/x509.go
ecordell: That seems like a good refactor to me; I assume the same for `ParsePEMPrivateKey`?

I was going to bring this up elsewhere but it seems relevant here: I think it's confusing that if you pass in a cert to `ParsePEMPrivateKey` you get back a `data.PublicKey` object containing the cert bytes in the `Public` field instead of the parsed out public key bytes.

I was planning to read through a bit more before suggesting a change, but personally I like the idea of a flow like: certs get loaded/verified -> translated to plain keys -> tuf only deals with plain keys. 

This seems like a very clean separation to me; the only issue would be that you would lose the property that certs get verified right before they're used every time. Perhaps it would make sense, after loading the certs into memory, to keep around a map of `parsedKeys -> originatingCerts` so that at various well-defined points during verification, you can do something like `reverifyCerts(currentlyReleventKeyList)` to find the certs you're using and check them again.

Just a couple of thoughts I've had on making the flow of notary easier to understand. I've just rambled a bit too much for an issue comment, so I'll pull this out into an issue later today...

tuf/utils/x509.go
ecordell: @endophage Also a good review note - not sure why I didn't do that originally.

tuf/utils/x509.go
endophage: I agree it's a little messy and would be good to clean up. Let's discuss how to approach it in the issue you file.

tuf/utils/x509.go
ecordell: I couldn't find one but I'm happy to switch it if you can point me to it.

ecordell:
Addressed review comments. 

@endophage held off on public key refactoring because of https://github.com/docker/notary/issues/977 

@cyli Didn't end up adding ed25519 support because the x509 library doesn't recognize it as a valid PEM public key type. If it's something that should be supported can always wrap the x509 lib and add ed25519 support.

cyli:
jenkins, test this please

trustpinning/certs_test.go
```diff
@@ -1007,12 +1003,79 @@ func generateExpiredTestingCertificate(rootKey data.PrivateKey, gun string) (*x5
 	return cryptoservice.GenerateCertificate(rootKey, gun, startTime, startTime.AddDate(1, 0, 0))
 }
 
+// Helper function for explicitly generating key IDs and unexported fields for equality testing
+func generateRootKeyIDs(r *data.SignedRoot) {
```
cyli: I'm not sure if this function is still used anywhere - should it be removed?  Although you do bring up a good point;  it would be good to useful to have equality test functions for some of our data structures that have unexported fields at some point.

cyli:
jenkins, test this please

cyli:
(Hmm... you can't edit the text of your previous review status comment :P  sorry for the spam!)

trustpinning/certs_test.go
```diff
@@ -1007,12 +1003,70 @@ func generateExpiredTestingCertificate(rootKey data.PrivateKey, gun string) (*x5
...
+		Type:    "PUBLIC KEY",
+		Headers: nil,
+		Bytes:   rsaPubKey.Public(),
+	})
+	_, err = utils.ParsePEMPublicKey(rsaPemBytes)
+	require.NoError(t, err)
+
+	// unsupported key type
+	unsupportedPemBytes := pem.EncodeToMemory(&pem.Block{
+		Type:    "UNSUPORTED KEY",
```
riyazdf: nit typo: `UNSUPPORTED`

riyazdf: I think this case errors because `pem.Decode` fails, not because of the type-switch.

This case is great, but It would be awesome if we could add another test case with a valid type that `ParsePEMPublicKey` does not support, like `CERTIFICATE REQUEST` or `PRIVATE KEY`

tuf/utils/x509.go
endophage: In `tuf/signed/errors.go` we have `ErrInvalidKeyType` which seems appropriate. It's used to indicate it the key type is invalid/unrecognized in the context of key types Notary accepts, not necessarily that the key is universally invalid.

trustpinning/certs_test.go
ecordell: Updated!

cyli:
jenkins, test this please

riyazdf:
thanks @ecordell for updating this, glad to have this feature :)

Issue #551, opened 2016-02-05
Build notary on Windows


michaKFromParis:

friism:
You should be able to cross-compile to Windows (eg. on CircleCI), all the other projects do that.

@mikedougherty is working on getting Windows CI running on the Jenkins setup: https://github.com/docker/engine-api/pull/136

dgageot:
ping @docker/core-notary-maintainers Is it something we can expect?

dmp42:
bump @docker/core-notary-maintainers :)

endophage:
We should be able to get this working without too much problem if we don't include yubikey support. I don't have a good feel for what it will be like getting yubikey support included on windows.

It's not scheduled in the 0.4.0 timeframe but we can look at it after that release is cut and release a first windows build at 0.4.1.

dmp42:
Thanks @endophage - what would be the approximate release date for 0.4 & 0.4.1 respectively?

endophage:
0.4.0 should get an RC cut sometime around mid-August but is dependent on other releases happening around then. If they are delayed, Notary will be as well.

We don't have a schedule for 0.4.1 yet. It would likely be a release just to add Windows builds.

riyazdf:
been poking around with @endophage, and found some issues that we're now fixing.  Reporting back for tracking purposes and fixing:

- [ ] `notary.NotarySupportedSignals` is not Windows-compatible (`syscall.SIGUSR1` and `syscall.SIGUSR2` do not exist on windows)
- [ ] Generally using `filepath.Join` wherever possible due to `/` vs `\` discrepancies
- [ ] Relative path logic is more complicated (`filepath.IsAbs` checks for drive letters) 
- [ ] Similarly, in tests, `ioutil.TempDir("/tmp", ...)` is not Windows-compatible because Windows does not have a `/tmp`
- [ ] Changelist ordering from the filesystem does not seem to be respected, investigating further

endophage:
As of 0.4.2 we started releasing windows builds.

PR #970, opened 2016-09-22
Do not support SIGUSR1 and SIGUSR2 syscall handling in windows

### LINKED ISSUE #791: Increase/decrease logging level on notary-signer
For people who would like to try some contribute to Notary :)

For now we have already supported increasing/decreasing logging level on notary server side #748 , and it would be nice to have that functionality also on notary signer side.

Tips: 
What you need to do would like to be:
- Embed  a `setupSignalTrap()` into the notary-signer booting process.
- Add a handler to deal with those captured signals.
- Add some test for your handler
- Add some docs about it  

The infra you might want to use for this feature are:
- `NotarySupportedSignals ` in `const.go`
- `AdjustLogLevel()` to increase/decrease logging in `utils/configuration.go`

cyli:
Since they don't exist in Windows.  While I was at it, since I had to create a bunch of windows-specific and non-windows-specific files, I moved the signal handling code to `utils` so that it can also be used by the signer (that way we don't have windows/non-windows files for both server and signer).

Technically I think the only change that would have been necessary to make the windows client build is the `const.go` change, and I don't think we care about supporting windows for server/signer, but the server and signer changes are added for completeness (and if we want to run full unit tests in windows).

Unfortunately we don't have any windows CI systems yet.  :|  Although I did add at least cross-compiling to windows to our CI runs.

Addresses part of #551.
Fixes #791.

Note: possibly we want to rebase this to 0.4.0 so that it can be added to 0.4.1

### DIFF
--- a/const_windows.go
+++ b/const_windows.go
@@ -1,0 +1,8 @@
+// +build windows
+
+package notary
+
+import "os"
+
+// NotarySupportedSignals does not contain any signals, because SIGUSR1/2 are not supported on windows
+var NotarySupportedSignals = []os.Signal{}
--- a/docs/reference/signer-config.md
+++ b/docs/reference/signer-config.md
@@ -210,6 +210,44 @@ The environment variables for the older passwords are optional, but Notary
 Signer will not be able to decrypt older keys if they are not provided, and
 attempts to sign data using those keys will fail.
 
+## Hot logging level reload
+We don't support completely reloading notary signer configuration files yet at present. What we support for Linux and OSX now is:
+- increase logging level by signaling `SIGUSR1`
+- decrease logging level by signaling `SIGUSR2`
+
+No signals and no dynamic logging level changes are supported for Windows yet.
+
+Example:
+
+To increase logging level
+```
+$ kill -s SIGUSR1 PID
+
+or
+
+$ docker exec -i CONTAINER_ID kill -s SIGUSR1 PID
+```
+
+To decrease logging level
+```
+$ kill -s SIGUSR2 PID
+
+or
+
+$ docker exec -i CONTAINER_ID kill -s SIGUSR2 PID
+```
+PID is the process id of `notary-signer` and it may not the PID 1 process if you are running
+the container with some kind of wrapper startup script or something.
+
+You can get the PID of `notary-signer` through
+```
+$ docker exec CONTAINER_ID ps aux
+
+or
+
+$ ps aux | grep "notary-signer -config" | grep -v "grep"
+```
+
 
 ## Related information
 
--- a/utils/configuration.go
+++ b/utils/configuration.go
@@ -5,6 +5,8 @@ package utils
 import (
 	"crypto/tls"
 	"fmt"
+	"os"
+	"os/signal"
 	"path/filepath"
 	"strings"
 
@@ -243,4 +245,21 @@ func AdjustLogLevel(increment bool) error {
 
 	logrus.SetLevel(lvl)
 	return nil
+}
+
+// SetupSignalTrap is a utility to trap supported signals hand handle them (currently by increasing logging)
+func SetupSignalTrap(handler func(os.Signal)) chan os.Signal {
+	if len(notary.NotarySupportedSignals) == 0 {
+		return nil
+
+	}
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, notary.NotarySupportedSignals...)
+	go func() {
+		for {
+			handler(<-c)
+		}
+	}()
+
+	return c
 }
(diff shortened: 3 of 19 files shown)


HuKeping:
Haven't tried notary on Windows, but changes LGTM

docs/reference/signer-config.md
```diff
@@ -210,6 +210,42 @@ The environment variables for the older passwords are optional, but Notary
 Signer will not be able to decrypt older keys if they are not provided, and
 attempts to sign data using those keys will fail.
 
+## Hot logging level reload
+We don't support completely reloading notary signer configuration files yet at present. What we support for now is:
```
riyazdf: It might be worth making a note about how this is not supported in Windows

utils/configuration.go
```diff
@@ -244,3 +246,20 @@ func AdjustLogLevel(increment bool) error {
 	logrus.SetLevel(lvl)
 	return nil
 }
+
+// SetupSignalTrap is a utility to trap supported signals hand handle them (currently by increasing logging)
+func SetupSignalTrap(handler func(os.Signal)) chan os.Signal {
+	if len(notary.NotarySupportedSignals) == 0 {
```
riyazdf: non-blocking-nit: can we test this case with some mocking?  I'm fine with assigning this const to empty for the test and deferring to return it back to its original value, to mock what Windows would do

cyli: 👍 I was actually going to see if I could try to set up CI for windows (not sure if we want to do it in this PR) so that we can get coverage numbers for it, in which case this should be tested. I'll follow up on this later today and if it proves to be a ton of work I'll stub out the const instead. :)

const_windows.go
```diff
@@ -0,0 +1,8 @@
+// +build windows
```
endophage: Fine with leaving it in for explicit-ness but for my own understanding, we don't technically need it when using the `*_windows.go` filename right?

cyli: yep, not needed.  I just like having the labels since we also have a `nowindows` that matches against everything else.

utils/configuration_nowindows.go
```diff
@@ -0,0 +1,29 @@
...
+// LogLevelSignalHandle will increase/decrease the logging level via the signal we get.
+func LogLevelSignalHandle(sig os.Signal) {
+	switch sig {
+	case syscall.SIGUSR1:
+		if err := AdjustLogLevel(true); err != nil {
+			fmt.Printf("Attempt to increase log level failed, will remain at %s level, error: %s\n", logrus.GetLevel(), err)
+			return
+		}
+	case syscall.SIGUSR2:
+		if err := AdjustLogLevel(false); err != nil {
```
endophage: Curious why the error case for SIGUSR1 is being covered but this one isn't. Can't work out why that would be that case.

riyazdf: I think it's this line, when it goes from `debug` --> `debug`: https://github.com/docker/notary/pull/970/files#diff-d39909358e7e39607c560d5d0fad0823R23

endophage:
Can run the client tests and build+run notary client on windows 👍

riyazdf:
LGTM on @endophage's commit, so overall LGTM from me 👍

utils/configuration.go
cyli: Ok, I think we can probably set this up. In case it goes horribly badly and doesn't work, I've added a mocking test for this in a commit we can revert once we have the CI set up.

docs/reference/signer-config.md
cyli: Great idea, done.  :)

utils/configuration_nowindows.go
cyli: Hopefully I've fixed this so we test the same set of log level adjustments for both `AdjustLogLevel` and the signal handler.  We'll see what coverage says.

utils/configuration_test.go
```diff
@@ -560,3 +560,18 @@ func TestSetSignalTrap(t *testing.T) {
...
+// a windows CI system running
+func TestSetSignalTrap(t *testing.T) {
+	testSetSignalTrap(t)
+}
+
+func TestSetSignalTrapMockWindows(t *testing.T) {
+	old := notary.NotarySupportedSignals
+	notary.NotarySupportedSignals = nil
+	defer func() {
+		notary.NotarySupportedSignals = old
```
endophage: I think not deferring this and moving it to after `testSetSignalTrap` will solve the race checker problem.

cyli: Cool, thanks!

cyli:
If I rebase to master, the race condition goes away... strange.

cyli:
Oh, no.  It's because covermode is still count on this branch, as opposed to atomic, which does more precise counting when there's parallelism.

riyazdf:
newest changes LGTM

PR #972, opened 2016-09-22
Fix auth login error messages and messaging when witnessing a delegation with 0 valid keys

### LINKED ISSUE #969: Clean up error message for incorrect passphrase
We get weird `(MISSING)` delimiters and other hard-to-read output when failing to login with correct credentials against a notary-server with auth (like `https://notary.docker.io`)

```
notary 🐳 $ bin/notary -s https://<URL> publish <REPO>
Pushing changes to <REPO>
Enter username: <USERNAME>
Enter password:

* fatal: token auth attempt for registry: https://<URL>/auth/token?account=<USERNAME>&scope=repository%!A(MISSING)d<URL>%!F(MISSING)<ORG>%!F(MISSING)<REPO>%!A(MISSING)push%!C(MISSING)pull&service=<URL> request failed with status: 401 Unauthorized
```

I've added in `<URL>`, `<REPO>`, `<USERNAME>`, and `<ORG>` myself, but the `%!F(MISSING)` output is straight from the terminal

cyli:
When I remove a key from a delegation that has a single key, then I try to witness that delegation, signing fails because there are no valid keys for that delegation.  The error message was:

`* fatal: signing keys not available, need 1 keys out of: ` - I've fixed the error message to say there are an insufficient number of valid signing keys for the threshold required, basically.  We might want to move this error check up further to when we apply a witness change, since I think that might be the only time when we might have an invalid role.

Fixes #969 - that format string was in distribution, so I re-vendored.  I also added logic to check if the login username/password is a terminal, otherwise fail.

Note: possibly we want to rebase this to 0.4.0 so that it can be added to 0.4.1


endophage:
Is there an easy way to get Godep to exclude the distribution `*.md` and build files?

cyli:
I can manually remove them - Godep tried to include the vendor dir :|

tuf/signed/errors.go
```diff
@@ -15,11 +15,17 @@ type ErrInsufficientSignatures struct {
 
 func (e ErrInsufficientSignatures) Error() string {
 	candidates := strings.Join(e.MissingKeyIDs, ", ")
-	if e.FoundKeys == 0 {
-		return fmt.Sprintf("signing keys not available, need %d keys out of: %s", e.NeededKeys, candidates)
+	switch {
+	case len(e.MissingKeyIDs) < e.NeededKeys:
+		return fmt.Sprintf(
+			"cannot sign because while %d signatures are needed, an insufficient number of valid signing keys have been specified",
```
endophage: This reads a little verbosely. Maybe something like: "insufficient signing keys available. %d signatures are required but only %d keys are available"

cyli: Hmm... that implies that we are missing keys.  We may not be missing any keys - it's just that the role doesn't specify enough keys for anyone to sign.

cyli: Maybe I should just say that the role is broken?

riyazdf: this case could be broken down further into `len(e.MissingKeyIDS) + e.FoundKeys < e.NeededKeys` and `len(e.MissingKeyIDS) + e.FoundKeys >= e.NeededKeys` right?

For `len(e.MissingKeyIDS) + e.FoundKeys < e.NeededKeys`: role is broken
For `len(e.MissingKeyIDS) + e.FoundKeys >= e.NeededKeys`: @endophage's suggestion of `"insufficient signing keys available. %d signatures are required but only %d keys are available"`  This could be the base case in `default`

cyli: As per discussion IRL, this is basically just a role validity check.  The only way we can get to this weird invalid state is with Witness, so we should just move this check to that functionality.

cyli:
@endophage @riyazdf sorry this took so long to update.  Tried to fix some of the messaging around witnessing an invalid delegation role as suggested above.

cmd/notary/integration_test.go
```diff
@@ -1637,6 +1639,12 @@ func TestWitness(t *testing.T) {
 	require.NoError(t, err)
 	require.Contains(t, output, targetName)
 	require.Contains(t, output, targetHash)
+
+	_, err = runCommand(t, tempDir, "-s", server.URL, "delegation", "remove", "-p", "gun", delgName, keyID, keyID2)
+	require.NoError(t, err)
+	_, err = runCommand(t, tempDir, "-s", server.URL, "witness", "-p", "gun", delgName)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "role does not specify enough valid signing keys to meet its required threshold")
```
riyazdf: awesome, thanks for adding this test 👍

cmd/notary/tuf.go
```diff
@@ -755,7 +757,8 @@ type passwordStore struct {
 }
 
 func (ps passwordStore) Basic(u *url.URL) (string, string) {
-	if ps.anonymous {
+	// if it's not a terminal, don't wait on input
+	if ps.anonymous || !terminal.IsTerminal(int(os.Stdin.Fd())) {
```
endophage: Shouldn't we be checking the input that was set on the command?

endophage: IRL: leave as is for now as we use `os.Stdin` elsewhere too. Issue open already for looking into this.

PR #976, opened 2016-09-23
[WIP] Try to set up Jenkins job for windows tests

cyli:
test test test


cyli:
closing this - I think there've been a bunch of jenkins changes lately, so this is outdated.  we may possibly want to run this on appveyor instead as well so that other maintainers can admin.

PR #981, opened 2016-09-28
Be able to read keys even if they are not in the root/non-root subdirs

cyli:
This way a 0.4.1 client can read a repo that a client built from master (with the flattened keystore #872) converts

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 + Preliminary Windows support for notary client [#970](https://github.com/docker/notary/pull/970)
 + Output message to CLI when repo changes have been successfully published [#974](https://github.com/docker/notary/pull/974)
 + Improved error messages for client authentication errors and for the witness command [#972](https://github.com/docker/notary/pull/972)
++ Support for finding keys that are anywhere in the notary directory's "private" directory, not just under "private/root_keys" or "private/tuf_keys" [#981](https://github.com/docker/notary/pull/981)
 
 ## [v0.4.0](https://github.com/docker/notary/releases/tag/v0.4.0) 9/21/2016
 + Server-managed key rotations [#889](https://github.com/docker/notary/pull/889)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 + Preliminary Windows support for notary client [#970](https://github.com/docker/notary/pull/970)
 + Output message to CLI when repo changes have been successfully published [#974](https://github.com/docker/notary/pull/974)
 + Improved error messages for client authentication errors and for the witness command [#972](https://github.com/docker/notary/pull/972)
++ Support for finding (but not writing) keys even if they are not in the "root_keys" or "tuf_keys" subdirectories [#981](https://github.com/docker/notary/pull/981)
```
endophage: Should probably be more specific because at the moment this reads like "we'll find keys anywhere!" It's that we'll also look one dir up in `private/` right?

cyli: Ah true, we only check in the private dir.  Fixed, and I've also add that test.

trustmanager/keystore_test.go
```diff
@@ -567,6 +582,42 @@ func TestRemoveKey(t *testing.T) {
...
+
+		err = store.RemoveKey(fmt.Sprintf("keyID%d", i))
+		require.NoError(t, err)
+
+		// File should no longer exist
+		_, err = ioutil.ReadFile(fp)
+		require.True(t, os.IsNotExist(err), "file should not exist")
+	}
+
+	// removing a non-existant key should not error
```
riyazdf: typo: `non-existent`

cyli: Thanks.  :)  I always misspell that - need to re-train my muscle memory for that word.  Fixed!

noderunner:
Will this allow users to use client certificates when connecting to a remote Docker registry? What kind of keys are read from this location?

riyazdf:
@noderunner this change relates to signing keys used for Notary and Docker Content Trust, for more info about docker registry certificates you might be interested in this documentation: https://docs.docker.com/engine/security/certificates/

PR #982, opened 2016-09-29
Client should fall back on cache if operating offline when updating

cyli:
If the client is in offline mode, or encounters some kind of network error when downloading a new timestamp, it should fall back on using the version of the repo in the cache.

Since there a huge number of possible network errors, just wrap any errors received when actually making the HTTP request using the remote store in a `NetworkError`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 + Output message to CLI when repo changes have been successfully published [#974](https://github.com/docker/notary/pull/974)
 + Improved error messages for client authentication errors and for the witness command [#972](https://github.com/docker/notary/pull/972)
 + Support for finding keys that are anywhere in the notary directory's "private" directory, not just under "private/root_keys" or "private/tuf_keys" [#981](https://github.com/docker/notary/pull/981)
++ Previously, on any error updating, the client would fall back on the cache.  Now we only do so if there is a network error or if the server is unavailable or missing the TUF data. Invalid TUF data will cause the update to fail - for example if there was an invalid root rotation. [#982](https://github.com/docker/notary/pull/982)
 
 ## [v0.4.0](https://github.com/docker/notary/releases/tag/v0.4.0) 9/21/2016
 + Server-managed key rotations [#889](https://github.com/docker/notary/pull/889)
--- a/client/client_update_test.go
+++ b/client/client_update_test.go
@@ -172,6 +172,45 @@ func TestUpdateSucceedsEvenIfCannotWriteExistingRepo(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 + Output message to CLI when repo changes have been successfully published [#974](https://github.com/docker/notary/pull/974)
 + Improved error messages for client authentication errors and for the witness command [#972](https://github.com/docker/notary/pull/972)
 + Support for finding keys that are anywhere in the notary directory's "private" directory, not just under "private/root_keys" or "private/tuf_keys" [#981](https://github.com/docker/notary/pull/981)
++ Previously, on any error updating, the client would fall back on the cache.  Now we only do so if there is a network error or if the server is unavailable or missing the TUF data (and fails on invalid root rotation). [#982](https://github.com/docker/notary/pull/982)
```
riyazdf: Can we split out the parenthesized bit for clarity?  Something like "Invalid TUF data will cause the update to fail, for example if there was an invalid root rotation`

storage/httpstore.go
```diff
@@ -174,22 +183,9 @@ func (s HTTPStore) GetSized(name string, size int64) ([]byte, error) {
...
-	if err != nil {
-		return err
-	}
-	resp, err := s.roundTrip.RoundTrip(req)
-	if err != nil {
-		return err
-	}
-	defer resp.Body.Close()
-	return translateStatusToError(resp, "POST "+name)
+	return s.SetMulti(map[string][]byte{name: blob})
```
riyazdf: this is an awesome cleanup 👍

client/client_update_test.go
```diff
@@ -172,6 +172,41 @@ func TestUpdateSucceedsEvenIfCannotWriteExistingRepo(t *testing.T) {
...
+	require.Error(t, err)
+	require.IsType(t, store.NetworkError{}, err)
+
+	// offline client: no cache - errors
+	tempBaseDir, err := ioutil.TempDir("", "notary-test-")
+	require.NoError(t, err, "failed to create a temporary directory: %s", err)
+	defer os.RemoveAll(tempBaseDir)
+	offlineRepo, err := NewNotaryRepository(tempBaseDir, "docker.com/notary", "https://nope",
+		nil, passphrase.ConstantRetriever("pass"), trustpinning.TrustPinConfig{})
+	require.NoError(t, err)
```
riyazdf: Should we call `offlineRepo.Update(false)` and verify that it errors?

cyli: Oops, yes I was intending to and just forgot.  :)  Thanks!

storage/httpstore.go
endophage: Oh man yeah! Nice one!

PR #985, opened 2016-10-01
custom autobuild script to create minimal build

endophage:
~~I need to test the builds but samples can be found on docker hub under endophage/notaryauto. These minimal images cut us down from ~160MB to ~5MB~~

~~Reminder to self: reset the docker-compose.yml to master's version before submitting this for review. Been using it to test the new images.~~

Tested and reset docker-compose.yml. These new hooks will be used by anyone setting up an autobuild of this on docker hub/cloud. They produce very minimal images based on busy box. Relative sizes are ~25MB each for server and signer vs a little over 500MB for the existing images built from `server.Dockerfile` and `signer.Dockerfile` (N.B. these are the sizes once unpacked on disk, hub shows much smaller sizes still, unsure if this is a miscalculation or they really compress that well).

p.s. as soon as we merge this to master, the dockersecurity/notary_autobuilds on docker hub will start using these images

### DIFF
--- a/hooks/push
+++ b/hooks/push
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+serverTag="server"
+signerTag="signer"
+
+docker push $DOCKER_REPO:${serverTag}-$DOCKER_TAG 
+docker push $DOCKER_REPO:${signerTag}-$DOCKER_TAG 
--- a/hooks/build
+++ b/hooks/build
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+serverTag="server"
+signerTag="signer"
+
+# build server and signer images
+docker build -t $serverTag -f server.Dockerfile .
+docker build -t $signerTag -f signer.Dockerfile .
+
+mkdir ./bin
+
+# ensure the containers are removed if they already exist
+docker rm -f server signer
+
+# copy binaries out of images
+loc=$(docker run --name=server --entrypoint=which $serverTag notary-server)
+docker cp server:$loc ./bin/notary-server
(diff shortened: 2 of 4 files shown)


cyli:
This is awesome, and we should use it as a reference for anyone that does the same.  LGTM, thanks!

HuKeping:
> (N.B. these are the sizes once unpacked on disk, hub shows much smaller sizes still, unsure if this is a miscalculation or they really compress that well).

AFAIK, they do compress it in registry

hooks/push
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+serverTag="server"
+signerTag="signer"
+
+docker push $DOCKER_REPO:${serverTag}-$DOCKER_TAG 
+docker push $DOCKER_REPO:${signerTag}-$DOCKER_TAG
```
HuKeping: and here

endophage:
Yeah, they definitely do compress, the ratios would just be slightly better than I expect (particularly for the old image at about 3.5x). Not gratuitously, but slightly, and hub has had miscalculation bugs in the past.

endophage:
Interesting. Where are you seeing the warning? Wondering if it's a bash version difference or if I just missed it.

HuKeping:
image

endophage:
Ah sorry, was replying from email which didn't make it obvious you were commenting on a specific line. I thought you were seeing it in an IDE or trying to test the scripts locally. I'll submit a follow up just to add those newlines. A quick google indicates its a legacy thing about how unix defines a "file"

PR #988, opened 2016-10-03
lint was updated per linked issues. This fixes new requirement

endophage:
https://github.com/golang/lint/pull/245
https://github.com/golang/go/issues/17293

### DIFF
--- a/notary.go
+++ b/notary.go
@@ -5,3 +5,8 @@ package notary
 // confirmation), createNew will be true. Attempts is passed in so that implementers
 // decide how many chances to give to a human, for example.
 type PassRetriever func(keyName, alias string, createNew bool, attempts int) (passphrase string, giveup bool, err error)
+
+// CtxKey is a wrapper type for use in context.WithValue() to satisfy golint
+// https://github.com/golang/go/issues/17293
+// https://github.com/golang/lint/pull/245
+type CtxKey string
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -126,12 +126,12 @@ func simpleTestServer(t *testing.T, roles ...string) (
 func fullTestServer(t *testing.T) *httptest.Server {
 	// Set up server
 	ctx := context.WithValue(
-		context.Background(), "metaStore", storage.NewMemStorage())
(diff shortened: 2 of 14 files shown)


endophage:
@cyli @riyazdf @HuKeping not sure why codecov failed but the diff has 100% coverage and coverage wouldn't have changed for this change anyway. If we can get this merged ASAP other PRs will be able to pass linting (once rebased)

notary.go
```diff
@@ -5,3 +5,8 @@ package notary
 // confirmation), createNew will be true. Attempts is passed in so that implementers
 // decide how many chances to give to a human, for example.
 type PassRetriever func(keyName, alias string, createNew bool, attempts int) (passphrase string, giveup bool, err error)
+
+// CtxKey is a wrapper type for use in context.WithValue() to satisy golint
+// https://github.com/golang/go/issues/17293
+// https://github.com/golang/lint/pull/245
+type CtxKey string
```
cyli: Non-blocking:  wondering if we can make constants for the 4 CTX keys we use:  `cryptoService`, `metaStore`, `repo` (maybe we should change this to "GUN" though), and `keyAlgorithm`.  

Even better is if we can unexport the type and just export the constants, but lint might complain.

At any rate, happy to have this done in another PR.

endophage: ooooh, I do like that idea! Lets unblock circle for now and I'll submit that tomorrow.

notary.go
```diff
@@ -5,3 +5,8 @@ package notary
 // confirmation), createNew will be true. Attempts is passed in so that implementers
 // decide how many chances to give to a human, for example.
 type PassRetriever func(keyName, alias string, createNew bool, attempts int) (passphrase string, giveup bool, err error)
+
+// CtxKey is a wrapper type for use in context.WithValue() to satisy golint
```
cyli: Oops, sorry one last non-blocking nitpick:  "satisfy"

endophage: Huh, wonder why misspell didn't catch that. Fixing

notary.go
endophage: https://github.com/docker/notary/issues/989

Issue #932, opened 2016-08-26
Improve Dependency Injection for NotaryRepository initialization


endophage:
Currently it takes a basedir and a RoundTripper, this is fine for `NewNotaryRepository` which should be a simplified initializer with assumptions made. However, for unusual use cases and testing, it would be helpful to be able to inject a cache and RemoteStore directly.

We think this should probably be done by exporting `repositoryFromKeystores` (and maybe renaming it to something like `NewRepositoryFromStores`), and having it take an initialized cache and RemoteStore, rather than `baseDir` and `rt`.

endophage:
Closed by #1094

PR #998, opened 2016-10-07
make cache injectable into NotaryRepository

endophage:
Reasonably minimal changes to make cache injectable while not having to move around much code. Also want to keep and add to "helper" constructors (e.g. `NewFileCachedNotaryRepository`) rather than push too much of that code up the call stack.

Also moved a bunch of errors that were in client/client.go into a new client/errors.go to keep things a little cleaner. client/client.go was a bit messy with a bunch of error types. They're still in the same scope (i.e. in the `client` package) so nothing else had to be changed by moving them.

### DIFF
--- a/client/repo.go
+++ b/client/repo.go
@@ -4,26 +4,15 @@ package client
 
 import (
 	"fmt"
-	"net/http"
 
 	"github.com/docker/notary"
 	"github.com/docker/notary/trustmanager"
-	"github.com/docker/notary/trustpinning"
 )
 
-// NewNotaryRepository is a helper method that returns a new notary repository.
-// It takes the base directory under where all the trust files will be stored
-// (This is normally defaults to "~/.notary" or "~/.docker/trust" when enabling
-// docker content trust).
-func NewNotaryRepository(baseDir, gun, baseURL string, rt http.RoundTripper,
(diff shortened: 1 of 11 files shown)


riyazdf:
also as a breadcrumb, this partially addresses https://github.com/docker/notary/issues/932

client/repo.go
```diff
@@ -4,26 +4,15 @@ package client
...
 
-// NewNotaryRepository is a helper method that returns a new notary repository.
-// It takes the base directory under where all the trust files will be stored
-// (This is normally defaults to "~/.notary" or "~/.docker/trust" when enabling
-// docker content trust).
-func NewNotaryRepository(baseDir, gun, baseURL string, rt http.RoundTripper,
-	retriever notary.PassRetriever, trustPinning trustpinning.TrustPinConfig) (
-	*NotaryRepository, error) {
-
+func getKeyStores(baseDir string, retriever notary.PassRetriever) ([]trustmanager.KeyStore, error) {
```
cyli: Non-blocking:  It might be nice to make these exportable; the CLI has a function that basically replicates this for listing keys, etc, and it would be awesome to replace it with these.  Can be done in a separate PR though.

endophage: 👍

PR #1003, opened 2016-10-13
Drop the executable bits for a regular file

### LINKED ISSUE #1002: Separate the perm between a regular file and directory
At present, the regular files are with the same perm as its directory.

image

HuKeping:
Closes #1002


storage/filestore.go
```diff
@@ -221,7 +221,7 @@ func (f *FilesystemStore) Set(name string, meta []byte) error {
 	os.RemoveAll(fp)
 
 	// Write the file to disk
-	if err = ioutil.WriteFile(fp, meta, f.perms); err != nil {
+	if err = ioutil.WriteFile(fp, meta, f.perms&notary.PermExecutableMask); err != nil {
```
riyazdf: I'm wondering if instead of having to perform this bitwise and `&` at every possible write point, that if we instead create a new unexported field for `FilesystemStore` for file permissions without the executable bits set.  

For example, this write could become `ioutil.WriteFile(fp, meta, f.nonExecutePerms)`

riyazdf: @HuKeping what do you think of @endophage's constants proposal?

With these constants, we could separate `.perms` to `.filePerms` and `.dirPerms` that are initialized to the appropriate constants (`PubExecPerms+PubNoExecPerms` or `PrivExecPerms+PrivNoExecPerms`) depending on which store it is?

I'm worried that adding bitwise masks makes it too easy to make a mistake or forget another place where we need to use it, especially when we're only writing files/directories with one of four permissions.

HuKeping: That's a good point, totally fair. It's quite easy to forget using it somewhere else. I'll change, thanks!

riyazdf: my understanding was that `NewPrivateSimpleFileStore` would use the private perms pair and `NewFilesystemStore` would use the public ones.

We would have to change the usage of `NewFilesystemStore` to `NewPrivateSimpleFileStore` in `client.go` to keep the existing permissions behavior.

endophage: The function naming is poor, but the intent was to have utility initializers to create public, `NewSimpleFileStore`, and private, `NewFilesystemStore`, file storage, with the ability to always set more explicit options via function they both call: `NewFileStore`.

It would be fine to create more wrappers if necessary, but I think _everything_ is using `NewFilesystemStore` at the moment (#1010 removes `NewSimpleFileStore`), so we should just modify the permissions for everyone there.

We're also thinking of adding https://gas.css.bio/#docker/notary as a lint check so we might want to cut the permissions down to `0700` / `0600` and have no concept of public files.

HuKeping: Thanks to the GFW I can't get access to the site  https://gas.css.bio/#docker/notary .

Do you think it make sense that removing the function `NewFilesystemStore()` because I find it was only  used to join the path , so we can use `NewFileStore()` to replace it.

Besides, if we decide to drop the concept of public files, do you think it also make sense that re-define the function  `NewFileStore(baseDir, fileExt string, perms os.FileMode)`   to just `NewFileStore(baseDir, fileExt string)` and set the permission to `0700/0600` wherever it needs (the only place is the filestore.Set()).

HuKeping: Sorry to have updated this PR before getting the feedback from you guys, please feel free to correct me if I'm doing wrong.

const.go
```diff
@@ -68,6 +70,11 @@ const (
 	HealthCheckKeyManagement = "grpc.health.v1.Health.KeyManagement"
 	HealthCheckSigner        = "grpc.health.v1.Health.Signer"
 	HealthCheckOverall       = "grpc.health.v1.Health.Overall"
+
+	// PrivExecPerms indicates the file permissions for directory
+	// and PrivNoExecPerms for file.
+	PrivExecPerms   = 0700
```
endophage: Ah good catch, yeah we should just consolidate the consts

HuKeping: Updated.

PR #1005, opened 2016-10-13
add newlines on build hooks so GH doesn't warn

endophage:
cc @HuKeping 


cyli:
LGTM on green

Issue #1007, opened 2016-10-14
Cleanup error message for auth challenge  


HuKeping:
There's some `MISSING` in the output error message when we failed the challenge for authentication
```
* fatal: Get https://ihub.com:443/index/notary?account=hukeping&scope=repository%!A(MISSING)huawei.com%!A(MISSING)push%!C(MISSING)pull&service=notary: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```

Seems it was caused by the error message we get from roundtripper contains some ASCII code like `%3A`, and when the whole string be used as an format output, the warning we mentioned above will show up.

riyazdf:
@HuKeping: which notary server version are you seeing this on?  I ask because we had a similar issue https://github.com/docker/notary/issues/969 and the fix was to revendor `docker/distribution`: https://github.com/docker/notary/pull/972

This should be included in notary v0.4.1+

HuKeping:
It should the latest version I suppose, but I'll check that the next Monday

HuKeping:
Hi @riyazdf  it was based on commit 303e1fe40f223ca944274a1129f88e11e0b4d0d3 upstream.

HuKeping:
The version info:
```
:~$ notary-server -debug
{"level":"info","msg":"Version: 0.5-dev, Git commit: 303e1fe","time":"2016-10-17T17:27:00+08:00"}
{"level":"info","msg":"Debug server listening on localhost:8080","time":"2016-10-17T17:27:00+08:00"}
{"level":"fatal","msg":"Could not read config at :, viper error: Unsupported Config Type \"\"","time":"2016-10-17T17:27:00+08:00"}
:~$ 
:~$ notary-signer -debug
{"level":"info","msg":"Version: 0.5-dev, Git commit: 303e1fe","time":"2016-10-17T17:27:10+08:00"}
{"level":"info","msg":"Debug server listening on localhost:8080","time":"2016-10-17T17:27:10+08:00"}
{"level":"fatal","msg":"Could not read config at :, viper error: Unsupported Config Type \"\"","time":"2016-10-17T17:27:10+08:00"}
:~$ 
:~$ notary version
notary
 Version:    0.5-dev
 Git commit: 303e1fe
```

HuKeping:
Steps for reproducing the problem:
```
# 1. start notary-signer with the local configuration file
$ notary-signer -config fixtures/signer-config-local.json 
[...]

PR #1008, opened 2016-10-14
Cleanup network error message

HuKeping:
Cleanup those error message with "%" followed by two
hexadecimal digits which will cause some "MISSING" warnning.


storage/httpstore.go
```diff
@@ -39,7 +39,14 @@ type NetworkError struct {
 }
 
 func (n NetworkError) Error() string {
-	return n.Wrapped.Error()
+	res, err := url.QueryUnescape(n.Wrapped.Error())
+	if nil != err {
```
endophage: `err != nil` for consistency

storage/httpstore.go
```diff
@@ -39,7 +39,14 @@ type NetworkError struct {
...
+		// QueryUnescape does the inverse transformation of QueryEscape,
+		// converting %AB into the byte 0xAB and '+' into ' ' (space).
+		// It returns an error if any % is not followed by two hexadecimal digits.
+		return fmt.Sprintf("Original error is %p, url Unescape error is %p", n.Wrapped.Error(), err.Error())
```
endophage: Not sure if this is necessary. Probably just want to return the original error and log the `QueryUnescape` error.

storage/httpstore.go
```diff
@@ -39,7 +39,14 @@ type NetworkError struct {
 }
 
 func (n NetworkError) Error() string {
-	return n.Wrapped.Error()
```
cyli: Following up on https://github.com/docker/notary/issues/1007#issuecomment-254635035:

Nonblocking:  In that thread I was suggesting that rather that we can check to see if ``n.Wrapped` is a `url.Error` and if so, convert the wrapped error to a new `url.Error` with the same params but with the URL escaped - that way we know that we'd always be applying the unescape specifically to a URL, and not some other string that might have an arbitrary `%` in it.

I don't know of any other network errors that have a `%` though - this is likely fine as it is, and is just my being pedantic.

HuKeping: Thank you @cyli ! You are right about this ! I debug it a little more deeper and find that the `n.Wrapped` here can also be other error types like  `errcode.Error` not only the `url.Error`. So I think we should make a judgement here to see if it was the `url.Error` and if so we do the unescape operation.

storage/httpstore.go
HuKeping: updated.

PR #1012, opened 2016-10-19
Redundant godep vendoring

riyazdf:
I removed redundant dependencies that were somehow picked up by godep inside `docker/distribution`'s vendor directory (like `github.com/docker/distribution/vendor/github.com/Sirupsen/logrus`).

Also bumped the major go version to 1.7 and did a `godep save ./...` that resorted and cleaned up the `Godeps.json` file.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -905,41 +905,6 @@
 		{
 			"ImportPath": "golang.org/x/net/lex/httplex",
 			"Rev": "6a513affb38dc9788b449d59ffed099b8de18fa0"
-		},
-		{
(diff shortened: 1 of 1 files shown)


endophage:
Only thing I'm seeing is the changed Godeps file but based on your (@riyazdf) comment I'm expecting to see some changes in `vendor/github.com/docker/distribution/vendor/...`. Am I misunderstanding?

riyazdf:
@endophage for some reason those files were never there, so I don't know why the Godeps file registered those paths

endophage:
Ah great, thanks for clarifying!

PR #1015, opened 2016-10-21
Add gas check to make lint, fix unnecessary assignment

riyazdf:
Removes the parent directory creation code from `client.go` because it's already handled by the underlying cache.

Also adds github.com/HewlettPackard/gas to the `make lint` check, though it's a little bit hacky until https://github.com/HewlettPackard/gas/issues/55 and https://github.com/HewlettPackard/gas/issues/50 are resolved.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -776,8 +776,7 @@ func (r *NotaryRepository) saveMetadata(ignoreSnapshot bool) error {
 	}
 
 	for role, blob := range targetsToSave {
-		parentDir := filepath.Dir(role)
-		os.MkdirAll(parentDir, 0755)
+		// If the parent directory does not exist, the cache.Set will create it
 		r.cache.Set(role, blob)
 	}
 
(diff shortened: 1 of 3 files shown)


client/client.go
```diff
@@ -777,8 +777,7 @@ func (r *NotaryRepository) saveMetadata(ignoreSnapshot bool) error {
 	}
 
 	for role, blob := range targetsToSave {
-		parentDir := filepath.Dir(role)
-		os.MkdirAll(parentDir, 0755)
```
HuKeping: Thanks for removing this!

client/client.go
```diff
@@ -777,8 +777,7 @@ func (r *NotaryRepository) saveMetadata(ignoreSnapshot bool) error {
...
+		// If the parent directory does not exist, the cache.Set will create it
```
HuKeping: I wonder if this code comment is suitable for all the other metadata store implementations like `memorystore` and `httpstore`, they may need not to "create" the parent directory.

HuKeping: Oh, it seems it will always init as FilesystemStore

riyazdf: @HuKeping: yup this function is only used for clients saving local copies of the repo metadata to files on disk

HuKeping:
Tested it locally and it works great! Have rebased to the master.  LGTM!

PR #1018, opened 2016-10-27
Rework auth

HuKeping:
```
The authentication challenge before a http request is really great to prevent some attacks like DDoS.

But the green channel for some Non-push operation like "notary list" will break that guarantees.

We should challenge all the request which brings data back from the server except "Ping".
#713 add an extra challenge(the basic) for private repo, but as I said above it's not enough, all the request for public repo should be challenged too, otherwise the attacker can create a script with infinite  loop to list the public repo information and we can't know who is doing that.

So I revert the commit in #713 and make all request must be challenged if it was configured as such .

I'm not sure if I have missed anything or any scenarios, please feel free to chime in.

/cc @liron-l 
```
EDIT: The above is not a good comments

The original purpose is about to add a basic authentication check even those not-push operations like `lookup` , `list`, etc.. #713 have done that with an additional RoundTripper which I think is not very necessary.  We can just make the client to ask for token everytime by just setting the `anonymous` to false.

### DIFF
--- a/cmd/notary/tuf.go
+++ b/cmd/notary/tuf.go
@@ -907,7 +907,7 @@ func tokenAuth(trustServerURL string, baseTransport *http.Transport, gun string,
 		return nil, err
 	}
 
-	ps := passwordStore{anonymous: permission == readOnly}
+	ps := passwordStore{anonymous: false}
 
 	var actions []string
 	switch permission {
--- a/cmd/notary/tuf.go
+++ b/cmd/notary/tuf.go
@@ -923,16 +923,8 @@ func tokenAuth(trustServerURL string, baseTransport *http.Transport, gun string,
 
 	tokenHandler := auth.NewTokenHandler(authTransport, ps, gun, actions...)
 	basicHandler := auth.NewBasicHandler(ps)
-
-	modifier := auth.NewAuthorizer(challengeManager, tokenHandler, basicHandler)
-
-	if permission != readOnly {
-		return newAuthRoundTripper(transport.NewTransport(baseTransport, modifier)), nil
-	}
-
-	// Try to authenticate read only repositories using basic username/password authentication
-	return newAuthRoundTripper(transport.NewTransport(baseTransport, modifier),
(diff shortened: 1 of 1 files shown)


liron-l:
@HuKeping I'm not sure I understand this change.
Notary is a client side application, requiring authentication for public endpoints from a client application does not prevent DDos attacker (since attacker can still call the API directly without the client application and auth)

HuKeping:
Oh I'm sorry for the misunderstanding, I think you are right about the DDos attack @liron-l  . 

With #713 it will still fail with error "unauthorized: authentication required" when `notary list xxx`  and this pr is meant to add authenticate checking even it was a "pull" permission request, so that when we typed `notary list xxx` it will also ask for username/password no matter it was a public or private repo.

endophage:
I agree with @liron-l, adding auth challenges to public endpoints only increases work on the server, it doesn't protect against DDoS given anyone can get a token for a public repo.

I think (and I know @diogomonica will have some good input) that the right way to protect against a DDoS is to put something like Varnish in front of Notary and crank up the cache lifetimes so most requests get served from cache. Varnish can then be horizontally scaled and we can push updates into the Varnish cache at our own speed.

None of that requires changes in Notary, it's an operational problem.

We may however want to apply some kind of rate limiting to pushes as a compromised/malicious account could use the authed push endpoints to continue to hit the notary servers directly, where verification and snapshot/timestamp signing requires a meaningful amount of CPU. I've added an issue for this here: https://github.com/docker/notary/issues/1020

HuKeping:
I agree all about the DDoS stuff.  What I want to floating to discuss is whether we should also challenge the public operations.  

But it seems like it's up to the implementation of the  JWT-Server, if the JWT-Server allow public images need no to be authenticated, it will always create a token back to the client. In the other hand, if the JWT-Server doesn't distinguish the public/private images, it will only create the token when it get the correct credential(like username/passwd).

What's the docker's auth server(JWT Server) looks like?

Btw, could you please  try `notary --debug list PRIVATE_IMAGES`,  it warns that it can not download the timestamp metafile on my machine (with the master branch).

HuKeping:
closing this since it only change the client :)

Issue #796, opened 2016-06-22
Define and document notary server API


riyazdf:
We should document the notary server endpoint API

HuKeping:
👍

PR #1019, opened 2016-10-28
Changefeed implementation

### LINKED ISSUE #436: Change feed
Add an endpoint that provides a feed of published versions of the repository (published timestamps from which a versioned repository tree can be walked).

endophage:
Only works for the supported relational databases. No intent to implement for RethinkDB in this PR.

Closes #436 

Implementation (~~some of this needs to be done/tweaked, only basic global feed implemented at time of writing and some query string params need to be brought inline with the below~~): 

A changefeed can be requested for a specific repo, or globally. Query string params are `change_id` and `records`, providing a starting index (exclusive) from which to query, and the number of results to return. The `change_id` may be `-1` indicating to start from the most recent change, and in this case, we will assume any value passed to `records` should be considered negative. `records` may be a negative number indicating that number of records ordered before the index should be returned. URL for the feed is `/v2/changefeed/_trust`, the final `_trust` segment being included to keep compatibility with existing url routing for running notary beside registry.

Signed-off-by: David Lawrence david.lawrence@docker.com (github: endophage)


endophage:
Also need to add image name filtering. Only going to implement global and single repo, not wildcarded `docker.io/foo/*`. Tests should pass this run.

ecordell:
@endophage This could be a great way to implement the historic fetching for #942 backwards-compatibility. Is that intentional or a happy coincidence?

endophage:
@ecordell this is going to provide versions over timestamps (because they represent actual published versions of the repo). I'd guess it would still be more efficient to query for root files by version, which are likely to change much less frequently.

I'm thinking more and more it may also be worth implementing a hard requirement server side that updates increase versions of each role monotonically so we can guarantee "root.X" exists for all X < latest version.

endophage:
OK, auth and image name filtering implemented. Circle won't pass until I get the new migration files committed when I'm back at my desktop.

server/storage/sql_models.go
```diff
@@ -26,8 +60,11 @@ func CreateTUFTable(db gorm.DB) error {
...
-	if query.Error != nil {
-		return query.Error
-	}
-	return nil
+	return query.Error
+}
+
+// CreateChangefeedTable creates the DB table for Changefeed
+func CreateChangefeedTable(db gorm.DB) error {
+	query := db.Set("gorm:table_options", "ENGINE=InnoDB DEFAULT CHARSET=utf8").CreateTable(&Changefeed{})
```
riyazdf: do we want any indexes on gun and/or version?

endophage: It's possible we may want an index on gun as we'll filter by gun for single repo updates. No need for an index on version though as unless something is desperately wrong, it'll order the same way as the id and we shouldn't be querying the changefeed table for a specific version.

server/handlers/changefeed.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	qs := r.URL.Query()
+	imageName := qs.Get("filter")
+	if imageName == "" {
+		// no image name means global feed
+		imageName = "*"
+	}
+	pageSizeStr := qs.Get("page_size")
+	pageSize, err := strconv.ParseInt(pageSizeStr, 10, 32)
```
endophage: 👍

server/handlers/changefeed.go
```diff
@@ -0,0 +1,75 @@
...
+
+	changes, err := store.GetChanges(changeID, int(pageSize), imageName, reversed)
+	if err != nil {
+		logger.Errorf("500 GET could not retrieve records: %s", err.Error())
+		return errors.ErrUnknown.WithDetail(err)
+	}
+
+	// if reversed, we need to flip the list order so oldest is first
+	if reversed {
+		for i, j := 0, len(changes)-1; i < j; i, j = i+1, j-1 {
```
riyazdf: nit: might be worth factoring this into `utils` but leaving this here is fine for now.

server/server.go
```diff
@@ -186,11 +193,18 @@ func RootHandler(ctx context.Context, ac auth.AccessController, trust signed.Cry
 		ServerHandler:       handlers.DeleteHandler,
 		PermissionsRequired: []string{"*"},
 	}))
+	r.Methods("GET").Path("/v2/changefeed/_trust").Handler(createHandler(_serverEndpoint{
+		OperationName:       "Changefeed",
+		ServerHandler:       handlers.Changefeed,
+		PermissionsRequired: []string{"*"},
```
riyazdf: Should this just be `pull`?  Since pull permission are enough to get TUF metadata in the first place?

server/storage/memory.go
```diff
@@ -8,6 +8,10 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"strconv"
```
riyazdf: super nit: move this up to the golang import blocks

server/storage/interface.go
```diff
@@ -39,4 +39,28 @@ type MetaStore interface {
...
+
+	// GetChanges returns an ordered slice of changes. It starts from
+	// the change matching changeID, but excludes this change from the results
+	// on the assumption that if a user provides an ID, they've seen that change.
+	// If changeID is 0, it starts from the
+	// beginning, and if changeID is -1, it starts from the most recent
+	// change. The number of results returned is limited by pageSize.
+	// Reversed indicates we are fetching pages going backwards in time, the
+	// default being to fetch pageSize from changeID going forwards in time.
+	GetChanges(changeID string, pageSize int, filterName string, reversed bool) ([]Change, error)
```
riyazdf: I think we should remove `reversed` from the signature since we do the reverse sorting in the server handler wrapper itself.

server/storage/memory.go
```diff
@@ -158,6 +196,25 @@ func (st *MemStorage) Delete(gun string) error {
 	return nil
 }
 
+// GetChanges returns a []Change starting from but excluding the record
+// identified by changeID. In the context of the memory store, changeID
+// is simply an index into st.changes. The ID of a change, and its index
+// are equal, therefore, we want to return results starting at index
+// changeID+1 to match the exclusivity of the interface definition.
+func (st *MemStorage) GetChanges(changeID string, pageSize int, filterName string, reversed bool) ([]Change, error) {
+	id, err := strconv.ParseInt(changeID, 10, 32)
```
riyazdf: `ParseUint` here too?  Generally wondering if we should change `pageSize` to a `uint` in signatures too

utils/http.go
```diff
@@ -28,74 +29,137 @@ type rootHandler struct {
...
-				// error level logging for 5XX http errors
-				log.Errorf("%s: %s: %v", httpErr.ErrorCode().Error(), httpErr.Message, httpErr.Detail)
-			} else {
-				log.Infof("%s: %s: %v", httpErr.ErrorCode().Error(), httpErr.Message, httpErr.Detail)
-			}
+		serveError(log, w, err)
+	}
+}
+
+func serveError(log ctxu.Logger, w http.ResponseWriter, err error) {
```
riyazdf: thanks for factoring this out :+1:

utils/http.go
```diff
@@ -112,6 +176,20 @@ func buildAccessRecords(repo string, actions ...string) []auth.Access {
 	return requiredAccess
 }
 
+func buildCatalogRecord(actions ...string) []auth.Access {
+	requiredAccess := make([]auth.Access, 0, len(actions))
+	for _, action := range actions {
+		requiredAccess = append(requiredAccess, auth.Access{
+			Resource: auth.Resource{
+				Type: "registry",
+				Name: "catalog",
```
riyazdf: what is `catalog` in this context?  All images?  Does notary know how to handle this already?

utils/http.go
```diff
@@ -28,74 +29,137 @@ type rootHandler struct {
...
+		return vars["imageName"], nil
+	case ImageInQueryString:
+		qs := r.URL.Query()
+		imageName := qs.Get("filter")
+		if imageName == "" {
+			// no image name means global feed
+			imageName = "*"
+		}
+		return imageName, nil
+	case NoImageName:
```
riyazdf: nit: I'm wondering if this should be renamed to something else, like `NoImageUsed`?  I'm proposing this because of the `if imageName == ""` in the case above

endophage: This doesn't indicate no image was passed, it indicates no image was expected. The `imageName == ""` above is to set a default value when we expected an image name and none was explicitly given (and in the changefeed situation I put that in for, it's desirable to make that assumption when no name is given).

I opened issue #1021 to look at refactoring how we do auth in general because this is only one of the things I wasn't very happy with, regardless of how it's named.

server/storage/memory.go
endophage: `pageSize` can be `ParseUint`, but the `changeID` can be `-1` to indicate "start from the most recent change", in which case, I'm doing to add a check to ensure if the `changeID` is `-1` the user has also requested the results to be reversed, otherwise it doesn't make sense.

riyazdf: ok that makes more sense, sounds good 👍

utils/http.go
endophage: The auth lib and the token servers know how to handle this. According to @jlhawn this is the correct resource to request for listing all repos.

endophage: After more discussion with @jlhawn, this resource should always be requested with the "*" permission, so I've hard coded that into this function. The configured "pull" action will now only be used when requesting access to the changefeed of a specific repo.

migrations/server/mysql/0005_changefeed.up.sql
```diff
@@ -0,0 +1,27 @@
...
+        `gun`,
+        `version`,
+        `sha256`
+    FROM
+        `tuf_files`
+    WHERE
+        `role` = "timestamp"
+    ORDER BY
+        `created_at` ASC
+);
```
HuKeping: `No newline` warning lol

migrations/server/postgresql/0002_changefeed.up.sql
```diff
@@ -0,0 +1,27 @@
...
+        "gun",
+        "version",
+        "sha256"
+    FROM
+        "tuf_files"
+    WHERE
+        "role" = 'timestamp'
+    ORDER BY
+        "created_at" ASC
+);
```
HuKeping: same

server/handlers/changefeed.go
```diff
@@ -0,0 +1,82 @@
+package handlers
+
+import (
+	"encoding/json"
+	"net/http"
+	"strconv"
+	"strings"
+
+	ctxu "github.com/docker/distribution/context"
```
HuKeping: how about putting this import line along with those `github` package.

server/storage/sql_models.go
```diff
@@ -17,6 +21,20 @@ func (g TUFFile) TableName() string {
 	return "tuf_files"
 }
 
+// Change defines the the fields required for an object in the changefeed
+type Change struct {
+	ID        uint `gorm:"primary_key" sql:"not null"`
+	CreatedAt time.Time
+	GUN       string `gorm:"column:gun" sql:"type:varchar(255);not null"`
+	Version   int    `sql:"not null"`
+	Sha256    string `sql:"type:varchar(64);"`
```
HuKeping: We might want to consistent whether to use SHAXXX or ShaXXX, we use both at present. But it can be totally leave to another separated PR.

utils/http.go
```diff
@@ -28,74 +29,137 @@ type rootHandler struct {
 	context context.Context
 	trust   signed.CryptoService
 	//cachePool redis.Pool
+	imageNameAt ImageNameLocation
+}
+
+// ImageNameLocation is used to define where the auth code should read the image name
+// from. For most repo operations, it will be ImageInURL, i.e. it's the segments between
+// `/v2/` and `/_trust`. For the changefeed however, an image name is optional, and if
```
HuKeping: i.e., it's the segments between "/v2/" and "/_trust" - `/v2/docker.io/official/busybox/_trust/_tuf/targets.json`.

utils/http.go
```diff
@@ -112,6 +176,21 @@ func buildAccessRecords(repo string, actions ...string) []auth.Access {
 	return requiredAccess
 }
 
+// buildCatalogRecord returns the only valid format for the catalog
+// resource. Only admins can get this access level from the token
+// server.
+func buildCatalogRecord(actions ...string) []auth.Access {
+	requiredAccess := []auth.Access{{
+		Resource: auth.Resource{
+			Type: "registry",
```
HuKeping: Why use a different `Type` here with what we have in `buildAccessRecords()`, and does the `Type` matter to the whole auth process?

migrations/server/mysql/0005_changefeed.up.sql
HuKeping: add a commit efdf807 to fix it

migrations/server/postgresql/0002_changefeed.up.sql
HuKeping: added a commit efdf807 to fix it

HuKeping:
It would be great if also update the docs!

utils/http.go
```diff
@@ -28,74 +29,137 @@ type rootHandler struct {
...
+// Constants for different image locations
+const (
+	ImageInURL = iota
+	ImageInQueryString
+	NoImageName
+)
+
+func parseImageName(loc ImageNameLocation, r *http.Request) (string, error) {
+	switch loc {
+	case ImageInURL:
```
HuKeping: Anyway we can also test the case since we have already tested the other two.

utils/http.go
```diff
@@ -28,74 +29,137 @@ type rootHandler struct {
...
+		vars := mux.Vars(r)
+		return vars["imageName"], nil
+	case ImageInQueryString:
+		qs := r.URL.Query()
+		imageName := qs.Get("filter")
```
HuKeping: nit: do we decide to use `filter` because it seems more like a general purporse

HuKeping:
seems ready moving to code review

HuKeping:
commit 4552ede adds a test for rethinkdb

utils/http.go
endophage: Yeah. I'm anticipating adding suffix wildcarding at some point, like `docker.io/library/*`. There are some token permissions stuff that needs to be solved in the token server before we can do that though.

utils/http.go
endophage: It has coverage as part of other tests. I can take another look, `mux` is just a pain to work with for tests.

utils/http.go
endophage: This is basically the only valid form for the catalog record as far as Docker's token server is concerned. Docker's token server will only ever return a catalog record with these Resource field values.

server/storage/sql_models.go
endophage: Yep, we should follow Go's standard. "SHA" is technically an acronym so it shold either be all caps, `SHA`, or all lower case, `sha`, depending on whether it needs to be exported or not. I'll fix it for this new struct and file an issue for us to go review other places.

migrations/server/mysql/0005_changefeed.up.sql
endophage: Thanks!

server/handlers/changefeed.go
endophage: Ah, actually when I went and looked at it, this is in the right place. We've been nominally (and we should document it somewhere) stealing Python's PEP8 import formatting, which requires imports be broken into 3 blocks of:
```
import (
    // Std library imports
    "os" // etc...

    // 3rd party dependencies
    "github.com/sirupsen/logrus" // etc...

    // dependencies within the same project
    "github.com/docker/notary" // etc...
)
```

server/storage/sql_models.go
endophage: For fixing other locations: #1024

endophage:
@HuKeping hmmm, I've just realized we don't actually have server API docs... Docker is also in the process of changing how it does docs and we need to setup a microsite for Notary (which will contain notary specific docs with no relationship to the docker integration).

server/handlers/changefeed.go
HuKeping: OK, then we might need to make another PR to do the cleanup, I've supposed it was like:
```
import (
        // std
        "os"

        // pre_1
        "github.com/..."

        // pre_2
        "golang.org/..."
)
```

HuKeping:
> @HuKeping hmmm, I've just realized we don't actually have server API docs... Docker is also in the process of changing how it does docs and we need to setup a (Github Pages) microsite for Notary (which will contain notary specific docs with no relationship to the docker integration

Maybe file an issue to track this?

server/storage/memory.go
```diff
@@ -158,6 +181,97 @@ func (st *MemStorage) Delete(gun string) error {
...
+		start     = int(id)
+		toInspect []Change
+	)
+	if err != nil {
+		return nil, err
+	}
+	if len(st.changes) <= int(id) && !reversed {
+		// no records to return as we're essentially trying to retrieve
+		// changes that haven't happened yet.
+		return nil, nil
```
endophage: I had that originally but I decided it should be OK to request a range that doesn't exist with the expected result being an empty list. I created `ErrBadChangeID` to indicate the ID is unrecognizable as an ID.

server/storage/memory.go
```diff
@@ -71,9 +75,25 @@ func (st *MemStorage) UpdateCurrent(gun string, update MetaUpdate) error {
...
+		st.writeChange(gun, update.Version, checksum)
+	}
 	return nil
 }
 
+// writeChange must only be called by a function already holding a lock on
+// the MemStorage. Behaviour is undefined otherwise
+func (st *MemStorage) writeChange(gun string, version int, checksum string) {
+	c := Change{
+		ID:        uint(len(st.changes) + 1),
```
endophage: Not quite, the migration sql backfills the changefeed table.

server/server.go
```diff
@@ -186,11 +193,18 @@ func RootHandler(ctx context.Context, ac auth.AccessController, trust signed.Cry
 		ServerHandler:       handlers.DeleteHandler,
 		PermissionsRequired: []string{"*"},
 	}))
+	r.Methods("GET").Path("/v2/changefeed/_trust").Handler(createHandler(_serverEndpoint{
```
endophage: `/v2/{imageName:.*}/_trust/changefeed/` -> what happens if imageName is `""`

server/storage/interface.go
```diff
@@ -39,4 +39,15 @@ type MetaStore interface {
...
+	// GetChanges returns an ordered slice of changes. It starts from
+	// the change matching changeID, but excludes this change from the results
+	// on the assumption that if a user provides an ID, they've seen that change.
+	// If changeID is 0, it starts from the
+	// beginning, and if changeID is -1, it starts from the most recent
+	// change. The number of results returned is limited by pageSize.
+	// Reversed indicates we are fetching pages going backwards in time, the
+	// default being to fetch pageSize from changeID going forwards in time.
+	// The returned []Change should always be ordered oldest to newest.
+	GetChanges(changeID string, pageSize int, filterName string, reversed bool) ([]Change, error)
```
endophage: ChangeID should be checksum, must have ID to request most recent (i.e. empty == oldest)

server/storage/sqldb.go
```diff
@@ -54,31 +56,62 @@ func (db *SQLStorage) UpdateCurrent(gun string, update MetaUpdate) error {
...
+	}
+
+	// If we're publishing a timestamp, update the changefeed as this is
+	// technically an new version of the TUF repo
+	if update.Role == data.CanonicalTimestampRole {
+		if err := db.writeChangefeed(tx, gun, update.Version, hexChecksum); err != nil {
+			return rb(err)
+		}
+	}
+	return tx.Commit().Error
```
endophage: Put queries into anon func and call rollback once at the end.

server/server.go
HuKeping: I suppose the mux.Vars()["imageName"] would be ""

server/handlers/changefeed.go
```diff
@@ -56,24 +56,23 @@ func changefeed(logger ctxu.Logger, store storage.MetaStore, imageName, changeID
...
 	if err != nil {
-		logger.Errorf("400 GET invalid pageSize: %s", ps)
+		logger.Errorf("400 GET invalid pageSize: %s", r)
 		err = errors.ErrInvalidParams.WithDetail("invalid pageSize parameter, must be an integer >= 0")
 		return
 	}
 	if pageSize == 0 {
 		pageSize = notary.DefaultPageSize
 	}
-	reversed = rev == "1" || strings.ToLower(rev) == "true"
```
HuKeping: The last ref to pkg `stings` cause the build error, will fix.

server/storage/interface.go
```diff
@@ -49,5 +49,5 @@ type MetaStore interface {
 	// Reversed indicates we are fetching pages going backwards in time, the
 	// default being to fetch pageSize from changeID going forwards in time.
 	// The returned []Change should always be ordered oldest to newest.
-	GetChanges(changeID string, pageSize int, filterName string, reversed bool) ([]Change, error)
+	GetChanges(changeID string, records int, filterName string) ([]Change, error)
```
HuKeping: seems forget updating the one for rethinkDB, will fix.

cyli:
Non-blocking: we mentioned storing delete repo events in the table, but since it doesn't reference the timestamp, we weren't sure how that would be stored.  Supporting delete repository events could also be addressed in another PR.  I just wanted to leave a note somewhere so we remember.

endophage:
@cyli I'll get that in here 👍

server/storage/interface.go
```diff
@@ -39,4 +39,15 @@ type MetaStore interface {
 	// Delete removes all metadata for a given GUN.  It does not return an
 	// error if no metadata exists for the given GUN.
 	Delete(gun string) error
+
+	// GetChanges returns an ordered slice of changes. It starts from
+	// the change matching changeID, but excludes this change from the results
+	// on the assumption that if a user provides an ID, they've seen that change.
+	// If changeID is 0, it starts from the
```
cyli: IIRC, we discussed not having a "latest" indicator - but a change ID can be empty for "the beginning"?  I wasn't sure if that was going to be the eventual goal or if it would apply to the current changes - if so, we should probably modify the comment.

server/server.go
```diff
@@ -144,21 +149,21 @@ func RootHandler(ctx context.Context, ac auth.AccessController, trust signed.Cry
 	r := mux.NewRouter()
 	r.Methods("GET").Path("/v2/").Handler(authWrapper(handlers.MainHandler))
 
-	r.Methods("POST").Path("/v2/{imageName:.*}/_trust/tuf/").Handler(createHandler(_serverEndpoint{
+	r.Methods("POST").Path("/v2/{imageName:[^*]+}/_trust/tuf/").Handler(createHandler(_serverEndpoint{
```
endophage: I'm inclined to use the simplest thing possible until we have a use case that requires partially wildcard GUNs.

server/handlers/changefeed_test.go
```diff
@@ -0,0 +1,196 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+type changefeedArgs struct {
+	logger    ctxu.Logger
+	store     storage.MetaStore
+	imageName string
+	changeID  string
+	pageSize  int64
+	reversed  bool
```
endophage: 👍 thanks!

server/storage/interface.go
endophage: Hmmm, I thought we were going to keep the latest indicator when we decided to not use the checksums as indices because "delete" wouldn't have a checksum.

And yes, an empty change should also mean "beginning", will fix that.

utils/http.go
```diff
@@ -27,7 +27,16 @@ type rootHandler struct {
 	actions []string
 	context context.Context
 	trust   signed.CryptoService
-	//cachePool redis.Pool
+}
+
+func parseImageName(r *http.Request) string {
+	vars := mux.Vars(r)
```
endophage: Slept on it, I think with the updates to the URL routes switching from `*` to `+` in the regexes, this is OK because only the changefeed will be able to produce an empty image name.

server/storage/interface.go
cyli: @endophage Ah right, forgot about that part of the discussion.  I'm not sure how to indicate that then with a string, if we want to keep them opaque IDs... if they always had to evaluate to an integer, to indicate "latest", "earliest", or otherwise a totally valid change ID, would it make sense for them to just be integers instead of strings, if they always had to evaluate that way?

endophage: I think we an probably safely say that `-1` will never be a valid ID so we just keep that as the official "latest" indicator. The literal characters in the query string don't change between `-1` the integer, and `-1` the string, it's up to the server to interpret it.

utils/http.go
endophage: Phew! I thought in my med induced malaise I may have done something really daft :-P

cyli:
Nope, it was my sleep-deprived haze :D

server/handlers/changefeed_test.go
```diff
@@ -0,0 +1,191 @@
...
+			args: args{
+				logger: logrus.New(),
+				s:      s,
+				ps:     "10",
+			},
+			wantStore:    s,
+			wantPageSize: 10,
+		},
+		{
+			name: "Reversed \"1\"",
```
cyli: I think this case is the same as the previous one and the next 3 - it looks like all the  "Reversed" cases, except for the last one which tests the negative page size, can probably be removed.

server/storage/sql_models.go
```diff
@@ -1,6 +1,15 @@
...
 
-import "github.com/jinzhu/gorm"
+import (
+	"time"
+
+	"github.com/jinzhu/gorm"
+)
+
+const (
+	changeCategoryUpdate   = "update"
```
cyli: Non-blocking:  Can be done in the future when we actually have a consumer for this as a library, but should these be a type and exported, so someone using the API can check one against the other?

server/storage/errors.go
```diff
@@ -40,3 +40,13 @@ type ErrNoKey struct {
...
 	return fmt.Sprintf("Error, no timestamp key found for %s", err.gun)
 }
+
+// ErrBadChangeID indicates the change ID provided by the user is not
+// valid for some reason, i.e. it is out of bounds
+type ErrBadChangeID struct {
+	id string
+}
+
+func (err ErrBadChangeID) Error() string {
```
riyazdf: I think I had commented somewhere previously but did we want to use this for the case when a user requested a yet-to-exist change ID?

endophage: Given the change ID is an index, I treated it as essentially "there's nothing in the range you requested", and returned an empty result rather than an error. Especially in the database cases, it's actually much easier to do that than run extra queries to determine if the ID is within the existing range.

server/storage/memory_test.go
```diff
@@ -90,3 +91,135 @@ func TestGetChecksumNotFound(t *testing.T) {
...
+			Version: 3,
+			Data:    []byte{},
+		},
+		{
+			Role:    data.CanonicalTimestampRole,
+			Version: 4,
+			Data:    []byte{},
+		},
+	}))
+	require.NoError(t, s.UpdateMany("busybox", []MetaUpdate{
```
cyli: Was wondering if we can insert a `Delete` here or after the busybox updates?  So that we can test that the delete events get returned, etc.  Also, I think this is the same as the sql tests?  (creating the same changes, testing the change feed, etc.) - would it make sense to move the implementation into `tuf_store_test.go` (or we can rename that file or something) as `testGetChanges`, and then invoke it with either a memory store or a sql store?

Similarly, can we add a `Delete` of a GUN that doesn't exist to the end to ensure that no change object is added in that case?

cyli: Also, maybe a `s.UpdateCurrent` with a timestamp update, and `s.UpdateCurrent` with a snapshot update?  To ensure the timestamp update creates a change log, but the snapshot update does not.

cyli: Oops sorry, I didn't look far enough - not `tuf_store_test` - `storage_test`.   Looks like we have a delete test in there - could add the check of the change feed in that test instead of in this one.  There is also a rollback test in there - could we add an assertion that in the rollback case, no change log is created for that update?

cyli:
>>@HuKeping hmmm, I've just realized we don't actually have server API docs... Docker is also in the process of changing how it does docs and we need to setup a (Github Pages) microsite for Notary (which will contain notary specific docs with no relationship to the docker integration

>Maybe file an issue to track this?

@HuKeping https://github.com/docker/notary/issues/796 from previous :)

@endophage Thank you for all your work on this!  It LGTM aside from a couple tests that would be awesome to add.

server/storage/memory_test.go
endophage: Done, done, and done! :-D

server/server.go
HuKeping: @cyli it seems `([^*]|.{2,})` will  just match out the first character if it is not `*`. It will not use the 2nd alternative unless the string is begin with `*`. For example `abcd`, only the `a` will be matched out, though `*abcd` will match the whole string.

server/storage/errors.go
endophage: Huh, not sure how I ended up not using this... I know I did at some point. I need to make one more tiny tweak to tell the JSON encoding to return the change ID as a string so I'll take this out at the same time.

HuKeping:
I'd like to  take a final review to it if it was not in a hurry to merge

server/handlers/changefeed.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	out, err := changefeed(logger, store, imageName, changeID, records)
+	if err == nil {
+		w.Write(out)
+	}
+	return err
+}
+
+func changefeed(logger ctxu.Logger, store storage.MetaStore, imageName, changeID string, records int64) ([]byte, error) {
+	changes, err := store.GetChanges(changeID, int(records), imageName)
```
HuKeping: I'm a little worried about the `int64` -> `int`, what about the records is over 2^32 - 1 and unfortunately we are on a 32 machine.

server/handlers/changefeed.go
```diff
@@ -0,0 +1,79 @@
...
+	store, ok := s.(storage.MetaStore)
+	if !ok {
+		logger.Error("500 GET unable to retrieve storage")
+		err = errors.ErrNoStorage.WithDetail(nil)
+		return
+	}
+	pageSize, err = strconv.ParseInt(r, 10, 32)
+	if err != nil {
+		logger.Errorf("400 GET invalid pageSize: %s", r)
+		err = errors.ErrInvalidParams.WithDetail("invalid records parameter, must be an integer >= 0")
```
HuKeping: I think the error will explain itself well, what we do now will lose the original error message which I think is worth to keep, for example it will return "value out of range" if the provided value is out of range.

server/handlers/changefeed.go
endophage: When I parse it from the query string I specifically parse as a 32 bit. It's not great but `strconv.ParseInt` always returned an `int64` :-(

server/handlers/changefeed.go
endophage: Updated to include the original error string. We need to add some detail otherwise it's not clear to the user which parameter was wrong.

server/handlers/changefeed.go
HuKeping: Ah right, I see the `strconv.ParseInt` , why not cast it in the function `checkChangefeedInputs` where we do the `strconv.ParseInt`

server/handlers/changefeed.go
HuKeping: Ah you are right :+1:

server/handlers/changefeed.go
endophage: I could, I don't have any particular reason why it ended up being done here.

endophage: The build is has been so flaky the last 48 hours I'm mildly afraid of making any more changes. I've filed issues to look at CI. Unless you feel really strongly about this I'm not going to change it with the tight timeline.

HuKeping: Nothing more at present , I think we can do it in the followed PRs if we would like to change anything.

HuKeping:
LGTM, thanks for all your work on this @endophage !

endophage:
@HuKeping @riyazdf @cyli Thanks for all the great review guys!

PR #1027, opened 2016-11-10
Try using vndr instead of Godeps, as it has a cleaner version pinning file

cyli:
[docker/docker](https://github.com/docker/docker) and [swarmkit](https://github.com/docker/swarmkit) both use [VNDR](https://github.com/LK4D4/vndr), and I've had a much easier time vendoring using it as opposed to Godeps, so this is an attempt to convert.

Mainly I wanted to see if CI passes, since vndr seems to clean up a lot of extra stuff in `vendor/`.  But also, if folks like `Godeps` instead, we can stay with godeps.  :)


CONTRIBUTING.md
```diff
@@ -87,4 +87,9 @@ Complying to these simple rules will greatly accelerate the review process, and
 ## Review and Development notes
 
 - All merges require LGTMs from any 2 maintainers.
-- We use the git flow model (as best we can) using the `releases` branch as the stable branch, and the `master` branch as the development branch.  When we get near a potential release, a release branch (`release/<semver>`) will be created from `master`.  Any PRs that should go into the release should be made against that branch.  Hotfixes for a minor release will be added to the branch `hotfox/<semver>`.
\ No newline at end of file
+- We use the git flow model (as best we can) using the `releases` branch as the stable branch, and the `master` branch as the development branch.  When we get near a potential release, a release branch (`release/<semver>`) will be created from `master`.  Any PRs that should go into the release should be made against that branch.  Hotfixes for a minor release will be added to the branch `hotfox/<semver>`.
```
riyazdf: nit typo that we seemed to have earlier `hotfix/<semver>`

cyli: oops, hahah!  thanks - that's a pretty funny typo though.

HuKeping:
I have a bad experience with `Godeps`, I'd like to have a quick taste for `vndr`

HuKeping:
what is blocking the status checking process?

cyli:
jenkins, test this please

cyli:
@HuKeping  :|  Not sure, it may have gotten stuck, even after I rebuilt.  Let me poke it again

cyli:
jenkins, retest this please

cyli:
retest this please

ecordell:
This seems like a good place to bring up: https://github.com/mattfarina/golang-broken-vendor

I have this problem when using notary as a lib - leaving `vendor` out of version control can fix it (but I might also argue this is a golang problem that should be fixed by the language). Just thought I'd bring it up in case you wanted to consider it.

cyli:
@HuKeping I think that particular checker is just broken for now :|  Or in the process of being migrated.

@ecordell:  By "using notary as a lib" do you mean vendoring it in to some other project?  I haven't done a lot with other go projects - I thought that most used some kind of vendoring tool that does the flattening, though.  It *would* be nice if golang provided a default vendoring tool that does this though, or yes, was able to handle the nested vendoring.

HuKeping:
@ecordell does `glide` can solve you problem? I know rkt is using `glide`.

endophage:
@ecordell that's an interesting problem. In that linked example, it's a legitimate problem when the "a" packages are at the same version, but what would be the correct behaviour if they were two different versions? I always figured the way vendoring was done to allow different packages to depend on different versions, but it clearly does cause problems even if the differences are only internal, within a strongly typed system.

cyli:
@ecordell I opened https://github.com/docker/notary/issues/1040 where we can continue discussing whether it's worthwhile to keep our vendor-ed code checked in.

Issue #1029, opened 2016-11-14
Speed up CI


endophage:
We need to work out how to make the CI run faster. It's at ~20 minutes right now which is especially painful with the flakiness we sometimes get.

Improvements so far:
- #1032 reduced CircleCI by ~ 2 mins per @cyli's estimate

endophage:
Considering this as a won't fix for now. We can open a new issue if and when we have time to look at it again.

PR #1032, opened 2016-11-14
Do not build the osx cross packages when running CI tests

cyli:
Hopefully this will shave at least 30 seconds from the CI times.
Also, bump up to go 1.7.3.

cc #1029


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.7.1
+FROM golang:1.7.3
```
endophage: If the OSXCross stuff is no longer in here, can this (and should this) use alpine now?

endophage: No worries, just noticed the comment about not using alpine due to osxcross.

cyli: Looks like it's https://github.com/golang/go/issues/14481 - so for our CI tests, which run with `-race` enabled, I don't think we can use alpine :|

PR #1034, opened 2016-11-14
fixing default records size for changefeed

endophage:


server/handlers/changefeed.go
```diff
@@ -67,15 +67,19 @@ func checkChangefeedInputs(logger ctxu.Logger, s interface{}, r string) (
...
+			err = errors.ErrInvalidParams.WithDetail(
+				fmt.Sprintf("invalid records parameter: %s", err.Error()),
+			)
+			return
+		}
+		if pageSize == 0 {
+			pageSize = notary.DefaultPageSize
+		}
+	} else {
 		pageSize = notary.DefaultPageSize
```
cyli: Non-blocking stylistic nit:  rather than an if/else, we just set `pageSize = notary.DefaultPageSize` above the if statement.  Happy to just get this in as is, though.

cyli: Do we also have a test for this when testing the changefeed inputs?

endophage: Will add test so don't merge yet :-)

Re. first comment, I can make that change too. Had this style drilled into me a long time ago so it's my go to. Happy to do it other ways.

cyli: @endophage I had the other one drilled into me.  :)  I don't feel particularly strongly.  This is fine too.  We can also do:

```
if r != "" {
  ...
}

if pageSize == 0 {
   pageSize = notary.DefaultPageSize
}
```

(moves the extra check for outside of the `if r != ""` block)

PR #1036, opened 2016-11-15
set max page size for changefeed

endophage:

### DIFF
--- a/const.go
+++ b/const.go
@@ -75,7 +75,7 @@ const (
 	// DefaultPageSize is the default number of records to return from the changefeed
 	DefaultPageSize = 100
 	// MaxPageSize is the maximum number of records that can be requested from
-	//  the changefeed in a single request
+	// the changefeed in a single request
 	MaxPageSize = 1000
 )
 
--- a/const.go
+++ b/const.go
@@ -74,6 +74,9 @@ const (
 
 	// DefaultPageSize is the default number of records to return from the changefeed
 	DefaultPageSize = 100
+	// MaxPageSize is the maximum number of records that can be requested from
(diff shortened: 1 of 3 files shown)


const.go
```diff
@@ -74,6 +74,9 @@ const (
 
 	// DefaultPageSize is the default number of records to return from the changefeed
 	DefaultPageSize = 100
+	// MaxPageSize is the maximum number of records that can be requested from
+	//  the changefeed in a single request
```
HuKeping: nit: how about just one space between the `//` and `the`

server/handlers/changefeed.go
```diff
@@ -77,9 +77,15 @@ func checkChangefeedInputs(logger ctxu.Logger, s interface{}, r string) (
...
 		}
-		if pageSize == 0 {
-			pageSize = notary.DefaultPageSize
-		}
+
+	}
+	switch {
+	case pageSize == 0:
+		pageSize = notary.DefaultPageSize
+	case pageSize > notary.MaxPageSize:
```
riyazdf: can we add tests to check for `MaxPageSize` and `-MaxPageSize`?

endophage: Definitely, was planning to. @cyli asked for them while I was on my way out of the office last night :)

endophage:
@HuKeping I'm also going to do you suggestion about having `checkChangefeedInputs` return an `int` rather than an `int64`

HuKeping:
Thanks for making that change @endophage !

endophage:
@cyli @riyazdf can we get another LGTM?

PR #1038, opened 2016-11-15
Pass the netgo tag when building static binaries

cyli:
- When building static binaries with CGO, cgo unix network segfaults
- When building dynamically linked binaries with CGO, everything is fine
- When building static binaries with CGO but with netgo enabled, dns lookups seem fine

So add the netgo tag when building static binaries.

https://github.com/prometheus/alertmanager/issues/304
https://github.com/golang/go/issues/7857

PR #1039, opened 2016-11-16
Force go networking instead of cgo for cross and static builds

riyazdf:
Identical fix to https://github.com/docker/notary/pull/1038 if we wanted to include it in a `v0.4.3` release.  Set up on a `hotfix/0.4.3` branch


riyazdf:
CircleCI will fail only on container 4 during the `$ buildscripts/circle_parallelism.sh (4)` due to golang lint updates -- I'd prefer not to keep them in this PR to minimize code changes for this hotfix.  

The rest of the tests pass, and here's an example output from `make cross`'s linux client to show the fix works:
```
# ./notary version
notary
 Version:    0.4.3
 Git commit: 485d5e1
# ./notary -D list test-repo
DEBU[0000] Configuration file not found, using defaults
DEBU[0000] Using the following trust directory: /root/.notary
ERRO[0000] could not reach https://notary-server:4443: Get https://notary-server:4443/v2/: dial tcp: lookup notary-server on 8.8.8.8:53: no such host
INFO[0000] continuing in offline mode
DEBU[0000] No yubikey found, using alternative key storage: no library found

* fatal: client is offline
```

cyli:
LGTM

PR #1065, opened 2016-12-12
notary microsite for stand-alone documentation

### LINKED ISSUE #1054: Create Notary gh-pages microsite
The goal is for the microsite to contain all notary specific documentation for people that want to use notary as an independent, stand-alone product.

The plan is:
- To use the /docs folder on the master branch method of deploying the gh-pages site. This allows people to update docs in their PRs when necessary.
- To use Jekyll so users do not have to build the site using some other site generator before submitting their PR, they only need to add/update .md files, which will subsequently get compiled into the site by GH.

endophage:
Closes #1054 

TODO:
- [X] Fix top nav nesting and layout
- [x] Fix table formatting on all pages
- [x] Fix H1 formatting
- [x] Fix code formatting
- [x] Fix bottom nav layout
- [X] Update docs readme


endophage:
This is not perfect in any way but it's a good start. I've made a couple of tweaks to the copy to distance it from the docker official docs but everything else is just layout and linking that had to be changed to work properly with Jekyll. I'm anticipating having to deal with some bugs once it's enabled for gh pages but I'm not sure how to test that right now...

Issue #1073, opened 2017-01-06
upgrade gorethink to v3


vikstrous:
gorethink v3 has some fixes that we would like to be able to use, but since we depend on notary, we can't use gorethink v3 ourselves. If both v2 and v3 are used at the same time, it causes a panic in protobufs:
```
panic: proto: duplicate enum registered: VersionDummy_Version
```
Please upgrade to gorethink v3.

cyli:
@vikstrous Does this need to be backported to a particular version series, or can it go into master?

cyli:
@vikstrous Nm, I checked slack and saw that it requires a backport.  :)

HuKeping:
Closes this via #1074 & #1075 , thanks @cyli for picking this up and thanks @vikstrous!

PR #1074, opened 2017-01-06
Bump gorethink dependency to v3

cyli:
This is for the master branch
Part of #1073 .


cmd/notary-server/config.go
```diff
@@ -25,7 +25,7 @@ import (
 	_ "github.com/lib/pq"
 	"github.com/spf13/viper"
 	"golang.org/x/net/context"
-	gorethink "gopkg.in/dancannon/gorethink.v2"
```
riyazdf: this is nifty, didn't realize we could omit this

server/storage/rethinkdb.go
```diff
@@ -32,7 +32,7 @@ func (r RDBTUFFile) TableName() string {
 	return "tuf_files"
 }
 
-// gorethink can't handle an UnmarshalJSON function (see https://github.com/dancannon/gorethink/issues/201),
+// gorethink can't handle an UnmarshalJSON function (see https://github.com/gorethink/gorethink/issues/201),
```
riyazdf: also thanks for updating this 👍

cmd/notary-server/config.go
cyli: I found out about it recently, because `goimports` keeps doing the wrong thing when I save (https://github.com/joefitzgerald/go-plus/issues/52), but apparently if there's a `.` or `-` or some other special character, everything after that gets ignored for naming?

HuKeping: Interesting :)

PR #1075, opened 2017-01-06
Backport bump gorethink to v3

### LINKED ISSUE #1073: upgrade gorethink to v3

cyli:
Fixes #1073.


cyli:
Thank you @HuKeping!

Issue #1069, opened 2017-01-04
Use GUN instead of ImageName to show that notary is not only for docker image


HuKeping:
Since Notary is not only used for signing Docker image, it can be used to allows anyone to have trust over arbitrary collections of data . So we should probably not use `imageName` as the  GUN ( Globally Unique Names ) to identify trust collections.

We should also add some other demo/document for the non-signing-docker-image use cases, although it is not the priority task.

See https://github.com/docker/notary/blob/master/server/server.go#L152  
and maybe some other place.

endophage:
I don't disagree with anything you've stated, but it's internal and most users will never see the URL templating. I'd accept a PR changing it to `gun` but I'm kind of indifferent.

If we mix it up in the docs anywhere, we'd definitely want to change it there.

HuKeping:
Aha, I'm not feeling very strong about it also, it's just when I reading some code and find that we are using the `imageName` where it seems should be better to use `gun`. Of course most user will not see that though.

ecordell:
This was done in #1081

PR #1081, opened 2017-01-10
Minor refactoring (nicer lib usage)

ecordell:
When using Notary as a library, I ran into some visibility issues that I've addressed here.

I'm a little uneasy about this PR because I can only justify it from my own experience, but I don't think there's anything egregious here. Personally, I think this changeset a slight readability improvement, regardless.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -98,15 +98,15 @@ func filterImagePrefixes(requiredPrefixes []string, err error, handler http.Hand
 	}
 
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		imageName := mux.Vars(r)["imageName"]
+		gun := mux.Vars(r)["gun"]
 
-		if imageName == "" {
+		if gun == "" {
 			handler.ServeHTTP(w, r)
 			return
 		}
 
 		for _, prefix := range requiredPrefixes {
-			if strings.HasPrefix(imageName, prefix) {
+			if strings.HasPrefix(gun, prefix) {
 				handler.ServeHTTP(w, r)
 				return
 			}
@@ -116,7 +116,7 @@ func filterImagePrefixes(requiredPrefixes []string, err error, handler http.Hand
 	})
 }
 
-// EndpointConfig stores settings used to create a Notary handler
+// EndpointConfig stores settings used to create a ServerHandler
 type EndpointConfig struct {
 	OperationName       string
 	ServerHandler       utils.ContextHandler
@@ -151,7 +151,7 @@ func RootHandler(ctx context.Context, ac auth.AccessController, trust signed.Cry
 
 	r := mux.NewRouter()
 	r.Methods("GET").Path("/v2/").Handler(authWrapper(handlers.MainHandler))
-	r.Methods("POST").Path("/v2/{imageName:[^*]+}/_trust/tuf/").Handler(CreateHandler(EndpointConfig{
+	r.Methods("POST").Path("/v2/{gun:[^*]+}/_trust/tuf/").Handler(CreateHandler(EndpointConfig{
 		OperationName:       "UpdateTUF",
 		ErrorIfGUNInvalid:   invalidGUNErr,
 		ServerHandler:       handlers.AtomicUpdateHandler,
@@ -159,7 +159,7 @@ func RootHandler(ctx context.Context, ac auth.AccessController, trust signed.Cry
 		AuthWrapper:         authWrapper,
 		RepoPrefixes:        repoPrefixes,
 	}))
-	r.Methods("GET").Path("/v2/{imageName:[^*]+}/_trust/tuf/{tufRole:root|targets(?:/[^/\\s]+)*|snapshot|timestamp}.{checksum:[a-fA-F0-9]{64}|[a-fA-F0-9]{96}|[a-fA-F0-9]{128}}.json").Handler(CreateHandler(EndpointConfig{
+	r.Methods("GET").Path("/v2/{gun:[^*]+}/_trust/tuf/{tufRole:root|targets(?:/[^/\\s]+)*|snapshot|timestamp}.{checksum:[a-fA-F0-9]{64}|[a-fA-F0-9]{96}|[a-fA-F0-9]{128}}.json").Handler(CreateHandler(EndpointConfig{
 		OperationName:       "GetRoleByHash",
 		ErrorIfGUNInvalid:   notFoundError,
 		IncludeCacheHeaders: true,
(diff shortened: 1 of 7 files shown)


cyli:
jenkins, test this please

server/server.go
```diff
@@ -133,84 +147,90 @@ func RootHandler(ctx context.Context, ac auth.AccessController, trust signed.Cry
...
-	}
-
 	invalidGUNErr := errors.ErrInvalidGUN.WithDetail(fmt.Sprintf("Require GUNs with prefix: %v", repoPrefixes))
 	notFoundError := errors.ErrMetadataNotFound.WithDetail(nil)
 
 	r := mux.NewRouter()
 	r.Methods("GET").Path("/v2/").Handler(authWrapper(handlers.MainHandler))
-
-	r.Methods("POST").Path("/v2/{imageName:[^*]+}/_trust/tuf/").Handler(createHandler(_serverEndpoint{
+	r.Methods("POST").Path("/v2/{imageName:[^*]+}/_trust/tuf/").Handler(CreateHandler(EndpointConfig{
```
HuKeping: How about changing the `imageName` to `gun` in this PR as well, see #1069

cyli:
jenkins, test this please

server/storage/sql_models.go
```diff
@@ -11,6 +11,16 @@ const (
 	changeCategoryDeletion = "deletion"
 )
 
+// TUFFileTableName returns the name used for the tuf file table
+var TUFFileTableName = func() string {
```
cyli: Non-blocking:  perhaps these should just be string constants?

utils/http.go
```diff
@@ -56,14 +59,13 @@ func (root *rootHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	ctx, w = ctxu.WithResponseWriter(ctx, w)
 	ctx = ctxu.WithLogger(ctx, log)
 	ctx = context.WithValue(ctx, notary.CtxKeyCryptoSvc, root.trust)
-
-	defer func() {
+	defer func(ctx context.Context) {
 		ctxu.GetResponseLogger(ctx).Info("response completed")
-	}()
+	}(ctx)
```
cyli: 👍 Thank you for this fix!

server/storage/sql_models.go
endophage: As they're not modifiable, yeah, they should be string consts.

endophage: Then again, they should also NEVER be used in more than one context (i.e. through the appropriate struct) so I actually prefer them being defined directly in the struct methods. It tightly binds them to the correct context.

utils/http.go
endophage: Is this correct? Don't we want to use the possibly modified context (which may contain additional info) when the function exits?

endophage:
I'm curious what the use case is here. My assumption is that you're adding additional http routes to the server (on the object returned by `RootHandler`), and want to apply the same `EntrypointConfig` over those additional routes?

utils/http.go
cyli: `doAuth` does not currently updates the logger, but I suppose it's possible in the future that it may.  I'm not sure if we necessarily care though about all the other fields in a "response completed" log.

cyli: You may be right and we may want the modified context if the logger gets updated in the future.  I have no strong opinions either way.  I had originally assumed we just wanted the base logger.  :)

riyazdf:
Piggy-backing off of @endophage's comment - I'm cool with making `createHandler` exported, but I want to make sure that if it's because Notary could benefit from a particular endpoint that we can add that functionality to the released server.

endophage:
On a related note, I'm wondering if `EntrypointConfig` has outlived its usefulness??? We construct a config inline to the `CreateHandler` function call only to immediately deconstruct it within the `CreateHandler` function. Should we just be passing the fields as arguments directly? Admittedly the argument list would be long-ish but it would be less obscure.

I think at some point (scraping my memory) we did it so you didn't end up with `CreateHandler(nil, nil, nil, thingICareAbout, nil, nil, nil, nil)`, but looking at the handlers, only one directly sets 5 fields, and the rest all appear to set 6+ out of a total 8 fields.

If we think it's still useful, maybe it makes sense to pull out the 3 fields every entrypoint is using: `OperationName`, `ServerHandler`, and `PermissionsRequired` and make those arguments to CreateHandler`

server/storage/sql_models.go
endophage: @ecordell I think these functions are the only thing that I really don't like in this PR. the rest of it I'm fine with (I've been waiting to see if you wanted to make the EndpointConfig updates here).

I really preferred them attached to the struct, it keeps it very clear what it's appropriate to use them for. However, if you want to break them out, it makes a lot more sense to just make them `const`'s.

HuKeping: Agreed with @endophage , I don't see the explicit benefit we can from these functions, it's not that directly and clear.

ecordell: Didn't mean to let this stagnate, I've been trying to evaluate which of these changes are really needed.

I was thinking that the functions might help for threshold "witnessing" (or whatever the term is now) by creating a different table for "staged" metadata. But since that design is still very up in the air I will change this and update the PR.

ecordell:
Took a while to get back around to this, but I believe I've addressed everything now!

server/server.go
```diff
@@ -116,13 +116,27 @@ func filterImagePrefixes(requiredPrefixes []string, err error, handler http.Hand
 	})
 }
 
-type _serverEndpoint struct {
+// EndpointConfig stores settings used to create a ServerHandler
+type EndpointConfig struct {
```
endophage: We can delete this now right?

server/server.go
```diff
@@ -116,13 +116,27 @@ func filterImagePrefixes(requiredPrefixes []string, err error, handler http.Hand
...
 	ServerHandler       utils.ContextHandler
 	ErrorIfGUNInvalid   error
 	IncludeCacheHeaders bool
 	CacheControlConfig  utils.CacheControlConfig
 	PermissionsRequired []string
+	AuthWrapper         utils.AuthWrapper
+	RepoPrefixes        []string
+}
+
+// CreateHandler creates a server handler from an EndpointConfig
```
endophage: Comment needs update now that we don't take an EndpointConfig any more

server/server.go
ecordell: Yup, I think the rebase added it back...

ecordell: fixed

server/server.go
ecordell: fixed

server/storage/sql_models.go
```diff
@@ -11,6 +11,12 @@ const (
 	changeCategoryDeletion = "deletion"
 )
 
+// TUFFileTableName returns the name used for the tuf file table
+var TUFFileTableName = "tuf_files"
```
endophage: `const`

ecordell: fixed

endophage:
Jenkins had a random Java failure. I've given it a kick. I'm good to go on this once Jenkins goes green.

endophage:
@cyli @riyazdf @HuKeping can we get another LGTM?

cyli:
LGTM!

Issue #997, opened 2016-10-05
Create types for GUN and role names, use them everywhere


endophage:
Create two new types, should probably be located in `tuf/data/?`

```
type GUN string
type RoleName string
```

Use these types everywhere!

We have GUNs and roles being used beside eachother in lots of locations and it would be beneficial  to have stronger differentiation, especially when a function return signature has something like `..., string, string, ...`. We've been relatively (completely?) consistent so far in putting gun first, role second, but it's non-obvious which is which to a newcomer.

HuKeping:
:+1:

grantleehoffman:
@endophage i've started tackling this issue. Realized there were more changes to be made than I originally perceived. I have replaced most of the String types with GUN and RoleName types. Having a few issues with a couple of tests. Hoping someone more familiar with the code base could help me finish this off. https://github.com/grantleehoffman/notary/tree/type_updates

endophage:
@grantleehoffman hello and thank you for picking this up. Can you submit your branch in as a PR with a "WIP: " (Work In Progress) prefix on the title. That'll help us collectively see which tests are having problems and coordinate feedback rather than multiple maintainers individually digging in to it.

ecordell:
Addressed by #1087

PR #1087, opened 2017-01-25
Type updates

grantleehoffman:
This PR is for Issue #997  
I'm having some issues narrowing down the test failures. I am also seeing a large number of output errors while running `make test`. Hoping some of you more familiar with the codebase could help me narrow down the cause of the issues. I apologize in advance for the large number of changes in this PR. What I thought would be a small change turned into a very large project, touching a lot of the existing code. Thanks in advance!


grantleehoffman:
Will fix the signed commits now

ecordell:
Haven't had a chance to check out the cause of the errors yet, but I do think that metadata store should be updated to take `RoleName` instead of `string`

```golang
type MetadataStore interface {
	GetSized(name string, size int64) ([]byte, error)
	Set(name string, blob []byte) error
	SetMulti(map[data.RoleName][]byte) error
	RemoveAll() error
	Remove(name string) error
}
```

This interface is pretty ubiquitous so that's not the smallest change.

tuf/data/types.go
```diff
@@ -17,6 +18,53 @@ import (
...
 
+// GUN type for specifying gun
+type GUN string
+
+func (g GUN) String() string {
+	return string(g)
+}
+
+// NewGUN generates a GUN object from the provided string
+func NewGUN(g string) GUN {
```
ecordell: I'm not sure I see how `NewGun("gun")` is better than `GUN("gun")`, which you get for free

tuf/data/types.go
```diff
@@ -262,10 +310,10 @@ func NewDelegations() *Delegations {
 
 // These values are recommended TUF expiry times.
 var defaultExpiryTimes = map[string]time.Duration{
```
ecordell: Shouldn't this be `map[RoleName]time.Duration`?

client/backwards_compatibility_test.go
```diff
@@ -278,7 +278,7 @@ func Test0Dot3RepoFormat(t *testing.T) {
 	delegations, err := repo.GetDelegationRoles()
 	require.NoError(t, err)
 	require.Len(t, delegations, 1)
-	require.Equal(t, "targets/releases", delegations[0].Name)
+	require.Equal(t, "targets/releases", string(delegations[0].Name))
```
ecordell: This is a nit, but since we're lifting types up anyway, I think this makes more sense as:

```golang
require.Equal(t, data.GUN("targets/releases"), delegations[0].Name)
```

Actually now that the datatype is `RoleName` I think `delegations[0].RoleName` would be clearer, but that could be a later PR.

client/backwards_compatibility_test.go
```diff
@@ -300,15 +300,15 @@ func Test0Dot3RepoFormat(t *testing.T) {
...
 	passwd := "randompass"
 
 	metaCache, err := store.NewFileStore(
-		filepath.Join("../fixtures/compatibility/notary0.1/tuf", filepath.FromSlash(gun), "metadata"),
+		filepath.Join("../fixtures/compatibility/notary0.1/tuf", filepath.FromSlash(gun.String()), "metadata"),
 		"json")
 	require.NoError(t, err)
 
-	ts := readOnlyServer(t, metaCache, http.StatusNotFound, gun)
+	ts := readOnlyServer(t, metaCache, http.StatusNotFound, gun.String())
```
ecordell: Why not make `readOnlyServer` take a `data.GUN` instead of a string?

client/changelist/change.go
```diff
@@ -37,15 +37,15 @@ type TUFChange struct {
...
+	RoleName data.RoleName `json:"role"`
 }
 
 // NewTUFChange initializes a TUFChange object
-func NewTUFChange(action string, role, changeType, changePath string, content []byte) *TUFChange {
+func NewTUFChange(action string, role data.RoleName, changeType, changePath string, content []byte) *TUFChange {
 	return &TUFChange{
 		Actn:       action,
-		Role:       role,
+		Role:       role.String(),
```
ecordell: `TUFChange` should probably use a `RoleName` here, and thread the type through the rest of the change operations.

client/client.go
```diff
@@ -392,10 +393,11 @@ func (r *NotaryRepository) ListTargets(roles ...string) ([]*TargetWithRole, erro
 	}
 
 	if len(roles) == 0 {
```
ecordell: I think this method should just take a list of `RoleName`s instead

client/client.go
```diff
@@ -568,7 +571,7 @@ func (r *NotaryRepository) ListRoles() ([]RoleWithSignatures, error) {
 		case data.CanonicalRootRole:
 			roleWithSig.Signatures = r.tufRepo.Root.Signatures
 		case data.CanonicalTargetsRole:
-			roleWithSig.Signatures = r.tufRepo.Targets[data.CanonicalTargetsRole].Signatures
+			roleWithSig.Signatures = r.tufRepo.Targets[data.CanonicalTargetsRole.String()].Signatures
```
ecordell: Same: `Targets` should be a map with `RoleName`s as Key

tuf/data/types.go
grantleehoffman: I will change to GUN("gun"), I was just trying to keep with convention. Will make the change thanks for the feedback :+1:

grantleehoffman:
@ecordell Thanks the feedback. I was starting to get a bit uneasy with the number of changes as I dug deeper. I decided to try to limit the number of changes I was making to make it manageable to review. I'll go ahead and continue changing out all of the gun and roles strings for the entire project.

endophage:
Agreed with @ecordell. Let's keep it to straight type changes here but hit all of them. Even if it's long that's still pretty easy for us to review. We can do separate PR for changes like the `Name` -> `RoleName` field change suggested.

grantleehoffman:
:+1:

client/client.go
grantleehoffman: done

grantleehoffman:
Still having issues with test `TestUpdateSucceedsEvenIfCannotWriteExistingRepo` client/client_update_test.go:143

tuf/testutils/swizzler.go
```diff
@@ -105,17 +104,19 @@ func signedFromStore(cache store.MetadataStore, role string) (*data.Signed, erro
...
 // mapping of roles to initial metadata bytes, and a cryptoservice
-func NewMetadataSwizzler(gun string, initialMetadata map[string][]byte,
+func NewMetadataSwizzler(gun data.GUN, initialMetadata map[data.RoleName][]byte,
 	cryptoService signed.CryptoService) *MetadataSwizzler {
 
-	var roles []string
-	for roleName := range initialMetadata {
-		roles = append(roles, roleName)
+	var roles []data.RoleName
+	metadata := make(map[data.RoleName][]byte)
```
cyli: @grantleehoffman The client update test that's failing (and possibly others) assume that as the data in the server data is swizzled, the metadata map is updated, because previously we were just mutating the initial metadata map.  Here a new one is created, so the original is never updated.

2 ways to fix it are to either make the original map mutable, or in the test, iterate over the roles in the swizzler and call `GetSized` on the swizzler's cache instead of iterating over the metadata map.

grantleehoffman: Thanks for the feedback. Fixing

client/backwards_compatibility_test.go
grantleehoffman: 👍

client/backwards_compatibility_test.go
grantleehoffman: 👍

client/changelist/change.go
grantleehoffman: 👍

grantleehoffman: @ecordell all of the mentioned changes have been made

cyli:
jenkins, test this please

client/client.go
grantleehoffman: done

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 + The keys on disk are now stored in the `~/.notary/private` directory, rather than in a key hierarchy that separates them by GUN and by role.  Notary will automatically migrate old-style directory layouts to the new style.  **This is not forwards-compatible against notary<0.4.2 and docker<=1.12** [#872](https://github.com/docker/notary/pull/872)
 + A new changefeed API has been added to Notary Server. It is only supported when using one of the relational database backends: MySQL, PostgreSQL, or SQLite.[#1019](https://github.com/docker/notary/pull/1019)
 
+## [v0.4.3](https://github.com/docker/notary/releases/tag/v0.4.3) 1/3/2017
```
endophage: Can you reset this file against master. There shouldn't be any changes in here and I'm not sure why GH is picking up a change here, it looks unchanged vs master:

```
## [v0.4.3](https://github.com/docker/notary/releases/tag/v0.4.3) 1/3/2017
+ Fix build tags for static notary client binaries in linux [#1039](https://github.com/docker/notary/pull/1039)
+ Fix key import for exported delegation keys [#1067](https://github.com/docker/notary/pull/1067)
```

MAINTAINERS
```diff
@@ -15,6 +15,7 @@
 			"diogomonica",
 			"dmcgowan",
 			"endophage",
+			"ecordell",
```
endophage: Same here, please reset this file against master.

endophage:
I'm not sure why GH is showing your branch as adding things unrelated to your change that are already in the master branch. Can you try a rebase against master and see if that clears those items out?

client/backwards_compatibility_test.go
```diff
@@ -80,7 +80,7 @@ func Test0Dot1Migration(t *testing.T) {
 	require.NoError(t, err)
 	require.NoError(t, recursiveCopy("../fixtures/compatibility/notary0.1", tmpDir))
 
-	gun := "docker.com/notary0.1/samplerepo"
+	gun := data.GUN("docker.com/notary0.1/samplerepo")
```
endophage: Rather than casting in cases like this, it's slightly more idiomatic to declare the type on the variable and go will deal with ensuring the literal has the correct type:

```
var gun data.GUN = "docker.io/notary0.1/samplerepo"
```

client/client.go
```diff
@@ -252,47 +252,48 @@ func (r *NotaryRepository) initializeRoles(rootKeys []data.PublicKey, localRoles
 
 	// we want to create all the local keys first so we don't have to
 	// make unnecessary network calls
-	for _, role := range localRoles {
+	for _, roleName := range localRoles {
```
endophage: Try to avoid variable name changes like this. The change in name doesn't really add much to the readability of the code and creates noise in the PR.

tuf/data/types.go
```diff
@@ -26,6 +64,15 @@ func (k SigAlgorithm) String() string {
 
 const defaultHashAlgorithm = "sha256"
 
+// NotaryDefaultExpiries is the construct used to configure the default expiry times of
+// the various role files.
+var NotaryDefaultExpiries = map[RoleName]time.Duration{
+	RoleName("root"):      notary.NotaryRootExpiry,
```
endophage: Don't need to do casts, Go works it out from the `map` definition, can just do `"root": notary.NotaryRootExpiry,`

https://play.golang.org/p/uzOhh7BDsN

endophage: Actually, if this is being touched, we should just convert it to use `CanonicalRootRole` etc...

tuf/data/types.go
```diff
@@ -108,7 +155,7 @@ type Signature struct {
 
 // Files is the map of paths to file meta container in targets and delegations
 // metadata files
-type Files map[string]FileMeta
+type Files map[RoleName]FileMeta
```
endophage: The string in this case is a target name, not a role name. This change needs to be reverted.

CHANGELOG.md
grantleehoffman: Sorry, I think I merged in master to pull in current as per my usual development cycle. Will fix, same for all other duplicated changes

client/backwards_compatibility_test.go
grantleehoffman: Changing

tuf/data/types.go
grantleehoffman: :+1:

client/changelist/change.go
```diff
@@ -58,8 +58,8 @@ func (c TUFChange) Action() string {
 }
 
 // Scope returns c.Role
-func (c TUFChange) Scope() string {
-	return c.Role
+func (c TUFChange) Scope() data.RoleName {
+	return data.RoleName(c.Role)
```
endophage: `c.Role` should already have the type `data.RoleName` per the struct definition.

client/client.go
```diff
@@ -33,14 +33,14 @@ const (
 )
 
 func init() {
-	data.SetDefaultExpiryTimes(notary.NotaryDefaultExpiries)
+	data.SetDefaultExpiryTimes(data.NotaryDefaultExpiries)
```
endophage: This actually needs to be `notary.NotaryDefaultExpires`. The values in `data.NotaryDefaultExpires` are the recommended TUF expiries, provided as a convenience for people that want to use Notary as a library. The values in `notary.NotaryDefaultExpires` are our opinionated expiry times that we're actively using.

client/client.go
```diff
@@ -274,25 +275,25 @@ func (r *NotaryRepository) initializeRoles(rootKeys []data.PublicKey, localRoles
 			)
 		}
 	}
-	for _, role := range remoteRoles {
+	for _, remoteRole := range remoteRoles {
```
endophage: Same comment I made above earlier. Changing variable names creates noise in the PR and should really only be done if it makes a significant difference to the readability of the code.

endophage: Same comment I made above earlier. Changing variable names creates noise in the PR and should really only be done if it makes a significant difference to the readability of the code.

client/client.go
grantleehoffman: @endophage I moved `NotaryDefaultExpiries` from the `notary` package to the `data` package to avoid a cyclic dependency, as `NotaryDefaultExpiries` is now using the `RoleName` type as the map key. tuf/data/types.go imports `github.com/docker/notary` so I could not import `github.com/docker/notary/tuf/data` into const.go (`notary` package) without a cyclic dependency.

https://github.com/docker/notary/pull/1087/files#diff-9d6b1e1df9b74f1f173c502269aa1ae3L89
https://github.com/docker/notary/pull/1087/files#diff-de0c95529058d827a3e5143442b9de1cR69

grantleehoffman: Let me know if you see any issues with this approach, or if you think NotaryDefaultExpiries belongs someplace else.

endophage: Ah my apologies. I was also confusing `NotaryDefaultExpiries` with `defaultExpiryTimes` (the var that was already in the data package). This is fine for now, we have some tech debt we need to pay off that will require reassessing where some of our types are defined. It goes beyond the scope of this PR so this will work for now.

grantleehoffman: :+1:

grantleehoffman:
I think I've addressed all of the current PR comments. Please take another look when you have the time. Thanks!

ecordell:
@grantleehoffman The only thing left I can see is the `MetadataStore` interface (and the types that satisfy it)

grantleehoffman:
@ecordell I was staying away from it for now, I believe @endophage suggested doing the name -> RoleName change in another PR. I can start working on the MetadataStore Interface if everyone is in agreement or we can create a new PR for that change.

endophage:
I need to dig into MetadataStore some more. My instinct was that it was always a generic data store but taking a brief look, it may be specific enough that it should be bound to the role types. I'll get back to you.

grantleehoffman:
It looks like FilesystemStore should implement MetadataStore, this starts to get messy as we are dealing with files and roles.

endophage:
FilesystemStore is definitely an implementer of MetadataStore. I think for now lets just leave the entire `storage` package untouched. We need to possibly do some more refactoring in there because MetadataStore implementers are also used for key storage. That was part of a big DRY-ing out of our storage code I did last year because we had two very overlapping sets of storage objects.

grantleehoffman:
@endophage @ecordell So it sounds like refactoring `MetadataStore` should be done in a different PR, since we can't make the changes we need to make to `MetadataStore` without touching `Storage`.

endophage:
Yep, and post refactoring, it think it probably won't make sense to push the `RoleName` type into the storage package as the interfaces will become more clearly generic.

grantleehoffman:
Ok sounds good. Let me know if you see any other outstanding changes that need to be made.

grantleehoffman:
This is ready to merge on my end. Looks like it just needs the yubikey test to run, is there a way to start this test?

ecordell:
jenkins, test this please

cyli:
Unfortunately our yubikey ci job is having some infra issues, so it won't work.  So one of us will be jenkins for now until it comes back up.  :)

```
`go test` commands that will be run:

	go test -tags pkcs11 -race -covermode atomic -coverprofile cmd/notary/coverage.pkcs11.txt -coverpkg github.com/docker/notary/storage,github.com/docker/notary,github.com/docker/notary/version,github.com/docker/notary/trustmanager/yubikey,github.com/docker/notary/cryptoservice,github.com/docker/notary/tuf/utils,github.com/docker/notary/server/snapshot,github.com/docker/notary/tuf,github.com/docker/notary/server/storage,github.com/docker/notary/tuf/data,github.com/docker/notary/server,github.com/docker/notary/client,github.com/docker/notary/trustmanager,github.com/docker/notary/passphrase,github.com/docker/notary/utils,github.com/docker/notary/tuf/signed,github.com/docker/notary/trustpinning,github.com/docker/notary/client/changelist,github.com/docker/notary/server/handlers,github.com/docker/notary/server/timestamp,github.com/docker/notary/tuf/validation,github.com/docker/notary/server/errors,github.com/docker/notary/cmd/notary,github.com/docker/notary/storage/rethinkdb github.com/docker/notary/cmd/notary
	go test -tags pkcs11 -race -covermode atomic -coverprofile trustmanager/yubikey/coverage.pkcs11.txt -coverpkg github.com/docker/notary/storage,github.com/docker/notary,github.com/docker/notary/trustmanager,github.com/docker/notary/passphrase,github.com/docker/notary/tuf/signed,github.com/docker/notary/tuf/utils,github.com/docker/notary/trustmanager/yubikey,github.com/docker/notary/tuf/validation,github.com/docker/notary/tuf/data github.com/docker/notary/trustmanager/yubikey

Results:

ok  	github.com/docker/notary/cmd/notary	123.668s	coverage: 63.2% of statements in <24 dependencies>
ok  	github.com/docker/notary/trustmanager/yubikey	18.690s	coverage: 23.6% of statements in <9 dependencies>
```

Yubikey tests 👍   

(I tried to also push coverage to codecov, but I think I'm doing it wrong to get it counted)

tuf/utils/utils.go
```diff
@@ -100,10 +110,10 @@ func FindRoleIndex(rs []*data.Role, name string) int {
 // ConsistentName generates the appropriate HTTP URL path for the role,
 // based on whether the repo is marked as consistent. The RemoteStore
 // is responsible for adding file extensions.
-func ConsistentName(role string, hashSHA256 []byte) string {
+func ConsistentName(role data.RoleName, hashSHA256 []byte) data.RoleName {
```
endophage: This returns a consistent filename for download purposes per TUF's rules, shouldn't be a `data.RoleName` return type.

tuf/utils/x509.go
```diff
@@ -428,18 +428,18 @@ func blockType(k data.PrivateKey) (string, error) {
...
+func KeyToPEM(privKey data.PrivateKey, role data.RoleName, gun data.GUN) ([]byte, error) {
 	bt, err := blockType(privKey)
 	if err != nil {
 		return nil, err
 	}
 
 	headers := map[string]string{}
-	if role != "" {
-		headers["role"] = role
+	if role.String() != "" {
```
endophage: No need to call `.String()`, Go's compiler will work out that you mean the literal `""` to be of type `data.RoleName`. Same applies to `if gun.String() != ""` a couple of lines down.

tuf/utils/x509.go
```diff
@@ -538,7 +538,7 @@ func CertBundleToKey(leafCert *x509.Certificate, intCerts []*x509.Certificate) (
 }
 
 // NewCertificate returns an X509 Certificate following a template, given a GUN and validity interval.
-func NewCertificate(gun string, startTime, endTime time.Time) (*x509.Certificate, error) {
+func NewCertificate(gun data.GUN, startTime, endTime time.Time) (*x509.Certificate, error) {
```
endophage: Should leave this as type string as it's not always a gun. I've filed an issue to fix the naming of the argument (https://github.com/docker/notary/issues/1093)

utils/keys.go
```diff
@@ -95,15 +95,15 @@ func ExportKeys(to io.Writer, s Exporter, from string) error {
 func ImportKeys(from io.Reader, to []Importer, fallbackRole string, fallbackGUN string, passRet notary.PassRetriever) error {
 	// importLogic.md contains a small flowchart I made to clear up my understand while writing the cases in this function
 	// it is very rough, but it may help while reading this piece of code
-	data, err := ioutil.ReadAll(from)
+	byteData, err := ioutil.ReadAll(from)
```
endophage: given we already had the import alias, this is an unnecessary variable name change.

server/storage/storage_test.go
```diff
@@ -129,27 +129,27 @@ func testUpdateCurrentVersionCheck(t *testing.T, s MetaStore, oldVersionExists b
 // GetVersion should successfully retrieve a version of an existing TUF file,
 // but will return an error if the requested version does not exist.
 func testGetVersion(t *testing.T, s MetaStore) {
-	_, _, err := s.GetVersion("gun", "role", 2)
+	_, _, err := s.GetVersion(data.GUN("gun"), data.RoleName("role"), 2)
```
grantleehoffman: Thanks for the great feedback, I will start working on fixing this.

tuf/data/roles.go
```diff
@@ -10,16 +10,16 @@ import (
 )
 
 // Canonical base role names
-const (
-	CanonicalRootRole      = "root"
-	CanonicalTargetsRole   = "targets"
-	CanonicalSnapshotRole  = "snapshot"
-	CanonicalTimestampRole = "timestamp"
+var (
```
grantleehoffman: changed

grantleehoffman:
It looks like CircleCI had an issue downloading an artifact

endophage:
I gave it a kick

endophage:
I'll have a chance to do final review of this tomorrow

cyli:
👍 yubikey tests pass
```
`go test` commands that will be run:

	go test -tags pkcs11 -race -covermode atomic -coverprofile cmd/notary/coverage.pkcs11.txt -coverpkg github.com/docker/notary/storage,github.com/docker/notary,github.com/docker/notary/version,github.com/docker/notary/trustmanager/yubikey,github.com/docker/notary/cryptoservice,github.com/docker/notary/tuf/utils,github.com/docker/notary/server/snapshot,github.com/docker/notary/tuf,github.com/docker/notary/server/storage,github.com/docker/notary/tuf/data,github.com/docker/notary/server,github.com/docker/notary/client,github.com/docker/notary/trustmanager,github.com/docker/notary/passphrase,github.com/docker/notary/utils,github.com/docker/notary/tuf/signed,github.com/docker/notary/trustpinning,github.com/docker/notary/client/changelist,github.com/docker/notary/server/handlers,github.com/docker/notary/server/timestamp,github.com/docker/notary/tuf/validation,github.com/docker/notary/server/errors,github.com/docker/notary/cmd/notary,github.com/docker/notary/storage/rethinkdb github.com/docker/notary/cmd/notary
	go test -tags pkcs11 -race -covermode atomic -coverprofile trustmanager/yubikey/coverage.pkcs11.txt -coverpkg github.com/docker/notary/storage,github.com/docker/notary,github.com/docker/notary/trustmanager,github.com/docker/notary/passphrase,github.com/docker/notary/tuf/signed,github.com/docker/notary/tuf/utils,github.com/docker/notary/trustmanager/yubikey,github.com/docker/notary/tuf/validation,github.com/docker/notary/tuf/data github.com/docker/notary/trustmanager/yubikey

Results:

ok  	github.com/docker/notary/cmd/notary	113.518s	coverage: 63.3% of statements in <24 dependencies>
ok  	github.com/docker/notary/trustmanager/yubikey	18.492s	coverage: 23.6% of statements in <9 dependencies>
```

endophage:
Hey @grantleehoffman, I hope you don't mind I pushed the remaining cleanup. We're starting to line up some feature updates and I want to get this merged first rather than asking you to rebase and deal with the changes we're making. You can see my commit for the final stuff I did, it's largely in line with the things I pointed out previously, though I also really didn't want the internals of MemoryStore to know about RoleNames so I tweaked the constructor there (and in doing that uncovered a test that was relying on side effects of shared mutable state so that got fixed too).

cc @ecordell you might want to review my final commit

cc @cyli @riyazdf @HuKeping given I now have skin in this PR, could one of you at least review my final commit.

client/client_test.go
```diff
@@ -194,7 +194,7 @@ func createRepoAndKey(t *testing.T, rootType, tempBaseDir, gun, url string) (
 		tempBaseDir, data.GUN(gun), url, http.DefaultTransport, rec.retriever, trustpinning.TrustPinConfig{})
 	require.NoError(t, err, "error creating repo: %s", err)
 
-	rootPubKey, err := repo.CryptoService.Create(data.RoleName("root"), repo.gun, rootType)
+	rootPubKey, err := repo.CryptoService.Create("root", repo.gun, rootType)
```
HuKeping: Why not `data.CanonicalRootRole` since we already have it lol

endophage: I was moving fast, didn't catch everything :-(

client/backwards_compatibility_test.go
```diff
@@ -278,7 +278,7 @@ func Test0Dot3RepoFormat(t *testing.T) {
 	delegations, err := repo.GetDelegationRoles()
 	require.NoError(t, err)
 	require.Len(t, delegations, 1)
-	require.Equal(t, "targets/releases", delegations[0].Name)
+	require.EqualValues(t, data.GUN("targets/releases"), delegations[0].Name)
```
cyli: Should this be `data.Role` instead of `data.GUN`?  This wouldn't require using `EqualValues` then, and could just check the role against the delegation.

client/client.go
```diff
@@ -674,7 +674,7 @@ func (r *NotaryRepository) publish(cl changelist.Changelist) error {
 	if snapshotJSON, err := serializeCanonicalRole(
 		r.tufRepo, data.CanonicalSnapshotRole, nil); err == nil {
 		// Only update the snapshot if we've successfully signed it.
-		updatedFiles[data.CanonicalSnapshotRole] = snapshotJSON
```
cyli: Non-blocking:  this is totally fine, but it might be easier to change `updatedFiles` to be a `map[data.RoleName][]byte`, so below we don't have to call `data.CanonicalRootRole.String()`.  This would also mean that `signTargets` wouldn't have to cast back to `data.RoleName` before calling `serializeCanonicalRole`.

client/client.go
```diff
@@ -791,11 +791,11 @@ func signTargets(updates map[string][]byte, repo *tuf.Repo, initialPublish bool)
 	// iterate through all the targets files - if they are dirty, sign and update
 	for roleName, roleObj := range repo.Targets {
 		if roleObj.Dirty || (roleName == data.CanonicalTargetsRole && initialPublish) {
-			targetsJSON, err := serializeCanonicalRole(repo, roleName, nil)
+			targetsJSON, err := serializeCanonicalRole(repo, data.RoleName(roleName), nil)
```
cyli: I think `repo.Targets` is already a map of `data.RoleName` to targets, so this cast may not be necessary.

client/client.go
```diff
@@ -844,14 +844,14 @@ func (r *NotaryRepository) saveMetadata(ignoreSnapshot bool) error {
...
+	err = r.cache.Set(data.CanonicalRootRole.String(), rootJSON)
 	if err != nil {
 		return err
 	}
 
-	targetsToSave := make(map[string][]byte)
+	targetsToSave := make(map[data.RoleName][]byte)
 	for t := range r.tufRepo.Targets {
-		signedTargets, err := r.tufRepo.SignTargets(t, data.DefaultExpires(data.CanonicalTargetsRole))
+		signedTargets, err := r.tufRepo.SignTargets(data.RoleName(t), data.DefaultExpires(data.CanonicalTargetsRole))
```
cyli: `tufRepo.Targets` is already a `map[data.RoleName]*data.SignedTargets`, so this cast may not be necessary.

client/witness.go
```diff
@@ -10,7 +10,7 @@ import (
 
 // Witness creates change objects to witness (i.e. re-sign) the given
 // roles on the next publish. One change is created per role
-func (r *NotaryRepository) Witness(roles ...string) ([]string, error) {
+func (r *NotaryRepository) Witness(roles ...data.RoleName) ([]string, error) {
```
cyli: Would it make sense to return a list of `data.RoleName` instead of strings here?

cryptoservice/crypto_service.go
```diff
@@ -159,7 +159,7 @@ func (cs *CryptoService) ListAllKeys() map[string]string {
 	res := make(map[string]string)
 	for _, ks := range cs.keyStores {
 		for k, r := range ks.ListKeys() {
-			res[k] = r.Role // keys are content addressed so don't care about overwrites
+			res[k] = r.Role.String() // keys are content addressed so don't care about overwrites
```
cyli: Should `ListAllKeys` return a `map[string]data.RoleName` instead?

grantleehoffman:
I think circleci needs a kick again

cyli:
jenkins, test this please

cyli:
Having some CI issues, but once that's fixed LGTM on green.

endophage:
Once Jenkins is green I'll manually merge this. I want to test there are no conflicts/breaking changes that will result from the merge before pushing master. This is so large it's possible that non-overlapping (and therefore non-conflicting) but interdependent changes may have occurred.

cyli:
jenkins, test this please

endophage:
@grantleehoffman thanks for sticking with this, we're very happy to have this merged :-)

Issue #914, opened 2016-08-12
Notary requires an X509 cert's common name to match the GUN exactly, but the length of a common name is restricted to 64 chars


dnwake:
In order to be valid, Notary requires that any cert must have a common name that matches the GUN.  See https://github.com/docker/notary/blob/master/trustpinning/certs.go#L178

However, RFC3280 imposes a maximum length of 64 characters on the common name of an X509 certificate.  See https://www.ietf.org/rfc/rfc3280.txt (search for ub-common-name).  

When using Notary with Docker, it is easy to end up with a GUN that is longer than 64 characters (e.g. "my_docker_registry.somedomain.com/my_docker_repo_part_1/my_docker_repo_part_2")

While Notary's own internal certificate-generating code may not enforce this limit, it is enforced by third-party tools such as openssl.  This is a big problem for anyone planning to generate their own certificates and then import them into Notary (see https://github.com/docker/notary/pull/821)

riyazdf:
Thank you @dnwake for pointing this out!  I'm going to link this issue to https://github.com/docker/notary/issues/883 since it's similar in nature, but different

dnwake:
For the benefit of future people running into this problem, I've found that the **certtool** command (supplied as part of the **gnutls-utils** package) ignores the 64 character limitation, so can be used as a workaround.

endophage:
Having merged wildcard CNs and there being workarounds, I'm going to close this. We don't have any specific action items to address GUN name length. Please re-open or file an issue with specific action items if more is needed.

PR #1088, opened 2017-01-26
Support wildcard CNs in root leaf certificates

### LINKED ISSUE #883: Support wildcard root certs
So that `--root-cert` can work for multiple repos.  This would mean that throwaway root keys can be generated when initializing a repo, and the real root key doesn't have to be online.

ecordell:
This adds support for explicit wildcard suffixes in root (leaf) certificates of the form `prefix*`. 

I considered doing something more flexible using `filepath.Match` but I think this more conservative form is safer and (probably) just as useful. It would be easy to change in the future if more flexibility (e.g. `docker.io/*/test`) is needed. 

Note the wildcard `*` indicator is required to avoid accidentally using wildcard certs (e.g. `docker.io/test` and `docker.io/test2`).

Fixes #883 
Possibly addresses #914 as well?

### DIFF
--- a/trustpinning/certs.go
+++ b/trustpinning/certs.go
@@ -12,6 +12,8 @@ import (
 	"github.com/docker/notary/tuf/utils"
 )
 
+const wildcard = "*"
+
 // ErrValidationFail is returned when there is no valid trusted certificates
 // being served inside of the roots.json
 type ErrValidationFail struct {
@@ -175,6 +177,17 @@ func ValidateRoot(prevRoot *data.SignedRoot, root *data.Signed, gun string, trus
 	return data.RootFromSigned(root)
 }
 
+// MatchCNToGun checks that the common name in a cert is valid for the given gun.
+// This allows wildcards as suffixes, e.g. `namespace/*`
+func MatchCNToGun(commonName, gun string) bool {
+	if strings.HasSuffix(commonName, wildcard) {
+		prefix := strings.TrimRight(commonName, wildcard)
+		logrus.Debugf("checking gun %s against wildcard prefix %s", gun, prefix)
+		return strings.HasPrefix(gun, prefix)
+	}
+	return commonName == gun
+}
+
 // validRootLeafCerts returns a list of possibly (if checkExpiry is true) non-expired, non-sha1 certificates
 // found in root whose Common-Names match the provided GUN. Note that this
 // "validity" alone does not imply any measure of trust.
@@ -183,8 +196,8 @@ func validRootLeafCerts(allLeafCerts map[string]*x509.Certificate, gun string, c
 
 	// Go through every leaf certificate and check that the CN matches the gun
 	for id, cert := range allLeafCerts {
-		// Validate that this leaf certificate has a CN that matches the exact gun
-		if cert.Subject.CommonName != gun {
+		// Validate that this leaf certificate has a CN that matches the gun
+		if !MatchCNToGun(cert.Subject.CommonName, gun) {
 			logrus.Debugf("error leaf certificate CN: %s doesn't match the given GUN: %s",
 				cert.Subject.CommonName, gun)
 			continue
--- a/trustpinning/certs_test.go
+++ b/trustpinning/certs_test.go
@@ -1299,3 +1299,79 @@ func TestValidateRootWithExpiredIntermediate(t *testing.T) {
 	)
 	require.Error(t, err, "failed to invalidate expired intermediate certificate")
 }
+
+func TestCheckingWildcardCert(t *testing.T) {
+	memStore := trustmanager.NewKeyMemoryStore(passphraseRetriever)
+	cs := cryptoservice.NewCryptoService(memStore)
+	testPubKey, err := cs.Create(data.CanonicalRootRole, "docker.io/notary/*", data.ECDSAKey)
+	require.NoError(t, err)
+	testPrivKey, _, err := memStore.GetKey(testPubKey.ID())
+	require.NoError(t, err)
+
+	testCert, err := generateTestingCertificate(testPrivKey, "docker.io/notary/*", notary.Year)
+	require.NoError(t, err)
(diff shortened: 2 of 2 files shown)


trustpinning/certs.go
```diff
@@ -175,6 +175,17 @@ func ValidateRoot(prevRoot *data.SignedRoot, root *data.Signed, gun string, trus
 	return data.RootFromSigned(root)
 }
 
+// MatchCNToGun checks that the common name in a cert is valid for the given gun.
+// This allows wildcards as suffixes, e.g. `namespace/*`
+func MatchCNToGun(commonName, gun string) bool {
+	wildcard := "*"
```
endophage: should this be a const?

endophage:
Glad you avoided `filepath.Match`, it has weird behaviour. It doesn't support `**` and a single `*` matches exactly one and only one path segment. (it also probably wouldn't be consistent on windows).

One thought, should we require the `*` comes after a ` /`? I don't have any instinct either way, just throwing it out there.

trustpinning/certs.go
```diff
@@ -175,6 +175,17 @@ func ValidateRoot(prevRoot *data.SignedRoot, root *data.Signed, gun string, trus
 	return data.RootFromSigned(root)
 }
 
+// MatchCNToGun checks that the common name in a cert is valid for the given gun.
+// This allows wildcards as suffixes, e.g. `namespace/*`
+func MatchCNToGun(commonName, gun string) bool {
+	const wildcard = "*"
```
endophage: I figured a const at probably package scope :-P This works too as far as giving us immutability though.

ecordell: figured the go compiler would do what I wanted here, didn't really verify though

ecordell:
> One thought, should we require the * comes after a /? I don't have any instinct either way, just throwing it out there.

Personally I think that's a little too restrictive; I was considering an organization using a wildcard cert for a subset of their packages, e.g. `docker/notary-*` to cover `docker/notary`, `docker/notary-server`, and `docker/notary-signer`

trustpinning/certs.go
endophage: It probably will as far as optimization, I was thinking more along code organization.

trustpinning/certs_test.go
```diff
@@ -1299,3 +1299,189 @@ func TestValidateRootWithExpiredIntermediate(t *testing.T) {
...
+	var wildcardTests = []struct {
+		CN  string
+		gun string
+		out bool
+	}{
+		{"docker.com/*", "docker.com/notary", true},
+		{"test/*/wild", "test/test/wild", false},
+		{"*/all", "test/all", false},
+		{"docker.com/*/*", "docker.com/notary/test", false},
+		{"*", "docker.com/any", true},
```
riyazdf: can we add the following test cases:

CN: `""`, gun: `*`, out: `false`
CN: `*`, gun: `""`, out: `true`
CN: `*abc*`, gun: `"abc"`, out: `false` (since we check against `*abc`, no longer wildcarded)
CN: `test/*/wild*`, gun: `"test/test/wild"`, out: `false` (since we check against `test/*/wild`, no longer wildcarded)

trustpinning/certs.go
```diff
@@ -175,6 +175,17 @@ func ValidateRoot(prevRoot *data.SignedRoot, root *data.Signed, gun string, trus
...
+	if strings.HasSuffix(commonName, wildcard) {
+		prefix := commonName[:len(commonName)-len(wildcard)]
```
riyazdf: nit style: I'd prefer using `strings.TrimSuffix` instead of indexing directly

riyazdf: Also, could we add a log debug to display which prefix we're comparing against?

My concern is if folks use something like `prefix**` and expect matching against `prefix` instead of `prefix*`

trustpinning/certs_test.go
cyli: Out of curiosity, if the common name is `**`, should we treat that as `*`?  If not, can we also add one more test?

CN: `**`, gun: `"docker.com/any"`, out: `false`

trustpinning/certs_test.go
```diff
@@ -1299,3 +1299,193 @@ func TestValidateRootWithExpiredIntermediate(t *testing.T) {
 	)
 	require.Error(t, err, "failed to invalidate expired intermediate certificate")
 }
+
+func TestCheckingWilcardCert(t *testing.T) {
```
cyli: nit:  "TestCheckingWil**d**cartCert"

trustpinning/certs_test.go
cyli: Non-blocking:  Since we're not actually checking any trust pinning here, do we need a CA+intermediate+leaf for this test?  Might make the test a little shorter to just include the leaf.

riyazdf: @cyli: good idea - I like treating `**` as `*` for now to keep things simple so that test would be great 👍

ecordell: I went ahead and switched `TrimSuffix` to `TrimRight` so that any number of trailing `*`s will collapse down to 1.

lewiada:
How do I enable this feature?  Today the root public key in the root.json file is a self-signed cert with CN == GUN.  How do I use this such that the CN is a wildcard such as www.example.com/* ?

endophage:
The functionality to leverage it hasn't been added yet :-( #821 was part of the work on that but I guess @dnwake has been otherwise engaged. I pinged him a few weeks ago checking if he'd get back around to it and he said he would.

lewiada:
@endophage - we are making the changes (actually already done, just testing now) as well as some others that you and I discussed. You should see a pull request in the next few days.  Once this is in place, we are planning to work on CA pinning, and then moving trust pinning into DCT.  Once that's all finished, we can work on getting generalizing the pkcs11 interface to support HSMs other than Yubikey :-)

PR #1089, opened 2017-01-30
Proposal: Protecting keystore passphrase

rhonnava:
Proposal to add provision in config and code to protect the passphrase used to wrap signing private keys in the keystore DB. It also includes a proposal to add implementation for PKCS11 implementation as one of the methods to protect passphrases. 

Please see the file and diagram for more description.


rhonnava:
A script to automatically update the my git config user.email to the correct one has messed up this pull request. I will close this and open another one.

PR #1090, opened 2017-01-30
Adding proposal document for protecting keystore passphrase. 

rhonnava:
Proposal to add provision in config and code to protect the passphrase used to wrap signing private keys in the keystore DB. It also includes a proposal to add implementation for PKCS11 implementation as one of the methods to protect passphrases

Please find more details in the added document.


cmd/notary-signer/protecting_keystore_passphrase.md
```diff
@@ -0,0 +1,90 @@
...
+Table of Contents
+----------
+[TOC]
+
+
+
+Background
+-------------
+
+Currently in Nortary Signer, the signing keys are maintained in memory or are encrypted using AES wrapping with a password and the cipher text is persisted in a MySQL database.
```
endophage: Just for the record, the in memory implementation should never be used outside of testing and development.

endophage:
Thanks for this submission. It would be better to submit this as an issue rather than a PR so we can have some discussion and a tracking issue. Either way, the current file locations are very much not the correct location for this type of content. `cmd/*` is for the `main` packages that we build, not documentation.

If we have the discussion here, the end result will probably be the creation of said tracking issue, and converting the content of this PR to actual documentation (which depending on the time to implement, it may just make more sense to close the PR and open a new one with the final docs).

As far as the feature in general, I think it's worth discussing and better password protection is certainly desirable, there are probably some parts of the proposed implementation we'd like to tweak.

rhonnava:
I agree that this is not a good location for documentation. I was blindly following point 5 in this document: https://docs.docker.com/opensource/workflow/advanced-contributing/

I will open an issue for this.

rhonnava:
Closing pull request as this is being discussed in issue #1091

PR #1092, opened 2017-02-01
Move most of the notary server/signer http/grpc server code to subpackages

cyli:
This is so that either can be imported and run outside the main package.

cc @endophage @vikstrous (I couldn't find an associated issue but I vaguely remember @vikstrous asking for this a while back as well)

### DIFF
--- a/cmd/notary-server/main.go
+++ b/cmd/notary-server/main.go
@@ -2,96 +2,16 @@ package main
 
 import (
 	_ "expvar"
-	"flag"
-	"fmt"
-	"net/http"
 	_ "net/http/pprof"
-	"os"
-	"os/signal"
 
-	"github.com/Sirupsen/logrus"
-	"github.com/docker/distribution/health"
-	"github.com/docker/notary/server"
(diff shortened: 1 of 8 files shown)


vikstrous:
Awesome! Yeah, this would allow us to build notary using our own main file.

endophage:
@vikstrous I still want to understand why in the world you would want to do that. If there's something you need to configure that isn't currently available, let's just make that configurable.

vikstrous:
We have a go package that watches rethinkdb for changes in configs and restarts services. We use it as the entrypoint of the container and exec notary server/signer from it. We need it for delivering configs and certs to the containers.

endophage:
@cyli some of this happened in the Client API PR but I think there will probably be a lot of conflicts between this PR and that one. I'm thinking we should close this and we can reassess against Viktor's requirements once Client API is in. Does that sound OK?

cyli:
@endophage Sure, I don't mind.  The original purpose I think was to make the client API PR smaller, but happy to have it wherever.  :)

PR #1096, opened 2017-02-13
Remote key escrow 

endophage:
Remote key escrow via a GRPC implementation of the trustmanager.Storage interface.

Still finishing it up but intend for Notary Signer to be able to act as an escrow backend with configurable APIs (so it can host either or both of: current key storage for server and/or the new GRPC trustmanager.Storage server).

By implementing the low level trustmanager.Storage interface, passwords, and therefore key encryption and decryption, still happen at the client, so compromise of the remote key storage does not equate to compromise of the escrowed keys.


endophage:
This is about as tested as it's going to get. The compiled .pb.go is hurting coverage but I'm not going to write tests for generated code.

trustmanager/errors.go
```diff
@@ -0,0 +1,31 @@
...
+
+// ErrKeyNotFound is returned when the keystore fails to retrieve a specific key.
+type ErrKeyNotFound struct {
+	KeyID string
+}
+
+// ErrKeyNotFound is returned when the keystore fails to retrieve a specific key.
+func (err ErrKeyNotFound) Error() string {
+	return fmt.Sprintf("signing key not found: %s", err.KeyID)
+}
```
riyazdf: thanks for consolidating this 👍

trustmanager/interfaces.go
```diff
@@ -34,32 +32,11 @@ type Storage interface {
...
-
-// ErrKeyNotFound is returned when the keystore fails to retrieve a specific key.
-type ErrKeyNotFound struct {
-	KeyID string
-}
-
-// ErrKeyNotFound is returned when the keystore fails to retrieve a specific key.
-func (err ErrKeyNotFound) Error() string {
-	return fmt.Sprintf("signing key not found: %s", err.KeyID)
+// KeyInfo stores the role, path, and gun for a corresponding private key ID
```
riyazdf: nit since you're moving things around: remove `path` from description

trustmanager/remoteks/client.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+// RemoteStore is a wrapper around the GRPC storage client, translating between
+// the Go and GRPC APIs.
+type RemoteStore struct {
+	client   StoreClient
+	location string
+}
+
+var _ trustmanager.Storage = &RemoteStore{}
```
riyazdf: why is this necessary?  Is this due to the nature of GRPC?

trustmanager/remoteks/client.go
```diff
@@ -0,0 +1,94 @@
...
+func (s *RemoteStore) Remove(fileName string) error {
+	fm := &FileNameMsg{
+		FileName: fileName,
+	}
+	_, err := s.client.Remove(context.Background(), fm)
+	return err
+}
+
+// Get returns the file content found at fileName relative to the base directory
+// of the file store. The path is cleaned before being made absolute to ensure
```
riyazdf: GH is unfortunately collapsing many diffs and so I'm having trouble finding the cleaning logic - could you please point me to it?

trustmanager/remoteks/client.go
```diff
@@ -0,0 +1,94 @@
...
+	return bm.Data, nil
+}
+
+// ListFiles returns a list of paths relative to the base directory of the
+// filestore. Any of these paths must be retrievable via the
+// Storage.Get method.
+func (s *RemoteStore) ListFiles() []string {
+	logrus.Infof("listing files from %s", s.location)
+	fl, err := s.client.ListFiles(context.Background(), &google_protobuf.Empty{})
+	if err != nil {
```
riyazdf: can we add a test case for empty `ListFiles()`?

trustmanager/remoteks/client.go
```diff
@@ -0,0 +1,94 @@
...
+		logrus.Errorf("error listing files from %s: %s", s.location, err.Error())
+		return nil
+	}
+	return fl.FileNames
+
+}
+
+// Location returns a human readable indication of where the storage is located.
+func (s *RemoteStore) Location() string {
```
riyazdf: nit: can we add a test case for `Location` in case of any regressions?

cmd/escrow/config.go
```diff
@@ -0,0 +1,56 @@
...
+	"fmt"
+	"github.com/docker/notary"
+	"github.com/docker/notary/storage"
+	"github.com/docker/notary/trustmanager"
+	"github.com/docker/notary/trustmanager/remoteks"
+	"github.com/docker/notary/utils"
+	"github.com/spf13/viper"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"net"
```
endophage: This is how gogland is arranging them for me :-/ I need to try breaking them up into our normal format and see what it does with new imports that get auto-added. If we're all having to constantly tweak the import list though, maybe we stop trying so hard to organize it. It looks nicer the way we've been doing it, but I think everyone also loves just having imports auto-managed.

cmd/escrow/config.toml
```diff
@@ -0,0 +1,9 @@
+[storage]
```
endophage: I'd actually prefer converting everything to toml, I just didn't want to do it in this PR. I'll file an issue for it. Viper will make it little more than just rewriting the json files and updating our Dockerfiles and docker-compose files to point at the toml versions.

trustmanager/remoteks/client.go
endophage: Hacky go compile time check to ensure a type implements the desired interface. You'll see it around distribution a lot too. It's relatively idiomatic if unofficial.

trustmanager/remoteks/client.go
endophage: These comments are just copied from the interface definition. The cleaning actually happens in the server side after passing through to the real underlying storage.

trustmanager/remoteks/client.go
endophage: Already do :-) https://github.com/docker/notary/pull/1096/files/ca60db8420cc08ec94536811fcdcc48e789b7f96#diff-fc36b6ceeef2f9d048a363a6eeab3b0cR69

endophage: The GRPC API for Go always adds an error, but ListFiles never actually errors. The only way this error case would be hit is an error at the GRPC level.

cmd/escrow/config.toml
riyazdf: sounds good to me

endophage: https://github.com/docker/notary/issues/1102

Makefile
```diff
@@ -176,6 +180,9 @@ client: ${PREFIX}/bin/notary
 binaries: ${PREFIX}/bin/notary-server ${PREFIX}/bin/notary ${PREFIX}/bin/notary-signer
 	@echo "+ $@"
 
+escrow: ${PREFIX}/bin/escrow
```
endophage: No strong feelings. It didn't seem necessary as "escrow" isn't quite as generic as "server". I mostly just got tired of typing `notary-*` on the front of everything :-P

trustmanager/remoteks/client_test.go
```diff
@@ -0,0 +1,79 @@
...
+	name := "testfile"
+	bytes := []byte{'1'}
+
+	s := setupTestServer(t)
+	defer s.Stop()
+
+	// can't just use NewRemoteStore because it correctly sets up tls
+	// config and for testing purposes it's easier for the client to just
+	// be insecure
+	cc, err := grpc.Dial(
```
ecordell: `NewRemoteStore` could swap out `grpc.WithTransportCredentials` for `grpc.WithInsecure` if the tlsconfig is nil?

trustmanager/remoteks/server_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/docker/notary/storage"
+)
+
+func TestNewGRPCStorage(t *testing.T) {
+	s := NewGRPCStorage(nil)
+	require.IsType(t, &GRPCStorage{}, s)
+}
+
+// Set, Get, List, Remove, List, Get
+func TestGRPCStorage(t *testing.T) {
```
ecordell: What about the error cases?

endophage:
@ecordell the full repo will be coming in follow up PRs, with something identical or at least very, very close to that interface. This was a nice self contained piece of functionality (that is also somewhat necessary for the remote full repository).

The dependency injection PR from @n4ss is also foundational work for the full remote repo.

trustmanager/remoteks/client_test.go
endophage: You are correct but it seems like `InsecureSkipVerify` and `WithInsecure` achieve the same thing so I figured it was easier to just use the well understood tls config and let users set that up however they please, rather than having 2 ways to set clients to insecure mode.

Makefile
cyli: Should the escrow binary also be added to the `binaries` and `static` targets?

endophage: @cyli I think once we have configurable integration with the NotaryRepository that would make sense. At the moment it doesn't really plug into anything and until @n4ss's dependency injection PR is merged, there's no way to plug it in short of forked code.

trustmanager/remoteks/client.go
```diff
@@ -0,0 +1,93 @@
...
+	}, nil
+}
+
+// Set stores the data using the provided fileName
+func (s *RemoteStore) Set(fileName string, data []byte) error {
+	sm := &SetMsg{
+		FileName: fileName,
+		Data:     data,
+	}
+	_, err := s.client.Set(context.Background(), sm)
```
cyli: Non-blocking - should we be setting timeouts for these by default?

trustmanager/remoteks/keystore.proto
```diff
@@ -0,0 +1,39 @@
...
+    string FileName = 1;
+}
+
+message ByteMsg {
+    bytes Data = 1;
+}
+
+message StringListMsg {
+    repeated string FileNames = 1;
+}
```
cyli: Non-blocking:  missing newline

trustmanager/remoteks/server.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		return &ByteMsg{}, err
+	}
+	return &ByteMsg{Data: data}, nil
+}
+
+// ListFiles returns all known identifiers in the storage backend.
+func (s *GRPCStorage) ListFiles(ctx context.Context, _ *google_protobuf.Empty) (*StringListMsg, error) {
+	lst := s.backend.ListFiles()
+	logrus.Debugf("found keys: %s", strings.Join(lst, ","))
```
cyli: If the signer is eventually going to be a remote escrow, should we actually list every key in the signer in the debug logs?  Perhaps we can debug log the number?

trustmanager/remoteks/server.go
```diff
@@ -0,0 +1,59 @@
...
+// backend.
+func NewGRPCStorage(backend trustmanager.Storage) *GRPCStorage {
+	return &GRPCStorage{
+		backend: backend,
+	}
+}
+
+// Set writes the provided data under the given identifier.
+func (s *GRPCStorage) Set(ctx context.Context, msg *SetMsg) (*google_protobuf.Empty, error) {
+	logrus.Debugf("storing: %s", msg.FileName)
```
cyli: Non-blocking:  Should the other RPC calls also be logged?  Or at least `Remove`, since both are writes?

cyli:
Could we also add `trustmanager/remoteks/*.pb.go` to the codecov ignore section?

trustmanager/remoteks/server.go
endophage: I really need an "ultra debug" mode. Listing keys has been useful but you're right that in this case it could be a lot.

trustmanager/remoteks/client_test.go
ecordell: would it be easier to just use the certs already included in the project for testing? it's not a big deal since anyone actually running this will certainly discover problems with tls config right away, but it would be nice to have it in the test path if possible.

endophage: Pushed

endophage:
I think I've addressed everything here. Just rebased, tests passed locally, hopefully circle agrees.

codecov.yml
```diff
@@ -20,5 +20,6 @@ coverage:
     - "tuf/testutils/*"
     - "vendor/*"
     - "proto/*.pb.go"
+    - "trustmanager/remoteks/keystore.proto"
```
cyli: Non-blocking - I think you may need to exclude the `*.pb.go` files, since the coverage will be calculated for the generated go files, not the proto files (I don't think codecov knows about protos)

cyli:
I've re-kicked the yubikey tests, and they past.  Codecov is being weird.  :(   This otherwise LGTM, modulo a codecov ignore change (which does not account for the codecov weirdness atm)

codecov.yml
endophage: Oh sorry, I totally crossed my wires. I meant to exclude the pb.go!

endophage: Ok fixed now. Must have been one of those occasions where I was thinking of something else while typing and the wrong thing came out.

cyli: I do that at least half the time :)

PR #1104, opened 2017-02-24
Do not require signing if no there are no targets changes

cyli:
`AddTarget` and `RemoveTarget` currently checks if the user has valid signing keys regardless of whether the target actually needs to be added or removed, because both require signing regardless of whether the target needs to be added or removed.

Updates both functions so that if the target being added does not actually change the role, or if the target being removed doesn't actually change the role, signing keys are not required and the roles are not necessarily marked as dirty.


tuf/tuf_test.go
```diff
@@ -863,16 +884,29 @@ func TestRemoveExistingAndNonexistingTargets(t *testing.T) {
 	// still no metadata
 	_, ok = repo.Targets["targets/test"]
 	require.False(t, ok)
-}
 
-// Removing targets from a role that exists but without metadata succeeds.
-func TestRemoveTargetsNonexistentMetadata(t *testing.T) {
```
cyli: Note - this test didn't actually seem to do what it claimed - the body was a duplicate of `TestRemoveTargetsRoleDoesntExist`.  Since the previous test was supposed to test adding existing and non-existing targets anyway, I just removed this test and updated the previous one.

tuf/tuf_test.go
```diff
@@ -917,8 +952,12 @@ func TestRemoveTargetsNoSigningKeys(t *testing.T) {
 	// now delete the signing key (all keys)
 	repo.cryptoService = signed.NewEd25519()
 
-	// now remove the target - it should fail
-	err = repo.RemoveTargets("targets/test", "f")
+	// remove a nonexistant target - it should not fail
```
riyazdf: nit typo: `nonexistent`

cyli:
@riyazdf I only changed `AddTarget` and `RemoveTarget` - it should still be possible to mark a targets file as dirty and cause it to sign, although we may have to add a new function for that.  Doing so seems like it'd be more explicit than calling `AddTarget` of the same target or `RemoveTarget` of a non-existent target, though.

riyazdf:
@cyli - makes sense, I agree with that.  This could be an extension of `witness` or a new command entirely.

tuf/data/types.go
```diff
@@ -181,6 +182,34 @@ type FileMeta struct {
...
+	// we don't care if these are valid hashes, just that they are equal
+	for key, val := range f.Hashes {
+		if !bytes.Equal(val, o.Hashes[key]) {
+			return false
+		}
+	}
+	if f.Custom == nil && o.Custom == nil {
+		return true
+	}
+	fBytes, err := f.Custom.MarshalJSON()
```
endophage: `Custom` is of type `*json.RawMessage` which is just a re-typing of `[]byte`. I'm wondering if it makes sense to unmarshal into a `map[string]interface{}` and remarshal canonically to determine actual equivalence vs face value byte for byte equality.

endophage: I guess on the other hand, there's no guarantee it's JSON inside the Custom field so maybe that's not worth doing.

tuf/data/types.go
```diff
@@ -181,6 +182,34 @@ type FileMeta struct {
 	Custom *json.RawMessage `json:"custom,omitempty"`
 }
 
+// Equals returns true if the other FileMeta object is equivalent to this one
+func (f FileMeta) Equals(o FileMeta) bool {
+	if o.Length != f.Length || len(f.Hashes) != len(f.Hashes) {
+		return false
+	}
+	if f.Custom == nil && o.Custom != nil || f.Custom != nil && o.Custom == nil {
```
endophage: This and the custom checks further down could probably all just be replaced with `if *f.Custom != *o.Custom { return false }`

cyli: If `f.Custom` is `nil`, won't we get a panic trying to get `*f.Custom`?  Also, if neither were nil, I think that still wouldn't work because the equality operator doesn't work on byte slices.  :|  For instance:  https://play.golang.org/p/jY4yrw7lrL

endophage: ugh, yeah. I was thinking of the `RawMessage` being a `[]byte` but we essentially have double indirection here. I remember we specifically put that in because of the whole "addressability" thing cause the `RawMessage` to get double base64 encoded, if that's not necessary any more (I feel like I remember something addressing RawMessage and this problem being merged in Go 1.7, possible 1.6) we could also make `Custom` a non-pointer `RawMessage`.

Issue #1071, opened 2017-01-05
duplicate init on gun causes corrupted state


lewiada:
After initializing a GUN and publishing the local trusted collection, repeating this process seems to result in a corrupted state.  For example: 

notary init server.example.com/test1
notary publish server.example.com/test1
Successfully published changes for repository docker.mot-solutions.com/adam/iwan412

notary init server.example.com/test1
notary publish server.example.com/test1
* fatal: could not rotate trust to a new trusted root: failed to validate data with current trusted certificates

I don't feel there should be a use case to initialize a GUN twice, so the client/server should do a check to make sure it hasn't already been initialized before proceeding.

endophage:
Hey @lewiada, it's only a locally corrupted state right? You should be able to fix it with a `notary remove server.example.com/test1`.

The fix I want to get in for this is lazy initialization where one no longer has to explicitly call `notary init`. As part of lazy initialization, we would always query the server for existing data before assuming it needs to be created locally.

A shorter term fix may be to check the server, or if network connectivity isn't available, the local cache, for existing data. At the moment I believe init assumes the repo doesn't exist and overwrites any existing cache.

lewiada:
@endophage, I can't do that because there isn't a target yet.  The GUN has been initialized, but nothing has been added to the collection.  

I think your last point is where the issue is.  I think init should first check the server to see if a collection has been initialized, and inform the user if it already has, and only create it if it hasn't.

endophage:
`notary remove` only communicates with the server if you pass the `-r` option. Without that it should just clear local cache.

lewiada:
but remove requires two params: 1) gun, 2) target
it's not an issue of local vs. server, it's that i've only done an init and nothing has been added.  so I have no target to pass

endophage:
Sorry, having a brainfart: `notary delete`

PR #1105, opened 2017-02-24
Notary repository lazy initialization

n4ss:
This work depends on the work done in https://github.com/docker/notary/pull/1094, allowing us to inject a `remote store` and a `cache` in a `NotaryRepository` object.
Since this PR is stacked on top of https://github.com/docker/notary/pull/1094 which just got merged, make sure you pull the latests changes. 

We are now initializing the repository when publishing, if needed. We do this for three reasons:

1. reduce the amount of operations necessary for notary users to setup their repo
2. avoid metadata corruption when calling init several times on the same repo - https://github.com/docker/notary/issues/1071
3. give more flexibility on how to initialize

 The lazy initialization is done either from scratch by `NotaryRepository.Initialize()` when starting from a blank state or by `NotaryRepository.bootstrapRepo()` if metadata is found on disk (example: when someone does a `notary key rotate` as a first operation).

Some updates have been added to the way change lists were handled, we can now inject one in a`NotaryRepository` object.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -623,10 +623,12 @@ func (r *NotaryRepository) publish(cl changelist.Changelist) error {
 		if _, ok := err.(ErrRepositoryNotExist); ok {
 			err := r.bootstrapRepo()
 			if _, ok := err.(store.ErrMetaNotFound); ok {
+				logrus.Debugf("No metadata found on disk, initializing repository %s from scratch", r.gun.String())
 				err = r.Initialize(nil)
 			}
 
 			if err != nil {
+				logrus.Debugf("Unable to initialize repository at publish-time: %s", err.Error())
 				return err
 			}
 
--- a/client/client.go
+++ b/client/client.go
@@ -834,6 +834,8 @@ func (r *NotaryRepository) bootstrapRepo() error {
 	return nil
 }
 
+// initializeFromCache looks for cached metadata to bootstrap from
+// and will initialize the repository from scratch otherwise.
 func (r *NotaryRepository) initializeFromCache() error {
 	metaCached, err := isMetaCached(r.cache)
 	if err != nil {
--- a/client/client.go
+++ b/client/client.go
@@ -120,6 +120,11 @@ func NewNotaryRepository(baseDir string, gun data.GUN, baseURL string, remoteSto
 	return nRepo, nil
 }
 
+// GetGUN is a getter for the GUN object from a NotaryRepository
+func (r *NotaryRepository) GetGUN() string {
+	return r.gun
+}
+
 // Target represents a simplified version of the data TUF operates on, so external
 // applications don't have to depend on TUF data types.
 type Target struct {
@@ -613,22 +618,20 @@ func (r *NotaryRepository) Publish() error {
 
 // publish pushes the changes in the given changelist to the remote notary-server
 // Conceptually it performs an operation similar to a `git rebase`
-func (r *NotaryRepository) publish(cl changelist.Changelist) error {
+func (r *NotaryRepository) publish(cl changelist.Changelist, rootKeyIDs []string) error {
 	var initialPublish bool
 	// update first before publishing
 	if err := r.Update(true); err != nil {
 		// If the remote is not aware of the repo, then this is being published
-		// for the first time.  Try to load from disk instead for publishing.
+		// for the first time.  Try to initialize the repository before publishing.
 		if _, ok := err.(ErrRepositoryNotExist); ok {
-			err := r.bootstrapRepo()
+			err := r.Initialize(rootKeyIDs)
 			if err != nil {
-				logrus.Debugf("Unable to load repository from local files: %s",
+				logrus.Debugf("Unable to initialize repository at publish-time: %s",
 					err.Error())
-				if _, ok := err.(store.ErrMetaNotFound); ok {
-					return ErrRepoNotInitialized{}
-				}
 				return err
 			}
+
 			// Ensure we will push the initial root and targets file.  Either or
 			// both of the root and targets may not be marked as Dirty, since
 			// there may not be any changes that update them, so use a
@@ -1029,7 +1032,7 @@ func (r *NotaryRepository) RotateKey(role data.RoleName, serverManagesKey bool, 
 	if err := r.rootFileKeyChange(cl, role, changelist.ActionCreate, pubKeyList); err != nil {
 		return err
 	}
-	return r.publish(cl)
+	return r.publish(cl, []string{r.GetGUN()})
 }
 
 // Given a set of new keys to rotate to and a set of keys to drop, returns the list of current keys to use
--- a/client/client.go
+++ b/client/client.go
@@ -621,7 +621,11 @@ func (r *NotaryRepository) publish(cl changelist.Changelist) error {
 		// If the remote is not aware of the repo, then this is being published
 		// for the first time.  Try to initialize the repository before publishing.
 		if _, ok := err.(ErrRepositoryNotExist); ok {
-			err := r.initializeFromCache()
+			err := r.bootstrapRepo()
+			if _, ok := err.(store.ErrMetaNotFound); ok {
+				err = r.Initialize(nil)
+			}
+
 			if err != nil {
 				return err
 			}
@@ -831,29 +835,6 @@ func (r *NotaryRepository) bootstrapRepo() error {
 	if err == nil {
 		r.tufRepo = tufRepo
 	}
-	return nil
-}
-
-// initializeFromCache looks for cached metadata to bootstrap from
-// and will initialize the repository from scratch otherwise.
-func (r *NotaryRepository) initializeFromCache() error {
-	metaCached, err := isMetaCached(r.cache)
-	if err != nil {
-		logrus.Debugf("Unable to verify on-disk cache: %s", err.Error())
-		return err
-	}
-
-	if metaCached {
-		err = r.bootstrapRepo()
-	} else {
-		err = r.Initialize(nil)
-	}
-	if err != nil {
-		logrus.Debugf("Unable to initialize repository at publish-time: %s",
-			err.Error())
-		return err
-	}
-
 	return nil
 }
 
--- a/client/client.go
+++ b/client/client.go
@@ -633,23 +633,6 @@ func (r *NotaryRepository) Publish() error {
 	return nil
 }
 
-func (r *NotaryRepository) isMetaCached() (bool, error) {
-	for _, role := range data.BaseRoles {
-		_, err := r.cache.GetSized(role.String(), store.NoSizeLimit)
-		if err != nil {
-			if _, ok := err.(store.ErrMetaNotFound); ok {
-				continue
-			}
-
-			return false, err
-		}
-
-		return true, nil
-	}
-
-	return false, nil
-}
-
 // publish pushes the changes in the given changelist to the remote notary-server
 // Conceptually it performs an operation similar to a `git rebase`
 func (r *NotaryRepository) publish(cl changelist.Changelist) error {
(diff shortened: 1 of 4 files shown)


client/client_test.go
```diff
@@ -1666,14 +1665,31 @@ func TestPublishUninitializedRepo(t *testing.T) {
...
-	requireRepoHasExpectedMetadata(t, repo, data.CanonicalRootRole, false)
-	requireRepoHasExpectedMetadata(t, repo, data.CanonicalSnapshotRole, false)
-	requireRepoHasExpectedMetadata(t, repo, data.CanonicalTargetsRole, false)
+	// metadata is created
+	requireRepoHasExpectedMetadata(t, repo, data.CanonicalRootRole, true)
+	requireRepoHasExpectedMetadata(t, repo, data.CanonicalSnapshotRole, true)
+	requireRepoHasExpectedMetadata(t, repo, data.CanonicalTargetsRole, true)
+}
+
+// Tnitializing a repo and republishing after should succeed
```
riyazdf: nit typo: `Initializing`

client/client_test.go
```diff
@@ -2650,8 +2647,10 @@ func TestRemoteRotationNoRootKey(t *testing.T) {
 	require.IsType(t, signed.ErrInsufficientSignatures{}, err)
 }
 
-// The repo hasn't been initialized, so we can't rotate
-func TestRemoteRotationNonexistentRepo(t *testing.T) {
+// The repo is initialized at publish time after
+// rotating the key. We should be denied the access
+// to metadata by the server when trying to retrieve it.
```
riyazdf: I'm not sure I follow why this is the case?

client/helpers.go
```diff
@@ -268,3 +269,59 @@ func serializeCanonicalRole(tufRepo *tuf.Repo, role data.RoleName, extraSigningK
...
+func isMetaCached(cache store.MetadataStore) (bool, error) {
+	if cache == nil {
+		return false, nil
+	}
+
+	for _, role := range data.BaseRoles {
+		_, err := cache.GetSized(role.String(), store.NoSizeLimit)
+		if err != nil {
+			if _, ok := err.(store.ErrMetaNotFound); ok {
+				continue
```
riyazdf: is this intended?  This function seems to return true if a cache exists but none of the roles have metadata in it

cmd/notary/tuf.go
```diff
@@ -142,7 +142,10 @@ func (t *tufCommander) AddToCommand(cmd *cobra.Command) {
 	cmdReset.Flags().BoolVar(&t.resetAll, "all", false, "Reset all changes shown in the status list")
 	cmd.AddCommand(cmdReset)
 
-	cmd.AddCommand(cmdTUFPublishTemplate.ToCommand(t.tufPublish))
+	cmdTUFPublish := cmdTUFPublishTemplate.ToCommand(t.tufPublish)
+	cmdTUFPublish.Flags().StringVar(&t.rootKey, "rootkey", "", "Root key to initialize the repository with")
```
riyazdf: nit: can the description indicate that it's only used if this is the first publish?

cmd/notary/tuf.go
```diff
@@ -1031,14 +1037,14 @@ func maybeAutoPublish(cmd *cobra.Command, doPublish bool, gun data.GUN, config *
 		return err
 	}
 
-	cmd.Println("Auto-publishing changes to", gun)
-	return publishAndPrintToCLI(cmd, nRepo, gun)
+	cmd.Println("Auto-publishing changes to", nRepo.GetGUN())
```
riyazdf: :+1:

endophage: `gun` is already an argument to this function. No need to create a new getter for it on NotaryRepository.

riyazdf: ah you're right, the getter isn't needed - for some reason thought I had seen it somewhere else 😅

riyazdf: I don't think you need the getter because we use the`gun` argument to get the repository [here](https://github.com/docker/notary/pull/1105/files/6a1f1c1f5a528906d486d3d0a21afef98658264f#diff-1409503b7b31865d44186c540b70fb43R1034)
```
nRepo, err := notaryclient.NewFileCachedNotaryRepository(
 		config.GetString("trust_dir"), gun, getRemoteTrustServer(config), rt, passRetriever, trustPin)
```

client/helpers.go
riyazdf: @n4ss - sorry yes, you're correct.  Could we add a docstring-style comment to this function?

client/client_test.go
n4ss: @endophage found the issue, we needed a `fullTestServer` rather than a `simpleTestServer`.

cmd/notary/tuf.go
riyazdf: hm it depends on whether we want lazy initialization to support this (we'd have to inject the root key into [this call to `Initialize`](https://github.com/docker/notary/pull/1105/files/6a1f1c1f5a528906d486d3d0a21afef98658264f#diff-1cd41a2611da49ce63630d5a51bd82c5R847)), and it will get messy with the auto-publish flag.

I'm ok with removing this for now, since `notary init` will still allow for this functionality.  We can figure out the best way to handle this for lazy init on a followup PR 👍

n4ss: sounds good, I tried to draft it but there is no clean way to propagate the root key to Initialize in this code path.. we should talk about it later indeed

client/helpers.go
n4ss: will do!

cmd/notary/tuf.go
n4ss: I was thinking about the case where this function might be called elsewhere in the same package, we have no "contract" on the fact that the gun in the argument is the same as the one used for this repo.

It's just less error-prone if we don't allow someone to provide an unrelated gun. For now it's only used properly because we get it from this same repo but people might call it in a different way..

client/client_test.go
```diff
@@ -1666,14 +1665,31 @@ func TestPublishUninitializedRepo(t *testing.T) {
...
-	requireRepoHasExpectedMetadata(t, repo, data.CanonicalSnapshotRole, false)
-	requireRepoHasExpectedMetadata(t, repo, data.CanonicalTargetsRole, false)
+	// metadata is created
+	requireRepoHasExpectedMetadata(t, repo, data.CanonicalRootRole, true)
+	requireRepoHasExpectedMetadata(t, repo, data.CanonicalSnapshotRole, true)
+	requireRepoHasExpectedMetadata(t, repo, data.CanonicalTargetsRole, true)
+}
+
+// Initializing a repo and republishing after should succeed
+func TestPublishInitializedRepo(t *testing.T) {
```
cyli: I think this test case may already be covered by `TestPublishBareRepo`

cmd/notary/tuf.go
cyli: > I was thinking about the case where this function might be called elsewhere in the same package, we have no "contract" on the fact that the gun in the argument is the same as the one used for this repo.

If the gun passed in to the constructor of the repo, wouldn't that gun and the gun on the repo by contract of the `NotaryRepository` have to be the same?

cyli: Update:  unless you mean the contract between the gun and `publishAndPrintToCLI`?

client/client.go
```diff
@@ -837,6 +834,29 @@ func (r *NotaryRepository) bootstrapRepo() error {
 	return nil
 }
 
+// initializeFromCache looks for cached metadata to bootstrap from
```
cyli: Alternately, in the `publish` logic, if `ErrMetadataNotFound` is returned by `bootstrapRepo`, we can then call `Initialize` rather than return the previous `ErrRepoNotInitialized`.

cmd/notary/tuf.go
n4ss: I mean the contract at the `publishAndPrintToCLI` level only, which can be called with different values for `repo.gun` and `gun`.

client/client_test.go
n4ss: indeed, will fix.

cmd/notary/tuf.go
cyli: Ah ok, that makes sense.

client/client.go
n4ss: Will fix with the second solution. Thanks that's way cleaner that way.

cmd/notary/tuf.go
```diff
@@ -142,7 +142,9 @@ func (t *tufCommander) AddToCommand(cmd *cobra.Command) {
 	cmdReset.Flags().BoolVar(&t.resetAll, "all", false, "Reset all changes shown in the status list")
 	cmd.AddCommand(cmdReset)
 
-	cmd.AddCommand(cmdTUFPublishTemplate.ToCommand(t.tufPublish))
```
riyazdf: nit: I think this can stay as is now that the flag is gone

client/client.go
```diff
@@ -618,17 +619,17 @@ func (r *NotaryRepository) publish(cl changelist.Changelist) error {
 	// update first before publishing
 	if err := r.Update(true); err != nil {
 		// If the remote is not aware of the repo, then this is being published
-		// for the first time.  Try to load from disk instead for publishing.
+		// for the first time.  Try to initialize the repository before publishing.
 		if _, ok := err.(ErrRepositoryNotExist); ok {
 			err := r.bootstrapRepo()
+			if _, ok := err.(store.ErrMetaNotFound); ok {
```
riyazdf: Great cleanup!  Could we add some debug logs for these cases?

cmd/notary/tuf.go
n4ss: sounds good!

client/client.go
n4ss: sure, good idea

n4ss: (fixed)

cyli: Non-blocking:  I'm wondering if this should be logged as `Info` so that users can see it.  Mainly because while trying this out, I'm seeing something like:

```
$  bin/notary -c cmd/notary/config.json delegation add repo2 targets/releases --all-paths fixtures/notary-server.crt 

Addition of delegation role targets/releases with keys [d67a3bf8194b6cccbce29c859dc3677fedfb0a5e236d24b57c37edc202cc73dd], with paths ["" <all paths>], to repository "repo2" staged for next publish.

$  bin/notary -c cmd/notary/config.json publish repo2
Pushing changes to repo2
Enter passphrase for root key with ID f6ff23d: 
Enter passphrase for new targets key with ID 4788565: 
Repeat passphrase for new targets key with ID 4788565: 
Passphrases do not match. Please retry.
Enter passphrase for new targets key with ID 4788565: 
Repeat passphrase for new targets key with ID 4788565: 
Enter passphrase for new snapshot key with ID b5c6cd5: 
Repeat passphrase for new snapshot key with ID b5c6cd5: 
Successfully published changes for repository repo2
```

And there's not a lot of difference between "passphrase for root key" and "passphrase for new targets key", etc.  It may be useful for users to see a note about what's going on, instead of wondering why they're being asked for all these passphrases (since it's hard to distinguish between the output of the lazy init vs publishing an actual change).

Possibly the log message could be "No TUF data found locally or remotely - initializing repository %s for the first time" or something?  (also totally cool with "from scratch", but "first time" might make it more clear to users since we document how notary generates keys when initializing repos for the first time.

cyli:
Other than minor bikeshedding about a log message, this LGTM!  Thanks for all your work on this!

client/client.go
n4ss: Very good point, I didn't realize how disturbing it might be regarding the passphrases.

Will fix!

endophage:
This came out a LOT shorter than I was anticipating! Great work. It looks good but I'm reviewing it late-ish at night so will take another look when I'm a bit more alert.

PR #1108, opened 2017-03-01
fixes #1107

### LINKED ISSUE #1107: tokenAuth pingClient should not assume the "v2" url is at the root.
If I host my notary-server under a sub path:
**x.x.x.x/notary**
I can use the command:
```
notary -s https://x.x.x.x/notary list ......
```
to list the targets, however when the "auth" is configured at the notary-server, the command will not challenge me for credentials.

Checked the code:
https://github.com/docker/notary/blob/master/cmd/notary/tuf.go#L892

the "/v2/" url is hard coded to the root, but in my case the notary client should "ping" the url **x.x.x.x/notary/v2** instead of **x.x.x.x/v2**

a quick fix:
```
var p string
if strings.HasSuffix(endpoint.Path, "/") {
     p = endpoint.Path + "v2/"
} else {
     p = endpoint.Path + "/v2/"
}
subPath, err := url.Parse(p)
```
can work around the problem.

reasonerjt:
When doing auth, add the path of "tuf URL" before "/v2" for ping client 

For more details please refer to #1107 

Signed-off-by: Tan Jiang <jiangd@vmware.com>

### DIFF
--- a/cmd/notary/tuf.go
+++ b/cmd/notary/tuf.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"path"
 	"strconv"
 	"strings"
 	"time"
@@ -883,7 +884,7 @@ func tokenAuth(trustServerURL string, baseTransport *http.Transport, gun data.GU
 	if endpoint.Scheme == "" {
 		return nil, fmt.Errorf("Trust server url has to be in the form of http(s)://URL:PORT. Got: %s", trustServerURL)
 	}
-	subPath, err := url.Parse("v2/")
+	subPath, err := url.Parse(path.Join(endpoint.Path, "/v2") + "/")
 	if err != nil {
 		return nil, fmt.Errorf("Failed to parse v2 subpath. This error should not have been reached. Please report it as an issue at https://github.com/docker/notary/issues: %s", err.Error())
 	}


cmd/notary/tuf.go
```diff
@@ -889,7 +889,13 @@ func tokenAuth(trustServerURL string, baseTransport *http.Transport, gun data.GU
 	if endpoint.Scheme == "" {
 		return nil, fmt.Errorf("Trust server url has to be in the form of http(s)://URL:PORT. Got: %s", trustServerURL)
 	}
-	subPath, err := url.Parse("v2/")
+	var p string
+	if strings.HasSuffix(endpoint.Path, "/") {
```
cyli: I think you can just use `filepath.Join` here, which would take care of de-duping the `/` for you (and even de-duping extra `/` as well:  https://play.golang.org/p/ALPqJzKDCq)

cyli:
jenkins, test this please

cmd/notary/tuf.go
cyli: Non-blocking:  If you wouldn't mind, could this URL validation be moved to a utility function in `storage/httpstore.go` so that it could be shared between this and the `HTTPStore`?  It'd be useful for library users of the notary client as well.

It would also make it easier to test, since you can just test that function alone (because the second thing I was going to ask is a test for this particular validation withsubpaths :))

endophage: @cyli should just be `path.Join` as `filepath` would do the wrong thing on Windows systems and these are HTTP URLs.

Could alternatively use the `url` package which will absolutely ensure it's a correct URL.

endophage: You have to be precise with the `/` but it works. https://play.golang.org/p/eVHFrPre4i

cyli: @endophage Ah good point, thanks.

reasonerjt: @endophage 
Are you saying user have to call 
```notary -s https://docker.com/notary/```??
Wouldn't it be better if the code can handle both cases, with or without the trailing ```/```

reasonerjt: @cyli 
Thanks for reviewing I'll check ```httpstore.go``` when I have time, probably early next week.

endophage: Well... I was thinking we could detect the trailing slash and add if necessary. I don't feel strongly though, what you have currently is fine.

endophage:
The circleCI failures appear legitimate. It appears `path.Join` is causing the trailing `/` on the `/v2/` to be dropped, while the existing code didn't do that. I believe that's causing a `404` from the server in the failing tests.

reasonerjt:
@endophage Thanks for the comment.
I'll fix it, sorry 'bout that.

endophage:
@reasonerjt no worries, but it does make me think we should detect the presence of a trailing `/` on the initial base url and add it there if necessary, then just use `net/url` rather than `path`.

endophage:
jenkins, test this please

endophage:
LGTM pending Jenkins but fully expect it to pass.

PR #1119, opened 2017-03-20
WIP: Client API implementation

endophage:
Huge feature list here and apologies for the PR bomb.

- New Interface type defined for client interaction with a repository (client/interface.go)
- GRPC implementation of the new interface (client_api/api/*)
- CLI integration with GRPC API if configured (cmd/notary/repo_factory.go and other necessary knock on changes in the cmd/notary package)
- JWT token auth for the GRPC API consistent with Notary Server and Docker Registry (auth/*). Some of this is to be upstreamed into docker/distribution when I get the chance.
  - If enabled "push" and "pull" actions will always be requested to guarantee the Client API can communicate with the upstream Notary Server to read repos and publish updates. The server configuration will also accept a map of `GRPC Endpoint Name : [permissions]` (documentation to be written) that allows users to arbitrarily configure in more permissions per GRPC API. Users would need their token server to issue those permissions as appropriate to their users.

Still needed:

- more tests
- documentation

Shout out to @n4ss for all the hard work on this. Go get some sleep!

cc @ecordell @lewiada


cmd/notary/repo_factory.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/docker/notary/client"
+	"github.com/docker/notary/client_api/api"
+	"github.com/docker/notary/tuf/data"
+	"github.com/docker/notary/utils"
+	"net/http"
+)
+
+const remoteConfigField = "api"
+
+type repoFactory func(gun data.GUN) (client.Repository, error)
```
ecordell: I think the naming here could be confusing, because `RepoBuilder` already exists. Maybe more explicitly `clientRepoFactory`? (same for filename as well)

cmd/clientapi/main.go
```diff
@@ -0,0 +1,75 @@
...
+		logrus.Fatal(err)
+	}
+
+	logrus.SetLevel(logrus.DebugLevel)
+
+	// FIXME: refactor with notary-signer/config.go:getAddrAndTLSConfig()
+	// and setupGRPCServer()
+	grpcAddr, tlsConfig, err := setup.GetAddrAndTLSConfig(vc)
+	if err != nil {
+		logrus.Fatal("unable to set up TLS: %s", err.Error())
```
ecordell: I think we should allow running without TLS/Auth. 

I know this isn't the main use-case being considered, but I was excited about the idea of running as a daemon locally. Your notary client would connect to your notary client daemon running locally, which you don't have ACLs on because it just uses whatever keys you gave it through normal notary operations.

This is cool because it allows other applications to interact with notary on the user's behalf (obviously they'd still need to unlock access for e.g. local keychain and verify the operations). I think it would be a powerful way for notary to integrate with other tools.

You could require TLS on the loopback but the Authorization part seems completely unnecessary - it doesn't make sense to auth with an external service to access the keys on my machine.

cmd/clientapi/main.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	upstreamAddr := vc.GetString("upstream.addr")
+	upstreamCAPath := utils.GetPathRelativeToConfig(vc, "upstream.tls_ca_file")
+
+	creds := credentials.NewTLS(tlsConfig)
+	auth, err := setup.Authorization(vc)
```
ecordell: Another auth note - for the case when the client server (is there not a better name?) is NOT running locally, it would really be great to be able to provide auth as a plugin. For Quay we use the same token format but verify it differently (the server keeps an updated list of valid signing keys instead of using a CA)...Serverside I had to wrap notary server to support running with Quay, would be great to just swap out the plugin.

Obviously Quay is just an example, for notary to be useful to non-container-registries it will need to be able to support different auth mechanisms.

auth/grpc/auth.go
```diff
@@ -0,0 +1,173 @@
...
+	google_protobuf "github.com/golang/protobuf/ptypes/empty"
+	"golang.org/x/net/context"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"net/http"
+	"net/url"
+)
+
+type guner interface {
```
ecordell: how about:

```go
type HasGUN interface {
  GetGUN() data.GUN
}
```
?

auth/grpc/auth.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	tok, errToken := c.getToken(headers["www-authenticate"])
+	if errToken != nil {
+		// couldn't get a token, log the error and return the original Unauthenticated error
+		// (the caller of the GRPC method may be relying on a grpc type error)
+		logrus.Error(err)
+		return err
+	}
+	logrus.Info("token")
```
ecordell: these log statements should be cleaned up

auth/client/session.go
```diff
@@ -0,0 +1,337 @@
...
+// NewTokenHandler creates a new AuthenicationHandler which supports
+// fetching tokens from a remote token server.
+func NewTokenHandler(transport http.RoundTripper, creds CredentialStore, clientID, scope string, actions ...string) *TokenHandler {
+	return &TokenHandler{
+		transport:     transport,
+		creds:         creds,
+		offlineAccess: false,
+		forceOAuth:    false,
+		clientID:      clientID,
+		//scopes: []Scope{
```
ecordell: these comments should be cleaned up

auth/client/session.go
```diff
@@ -0,0 +1,337 @@
...
+// using the scope grammar
+func (rs RepositoryScope) String() string {
+	repoType := "repository"
+	if rs.Class != "" {
+		repoType = fmt.Sprintf("%s(%s)", repoType, rs.Class)
+	}
+	return fmt.Sprintf("%s:%s:%s", repoType, rs.Repository, strings.Join(rs.Actions, ","))
+}
+
+//// RegistryScope represents a token scope for access
```
ecordell: if this is unused it should be removed

auth/client/session.go
```diff
@@ -0,0 +1,337 @@
...
+		Timeout:   15 * time.Second,
+	}
+}
+
+func (th *TokenHandler) Scheme() string {
+	return "bearer"
+}
+
+func (th *TokenHandler) AuthorizeRequest(params map[string]string, scopes ...string) (string, error) {
+	//th.tokenLock.Lock()
```
ecordell: remove these? can't think of a reason there would need to be a lock here.

client_api/api/server.go
```diff
@@ -0,0 +1,799 @@
...
+	r, err := srv.initRepo(data.GUN(t.GetGun()))
+	if err != nil {
+		return nil, err
+	}
+	if err := r.RemoveTarget(
+		t.GetName(), "targets",
+	); err != nil {
+		return nil, err
+	}
+	if err := publishRepo(r); err != nil {
```
ecordell: Should the responsibility to publish be pushed back to the "clientapi client" instead of in the "clientapi server"?

IIRC this mirrors the current cli behavior, just wondering if there's a good reason to keep them coupled together.

client_api/api/server.go
```diff
@@ -0,0 +1,799 @@
...
+	for index, role := range message.Roles {
+		roles[index] = data.RoleName(role)
+	}
+
+	targets, err := r.ListTargets(roles...)
+	if err != nil {
+		return nil, err
+	}
+
+	resTargets := make([]*TargetWithRole, len(targets))
```
ecordell: Can these be factored out? Maybe something like `NewTargetWithRole(Target)` and `NewTargetWithRoles([]Target)`

They only really need to be serialized here so I understand not separating it out, but if they were separate it would be *very* clear that all this server does is proxy commands to a repo object and then serialize the results.

(Same comment for all of the other serialization happening in this file)

utils/token/accesscontroller.go
```diff
@@ -0,0 +1,215 @@
+package token
```
endophage: I'm actually going to try and upstream these auth pieces back into distribution. I don't particularly want a forked version I have to keep up to date if somebody changes something. The version in docker/distribution is basically the canonical version for Docker but it was too tightly coupled with net/http and I needed to make progress.

cmd/clientapi/main.go
endophage: Worth noting HTTP2 _requires_ TLS. It'll generate self signed certs or some such if none are provided. It's entirely up to the client to set itself to insecure mode.

endophage: and I forgot to include: GRPC is only implemented over HTTP2

cmd/clientapi/main.go
endophage: Makes sense

client_api/api/server.go
endophage: In the case where somebody horizontally scales Client API's (I'm just using that as the name for this new service), there's no guarantee you'll hit the same instance again. I think it's a lot cleaner and easier to implement if every operation is a one shot that publishes (where appropriate, we intend to support things like letting clients add keys, which would still require some kind of synchronization between instances).

cmd/notary/repo_factory.go
endophage: I have some stuff to discuss at our maintainers meeting at DockerCon that may obsolete RepoBuilder :-)

auth/grpc/auth.go
endophage: Totally :-P Quite a few were me debugging because Gogland truncates how much of a value you can view in its debugger.

auth/client/session.go
endophage: There was a token cache (`map[something]something`) I removed that this was locking.

endophage:
Superseded by #1139

PR #1126, opened 2017-03-30
wildcarded trustpinning for cert ids

### LINKED ISSUE #1123: Trustpinning should accept wildcarded specific keys
Now that we have wild carded certs, we need trustpinning to accept a wildcarded GUN for a specific key. Currently only CAs can be matched against a wildcarded GUN.

endophage:
Closes #1123 


endophage:
@ecordell figure you'll be interested in this. It basically makes individual key pinning useful in light of your wildcarded root cert CNs PR we merged. It'll also be useful when we roll back the requirement for root keys to be certs in the first place.

endophage:
@riyazdf you did trustpinning originally so would like your review too.

trustpinning/certs_test.go
```diff
@@ -363,6 +363,26 @@ func TestValidateRootWithPinnedCertAndIntermediates(t *testing.T) {
 		}
 	}
 	require.Equal(t, typedSignedRoot, validatedRoot)
+
+	// test is also works with a wildcarded gun in certs
```
riyazdf: super nit: s/is/it/

trustpinning/trustpin.go
```diff
@@ -37,6 +37,11 @@ func NewTrustPinChecker(trustPinConfig TrustPinConfig, gun data.GUN, firstBootst
 		t.pinnedCertIDs = pinnedCerts
 		return t.certsCheck, nil
 	}
+	var ok bool
+	t.pinnedCertIDs, ok = wildcardMatch(gun, trustPinConfig.Certs)
```
riyazdf: nit: I think you can get away with `:=` here and avoiding the extra var, but only because this is the last use of `t.pinnedCertIDs`.  I'm generally of the opinion of the way you wrote this is more defensive though, so 👍

trustpinning/trustpin.go
```diff
@@ -121,3 +126,23 @@ func getPinnedCAFilepathByPrefix(gun data.GUN, t TrustPinConfig) (string, error)
...
+// wildcardMatch will attempt to match the most specific (longest prefix) wildcarded
+// trustpinning option for key IDs. Given the simple globbing and the use of maps,
+// it is impossible to have two different prefixes of equal length.
+// This logic also solves the issue of Go's randomization of map iteration.
+func wildcardMatch(gun data.GUN, certs map[string][]string) ([]string, bool) {
+	var (
+		longest = ""
+		ids     []string
+	)
+	for k, v := range certs {
```
riyazdf: can we rename `k` and `v` to something more descriptive like `gunPrefix` and `keyIDs`?

trustpinning/certs_test.go
```diff
@@ -363,6 +363,26 @@ func TestValidateRootWithPinnedCertAndIntermediates(t *testing.T) {
...
+		nil,
+		signedRoot,
+		"docker.io/notary/test",
+		trustpinning.TrustPinConfig{
+			Certs: map[string][]string{
+				"docker.io/notar*": {ecdsax509Key.ID()},
+			},
+			DisableTOFU: true,
+		},
+	)
```
riyazdf: can we add another test case or two for the precedence between globbed and non-globbed matches?

Ex:
this fails
```
Certs: map[string][]string{
    "docker.io/notary": {"badID"},
    "docker.io/notar*": {ecdsax509Key.ID()},
 },
```

this succeeds
```
Certs: map[string][]string{
    "docker.io/notary": {ecdsax509Key.ID()},
    "docker.io/notar*": {"badID"},
 },
```

trustpinning/trustpin.go
endophage: Will double check but I think one of the go tools was shouting at me when I originally had `:=`

trustpinning/certs_test.go
endophage: Those are the other way around right? The first should succeed and the second should fail?

riyazdf: I thought that exact matches had precedence over wildcards?

trustpinning/trustpin.go
endophage: Yeah, it complains about `non-name t.pinnedCertIDs on left side of :=`

trustpinning/certs_test.go
endophage: nevermind, misunderstood because the `docker.io/notary` was short, you meant it to be `docker.io/notary/test` right? (i.e. the gun in use in the test)

endophage:
@riyazdf I think I addressed all your comments.

trustpinning/certs_test.go
riyazdf: sorry, yes that's what I meant!

endophage:
@riyazdf `/go/src/github.com/docker/notary/trustpinning/certs_test.go:388:2: ineffectual assignment to validatedRoot` fixing it now :-P

trustpinning/certs_test.go
cyli: Similar to this, can we also add a test that the more specific glob takes precedence over the less specific glob?
e.g. 

```
			Certs: map[string][]string{
 				"docker.io/notary/te*": {"badID"},
 				"docker.io/notar*":      {ecdsax509Key.ID()},
 			},
```

fails while 

```
			Certs: map[string][]string{
 				"docker.io/notary/te*": {ecdsax509Key.ID()},
 				"docker.io/notar*":      {"badID"},
 			},
```

succeeds?

cyli: Oh hm, nm - looks like https://github.com/docker/notary/pull/1126/files#diff-46adf7b1ba08ec329d671042ce0c84efR29 provides the test for the longest.

trustpinning/trustpin_test.go
```diff
@@ -0,0 +1,33 @@
...
+	require.False(t, ok)
+
+	// wildcard match should match on segment boundaries
+	res, ok = wildcardMatch("docker.io/endophage/foo", testCerts)
+	require.Len(t, res, 1)
+	require.Equal(t, "xyz", res[0])
+	require.True(t, ok)
+
+	// wildcardMatch should also match between segment boundaries, and take
+	// the first match is finds as the ONLY match.
```
cyli: Is `docker.io/endophage/b*` actually the first match it finds?  Since we can't depend on `map` order, couldn't the keys have come in any order?  It chooses the longest matching one, right?

cyli:
Minor nitpick on a comment, but other than that LGTM!

trustpinning/trustpin_test.go
endophage: ah sorry, I need to update the comment. I originally did first and was just going to have the user prioritize in their config file however they want, but that was before I remembered the random map ordering. Will update the comment.

Issue #1120, opened 2017-03-21
Allow actions to be taken against a specific file provided on CLI


endophage:
For example, allow something akin to `notary add file://./10.targets.json target_name a_file.txt` which would take the file `./10.targets.json` and add a new entry `target_name` with the checksum and length of `a_file.txt`, then sign and write out a `./11.targets.json`. 

Exactly how signing keys would be passed/detected TBD.

PR #1128, opened 2017-04-04
`notary add` against a specific file

endophage:
- Adds a `-i/--input` flag to `notary add` and `notary addhash`. When provided, rather than creating a changelist item, the target will be added to the file provided.
  - Setting the value of `-i/--input` to `"-"` will cause a new targets file to be created. This also requires that at least one of `--keys` or `--certs` are provided as the new targets file will have no signatures from which we can look up existing keys.
- Adds `--keys` and `--certs` flags to `notary add` and `notary addhash`. 
  - If `--keys` are provided, they will be used for signing the modified file.
  - If `--certs` are provided, they are cross referenced with any items in `--keys` and where the Canonical IDs match, the _Non_-Canonical ID is used when adding the resultant signature. Any `--certs` that don't match a `--keys` entry are looked up in the CryptoService and merged with the entries from `--keys`.
  - If neither `--keys` or `--certs` are provided, we will attempt to re-sign the file with any keys for which there is an existing signature on the provided `-i/--input`.

Result is output to the file provided by the `-o/--output` option. The flag is required in the case `-i/--input` is used. 

Begins to address #1120 

cc @riyazdf (if you have time) and @n4ss so you can test it with your export/import code.

### DIFF
--- a/client/isolated.go
+++ b/client/isolated.go
@@ -1,0 +1,124 @@
+package client
+
+import (
+	"encoding/json"
+	"errors"
+	"github.com/Sirupsen/logrus"
+	"github.com/docker/notary"
+	"github.com/docker/notary/cryptoservice"
+	"github.com/docker/notary/tuf/data"
+	"github.com/docker/notary/tuf/signed"
+	"io"
+	"io/ioutil"
+)
+
+// AddTargetToFile adds the target to the file found at filePath, and writes the updated and re-signed version
+// to the outBuf.
+// If signingKeys are provided, they are used to sign the file. If lookupKeys are provided, they
+// will be looked up in the CryptoService and used for signing where found.
+// If _both_ signingKeys, and lookupKeys are provided, the public keys in lookupKeys will be added to signingKeys
+// and the merged list will be used for signing.
+// If _neither_ signingKeys, no lookupKeys are provided, the key IDs associated with any existing signatures are
+// used to search for keys in the CryptoService and used to sign.
+func AddTargetToFile(baseDir string, retriever notary.PassRetriever, outBuf io.Writer, filePath string, target *Target, signingKeys map[string]data.PrivateKey, lookupKeys map[string]data.PublicKey) error {
+	// do some setup
+	var fakeRole data.RoleName = data.CanonicalTargetsRole
+	ks, err := getKeyStores(baseDir, retriever)
+	if err != nil {
+		return err
+	}
+	cs := cryptoservice.NewCryptoService(ks...)
+
+	var parsed *data.SignedTargets
+	if filePath == "-" {
+		// setting the filePath to "-" indicates we have no starting file, a new one should be created
+		parsed = data.NewTargets()
+	} else {
+		// read and parse input file
+		file, err := ioutil.ReadFile(filePath)
+		if err != nil {
+			return err
+		}
+
+		parsed = &data.SignedTargets{}
+		err = json.Unmarshal(file, parsed)
+		if err != nil {
+			return err
+		}
+	}
+
+	// add target
+	parsed.Signed.Targets[target.Name] = data.FileMeta{
+		Length: target.Length,
+		Hashes: target.Hashes,
+	}
+
+	// sign file
+	parsed.Signed.Expires = data.DefaultExpires(fakeRole)
+	parsed.Signed.Version = parsed.Signed.Version + 1
+	signedObj, err := parsed.ToSigned()
+	if err != nil {
+		return err
+	}
+
+	if signingKeys == nil {
+		// just in case, we don't want to panic when we try and assign
+		signingKeys = make(map[string]data.PrivateKey)
+	}
+	for canonID, pubKey := range lookupKeys {
+		privKey, _, err := cs.GetPrivateKey(canonID)
+		if err != nil {
+			// log the canonical ID as this would be the filename the person should look for
+			logrus.Errorf("key with ID %s not found", canonID)
+			continue
+		}
+		signingKeys[pubKey.ID()] = privKey
+	}
+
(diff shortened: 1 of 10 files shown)


endophage:
I don't have time to do it right now but this PR has also made me see a lot of places we could further deconstruct and make more "functional" a quite a few things for better composition when we want to start adding more interesting use cases like this.

endophage:
Don't merge just yet. I wanted to just output the result to stdout but that's kind of a pain when we ask for a password. You can't just `notary ... > file`. Adding the `-o` flag to specify a file for writing the result.

client/isolated.go
```diff
@@ -0,0 +1,124 @@
...
+	// do some setup
+	var fakeRole data.RoleName = data.CanonicalTargetsRole
+	ks, err := getKeyStores(baseDir, retriever)
+	if err != nil {
+		return err
+	}
+	cs := cryptoservice.NewCryptoService(ks...)
+
+	var parsed *data.SignedTargets
+	if filePath == "-" {
```
cyli: Non-blocking:  It totally makes sense for `-` to mean "no file" when provided on the CLI, but in the library code, but it seems more idiomatic in go to represent "no value provided" with empty string.

endophage: Hmmm, I think I'd probably have to switch to a `*string` then to differentiate "no value" (i.e. use traditional `notary add` logic) vs "empty value" (use new `notary add` logic, creating new empty targets file). Will take a look.

client/isolated.go
```diff
@@ -0,0 +1,124 @@
...
+	"io/ioutil"
+)
+
+// AddTargetToFile adds the target to the file found at filePath, and writes the updated and re-signed version
+// to the outBuf.
+// If signingKeys are provided, they are used to sign the file. If lookupKeys are provided, they
+// will be looked up in the CryptoService and used for signing where found.
+// If _both_ signingKeys, and lookupKeys are provided, the public keys in lookupKeys will be added to signingKeys
+// and the merged list will be used for signing.
+// If _neither_ signingKeys, no lookupKeys are provided, the key IDs associated with any existing signatures are
```
cyli: Non-blocking:  Not sure if we've decided against it, but there was a discussion in slack about maybe having the CLI look up the existing signatures, and the library function itself takes the signing keys or the lookup keys.

endophage: the function internally looks up keys based on signatures if nothing else was provided. I felt that was better than having to pre-parse the file and do the lookup of keys based on signatures. I figured in some cases one might only have the public key (for example, in the case the private key is in a Yubikey), so `lookupKeys` is basically the list of public keys provided on the CLI.

client/isolated.go
cyli: Ah ok, I just meant if you got a "-" in the CLI, you could pass a "" to the library code, but I can see why it may be more convenient to just pass it through as is for now, since we might want to refactor the client and/or the tuf repo object stuff later anyway.

endophage:
@cyli @riyazdf @ecordell @HuKeping is there anything I need to change here to get sign off?

cyli:
@endophage Should this just be a different command?  The `add` and `addhash` commands require a GUN, which is extraneous when just creating a file (we aren't looking up the keys based on role, etc.)

cmd/notary/tuf.go
```diff
@@ -154,7 +165,11 @@ func (t *tufCommander) AddToCommand(cmd *cobra.Command) {
 
 	cmdTUFAdd := cmdTUFAddTemplate.ToCommand(t.tufAdd)
 	cmdTUFAdd.Flags().StringSliceVarP(&t.roles, "roles", "r", nil, "Delegation roles to add this target to")
+	cmdTUFAdd.Flags().StringVarP(&t.input, "input", "i", "", "Specific file to add the target to")
+	cmdTUFAdd.Flags().StringVarP(&t.output, "output", "o", "", "File to output result to. Only used in tandem with --input.")
 	cmdTUFAdd.Flags().BoolVarP(&t.autoPublish, "publish", "p", false, htAutoPublish)
+	cmdTUFAdd.Flags().StringSliceVar(&t.privKeyPaths, "keys", nil, "Private keys to sign with. Only valid if used with the --file flag")
```
cyli: For these, the flag seems to be `--input` now, now `--file`.

cyli:
(as a note, other than the CLI docstring changes to be `--input` instead of `--file`, my other comment isn't blocking, just a suggestion because I thought it could be confusing for users who might expect the valid signing keys to be looked up in the existing repo)

endophage:
@cyli I agree GUN is confusing... at the same time I didn't want to add more commands that performed the same functions as existing commands... We'd end up with `add_to_file` and `addhash_to_file` or some equally ugly precedent. I could however drop the GUN from the CLI in the case where an input file is provided. It would require some extra logic but isn't terrible. Do you think that would be better? I was initially worried it might be more confusing.

cyli:
@endophage Agreed about `add_to_file` and `addhash_to_file` - maybe making the GUN optional is a good solution?  I wonder if maybe we should add one more layer of subcommands, so the repo stuff could be `notary repo list` etc?  I'm not entirely sure what would be the best UI.  

The only reason I worry about the GUN being optional is that "this flag only applies if this other flag is used, otherwise, use this third flag" might be confusing to explain.  :|  

I guess one question to ask is where do we see this going in the future?  Do we eventually want to have file-level versions of all the commands, or just the targets ones?  If it's just wrt to targets (`add`, `addhash`, and `remove`), maybe it's just fine to special case these, but if we wanted to add this to delegations too (for instance when the delegation name does not need to include the full path including parents) maybe it's worth considering splitting out file-scoped commands from repo-scoped commands?

riyazdf:
jenkins, test this please

PR #1129, opened 2017-04-04
update gorm to reduce size of #1110

endophage:
@ecordell this is just to make your channels PR appear a little smaller. I want to get a feel for the amount of churn without the vendoring distorting it.

This just updates gorm and adds the associated inflection package. There are also a couple of necessary derefs as gorm's initialization now return a `*`.


cyli:
I think you also have to pull in the changes from the server and signer config tests, as per https://github.com/docker/notary/pull/1110/files#r105954902.

endophage:
Ah thanks! I'm now getting the `killed` thing when just running tests on MacOS so I pushed to see what was failing :-P

server/storage/sql_models.go
```diff
@@ -50,7 +50,7 @@ func (c Change) TableName() string {
 // CreateTUFTable creates the DB table for TUFFile
 func CreateTUFTable(db gorm.DB) error {
 	// TODO: gorm
-	query := db.Set("gorm:table_options", "ENGINE=InnoDB DEFAULT CHARSET=utf8").CreateTable(&TUFFile{})
+	query := db.AutoMigrate(&TUFFile{})
```
riyazdf: seems like we can still have the `Set()` call here and chain that into an `AutoMigrate`: http://jinzhu.me/gorm/database.html#migration

endophage: The switch to sqlite in a couple of tests mean the `ENGINE` options is no longer valid.

cyli:
LGTM

endophage:
Retrying the yubikey tests, it was another random connectivity failure (these are happening a lot recently). Will merge if they pass

ecordell:
Thanks! I'll rebase the other PR on this.

ecordell:
LGTM a little late :)

PR #1134, opened 2017-04-07
`notary key generate` shouldn't be restricted to root keys

endophage:
It also gains the `-o`/`--output` flag, which if provided will cause the public and private key to be output to the filepath specified in the flag with the .pub and .priv file extensions respectively.

### DIFF
--- a/cmd/notary/keys.go
+++ b/cmd/notary/keys.go
@@ -46,13 +46,14 @@ var cmdKeyGenerateKeyTemplate = usageTemplate{
 	Use:   "generate [ algorithm ]",
 	Short: "Generates a new key with a given algorithm.",
 	Long: "Generates a new key with a given algorithm. If hardware key " +
-		"storage (e.g. a Yubikey) is available, the key will be stored both " +
+		"storage (e.g. a Yubikey) is available, generated root keys will be stored both " +
 		"on hardware and on disk (so that it can be backed up).  Please make " +
 		"sure to back up and then remove this on-key disk immediately" +
 		"afterwards. If a `--output` file name is provided, two files will " +
-		"be written, <output>.pub and <output>.priv, containing the public" +
+		"be written, <output>.pem and <output>-key.pem, containing the public" +
 		"and private keys respectively (the key will not be stored in Notary's " +
-		"key storage). If no `--role` is provided, \"root\" will be assumed.",
+		"key storage, including any connected hardware storage). If no `--role` " +
+		"is provided, \"root\" will be assumed.",
 }
 
 var cmdKeyRemoveTemplate = usageTemplate{
@@ -223,14 +224,14 @@ func (k *keyCommander) keysGenerate(cmd *cobra.Command, args []string) error {
 
 		pubKey, err := cs.Create(data.RoleName(k.generateRole), "", algorithm)
 		if err != nil {
-			return fmt.Errorf("Failed to create a new root key: %v", err)
+			return fmt.Errorf("Failed to create a new %s key: %v", k.generateRole, err)
 		}
 
 		cmd.Printf("Generated new %s %s key with keyID: %s\n", algorithm, k.generateRole, pubKey.ID())
 		return nil
 	}
 
-	// if we had an outfile set, we'll write 2 files with the given name, appending .pub and .priv for the
+	// if we had an outfile set, we'll write 2 files with the given name, appending .pem and -key.pem for the
 	// public and private keys respectively
 	return generateKeyToFile(k.generateRole, algorithm, k.getRetriever(), k.outFile)
 }
@@ -267,8 +268,9 @@ func generateKeyToFile(role, algorithm string, retriever notary.PassRetriever, o
 		return errors.New("no password provided")
 	}
 
-	privFile := strings.Join([]string{outFile, "priv"}, ".")
-	pubFile := strings.Join([]string{outFile, "pub"}, ".")
+	privFileName := strings.Join([]string{outFile, "key"}, "-")
+	privFile := strings.Join([]string{privFileName, "pem"}, ".")
+	pubFile := strings.Join([]string{outFile, "pem"}, ".")
 
 	err = ioutil.WriteFile(privFile, pemPrivKey, notary.PrivNoExecPerms)
 	if err != nil {
--- a/cmd/notary/keys.go
+++ b/cmd/notary/keys.go
@@ -1,12 +1,18 @@
 package main
 
 import (
+	"encoding/pem"
+	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 
 	"github.com/docker/notary"
 	notaryclient "github.com/docker/notary/client"
@@ -14,10 +20,8 @@ import (
 	store "github.com/docker/notary/storage"
 	"github.com/docker/notary/trustmanager"
 	"github.com/docker/notary/tuf/data"
+	tufutils "github.com/docker/notary/tuf/utils"
 	"github.com/docker/notary/utils"
-
-	"github.com/spf13/cobra"
-	"github.com/spf13/viper"
 )
 
 var cmdKeyTemplate = usageTemplate{
@@ -38,10 +42,17 @@ var cmdRotateKeyTemplate = usageTemplate{
 	Long:  `Generates a new key for the given Globally Unique Name and role (one of "snapshot", "targets", "root", or "timestamp").  If rotating to a server-managed key, a new key is requested from the server rather than generated.  If the generation or key request is successful, the key rotation is immediately published.  No other changes, even if they are staged, will be published.`,
 }
 
-var cmdKeyGenerateRootKeyTemplate = usageTemplate{
+var cmdKeyGenerateKeyTemplate = usageTemplate{
 	Use:   "generate [ algorithm ]",
-	Short: "Generates a new root key with a given algorithm.",
-	Long:  "Generates a new root key with a given algorithm. If hardware key storage (e.g. a Yubikey) is available, the key will be stored both on hardware and on disk (so that it can be backed up).  Please make sure to back up and then remove this on-key disk immediately afterwards.",
+	Short: "Generates a new key with a given algorithm.",
+	Long: "Generates a new key with a given algorithm. If hardware key " +
+		"storage (e.g. a Yubikey) is available, the key will be stored both " +
+		"on hardware and on disk (so that it can be backed up).  Please make " +
+		"sure to back up and then remove this on-key disk immediately" +
+		"afterwards. If a `--output` file name is provided, two files will " +
+		"be written, <output>.pub and <output>.priv, containing the public" +
+		"and private keys respectively (the key will not be stored in Notary's " +
+		"key storage). If no `--role` is provided, \"root\" will be assumed.",
 }
 
 var cmdKeyRemoveTemplate = usageTemplate{
@@ -80,17 +91,29 @@ type keyCommander struct {
 	legacyVersions         int
 	input                  io.Reader
 
-	keysImportRole string
-	keysImportGUN  string
-	exportGUNs     []string
-	exportKeyIDs   []string
-	outFile        string
+	importRole    string
+	generateRole  string
+	keysImportGUN string
+	exportGUNs    []string
+	exportKeyIDs  []string
+	outFile       string
 }
 
 func (k *keyCommander) GetCommand() *cobra.Command {
 	cmd := cmdKeyTemplate.ToCommand(nil)
 	cmd.AddCommand(cmdKeyListTemplate.ToCommand(k.keysList))
-	cmd.AddCommand(cmdKeyGenerateRootKeyTemplate.ToCommand(k.keysGenerateRootKey))
+	cmdGenerate := cmdKeyGenerateKeyTemplate.ToCommand(k.keysGenerate)
+	cmdGenerate.Flags().StringVarP(
+		&k.outFile,
+		"output",
+		"o",
+		"",
+		"Filepath to write export output to",
+	)
+	cmdGenerate.Flags().StringVarP(
+		&k.generateRole, "role", "r", "root", "Role to generate key with, defaulting to \"root\".",
+	)
+	cmd.AddCommand(cmdGenerate)
 	cmd.AddCommand(cmdKeyRemoveTemplate.ToCommand(k.keyRemove))
 	cmd.AddCommand(cmdKeyPasswdTemplate.ToCommand(k.keyPassphraseChange))
 	cmdRotateKey := cmdRotateKeyTemplate.ToCommand(k.keysRotate)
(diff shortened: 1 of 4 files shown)


cmd/notary/keys.go
```diff
@@ -14,10 +14,13 @@ import (
 	store "github.com/docker/notary/storage"
 	"github.com/docker/notary/trustmanager"
 	"github.com/docker/notary/tuf/data"
+	tufutils "github.com/docker/notary/tuf/utils"
```
ecordell: I think in other packages this is `tufUtils`

cmd/notary/keys.go
```diff
@@ -38,7 +41,7 @@ var cmdRotateKeyTemplate = usageTemplate{
 	Long:  `Generates a new key for the given Globally Unique Name and role (one of "snapshot", "targets", "root", or "timestamp").  If rotating to a server-managed key, a new key is requested from the server rather than generated.  If the generation or key request is successful, the key rotation is immediately published.  No other changes, even if they are staged, will be published.`,
 }
 
-var cmdKeyGenerateRootKeyTemplate = usageTemplate{
+var cmdKeyGenerateKeyTemplate = usageTemplate{
 	Use:   "generate [ algorithm ]",
 	Short: "Generates a new root key with a given algorithm.",
```
ecordell: Should update the description

cmd/notary/keys.go
```diff
@@ -90,7 +93,15 @@ type keyCommander struct {
 func (k *keyCommander) GetCommand() *cobra.Command {
 	cmd := cmdKeyTemplate.ToCommand(nil)
 	cmd.AddCommand(cmdKeyListTemplate.ToCommand(k.keysList))
-	cmd.AddCommand(cmdKeyGenerateRootKeyTemplate.ToCommand(k.keysGenerateRootKey))
+	cmdGenerate := cmdKeyGenerateKeyTemplate.ToCommand(k.keysGenerate)
+	cmdGenerate.Flags().StringVarP(
+		&k.outFile,
```
ecordell: Should `--output` only output the public key? Or at least have different flags for pub/priv out?

I'm just thinking about the case where you're storing in a yubikey but still want the public part output to a file.

cmd/notary/keys.go
```diff
@@ -159,7 +170,7 @@ func (k *keyCommander) keysList(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
-func (k *keyCommander) keysGenerateRootKey(cmd *cobra.Command, args []string) error {
+func (k *keyCommander) keysGenerate(cmd *cobra.Command, args []string) error {
```
ecordell: Shouldn't this take a `role` now?

cmd/notary/keys.go
```diff
@@ -189,19 +200,69 @@ func (k *keyCommander) keysGenerateRootKey(cmd *cobra.Command, args []string) er
...
+
+	// if no outFile is provided, use the known key stores
+	if k.outFile == "" {
+		ks, err := k.getKeyStores(config, true, true)
+		if err != nil {
+			return err
+		}
+		cs := cryptoservice.NewCryptoService(ks...)
+
+		pubKey, err := cs.Create(data.CanonicalRootRole, "", algorithm)
```
ecordell: I think the UX is a little confusing - it generates a root key if no output is defined?

cmd/notary/keys.go
endophage: This is to work in tandem with the PR to update `notary add` which allows keys to be provided on the CLI. My expectation is this would be used to generate delegation keys, which at the moment is a pain as they have to be generated with OpenSSL.

cmd/notary/keys.go
endophage: The command functions have a set signature required to meet one of the Cobra interfaces. It can't be changed.

cmd/notary/keys.go
endophage: Yeah, I wasn't sure about how to handle that. I didn't want to add a flag while I was uncertain about whether we'd be following up the PKCS#8 stuff with a PR to remove roles and guns from the PEMs.

Having seen how people use Notary, and have trouble using Notary, one of the things I wanted to discuss at DockerCon was how much we try to farm out some of the data management. We're not doing any special with keys so would it be more obvious for users if we just made them manage their own keys? Key management right now is almost too magic... We'd still need integrations for things like Yubikeys, but maybe a user would provide CLI args saying "Use Slot 3", rather than all the automatic detection we do right now.

cmd/notary/keys.go
endophage: Did a search in the project, it's consistently `tufutils` (all lower case)

cmd/notary/keys.go
ecordell: Right! I meant as one of the command args, which you took care of.

endophage: Ah sorry, misunderstood!

ecordell:
@endophage this looks good to me if there's an extra test or two coming

endophage:
@ecordell test added to cover the new functionality.

cmd/notary/keys.go
```diff
@@ -38,10 +42,17 @@ var cmdRotateKeyTemplate = usageTemplate{
...
+var cmdKeyGenerateKeyTemplate = usageTemplate{
 	Use:   "generate [ algorithm ]",
-	Short: "Generates a new root key with a given algorithm.",
-	Long:  "Generates a new root key with a given algorithm. If hardware key storage (e.g. a Yubikey) is available, the key will be stored both on hardware and on disk (so that it can be backed up).  Please make sure to back up and then remove this on-key disk immediately afterwards.",
+	Short: "Generates a new key with a given algorithm.",
+	Long: "Generates a new key with a given algorithm. If hardware key " +
+		"storage (e.g. a Yubikey) is available, the key will be stored both " +
+		"on hardware and on disk (so that it can be backed up).  Please make " +
+		"sure to back up and then remove this on-key disk immediately" +
+		"afterwards. If a `--output` file name is provided, two files will " +
```
cyli: Non-blocking grammar nit:  Maybe something like:  "If the `--output` flag is provided, two files will be written to disk:  the public key as `<output>.pub` and a private key as `<output>.priv` relative to the current working directory?

endophage: If `--output` is absolute is won't be relative to the cwd :-P I didn't want to get into the intricacies and was relying on user general knowledge of CLI tools to make sensible assumptions about providing paths to a tool.

cyli:
Since there are no restrictions on which roles can be used for `generateRole`, we can generate targets/timestamp/snapshot keys as well.  For those, should we also accept a GUN, or do we basically not really care about GUNs for those for now?

Also, could we add one more test for generating a delegation key without passing `--output`?

cmd/notary/keys.go
cyli: I think that still may be the term, though.  If you give something a relative link, it's relative to X, but absolute is always just absolute.  I totally understand why we may want to leave that part out, though.  

But would it make sense to include the rest of the above?  I just mainly wanted to directly state that the extensions will be appended to whatever output file name the user gives us.

cmd/notary/keys.go
```diff
@@ -189,19 +212,77 @@ func (k *keyCommander) keysGenerateRootKey(cmd *cobra.Command, args []string) er
...
+			return fmt.Errorf("Failed to create a new root key: %v", err)
+		}
+
+		cmd.Printf("Generated new %s %s key with keyID: %s\n", algorithm, k.generateRole, pubKey.ID())
+		return nil
+	}
+
+	// if we had an outfile set, we'll write 2 files with the given name, appending .pub and .priv for the
+	// public and private keys respectively
+	return generateKeyToFile(k.generateRole, algorithm, k.getRetriever(), k.outFile)
```
cyli: This seems to only write the keys to a file, but does not store the key in the yubikey, if it's available.

cyli: Oh is that what `the key will not be stored in Notary's key storage` means?  Neither the key storage disk or the yubikey?  Could we clarify that in the docstring for the command?  I thought providing an outfile would back up the keys to a different location, but also write to the yubikey.

endophage: Yeah, it _only_ outputs to the provided file and doesn't store it anywhere else. I'm trying to reduce the magic of the commands. It would still be possible to subsequently `notary key import` to load the key in to a yubikey.

cmd/notary/keys.go
endophage: I can rephrase to make that clearer

cmd/notary/keys.go
cyli: At some point, not in this PR, I wonder if it'd make sense to just generate the key directly in the yubikey if available unless `--output` is provided (no backups - the `output` can be your backup, since we've had issues where folks would prefer not having the key anywhere else).

endophage:
@cyli it came up in the PCKS#8 PR, I think GUN's in the key PEMs are going to be deprecated. Especially as keys are used across repos it become less relevant.

cyli:
>  I think GUNs in the key PEMs are going to be deprecated. Especially as keys are used across repos it become less relevant.

That makes sense, thanks

endophage:
@ecordell @riyazdf @cyli was there anything else I can do here? (will rebase it now)

cyli:
@endophage Are we expecting that this PR allows non-root keys to be imported into the yubikey?

endophage:
I wasn't planning on doing that here.

cyli:
@endophage Er sorry, not import, but that the generated non-root keys will be stored in the yubikey?

endophage:
Same answer, wasn't planning to make any changes to the yubikey logic in this PR.

cyli:
@endophage This LGTM with the caveat of updating the CLI docstring to make it clear that only root keys will be generated and put into the yubikey for now.  I know the next one is probably coming soon, but if it gets delayed or if we have a release between this one and when that one comes, it may be confusing for users who may expect otherwise.

cmd/notary/keys.go
```diff
@@ -38,10 +42,17 @@ var cmdRotateKeyTemplate = usageTemplate{
...
 }
 
-var cmdKeyGenerateRootKeyTemplate = usageTemplate{
+var cmdKeyGenerateKeyTemplate = usageTemplate{
 	Use:   "generate [ algorithm ]",
-	Short: "Generates a new root key with a given algorithm.",
-	Long:  "Generates a new root key with a given algorithm. If hardware key storage (e.g. a Yubikey) is available, the key will be stored both on hardware and on disk (so that it can be backed up).  Please make sure to back up and then remove this on-key disk immediately afterwards.",
+	Short: "Generates a new key with a given algorithm.",
+	Long: "Generates a new key with a given algorithm. If hardware key " +
+		"storage (e.g. a Yubikey) is available, the key will be stored both " +
```
riyazdf: Proposed amendment:

`Generates a new key with a given algorithm. If the key is a root key and hardware key storage (e.g. a Yubikey) is available...`

cmd/notary/keys.go
```diff
@@ -189,19 +212,77 @@ func (k *keyCommander) keysGenerateRootKey(cmd *cobra.Command, args []string) er
...
+	if k.outFile == "" {
+		ks, err := k.getKeyStores(config, true, true)
+		if err != nil {
+			return err
+		}
+		cs := cryptoservice.NewCryptoService(ks...)
+
+		pubKey, err := cs.Create(data.RoleName(k.generateRole), "", algorithm)
+		if err != nil {
+			return fmt.Errorf("Failed to create a new root key: %v", err)
```
riyazdf: s/"root key"/k.generateRole

cmd/notary/keys.go
```diff
@@ -189,19 +212,77 @@ func (k *keyCommander) keysGenerateRootKey(cmd *cobra.Command, args []string) er
...
+
+	// if no outFile is provided, use the known key stores
+	if k.outFile == "" {
+		ks, err := k.getKeyStores(config, true, true)
+		if err != nil {
+			return err
+		}
+		cs := cryptoservice.NewCryptoService(ks...)
+
+		pubKey, err := cs.Create(data.RoleName(k.generateRole), "", algorithm)
```
riyazdf: is empty GUN intended?  I noticed a comment about GUNs in PEMs being potentially deprecated but lost track if it was referring to this.

cmd/notary/keys.go
```diff
@@ -189,19 +212,77 @@ func (k *keyCommander) keysGenerateRootKey(cmd *cobra.Command, args []string) er
...
+	if chosenPassphrase != "" {
+		pemPrivKey, err = tufutils.EncryptPrivateKey(privKey, data.RoleName(role), "", chosenPassphrase)
+		if err != nil {
+			return err
+		}
+	} else {
+		return errors.New("no password provided")
+	}
+
+	privFile := strings.Join([]string{outFile, "priv"}, ".")
```
riyazdf: I'm wondering if we should use `.key` instead, so that manually moving the key and importing should just work?  `.priv` is conceptually fine with me, just that I'd like to add a `notary key import` test for it if we want to keep it that way.

ecordell: the cfssl defaults are:

private: `<name>-key.pem`
public: `<name>.pem`

which seems reasonable to me, though I don't have a strong preference

cmd/notary/keys.go
ecordell: This would be necessary for the "templating", right?

Perhaps this PR should remove the GUN checks when parsing PEM files, too?

ecordell:
To clarify my approval: LGTM after @riyazdf's comments are addressed

Excited to see this in! This is a solid UX improvement.

cmd/notary/keys.go
endophage: I'm happy to go with the cfssl filenames for consistency.

cmd/notary/keys.go
endophage: I think but I'm not certain we have those checks in a few scattered places. With the PRs coming in for providing certs and keys on the CLI, I'd rather hold off on making those changes here to keep potential conflicts down.

cmd/notary/keys.go
```diff
@@ -189,19 +213,78 @@ func (k *keyCommander) keysGenerateRootKey(cmd *cobra.Command, args []string) er
...
+		pubKey, err := cs.Create(data.RoleName(k.generateRole), "", algorithm)
+		if err != nil {
+			return fmt.Errorf("Failed to create a new %s key: %v", k.generateRole, err)
+		}
+
+		cmd.Printf("Generated new %s %s key with keyID: %s\n", algorithm, k.generateRole, pubKey.ID())
+		return nil
+	}
+
+	// if we had an outfile set, we'll write 2 files with the given name, appending .pub and .priv for the
```
riyazdf: nit: this should be `.pem` and `-key.pem`

cmd/notary/keys_test.go
```diff
@@ -30,6 +30,7 @@ import (
 	"github.com/docker/notary/trustpinning"
 	"github.com/docker/notary/tuf/data"
 	"github.com/docker/notary/tuf/utils"
+	"path/filepath"
```
riyazdf: nit: can we put this import into the first block?

cmd/notary/keys_test.go
```diff
@@ -806,3 +807,31 @@ func TestImportKeysNonexistentFile(t *testing.T) {
...
+
+	_, err := runCommand(t, tempDir, "key", "generate", data.ECDSAKey, "--role", "targets")
+	require.NoError(t, err)
+	assertNumKeys(t, tempDir, 0, 1, false)
+
+	_, err = runCommand(t, tempDir, "key", "generate", data.ECDSAKey, "--role", "targets", "-o", filepath.Join(tempDir, "testkeys"))
+	require.NoError(t, err)
+	assertNumKeys(t, tempDir, 0, 1, false) // key shouldn't be written to store and won't show up in keylist
+
+	// test that we can read the keys we created
```
riyazdf: could we additionally test a `key rotate` and/or `key import` to check that the notary CLI can ingest the key directly?

riyazdf:
jenkins, test this please

cmd/notary/keys_test.go
endophage: Added a test for import.

riyazdf:
@rubenmachuca FYI - you might have posted some personal information on a comment you left here.  I've deleted the original comment, please reach out (my email is in my signoffs) if you have additional questions

PR #1140, opened 2017-04-25
Pin migrate binary

riyazdf:
temporary workaround for https://github.com/mattes/migrate/issues/202 since we have not tested v3 nor does it support all drivers yet

Closes https://github.com/docker/notary/issues/1142

### DIFF
--- a/server.Dockerfile
+++ b/server.Dockerfile
@@ -3,8 +3,10 @@ MAINTAINER David Lawrence "david.lawrence@docker.com"
 
 RUN apk add --update git gcc libc-dev && rm -rf /var/cache/apk/*
 
-# Install SQL DB migration tool
-RUN go get github.com/mattes/migrate
+# Pin to the specific v1 version
+RUN git clone -b v1 https://github.com/mattes/migrate.git /go/src/github.com/mattes/migrate/ && \
+    go get github.com/mattes/migrate && \
+    go build -tags 'mysql' -o /usr/local/bin/migrate github.com/mattes/migrate
 
 ENV NOTARYPKG github.com/docker/notary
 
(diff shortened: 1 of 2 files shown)


server.Dockerfile
```diff
@@ -3,8 +3,10 @@ MAINTAINER David Lawrence "david.lawrence@docker.com"
 
 RUN apk add --update git gcc libc-dev && rm -rf /var/cache/apk/*
 
-# Install SQL DB migration tool
-RUN go get github.com/mattes/migrate
+# Pin to the specific v1 version
+RUN git clone -b v1 https://github.com/mattes/migrate.git /go/src/github.com/mattes/migrate/
+RUN go get -u -v github.com/mattes/migrate && \
```
endophage: Fewer layers (for all that it matters in the weeks until we get 17.05 multi-stage builds):
```
RUN git clone ... \
    && go get ...
```
Also, not sure we need the `-u` flag. If we just cloned, it shoudn't be necessary to ask go to also do a network update.

endophage: Also, if it's already cloned, you might just be able to do a `go install -tags 'mysql' =o ...` and skip the `go get` entirely

riyazdf: Removed `-u` flag and collapsed into one layer.

I tried skipping the `go get` and ran into dependency issues on the build, so left as is:
```
Cloning into '/go/src/github.com/mattes/migrate'...
src/github.com/mattes/migrate/driver/ql/ql.go:13:2: cannot find package "github.com/cznic/ql/driver" in any of:
	/usr/local/go/src/github.com/cznic/ql/driver (from $GOROOT)
	/go/src/github.com/cznic/ql/driver (from $GOPATH)
...
```

endophage: Ah yeah, that makes sense if migrate isn't vendoring dependencies (which it isn't)

riyazdf:
jenkins needed a kick, repushed the same commit.

cyli:
LGTM!  Thanks for fixing this so quickly!

PR #1146, opened 2017-05-04
Expose targets custom data in the NotaryRepository API

### LINKED ISSUE #1143: Expose targets custom data in the NotaryRepository API.

ashfall:
Closes #1143

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/Sirupsen/logrus"
+	canonicaljson "github.com/docker/go/canonical/json"
 	"github.com/docker/notary"
 	"github.com/docker/notary/client/changelist"
 	"github.com/docker/notary/cryptoservice"
@@ -128,9 +129,10 @@ func (r *NotaryRepository) GetGUN() data.GUN {
 // Target represents a simplified version of the data TUF operates on, so external
 // applications don't have to depend on TUF data types.
 type Target struct {
-	Name   string      // the name of the target
-	Hashes data.Hashes // the hash of the target
-	Length int64       // the size in bytes of the target
+	Name   string                   // the name of the target
+	Hashes data.Hashes              // the hash of the target
+	Length int64                    // the size in bytes of the target
+	Custom canonicaljson.RawMessage // the custom data provided to describe the file at TARGETPATH
 }
 
 // TargetWithRole represents a Target that exists in a particular role - this is
@@ -141,7 +143,7 @@ type TargetWithRole struct {
 }
 
 // NewTarget is a helper method that returns a Target
-func NewTarget(targetName string, targetPath string) (*Target, error) {
+func NewTarget(targetName, targetPath string, targetCustom canonicaljson.RawMessage) (*Target, error) {
 	b, err := ioutil.ReadFile(targetPath)
 	if err != nil {
 		return nil, err
@@ -152,7 +154,7 @@ func NewTarget(targetName string, targetPath string) (*Target, error) {
 		return nil, err
 	}
 
-	return &Target{Name: targetName, Hashes: meta.Hashes, Length: meta.Length}, nil
+	return &Target{Name: targetName, Hashes: meta.Hashes, Length: meta.Length, Custom: targetCustom}, nil
 }
 
 func rootCertKey(gun data.GUN, privKey data.PrivateKey) (data.PublicKey, error) {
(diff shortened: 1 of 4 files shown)


client/client.go
```diff
@@ -360,7 +362,14 @@ func (r *NotaryRepository) AddTarget(target *Target, roles ...data.RoleName) err
 	}
 	logrus.Debugf("Adding target \"%s\" with sha256 \"%x\" and size %d bytes.\n", target.Name, target.Hashes["sha256"], target.Length)
 
-	meta := data.FileMeta{Length: target.Length, Hashes: target.Hashes}
+	var customData *canonicaljson.RawMessage
+	if target.Custom != nil {
+		customData = new(canonicaljson.RawMessage)
+		if err := customData.UnmarshalJSON(target.Custom); err != nil {
```
cyli: Is this just to validating that `Target.Custom` is actually valid `json` data and not a cast `string`?

ashfall: Awesome catch, thanks! In my first iteration, `Target.Custom` was a `[]byte`, and this line was meant to do the correct type conversion. We shouldn't need this anymore. Updating!

ashfall: Pressed enter too soon. This was because I was getting `"No error is expected but got json: error calling MarshalJSON for type *json.RawMessage: unexpected end of JSON input"` in the cases where there was no custom data in the target, in the following line where we do `metaJSON, err := json.Marshal(meta)`.

client/client_test.go
```diff
@@ -1238,6 +1248,48 @@ func testListTarget(t *testing.T, rootType string) {
 	require.True(t, reflect.DeepEqual(*currentTarget, newCurrentTarget.Target), "current target does not match")
 }
 
+func testListTargetWithCustom(t *testing.T, rootType string) {
```
cyli: Non-blocking:  I wonder if it'd make sense to just fold this into `testListTarget` by adding custom metadata to one of the targets (either `latest` or `custom`)?

ashfall: Sure, that's a good idea. Updating, thanks!

cmd/notary/tuf.go
```diff
@@ -249,6 +253,21 @@ func getTargetHashes(t *tufCommander) (data.Hashes, error) {
 	return targetHash, nil
 }
 
+// Open and read a file containing the targetCustom data
+func getTargetCustom(targetCustomFilename string) (json.RawMessage, error) {
+	var nilCustom json.RawMessage
+	targetCustomFile, err := os.Open(targetCustomFilename)
```
cyli: Non-blocking:  `ioutil.ReadFile` might save some typing here, because it will open and close the file for you.

client/client.go
```diff
@@ -128,9 +129,10 @@ func (r *NotaryRepository) GetGUN() data.GUN {
 // Target represents a simplified version of the data TUF operates on, so external
 // applications don't have to depend on TUF data types.
 type Target struct {
-	Name   string      // the name of the target
-	Hashes data.Hashes // the hash of the target
-	Length int64       // the size in bytes of the target
+	Name   string                   // the name of the target
+	Hashes data.Hashes              // the hash of the target
+	Length int64                    // the size in bytes of the target
+	Custom canonicaljson.RawMessage // the custom data provided to describe the file at TARGETPATH
```
cyli: Non-blocking: would it be easier to make this a pointer, so that we can just assign to and from `FileMeta` without having check for `nil`?

cmd/notary/tuf.go
```diff
@@ -155,6 +157,7 @@ func (t *tufCommander) AddToCommand(cmd *cobra.Command) {
 	cmdTUFAdd := cmdTUFAddTemplate.ToCommand(t.tufAdd)
 	cmdTUFAdd.Flags().StringSliceVarP(&t.roles, "roles", "r", nil, "Delegation roles to add this target to")
 	cmdTUFAdd.Flags().BoolVarP(&t.autoPublish, "publish", "p", false, htAutoPublish)
+	cmdTUFAdd.Flags().StringVar(&t.custom, "custom", "", "Name of the file containing custom data for this target")
```
cyli: Non-blocking nit:  Maybe "path to the file..." instead of "name of the file"?

cmd/notary/tuf.go
```diff
@@ -249,6 +253,20 @@ func getTargetHashes(t *tufCommander) (data.Hashes, error) {
...
 }
 
+// Open and read a file containing the targetCustom data
+func getTargetCustom(targetCustomFilename string) (*json.RawMessage, error) {
+	var targetCustom json.RawMessage
+	rawTargetCustom, err := ioutil.ReadFile(targetCustomFilename)
+	if err != nil {
+		return nil, err
+	}
+	err = json.Unmarshal(rawTargetCustom, &targetCustom)
```
endophage: We should be doing `targetCustom.Unmarshal(rawTargetCustom)`. That will also require targetCustom to have been initialized, so line 258 needs to be `targetCustom := new(json.RawMessage)` then you'll also be able to remove the `&` when you reference it because `new` will already return a reference.

cmd/notary/tuf.go
```diff
@@ -21,6 +21,7 @@ import (
 	"github.com/docker/distribution/registry/client/auth/challenge"
 	"github.com/docker/distribution/registry/client/transport"
 	"github.com/docker/go-connections/tlsconfig"
+	"github.com/docker/go/canonical/json"
```
endophage: let's import this aliased to `canonicaljson` for clarity. It reduces confusion when/if somebody is having type errors related to the `encoding/json` vs `.../canonical/json` libraries

Issue #1118, opened 2017-03-20
Notary key rotate should take key ID/file 


endophage:
Currently `notary key rotate` always generates keys for you. For local keys, we should allow a key ID (of a key already existing in the key storage), or a key file (in the same format as `import` accepts) to be provided.

There can (should?) be separate flags for an ID vs a file. It might also be useful and to some degree future proofing to allow multiple IDs/files (in any combination) to be provided. For now all the keys would be added as signing key with a threshold of 1. In the future, `key rotate` could also take a threshold flag.

cyc115:
Hi @endophage good idea. we are thinking about working on this feature. Currently `notary init` has a flag `--rootkey path/to/key.key` implemented allowing a root key to specified during initialization. A similar flag for key rotate could be desirable. Additionally a `--rootcert` flag could also be useful in specifying the path to the certificate corresponding to the key.

endophage:
Agreed on all points!

PR #1147, opened 2017-05-04
adding `--rootcert` flag to key rotation

cyc115:
Adding a  `rootcert` flag to `key rotation`, making rotating to a give key pair possible for the `root` role. This feature is added to make certificate trust pinning more useful, because then we can rotate a repository's root key to pinned trust.

an example usage would be : 
`notary  key rotate [GUN] root --key path/to/key.key --rootcert path/to/rootcert.pem`

related issues: #1144, #1118, #731


endophage:
Jenkins test this please

cyli:
jenkins, ok to test

cyc115:
@cyli not sure who to ask, as I am working on this more, it seems that rotating root role requires extra confirmation on the command line which make it difficult to write integration tests. Would it be desirable to add a `terminal.IsTerminal(os.Stdin.Fd())` check to by pass this confirmation when testing? There seems to be one or 2 tests that test or depend on this behavior.

cyli:
cc @cyc115 @endophage How would you feel about requiring an explicit `-y` or some such for the rotation?  If we check `termina.IsTerminal(os.Stdin.Fd())` that would mean that scripting that command would probably also not require any confirmation, and it seems like a maybe dangerous thing to accidentally call by accident?

endophage:
I think the `-y` is probably safer than an implicit check 👍

cyc115:
@cyli Good idea :)

cyc115:
@cyli, circleCI seems to be failing with but passing on my laptop, do you know if there is anything that I might be missing? 
Thanks!
```
FAIL: TestEncryptedKeyImportSuccess (0.00s)
	Error Trace:	keys_test.go:534
	Error:		Not equal: &errors.errorString{s:"could not parse DER encoded private key: x509: failed to parse EC private key: asn1: structure error: tags don't match (16 vs {class:1 tag:9 length:12 isCompound:false}) {optional:false explicit:false application:false defaultValue:<nil> tag:<nil> stringType:0 timeType:0 set:false omitEmpty:false} ecPrivateKey @2"} (expected)
			        != &errors.errorString{s:"could not decrypt private key"} (actual)
```

cyli:
@cyc115 Yikes, I've not seen that before.  It passes for me locally on my machine too.  I'll look into it some more.

cyli:
And it strangely seems to have passed now?

cyc115:
:) machine gaining consciousness and deciding to use a different error message?

cyc115:
@cyli Hey there Ying! The yubikey test seems to be stuck again :P, can you give me a hand in fixing it?

client/client.go
```diff
@@ -1115,91 +1113,107 @@ func (r *NotaryRepository) bootstrapClient(checkInitialized bool) (*TUFClient, e
...
 			pubKeyList = append(pubKeyList, pubKey)
 		}
+	case len(pubKeys) != 0 && len(privKeys) != 0:
+		if len(pubKeys) != len(privKeys) {
+			return nil, fmt.Errorf("number of public keys does not match the number of private keys")
+		}
+		// validate key pairs
+		for i := 0; i < len(pubKeys); i++ {
+			id, err := utils.CanonicalKeyID(pubKeys[i])
+			if privKeys[i] != id {
```
endophage: Would be good if we can do this without requiring matching orders. Doesn't need to be a blocker but it's one more annoyance for a user that's relatively easily solvable in code.

client/client.go
```diff
@@ -1115,91 +1113,107 @@ func (r *NotaryRepository) bootstrapClient(checkInitialized bool) (*TUFClient, e
...
-			pubKey = r.CryptoService.GetKey(keyID)
-			if pubKey == nil {
+	case len(pubKeys) == 0 && len(privKeys) != 0:
+		for _, keyID := range privKeys {
+			if pubKey = r.CryptoService.GetKey(keyID); pubKey == nil {
 				return nil, fmt.Errorf("unable to find key: %s", keyID)
 			}
 			pubKeyList = append(pubKeyList, pubKey)
 		}
+	case len(pubKeys) != 0 && len(privKeys) != 0:
```
endophage: We should probably allow there to be more public keys than private keys. I am only expected to have access to my own private key, but I may want to include alternate signing keys for other members of the org. I might for example, use an ephemeral key to initialize the repo issued off some CA, while also wanting to include a super secure root key that is used for any future operations in the case of a compromise.

endophage: Also, if we assume len(pub) != len(priv) this case and the one above collapse in to a single thing.

endophage:
Other than my two comments, and for now I'm OK with those changes not making it in to this PR (let me know and I'll make sure we have issues for them to track), I'm LGTM on this. Will kick the yubikey tests momentarily

cyli:
Sorry, yubikey tests seem to be super unreliable every time I've attempted to kick them.  :|  Should we just remove it as a requirement?  cc @endophage @riyazdf

endophage:
@cyli they have become so flaky I think that it probably makes sense at this point, especially if we intend to move the yubikey code in to its own Go plugin. Would like to get @ecordell's opinion on removing the CI test too (we haven't confirmed the maintainers rules yet but I did propose a 2/3 majority in there for removing a PR check).

client/client.go
cyc115: Thanks David that makes sense. I will make an update to allow more public keys than private keys and also loose the restriction on ordering keys. I have kept the 4 cases distinct because it might be easier to understand?? Let me know what you think!

client/client.go
```diff
@@ -1115,91 +1113,120 @@ func (r *NotaryRepository) bootstrapClient(checkInitialized bool) (*TUFClient, e
...
+			return fmt.Errorf("invalid public key: %v", err)
+		}
+		pubKeyIDs[kid] = true
 	}
 
-	return pubKeyList, nil
+	// check if each private key has corresponding public key
+	for _, id := range privKIDs {
+		if exist := pubKeyIDs[id]; !exist {
+			return fmt.Errorf("require a public key matching private key ID: %v", id)
```
riyazdf: this case seems to have a different behavior from `case len(pubKeys) == 0 && len(privKeys) != 0` where we try to get the pubkey from the private key.  Should we be doing the same here?  cc @endophage

client/client.go
```diff
@@ -1115,91 +1113,120 @@ func (r *NotaryRepository) bootstrapClient(checkInitialized bool) (*TUFClient, e
...
+			return nil, fmt.Errorf("error converting public key with id %v to private key: %v", kid, err)
 		}
 		if loadedRole != role {
 			return nil, fmt.Errorf("attempted to load root key but given %s key instead", loadedRole)
 		}
-		pubKey, err = rootCertKey(r.gun, privKey)
-		if err != nil {
-			return nil, err
+		// do nothing in case pubKey is already a x509 pubKey
+		if pubKey.Algorithm() != data.RSAx509Key && pubKey.Algorithm() != data.ECDSAx509Key {
```
riyazdf: nit: can this check be a separate helper function ex:  `isx509Key`?  We have similar checking logic in `CanonicalKeyID` that could benefit from this too.  This can be a followup PR, though.

cmd/notary/keys.go
```diff
@@ -130,6 +132,8 @@ func (k *keyCommander) GetCommand() *cobra.Command {
 		nil,
 		"New key(s) to rotate to. If not specified, one will be generated.",
 	)
+	cmdRotateKey.Flags().StringSliceVar(&k.rootCert, "rootcert", nil, "Public key used in key rotation")
```
riyazdf: reword: `"Public key(s) to rotate to, must correspond to private key(s).`

cmd/notary/keys.go
```diff
@@ -130,6 +132,8 @@ func (k *keyCommander) GetCommand() *cobra.Command {
...
+	cmdRotateKey.Flags().BoolVarP(&k.autoConfirm, "yes", "y", false, "rotating root role does not require confirmation")
```
riyazdf: reword: `Auto-confirm rotating root role`

cmd/notary/tuf.go
```diff
@@ -428,19 +428,10 @@ func importRootKey(cmd *cobra.Command, rootKey string, nRepo *notaryclient.Notar
...
-		}
-		// add root key to repo
-		err = nRepo.CryptoService.AddKey(data.CanonicalRootRole, "", privKey)
-		if err != nil {
-			return nil, fmt.Errorf("Error importing key: %v", err)
-		}
-		rootKeyList = []string{privKey.ID()}
-	} else {
-		rootKeyList = nRepo.CryptoService.ListKeys(data.CanonicalRootRole)
+		cmd.Printf("import root key: ")
```
riyazdf: I think this should be removed

client/client.go
riyazdf: for this to match we'd have to change this function to return a new pubKeyList in addition to the error, which I think makes sense

client/helpers.go
```diff
@@ -237,12 +237,12 @@ func getRemoteKey(role data.RoleName, remote store.RemoteStore) (data.PublicKey,
 func rotateRemoteKey(role data.RoleName, remote store.RemoteStore) (data.PublicKey, error) {
 	rawPubKey, err := remote.RotateKey(role)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("unable to rotate remote key: %s", err)
```
riyazdf: I don't think we need these because of the `errFmtMsg` in `client/client.go` already prepends this string to the error before passing on

client/helpers.go
```diff
@@ -237,12 +237,12 @@ func getRemoteKey(role data.RoleName, remote store.RemoteStore) (data.PublicKey,
...
 	}
 
 	pubKey, err := data.UnmarshalPublicKey(rawPubKey)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("unable to rotate remote key: %s", err)
```
riyazdf: same here re: `errFmtMessage`

cmd/notary/tuf.go
riyazdf: Or this should be changed to `cmd.Printf(Using root key from command input: %s\n, rootKeyID)`, and then the test should be updated accordingly

cmd/notary/integration_test.go
```diff
@@ -256,7 +256,7 @@ i85wnaTwOgWv8n6q3tavmnIA/v2QqsTpmI+bhwrPNKQ=
 		"--rootkey", privKeyFilename,
 		"--rootcert", certFilename)
 	require.NoError(t, err)
-	require.Contains(t, output, "Root key found")
+	require.Contains(t, output, "import root key")
```
riyazdf: this should change with the feedback below

client/client.go
cyc115: @riyazdf @endophage thanks for the feedback 😄 . The logic can definitely be made similar to the case `len(pubKeys) == 0 && len(privKeys) != 0`.

`privateKeyIDsMatchPublicKeys` confirms that all the private keys are provided with one matching public key. other unmatched public keys are assumed to be the alternate signing keys for other members of the organization (proposed by @endophage). 

I'm not sure if there are cases when we need more private keys than public keys. If there are then I can change it to accommodate that too.

client/client.go
cyc115: 👍

cmd/notary/keys.go
cyc115: maybe remove "must"? For cases when we provide more public than private keys.

riyazdf: @cyc115: I think it's fine as-is, since we can only have `# public keys <= # private keys` since we cannot generate private keys corresponding to an existing public key?  See David's [previous comment](https://github.com/docker/notary/pull/1147/files/1974573c7499e00a04f20ef20269c1a641dfb16e#r122272437) - please let me know if I'm getting it backwards!

client/helpers.go
cyc115: I've changed `%s` to `%v` but left the message intact because I had to reduce the cyclomatic complexity of `pubKeyListForRotation` by removing the error handling logic. 

https://github.com/docker/notary/pull/1147/commits/61f6fbe9b7a73842316a6c34164147a211f31d99#diff-1cd41a2611da49ce63630d5a51bd82c5L1162

client/client.go
cyc115: @riyazdf on L1193 `pubKeyFromPrivKIDs()` I can't think of a way to simplify the logic further because we need to accomodate for the case when there are more public than private keys. :) any ideas?

client/client.go
```diff
@@ -1115,91 +1113,121 @@ func (r *NotaryRepository) bootstrapClient(checkInitialized bool) (*TUFClient, e
...
+		if err != nil {
+			return nil, fmt.Errorf("invalid public key: %v", err)
+		}
+		pubKeyIDs[kid] = true
 	}
-
-	return pubKeyList, nil
+	for _, id := range privKIDs {
+		if exist := pubKeyIDs[id]; !exist {
+			return nil, fmt.Errorf("private key %v require matching public key", id)
```
riyazdf: nit typo: `s/require/requires`

cmd/notary/tuf.go
```diff
@@ -428,19 +428,10 @@ func importRootKey(cmd *cobra.Command, rootKey string, nRepo *notaryclient.Notar
...
-		}
-		// add root key to repo
-		err = nRepo.CryptoService.AddKey(data.CanonicalRootRole, "", privKey)
-		if err != nil {
-			return nil, fmt.Errorf("Error importing key: %v", err)
-		}
-		rootKeyList = []string{privKey.ID()}
-	} else {
-		rootKeyList = nRepo.CryptoService.ListKeys(data.CanonicalRootRole)
+		cmd.Printf("using root key from command input: %s\n", rootKey)
```
riyazdf: nit: let's capitalize this

cmd/notary/keys.go
```diff
@@ -343,10 +338,82 @@ func (k *keyCommander) keysRotate(cmd *cobra.Command, args []string) error {
...
+	keyList, certList, err := keysFromArgs(cmd, k, nRepo, gun, rotateKeyRole)
+	if err != nil {
+		return err
+	}
+	// rotate keys
+	if err := nRepo.RotateKeyWithCert(rotateKeyRole, k.rotateKeyServerManaged, keyList, certList); err != nil {
 		return err
 	}
-	cmd.Printf("Successfully rotated %s key for repository %s\n", rotateKeyRole, gun)
+	cmd.Printf("successfully rotated %s key for repository %s\n", rotateKeyRole, gun)
```
riyazdf: nit: let's capitalize this

client/client.go
cyc115: Thanks for the countless reviews and tips @riyazdf 😄 . I really appreciate it! 

I have corrected the typos and squashed everything into one commit.

cyc115:
@cyli the CI yubikey seems to be failing again :) Can you help me to check up on the error when you have time ？

riyazdf:
jenkins, test this please.

riyazdf:
@cyc115: I'm seeing the following error, which is odd because circle passes
```
23:09:14 Results:
23:09:14 
23:09:24 # github.com/docker/notary/cmd/notary
23:09:24 cmd/notary/integration_test.go:1673: undefined: "github.com/docker/notary/tuf/utils".EncryptPrivateKey
23:09:24 FAIL	github.com/docker/notary/cmd/notary [build failed]
23:09:24 
```

cyc115:
Hmm the error does indeed look odd. why is `"github.com/docker/notary/tuf/utils".EncryptPrivateKey` in quotes? There are multiple instances where EncryptPrivateKey is called but the error doesn't seem to occur in those places. I've also run `make test` and saw no error there either.

riyazdf:
@cyc115: you'll need to run `make yubikey-tests` to run the yubikey suite.  I'm able to reproduce the error.

It appears that we removed that helper function  – could you please rebase against the latest master?

cyc115:
Ahh thanks @riyazdf, I have made the fix and now it should be fine. Now there's only the error about storage not found (I assume it's ok to leave it like that?). The yubikey HW test is still failing so I wonder if there's still some more failures that I have not found?

riyazdf:
@cyc115: I re-ran the job and got the same error:
```
16:28:19 --- FAIL: TestRootKeyRotationWithCert (2.39s)
16:28:19 	<autogenerated>:12: 
                        
	Error Trace:	integration_test.go:1433
16:28:19 		
			integration_test.go:1757
16:28:19 		
	Error:		Not equal: true (expected)
16:28:19 			        != false (actual)
16:28:19 		
```

I haven't been able to repro but my guess is that you would need a physical yubikey to do so

cyc115:
Ah thanks! :) I think it's fixed now! I've noticed that the first time we do `notary init` with nothing in the key store (having a yubikey plugged in) we will write the new root key to both disk as well as yubikey. But when we do `notary key import --key key.key` only yubikey is updated with this new key. Is this an intended feature? If so should the behavior of `notary init` be the same?

endophage:
@cyc115 that's the intended behavior. When running `init`, the point of writing the key to disk is so you can back it up, as it can't be extracted from the yubikey later. However when doing an `import` you clearly have the key on disk already, so there's no need to write it to disk again.

cyc115:
@endophage Thanks for answering! I thought I would ask just to make sure :)

client/client.go
```diff
@@ -1122,91 +1120,121 @@ func (r *NotaryRepository) bootstrapClient(checkInitialized bool) (*TUFClient, e
...
-	// If no new keys are passed in, we generate one
-	if len(newKeys) == 0 {
-		pubKeyList = make(data.KeyList, 0, 1)
-		pubKey, err = r.CryptoService.Create(role, r.gun, data.ECDSAKey)
 		pubKeyList = append(pubKeyList, pubKey)
-	}
-	if err != nil {
-		return nil, fmt.Errorf("unable to generate key: %s", err)
+	case len(privKeys) != 0:
+		if pubKeyList, err = r.pubKeyFromPrivKIDs(privKeys, pubKeys); err != nil {
```
endophage: This will cause the content of `pubKeys` to be ignored if there are any private keys provided. Was this PR going to allow for a user to provide more (and paired) public keys vs private keys?

riyazdf: @endophage: I don't think it's being ignored, `pubKeyFromPrivKIDs` only returns pubkeys that match private keys?

cyc115: @riyazdf  just to clarify, you are suggesting to take all provided public keys and generate public keys for the other unmatched private keys.

riyazdf: > @riyazdf just to clarify, you are suggesting the rotate command to accept all provided public keys and for the unmatched private keys, generate corresponding keys.

We should generate public keys for unmatched private keys, and also allow rotating to unmatched public keys.

It's a real (and likely common) use case to rotate multiple public keys for a role and only personally sign with a threshold-meeting subset of the private keys in that set.

cyc115:
Hey there @riyazdf. Looks like this PR has been dragging on for too long. Would you guys prefer if I start a new pr for this feature?

riyazdf:
@cyc115: if you could rebase on this PR that would be best - sorry for the inconvenience!

cyc115:
@riyazdf, sounds good :) no worries.

cyc115:
@riyazdf It took a bit more time than expected but just pushed the change :), let me know if there's any change I should make! 
I've squashed the commits to make the entire change more obvious and made some change to how the keys are verified during test in the second commit to make sure yubikey test will pass as non-yubikey tests.

cyc115:
@endophage 😄

endophage:
Hey sorry for the delay, we've been away at DockerCon. Taking a look now.

client/client.go
```diff
@@ -1255,32 +1265,68 @@ func (r *repository) pubKeyListForRotation(role data.RoleName, serverManaged boo
...
+			pubKey := r.GetCryptoService().GetKey(keyID)
 			if pubKey == nil {
 				return nil, fmt.Errorf("unable to find key: %s", keyID)
 			}
 			pubKeyList = append(pubKeyList, pubKey)
 		}
+		return pubKeyList, nil
 	}
 
 	// Convert to certs (for root keys)
```
endophage: Comment is out of date. It can be removed.

client/client.go
```diff
@@ -1255,32 +1265,68 @@ func (r *repository) pubKeyListForRotation(role data.RoleName, serverManaged boo
...
+		fmt.Println("key id : " + kid)
+		if err != nil {
+			return nil, fmt.Errorf("public key is invalid, %v", err)
+		}
+		pubKeyIDs[kid] = true
+		pubKeyList = append(pubKeyList, k)
 	}
 
+	for _, id := range privKIDs {
+		if exist := pubKeyIDs[id]; !exist {
```
endophage: `if pubKeyIDs[id] {`

`pubKeyIDs` is a `map[string]bool` so indexing will return `true` or `false`, no need for the extra assignment.

client/client.go
```diff
@@ -1255,32 +1265,68 @@ func (r *repository) pubKeyListForRotation(role data.RoleName, serverManaged boo
...
+
+	return pubKeyList, nil
+}
+
+// pubKeyListFromPrivAndPubKeys returns a list of consolidated public keys obtained by
+// merging the pubKeys with the public keys of the private keys
+func (r *repository) pubKeyListFromPrivAndPubKeys(privKIDs []string, pubKeys data.KeyList) (pubKeyList data.KeyList, err error) {
+
+	// if pubKeys are not provided, generate pubKeyList from provided private keys
+	if len(pubKeys) == 0 {
```
endophage: I think you meant to delete this entire `if` block. I believe with the remaining code in this method it's unnecessary and will still cause bad behavior in that the return within the end of this block will cause an early exit from the function returning only public keys associated with the `privKIDs` slice.

cyc115: Hi @endophage. Good pick!

endophage:
Jenkins test this please

endophage:
Rebased this against the update imports for the new repo location.

endophage:
@riyazdf can you take a final look. There have been some final updates since you last approved.

endophage:
Also, wondering if it should be `--cert` or maybe even `--public` rather than `--rootcert`? Happy to make that change in a small follow up PR. Seems like it might have broader use cases.

client/client.go
```diff
@@ -1291,16 +1323,19 @@ func (r *repository) pubKeysToCerts(role data.RoleName, pubKeyList data.KeyList)
...
-			return nil, fmt.Errorf("attempted to load root key but given %s key instead", loadedRole)
-		}
-		pubKey, err = rootCertKey(r.gun, privKey)
-		if err != nil {
-			return nil, err
+		// convert to x509 if pubKey is not already one
+		if !utils.IsX509Key(pubKey) {
+			privKey, loadedRole, err := r.GetCryptoService().GetPrivateKey(pubKey.ID())
+			if err != nil {
+				return nil, fmt.Errorf("error converting public key with id %v to private key: %v", pubKey.ID(), err)
```
riyazdf: let's change this message: `fmt.Errorf("error retrieving private key for corresponding public key with ID %s: %v", pubKey.ID(), err)`

client/client.go
```diff
@@ -1291,16 +1323,19 @@ func (r *repository) pubKeysToCerts(role data.RoleName, pubKeyList data.KeyList)
...
+			}
+			if loadedRole != role {
+				return nil, fmt.Errorf("attempted to load root key but given %s key instead", loadedRole)
```
riyazdf: let's change this message to account for the proper role: `fmt.Errorf("attempted to load %s key but given %s key instead", role, loadedRole)`

cmd/notary/keys.go
```diff
@@ -340,14 +330,78 @@ func (k *keyCommander) keysRotate(cmd *cobra.Command, args []string) error {
...
+// 2. n keys m cert
+// 3. n keys 0 cert
+// 4. n cert (stored key)
+func checkRootCert(k *keyCommander, rotateKeyRole data.RoleName) error {
+	if len(k.rootCert) != 0 {
+		//role is not "root" then error
+		if rotateKeyRole != data.CanonicalRootRole {
+			return fmt.Errorf("error: importing certificate is only supported for root role")
+		} else if len(k.rotateKeyFiles) != 0 && len(k.rootCert) < len(k.rotateKeyFiles) {
+			return fmt.Errorf("error: each key must have a matching certificate but received %v keys and %v certificates", len(k.rotateKeyFiles), len(k.rootCert))
```
riyazdf: nit: let's change both `%v` to `%d`

client/client.go
```diff
@@ -1255,32 +1265,54 @@ func (r *repository) pubKeyListForRotation(role data.RoleName, serverManaged boo
...
+	return pubKeyList, nil
+}
+
+// pubKeyListFromPrivAndPubKeys returns a list of consolidated public keys obtained by
+// merging the pubKeys with the public keys of the private keys
+func (r *repository) pubKeyListFromPrivAndPubKeys(privKIDs []string, pubKeys data.KeyList) (pubKeyList data.KeyList, err error) {
+
+	// Obtain a list of public keys corresponding to privKIDs and merge the list with pubKeys to
+	// Obtain a combined list of public keys
+	pubKeyIDs := make(map[string]bool)
```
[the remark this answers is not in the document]
cyc115: Hi @riyazdf Thank you for taking the time to help me with the reviews :). Interesting way to use empty struct. Do you know if there's any advantage over using a map of bool?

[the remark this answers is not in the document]
riyazdf: it's more consistent with idiomatic Go I've encountered in Notary and other projects, also I think there's a small memory optimization (see: https://stackoverflow.com/questions/22770114/any-difference-in-using-an-empty-interface-or-an-empty-struct-as-a-maps-value-i).

That said, this is non-blocking for me, so if you'd prefer to leave as-is that's fine :)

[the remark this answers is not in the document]
cyc115: Gotcha, thanks for the link!

cyc115:
@riyazdf :) Missed your comment before :P, I am completely ok if you want to make changes :P

client/client.go
```diff
@@ -97,6 +97,7 @@ import (
 	"regexp"
 	"time"
 
+	"github.com/cloudflare/cfssl/log"
```
riyazdf: we should remove this import and all other logs in this file should be using `logrus`

client/client.go
```diff
@@ -1293,19 +1294,19 @@ func (r *repository) pubKeyListFromPrivAndPubKeys(privKIDs []string, pubKeys dat
 
 	// Obtain a list of public keys corresponding to privKIDs and merge the list with pubKeys to
 	// Obtain a combined list of public keys
-	pubKeyIDs := make(map[string]bool)
+	pubKeyIDs := make(map[string]struct{})
 	for _, k := range pubKeys {
 		kid, err := utils.CanonicalKeyID(k)
-		fmt.Println("key id : " + kid)
+		log.Debug("obtained public key ID %s for rotation", kid)
```
riyazdf: let's change `log` to `logrus`

client/client.go
```diff
@@ -1392,7 +1393,7 @@ func DeleteTrustData(baseDir string, gun data.GUN, URL string, rt http.RoundTrip
 	if deleteRemote {
 		remote, err := getRemoteStore(URL, gun, rt)
 		if err != nil {
-			logrus.Error("unable to instantiate a remote store: %v", err)
+			log.Errorf("unable to instantiate a remote store: %v", err)
```
riyazdf: let's change `log` to `logrus`

riyazdf:
@cyc115 I went ahead and updated with the logrus fix.  cc @endophage: I think this is good to go now

PR #1150, opened 2017-05-08
Auto-parse the DB URL and add parseTime=true if the backend is mysql

### LINKED ISSUE #328: Auto-parse the DB URL and add parseTime=true if the backend is mysql
Since that is required, and we want to prevent users from shooting themselves in the foot when they provide their own URLs, since the errors that occur if it's not passed are confusing (see #322).

https://github.com/go-sql-driver/mysql#dsn-data-source-name

The config files already provide the right urls as of #323.

ashfall:
Closes #328 

Also updates the mysql driver to a newer release

Signed-off-by: Ashwini Oruganti <ashwini.oruganti@gmail.com>

### DIFF
--- a/utils/configuration.go
+++ b/utils/configuration.go
@@ -114,12 +114,12 @@ func ParseSQLStorage(configuration *viper.Viper) (*Storage, error) {
 		// Parse the url into a  config object, and grab the string from it.
 
 		// cfg, err := sql.Open("mysql", store.Source)
-		url_config, err := mysql.ParseDSN(store.Source)
+		urlConfig, err := mysql.ParseDSN(store.Source)
 		if err != nil {
 			return nil, err
 		}
 
-		url_config.ParseTime = true
+		urlConfig.ParseTime = true
 		store.Source = url_config.FormatDSN()
 	}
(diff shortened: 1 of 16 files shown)


utils/configuration.go
```diff
@@ -109,6 +110,18 @@ func ParseSQLStorage(configuration *viper.Viper) (*Storage, error) {
...
 		)
+	case store.Backend == notary.MySQLBackend:
+		// Parse the url into a  config object, and grab the string from it.
+
+		// cfg, err := sql.Open("mysql", store.Source)
+		url_config, err := mysql.ParseDSN(store.Source)
+		if err != nil {
+			return nil, err
+		}
+		if !url_config.ParseTime {
```
cyli: Non-blocking:  We can probably just set `url_config.ParseTime` to true here, since if it was true before, setting it to true again doesn't hurt anything.  :)

Would be awesome if we could add a test for this though!

endophage:
Re-iterating @cyli, we need to add a test :-)

cyli:
LGTM!  Thanks for fixing this!

Issue #861, opened 2016-07-20
Notary auth PW keychain integration or OAUTH flow


cyli:
So you don't have to put in the password for the registry you are authing to for every notary command.

PR #1155, opened 2017-05-09
Allow scripting the terminal in TUF commands

ecordell:
This change came in here: https://github.com/docker/notary/commit/25cd29c8b0df3e8ae3cc0d8427c213f760b5e76f

But I can't really tell why. It breaks scripting tuf commands (if you're not running in a terminal, all requests are anonymous to the server).

I think #861 is the proper solution, but for now it's nice to be able to script it via stdin.

### DIFF
--- a/cmd/notary/tuf.go
+++ b/cmd/notary/tuf.go
@@ -14,8 +14,6 @@ import (
 	"strings"
 	"time"
 
-	"golang.org/x/crypto/ssh/terminal"
-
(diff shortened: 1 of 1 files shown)


ecordell:
Can you use ENV vars for setting username/password and not just key passphrases?

cyli:
@ecordell Ah blah, sorry, you're right the auth info cannot be set via env passphrases.

PR #1156, opened 2017-05-09
update setuptools required by codecov

cyc115:
`docker build -t notary_client` complains that  "platform_system" is required, when installing codecov. It looks like codecov's dependency `setuptools` needs upgrading.  

```
Step 4/9 : RUN pip install codecov
 ---> Running in 4c5abc930d19
Downloading/unpacking codecov
  Downloading codecov-2.0.9-py2.py3-none-any.whl
Downloading/unpacking requests>=2.7.9 (from codecov)
Cleaning up...
Exception:
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/pip/basecommand.py", line 122, in main
    status = self.run(options, args)
  File "/usr/lib/python2.7/dist-packages/pip/commands/install.py", line 290, in run
    requirement_set.prepare_files(finder, force_root_egg_info=self.bundle, bundle=self.bundle)
  File "/usr/lib/python2.7/dist-packages/pip/req.py", line 1266, in prepare_files
    req_to_install.extras):
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 2401, in requires
    dm = self._dep_map
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 2597, in _dep_map
    self.__dep_map = self._compute_dependencies()
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 2630, in _compute_dependencies
    common = frozenset(reqs_for_extra(None))
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 2627, in reqs_for_extra
    if req.marker_fn(override={'extra':extra}):
  File "/usr/lib/python2.7/dist-packages/_markerlib/markers.py", line 113, in marker_fn
    return eval(compiled_marker, environment)
  File "<environment marker>", line 1, in <module>
NameError: name 'platform_system' is not defined
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,7 @@ RUN apt-get update && apt-get install -y \
 	&& rm -rf /var/lib/apt/lists/*
 
 RUN useradd -ms /bin/bash notary \
+	&& pip install setuptools --upgrade \
 	&& pip install codecov \
 	&& go get github.com/golang/lint/golint github.com/fzipp/gocyclo github.com/client9/misspell/cmd/misspell github.com/gordonklaus/ineffassign github.com/HewlettPackard/gas
 


cyli:
Thank you so much for fixing this @cyc115 - I'm wondering if it'd be better if we just used their bash script instead?  I know we made the switch at some point because there were some issues connectivity/instability issues originally, but it seems like it's been pretty stable for a while now and may require less dependencies than their python tool.

cyc115:
Heyy there @cyli ,thanks for your comment! I agree, it would be nice to remove pip as dependcy and move to codecov bash :). I am not familiar with codecov though so it would take a bit of time for the fix if no one get around to it. Is it possible to check in this quick fix to allow further development on notary while making another pr to remove pip as dependcy and use bash version of codecov?

cyc115:
@cyli Actually the problem no longer persist. I am going to close this PR as it is no longer needed. :) Thanks cyli!

PR #1157, opened 2017-05-11
Change RSA key size to at least 1024 bits

alyyousuf7:
OpenSSL-FIPS does not allow RSA keys to be generated if the key size is less than 1024. So the RSA key sizes in tests have been updated to prevent them from failing if Notary is made FIPS compliant.

Signed-off-by: Ali Yousuf <aly.yousuf7@gmail.com>


cyli:
This LGTM.  

Non-blocking:  we disallow key sizes < 2048 for certicates, but we may want to disallow any key size, certificate or not, < 2048.

PR #1160, opened 2017-05-13
Enable client and server cert auth in postgresql

ashfall:
`./buildscripts/integrationtest.sh postgresql` should run cleanly after this change.


ashfall:
Some thoughts:

- Should consider `verify-full` instead of `verify-ca`. (https://www.postgresql.org/docs/9.5/static/libpq-ssl.html)
- We should consider upgrading to 9.6.

TODO:
- https://github.com/docker-library/postgres/blob/master/9.5/docker-entrypoint.sh#L92 creates a `pg_hba.conf` that my changes currently overwrite. Should do a careful sed here instead of rewriting a new file.

cyli:
Could we add generating the certs to `fixtures/regenerateTestingCerts.sh`?  These expire in a year, and having an automatic script to renew the certs would be awesome.  (if we could make them expire in 2 or 3 years instead of 1, so that we have more time to notice and prevent tests from starting to inexplicably fail, that would be also great) :)

cyli:
Looks like tests are failing because we have one more place we specify the DB_URL for db-specific unit tests :|  https://github.com/docker/notary/blob/master/buildscripts/dbtests.sh#L19

ashfall:
Ah, thanks!

notarysql/postgresql-initdb.d/tls-setup.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+# Setup the server so it knows where to find certs so that server can be
+# started with TLS enabled.
+set -e
+
+sed -i "s/#ssl = off/ssl = on/" "$PGDATA"/postgresql.conf
```
riyazdf: for my own understanding: where does `$PGDATA` come from?

ashfall: It is set in the Dockerfile for the image: https://github.com/docker-library/postgres/blob/master/9.5/Dockerfile#L60

ashfall:
@riyazdf Thanks for reviewing! It appears that the DBURL per https://github.com/docker/notary/pull/1160#issuecomment-301217412 doesn't lead to a su=ccessful connection, I am trying to figure out what might be different. (Since this script/setup is specifically for our CI, I'm unable to debug much locally. Any thoughts/insights?)

riyazdf:
jenkins is back up now, I kicked off a new test run

riyazdf:
jenkins, test this please

development.postgresql.yml
```diff
@@ -40,6 +40,7 @@ services:
       - mdb 
     volumes:
       - ./notarysql/postgresql-initdb.d:/docker-entrypoint-initdb.d
+    command: -l
```
riyazdf: for my own understanding: where does this come from?  I didn't find this option on the official image description...tried searching around but couldn't find the flag elsewhere

ashfall: Per https://www.postgresql.org/docs/9.5/static/app-postgres.html, `-l` "Enables secure connections using SSL.". 

FWIW, I think `ssl=on` is implicit after passing `-l`, but since I was editing the conf for `ssl_ca_file` path, I thought it would be clearer to set `ssl=on` explicitly, in case anything changes silently in the future versions. (e.g. Not sure what happens if we use `-l` and `ssl=off` -- which one takes a higher priority is never mentioned. :( In fact, the docs that mention `ssl-on` never refer to `-l`, and vice versa.)

riyazdf: awesome that helps a lot, thank you @ashfall!

Issue #1091, opened 2017-02-01
Option for protecting keystore passphrase


rhonnava:
Currently in Nortary Signer, the signing keys are encrypted using AES wrapping with a password and the cipher text is persisted in a key store. However, the password used to AES wrap the private keys is maintained in clear text in an environment variable. 

So, even though the private keys are maintained as cipher text in the keystore, they are as secure as the environment variable. There needs to be a option in the configuration and code to callback a protection mechanism for the password itself. The interface should allow any custom protection mechanism to be plugged into. 

Please find the attached document for more details:
[Keystore_protection_proposal.docx](https://github.com/docker/notary/files/745505/Keystore_protection_proposal.docx)

endophage:
In principle I agree with adding better ways to retrieve passwords. My initial suggestion would be to not simply add callbacks, but define an interface and implement the appropriate concrete types implementing that interface for each mechanism.

```
type KeyWrapper interface {
    Encrypt(key []byte) ([]byte, error)
    Decrypt(ciphertext []byte) ([]byte, error)
}

type PasswordWrapper struct {
    password string
}

func NewPasswordWrapper(alias string) PasswordWrapper {
    return PasswordWrapper{
        password: os.Getenv(alias)
    }
}

func (pw PasswordWrapper) encrypt(key []byte) ([]byte, error) {
    // encrypt and return ciphertext
}

func (pw PasswordWrapper) decrypt(ciphertext []byte) ([]byte, error) {
    // decrypt and return plaintext key
}
```
IMO this is clearer than using a bunch of closures to maintain state. There was a reason we did the existing password retriever as a closure, but as we propose alternative mechanisms, it's becoming clearer that the closure has outlived it's loc brevity.

rhonnava:
Thanks for the feedback. I have now updated it to use interfaces. Please let me know your thought on the below code snippet.

I was thinking in terms of having a PasswordStore interface that will abstract away all the password protection details from the config code. 

The password store code will look like this:

```
package password

import "os"

type PasswordStore interface {
	getPassword() (string, error)
	setPassword(newPassword string) error
}

func NewPasswordStore(protectType string, alias string) PasswordStore {
	switch protectType {
	case "HSM":
		return HSMPasswordStore{
			alias:             alias,
			encryptedPassword: os.Getenv(alias),
			//Initialize the HSM specific variables here
		}
	case "VAULT":
		return VaultPasswordStore{
			alias:             alias,
[...]

PR #1174, opened 2017-05-29
Protecting passphrase used to wrap private keys

rhonnava:
Pull request that addresses the issue: #1091

 - Currently on the default storage and protection mechanisms are implemented.
 - Have not included configuration changes as they can be brought in when mechanism to protect at run time is implemented.
 - Once there is more clarity on using plugins details of configuration can be implemented.


rhonnava:
Will address test failures and open another one

PR #1176, opened 2017-05-31
at some point we chopped the first line off the license

endophage:
This is just bookkeeping/cleanup.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Apache License
+                                 Apache License
                            Version 2.0, January 2004
                         http://www.apache.org/licenses/
 


ashfall:
LGTM on green. Thanks for catching this!

LICENSE
```diff
@@ -1,4 +1,5 @@
 Apache License
```
riyazdf: I think this just needs to be centered, and brought down a line?  see: https://github.com/linuxkit/linuxkit/blob/master/LICENSE

endophage:
@riyazdf lol... I guess my mouse wasn't over the correct sub-window when I was trying to scroll up. I thought we'd literally chopped off the "Apache License" line. Anyway, fixed it now so it's just in there once, and centered.

PR #1177, opened 2017-06-01
Protecting passphrase used to wrap private keys

rhonnava:
Pull request that addresses the issue: #1091

 - Currently only the default storage and protection mechanisms are implemented.
 - Have not included configuration changes as they can be brought in when mechanism to protect at run time is implemented.
 - Once there is more clarity on using plugins details of configuration can be implemented.

### DIFF
--- a/cmd/notary-signer/config.go
+++ b/cmd/notary-signer/config.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"net"
 	"os"
-	"strings"
 	"time"
 
 	"google.golang.org/grpc"
@@ -16,6 +15,7 @@ import (
 	"github.com/docker/distribution/health"
 	"github.com/docker/go-connections/tlsconfig"
 	"github.com/docker/notary"
+	signerpassp "github.com/docker/notary/cmd/notary-signer/passphrase"
 	"github.com/docker/notary/cryptoservice"
 	"github.com/docker/notary/passphrase"
 	pb "github.com/docker/notary/proto"
@@ -82,20 +82,31 @@ func parseSignerConfig(configFilePath string, doBootstrap bool) (signer.Config, 
 	}, nil
 }
 
-func getEnv(env string) string {
-	v := viper.New()
-	utils.SetupViper(v, envPrefix)
-	return v.GetString(strings.ToUpper(env))
-}
-
 func passphraseRetriever(keyName, alias string, createNew bool, attempts int) (passphrase string, giveup bool, err error) {
-	passphrase = getEnv(alias)
 
-	if passphrase == "" {
-		return "", false, errors.New("expected env variable to not be empty: " + alias)
+	// Construct a default passphrase store and passphrase protector.
+	// When more than one passphrase stores or passphrase protect methods are implemented,
+	// a decision will be made here, based on a config variable, as to which type to construct.
+	var passphraseStore signerpassp.Storage = signerpassp.NewDefaultPassphraseStore()
+	var passphraseProtect signerpassp.Protector = signerpassp.NewDefaultPassphraseProtector()
+
+	// Get the passphrase from the store
+	psswdCipherText, err := passphraseStore.GetPassphrase(alias)
+
+	if err != nil || psswdCipherText == "" {
+		logrus.Error("Error retrieving passphrase from passphrase store.")
+		return "", false, errors.New("error retrieving passphrase from passphrase store")
 	}
 
-	return passphrase, false, nil
+	// Unwrap the passphrase
+	psswdClear, err := passphraseProtect.Decrypt(psswdCipherText)
+
+	if err != nil || psswdClear == "" {
+		logrus.Error("Error decrypting the passphrase.")
+		return "", false, errors.New("error decrypting the passphrase")
+	}
+
+	return psswdClear, false, nil
 }
 
 // Reads the configuration file for storage setup, and sets up the cryptoservice
--- a/cmd/notary-signer/passphrase/default_storage.go
+++ b/cmd/notary-signer/passphrase/default_storage.go
@@ -1,0 +1,35 @@
+package passphrase
+
+import (
+	"os"
+	"strings"
+)
+
+const (
+	envPrefix = "NOTARY_SIGNER_"
+)
+
+// NewDefaultPassphraseStore instantiates a default passphrase store
+func NewDefaultPassphraseStore() Storage {
+	return DefaultPassphraseStore{}
+}
+
+// DefaultPassphraseStore implements a basic passphrase store which just stores and
+// retrieves passphrase from an environment variable.
+type DefaultPassphraseStore struct {
+}
+
+// SetPassphrase stores the clear passphrase in the ENV variable as is in clear text.
+func (pw DefaultPassphraseStore) SetPassphrase(alias string, newPassphrase string) error {
+	envVariable := strings.ToUpper(envPrefix + alias)
+	error := os.Setenv(envVariable, newPassphrase)
+
+	return error
+}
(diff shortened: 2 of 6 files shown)


cyli:
jenkins, test this please

rhonnava:
Please provide details of yuikey hardware test failure and I will fix it.

riyazdf:
@rhonnava - there was a jenkins-specific issue that's now fixed, so tests are passing!  🎉   I'm taking a look at the code now.

cmd/notary-signer/passphrase/default_storage.go
```diff
@@ -0,0 +1,35 @@
...
+	envVariable := strings.ToUpper(envPrefix + alias)
+	error := os.Setenv(envVariable, newPassphrase)
+
+	return error
+}
+
+// GetPassphrase retrieves the clear passphrase from the ENV variable.
+func (pw DefaultPassphraseStore) GetPassphrase(alias string) (string, error) {
+	envVariable := strings.ToUpper(envPrefix + alias)
+	pass := os.Getenv(envVariable)
```
endophage: I'd also get once and cache in a `map[alias]passphrase` defined on the DefaultPassphraseStore

rhonnava: @endophage I suppose we will also need a sync.Mutex lock around access to this map to avoid race conditions among concurrent requests with one updating the cache (on first access for a given alias) and the other one trying to read it. Please let me know your thought.

cmd/notary-signer/config.go
```diff
@@ -82,20 +82,31 @@ func parseSignerConfig(configFilePath string, doBootstrap bool) (signer.Config,
...
-
 func passphraseRetriever(keyName, alias string, createNew bool, attempts int) (passphrase string, giveup bool, err error) {
-	passphrase = getEnv(alias)
 
-	if passphrase == "" {
-		return "", false, errors.New("expected env variable to not be empty: " + alias)
+	// Construct a default passphrase store and passphrase protector.
+	// When more than one passphrase stores or passphrase protect methods are implemented,
+	// a decision will be made here, based on a config variable, as to which type to construct.
+	var passphraseStore signerpassp.Storage = signerpassp.NewDefaultPassphraseStore()
```
rhonnava: @riyazdf yes as-is it does not deviate from the current behavior but is a step in the direction towards go plugins.

rhonnava:
Addressed review comments from @riyazdf and @endophage

cmd/notary-signer/passwordwrap/interface.go
```diff
@@ -0,0 +1,19 @@
+package passwordwrap
+
+// Storage is the interface to store/retrieve password from any specific implementation type.
+type Storage interface {
+	// Get password from storage.
+	GetPassword(alias string) (string, error)
```
endophage: given the module namespacing these can probably be just `Get` and `Set`.

cmd/notary-signer/passwordwrap/interface.go
```diff
@@ -0,0 +1,19 @@
+package passwordwrap
```
endophage: Thinking all of this should just be under the `passphrase` module. Thoughts @riyazdf @cyli @ecordell ??

cmd/notary-signer/passwordwrap/interface.go
```diff
@@ -0,0 +1,19 @@
...
+type Storage interface {
+	// Get password from storage.
+	GetPassword(alias string) (string, error)
+
+	// Set password in storage.
+	SetPassword(alias string, newPassword string) error
+}
+
+// Protector is the interface to wrap/unwrap password using any specific implementation type.
+type Protector interface {
```
endophage: Now that I'm seeing an implementation, I'm wondering if it makes sense for a `Protector` to really just be a wrapping struct that also implements the `Storage` interface. Am I missing something that would make that impossible?

i.e. (over simplified example just for demonstration, also only implemented Get side, hopefully Set can be inferred):
```
type PassStore struct {
    passwds map[string]string
}

func (ps PassStore) Get(alias) (string,error) {
    if p, ok := ps.passwds[alias]; ok {
        return p, nil
    }
    return "", fmt.Errorf("password not found for alias: %s", alias)
}

type Protector struct {
    store Storage
}

func NewProtector(s Storage) Storage {
   return &Protector{store: s}
}

func (p Protector) Get(alias) (string, error) {
    pass, err := p.store.Get(alias) 
    if err != nil {
        return "", err
    }
    return p.decrypt(pass) // where func (p Protector) decrypt(pass string) (string, error)
}
```
This will simplify the dependency injection later by wrapping up the storage and protector in to a single object that needs to be injected.

rhonnava: @endophage 
Here is what I infer from the example above:
 - There will only be one interface, the `Storage` interface.
 - Protector will implement `Encrypt/Decrypt` apart from implementing the `Storage` interface.
 - So, in future all plugins will only implement the `Storage` interface.

Question:
 - My question is WRT, the implementation of `func (p Protector) decrypt(pass string) (string, error)`. In the default implementation, there is no need to implement `Decrypt` if the passphrase continues to be stored in clear in an environment variable. In case of a plugin implementing the `Storage` interface in future, it may or may not be a secure storage on its own. However, that decision has to be made in the plugin. So, in such case, the `Encrypt/Decrypt` in the `Protector` implementation will not be needed. (Which would not make it a Protector anymore). Please let me know if I am missing something in my understanding here.

endophage: In the default case, where passwords are in plaintext, one would simply not use the `Protector` at all and only use an instance of `PassStore`. This would be part of the initialization/configuration logic that happens in the main package with the instantiated objects being injected down in to the required locations.

rhonnava: Got it, thank! I'll update the implementation as per this conversation.

rhonnava:
Addressed review comments from @endophage . I have now removed the `Protector` interface and changed the `SetPassword/GetPassword` methods in `Storage` to `Get/Set`. 

I have not implemented the logic to construct `Storage` instance based on config as we only have the default type now. When we implement more `Storage` types, I was thinking of constructing them based on configuration, in `setUpCryptoservices` function in config.go and passing down the `Storage` object to passphraseRetriever. However, if you think it is important to implement the construction and dependency injection now, I can implement construction of only the default type, by calling from setUpCryptoservices and passing it down. Please let me know your thoughts.

I have also not yet implemented changes for the following two comments as I am waiting on consensus from the other contributors that @endophage requested for:
 - Setting back to the environment.
 - Moving all of this code under passphrase package.

cmd/notary-signer/passwordwrap/default_storage.go
```diff
@@ -0,0 +1,61 @@
...
+	if ok {
+		return passwd, nil
+	}
+
+	//If not in the cache, get the password from the environment variable
+	envVariable := strings.ToUpper(envPrefix + alias)
+	passwd = os.Getenv(envVariable)
+
+	//Cache it and return out the password
+	pw.lock.Lock()
```
rhonnava: Have not addressed this comment yet, as I am waiting on consensus for whether we will have a need to set back to the environment.

cmd/notary-signer/passwordwrap/interface.go
rhonnava: Any thoughts on this @endophage  @riyazdf @cyli @ecordell ?

brutuscat:
I always wondered why on linux people dont use things like [Kernel key management](https://lwn.net/Articles/210502/)

Issue #1178, opened 2017-06-02
failure to import key should have a higher logging level than INFO


cyc115:
At the moment importing malformed keys with the notary client raises no error. using the debug flag `-D` we can see the error is being logged at INFO level.

eg. 
```
notary key import priv.key -D
> DEBU[0000] Using the following trust directory: /Users/XXX/.docker/trust
> INFO[0000] failed to import key to store: Invalid key generated, key may be encrypted and does not contain path header
```

endophage:
looks like this was closed by #1179?

riyazdf:
@endophage: 👍  yes!

PR #1179, opened 2017-06-05
Bump up default log level, key import/export logs

### LINKED ISSUE #1162: Using cached metadata should not be a silent error
The notary client being unable to update and re-using cached local metadata instead should not be a silent warning that's only visible with debug logging:

```
WARN[0002] Error while downloading remote metadata, using cached timestamp - this might not be the latest version available remotely
```

Depending on what you're doing it may be okay to use the cached metadata, but it probably shouldn't be silent.

riyazdf:
- makes `warn` the default log level on notary CLI
- increases the `-v` log level from `error` to `info`
- bumps up key import/export failure logs from `info` to `warn`

cc @cyc115 @ecordell 

Closes #1162, #1178

Ex: here's what an offline `notary list` looks like now if there's a cache with a valid timestamp:
```
🐳 $ ./notary -s https://notary.docker.io list docker.io/library/alpine
ERRO[0000] could not reach https://notary.docker.io: Get https://notary.docker.io/v2/: dial tcp: lookup notary.docker.io on [::1]:53: read udp [::1]:56941->[::1]:53: read: connection refused
WARN[0000] Error while downloading remote metadata, using cached timestamp - this might not be the latest version available remotely
NAME               DIGEST                                                              SIZE (BYTES)    ROLE
----               ------                                                              ------------    ----
2.6                9ace551613070689a12857d62c30ef0daa9a376107ec0fff0e34786cedb3399b    528             targets
2.7                9f08005dff552038f0ad2f46b8e65ff3d25641747d3912e3ea8da6785046561a    1374            targets
<..snip..>
```

</img>

### DIFF
--- a/cmd/notary/integration_test.go
+++ b/cmd/notary/integration_test.go
@@ -1725,12 +1725,12 @@ func TestLogLevelFlags(t *testing.T) {
 	// Test default to fatal
 	n := notaryCommander{}
 	n.setVerbosityLevel()
-	require.Equal(t, "fatal", logrus.GetLevel().String())
+	require.Equal(t, "warning", logrus.GetLevel().String())
 
 	// Test that verbose (-v) sets to error
 	n.verbose = true
 	n.setVerbosityLevel()
-	require.Equal(t, "error", logrus.GetLevel().String())
+	require.Equal(t, "info", logrus.GetLevel().String())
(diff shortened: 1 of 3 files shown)


cyli:
LGTM

PR #1187, opened 2017-07-08
WIP:  When reading a file or stdin to do checksumming, do not read all at once

cyli:
This fixes `data.NewFileMeta` and attempts to update the `notary verify` to not read in everything at once, in case notary is passed something very large.

However, this is currently broken because `notary verify` might need to spit whatever the input was back out, which defeats the purpose of reading in chunks (if we have to load the whole thing into memory anyway in order to dump it back out on the command line):

- If the input is a file, that's fine because we can always just read the file in chunks again to either output to disk or to stdout.
- If `--output` is provided, that works too because we can write to a temporary file that gets removed if validation fails, or moved to the right location if validation succeeds.
- If `--quiet` is passed, we are also good

The problem lies in if neither `--quiet` nor `--output` are passed, and the input is STDIN - my thought was that maybe we can buffer a limited amount in memory (maybe 100MiB), and if STDIN exceeds that then print an error saying that the input was too large to report back on output.  I didn't want to write to a temporary file, because it might be sensitive data that we don't necessarily want to duplicate all over the place.


tuf/data/types.go
```diff
@@ -210,23 +210,18 @@ func (f FileMeta) Equals(o FileMeta) bool {
...
-			}
-			cnt++
-		case notary.SHA512:
-			checksum := sha512.Sum512(payload)
-			if subtle.ConstantTimeCompare(checksum[:], v) == 0 {
-				return ErrMismatchedChecksum{alg: notary.SHA512, name: name, expected: hex.EncodeToString(v)}
+		case notary.SHA256, notary.SHA512:
+			// skipping the boolean check, because all internal usages ensure that the correct
+			// hash algorithms are produced
+			checksum, _ := s.GetChecksum(k)
```
riyazdf: this works, but I'm wondering if we should be more defensive here and skip the constant time compare if  `s.GetChecksum` returns false in the second return value?

cyli:
Sorry for not replying for a while - `io.Copy` does copy in chunks, so merging this would not be super useful if we didn't want to change how `verify`.  I mostly replaced it to avoid duplicate code, if we needed the chunked hashing in verify.

Issue #481, opened 2016-01-19
Use Canonical JSON for all TUF serialization


endophage:
Currently the final serialization of `data.Signed` objects does not use canonical serialization. We can't reserialize data sent to us by clients so the current plan is to store the raw JSON originally received (both client and server) as a new field on `Signed`  and `Signed<Role>` objects and use this original JSON to create checksums for snapshot/timestamp files.

With that in place, we can begin to do canonical serialization for `data.Signed` while operating correctly with both old and new clients pulling and pushing data.

riyazdf:
Noting that we should also investigate `json.RawMessage` (used for `Custom` in `FileMeta`), in particular how it is handled with base64 encoding/decoding when used with and without a receiver

PR #1190, opened 2017-07-13
Make testutils metadata generation consistent with how metadata is produced by the client

cyli:
In the test repo generation code, when we serialized delegation metadata, we [serialized using canonical JSON serialization](https://github.com/docker/notary/blob/master/tuf/testutils/repo.go#L165), which does not match what actually gets [done in the client](https://github.com/docker/notary/blob/master/client/helpers.go#L270), or the rest of the testutil non-delegation serialization, or for [calculating the checksum in the snapshot](https://github.com/docker/notary/blob/master/tuf/tuf.go#L867).

Fix this to make it consistent with actual behavior.

(We want to migrate to doing canonical JSON everywhere at some point - see https://github.com/docker/notary/issues/481 - but until that happens we should keep the test generation and actual generation consistent).

Note that this should allow us to bump our canonical JSON vendor once https://github.com/docker/go/pull/6 gets merged.

### DIFF
--- a/tuf/testutils/repo.go
+++ b/tuf/testutils/repo.go
@@ -162,7 +162,7 @@ func SignAndSerialize(tufRepo *tuf.Repo) (map[data.RoleName][]byte, error) {
 		if err != nil {
 			return nil, err
 		}
-		metaBytes, err := json.MarshalCanonical(signedThing)
+		metaBytes, err := json.Marshal(signedThing)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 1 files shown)


cyli:
(rebased)

endophage:
Ugh, we so need to work out a fix for this so we can get all the TUF stuff marshalled canonically. LGTM on this change. We should take another look at #481 when we get a chance

PR #1191, opened 2017-07-13
Drop support for RSA key generation

umayr:
After this change notary won't be able to generate RSA keys, although it can import and parse keys if provided externally. 

CC: @endophage


dingwilson:
Is there any timeline for when RSA support will be dropped?

endophage:
@dingwilson we have no intent to drop support for using externally provided RSA keys. We are only ending support for internally generating RSA keys.

client/client_test.go
```diff
@@ -194,7 +195,7 @@ func createRepoAndKey(t *testing.T, rootType, tempBaseDir, gun, url string) (*No
 		tempBaseDir, data.GUN(gun), url, http.DefaultTransport, rec.retriever, trustpinning.TrustPinConfig{})
 	require.NoError(t, err, "error creating repo: %s", err)
 
-	rootPubKey, err := repo.CryptoService.Create(data.CanonicalRootRole, repo.gun, rootType)
+	rootPubKey, err := testutils.CreateOrAddKey(repo.CryptoService, data.CanonicalRootRole, repo.gun, rootType)
```
riyazdf: ah of course – because the cryptoservice cannot create RSA keys?  If that is the exact reason, this is perfectly fine then.

umayr: @riyazdf Exactly. 

To add more on this, I tried to add keys to crypto service in case of RSA, while for other sort of keys I created them with crypto service which later adds them internally.

umayr:
@riyazdf Rebased and fixed the conflicts. Let me know if there's anything else.

umayr:
@endophage I have added an explicit error message that tells rsa key generation is not supported, I have also added a negative case that verifies this case moreover, I have rebased it with latest origin/master. Please let me know it requires anything else.

PR #1195, opened 2017-07-18
Fix trustpinning tests because certs have expired

cyli:
Rather than pin hard-coded certs for testing, just dynamically generate them to avoid expiry issues.  There was also possibly an issue with `TestValidateRoot` where the root data was being templated and replaced, but the signature didn't change, which could cause false-positive errors, so updated the test to re-sign after every manipulation.

This vendors CFSSL for easier cert chain generation.

cc @umayr @riyazdf

### DIFF
--- a/trustpinning/certs_test.go
+++ b/trustpinning/certs_test.go
@@ -7,7 +7,6 @@ import (
 	"crypto/rand"
 	"crypto/x509"
 	"crypto/x509/pkix"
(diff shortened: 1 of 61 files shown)


cyli:
Ok, vendor finally fixed, thanks @riyazdf.   Going to merge since the yubikey tests passed before, and currently the worker is offline.  >:(

Issue #1194, opened 2017-07-16
Potential race condition in cachedKeyService


rhonnava:
The implementation of `cachedKeyService` in signer/keydbstore/cachedcryptoservice.go uses a sync.Mutex{} to synchronize write access to the cache map. However, reading from the map in `GetPrivateKey` in this line: `cachedKeyEntry, ok := s.cachedKeys[keyID]` is not protected using a read lock. So, there can be a race condition when the map is being updated at the same time when it is being read.

I would suggest using `sync.RWMutex` instead of `sync.Mutex` and using a read lock around access:

`s.lock.RLock()`
`cachedKeyEntry, ok := s.cachedKeys[keyID]`
`s.lock.RUnlock()`

I can submit a pull request if you are okay with this.

May also be worth considering sync.map: https://github.com/golang/go/issues/18177 when Notary is on the golang version that supports it.

cyli:
@rhonnava Thanks for pointing this out - sure a PR would be great, thanks!

PR #1198, opened 2017-07-20
Fixed potential race condition in cachedKeyService

rhonnava:
Changed `lock` to be of type `sync.RWMutex` in `cachedKeyService` in  notary/signer/keydbstore/cachedcryptoservice.go and introduced a read lock around read access to the map `cachedKeys` to avoid race conditions when the map is being read and written at the same time.

This fixes the issue https://github.com/docker/notary/issues/1194.


riyazdf:
jenkins, test this please.

cyli:
LGTM, thanks for fixing this @rhonnava!

Issue #1184, opened 2017-06-28
Notary should fail when being run in a script and no credentials provided


dave-tucker:
I'd like to use `notary` in a `Makefile` to check that an specific org/image:tag is signed.
There are cases where the image may not have yet been pushed and in these cases I would expect `notary` to `exit 1` so I can push/sign the image in question.

For example, in LinuxKit, if a contributor where to add a new package called "foo", our `Makefile` would check:
```
notary -s https://notary.docker.io lookup docker.io/linuxkit/foo $(HASH)
```
but `linuxkit/foo` would not yet exist.

This appeared to hang forever, but when I ran the command manually I saw that `notary` was prompting for a `Username:` and `Password:`. My expection would be that if no credentials were provided then this command should fail.

/cc @riyazdf

riyazdf:
closed by #1200

PR #1200, opened 2017-07-24
add timeout on docker username

eiais:
This should fix scripts that never end due to a user/pass prompt.

### DIFF
--- a/cmd/notary/tuf.go
+++ b/cmd/notary/tuf.go
@@ -851,6 +851,16 @@ type passwordStore struct {
 	anonymous bool
 }
 
+func getUsername(input chan string) {
+	in := bufio.NewReader(os.Stdin)
+	result, err := in.ReadString('\n')
+	if err != nil {
+		logrus.Errorf("error processing username input: %s", err)
+		input <- ""
+	}
+	input <- result
+}
+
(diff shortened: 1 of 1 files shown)


eiais:
closes #1184

cmd/notary/tuf.go
```diff
@@ -851,23 +851,37 @@ type passwordStore struct {
...
+	go getUsername(input)
+	var username string
+	select {
+	case i := <-input:
+		username = strings.TrimSpace(i)
+		if username == "" {
+			return "",""
+		}
+	case <-time.After(30 * time.Second):
+		return "",""
```
riyazdf: can we add an error log here?  Something like `timeout when retrieving username input`?

cmd/notary/tuf.go
```diff
@@ -851,23 +851,38 @@ type passwordStore struct {
...
+	go getUsername(input)
+	var username string
+	select {
+	case i := <-input:
+		username = strings.TrimSpace(i)
+		if username == "" {
+			return "",""
+		}
+	case <-time.After(30 * time.Second):
+		logrus.Errorf("timeout when retrieving username input")
```
riyazdf: nit: this can just be `logrus.Error` (no need for formatting)

PR #1212, opened 2017-08-14
type the No valid trust data error

eiais:

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -584,7 +584,7 @@ func (r *NotaryRepository) GetTargetByName(name string, roles ...data.RoleName) 
 			return &TargetWithRole{Target: Target{Name: name, Hashes: resultMeta.Hashes, Length: resultMeta.Length, Custom: resultMeta.Custom}, Role: resultRoleName}, nil
 		}
 	}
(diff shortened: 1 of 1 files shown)


client/client.go
```diff
@@ -595,6 +595,12 @@ type TargetSignedStruct struct {
 	Signatures []data.Signature
 }
 
+type NoValidTrustDataError string
```
cyli: Do we want to name this something more like `NoSuchTarget`?  Because there is already a `ErrRepositoryNotExist` which returns text similar to "... does not have trust data for ...", and the error type name would be clearer I think.

cyli:
Can we also return the same error in `GetTargetByName`?  The error text is currently slightly different, but that function and `GetAllTargetMetadataByName` should probably return the same error type for the same failure.

client/client.go
```diff
@@ -595,6 +595,12 @@ type TargetSignedStruct struct {
 	Signatures []data.Signature
 }
 
+type ErrNoSuchTarget string
```
cyli: There seems to be a lint error because this error doesn't have a comment.  (You can verify with `make lint`).

PR #1214, opened 2017-08-24
changefeed for rethinkdb

endophage:

### DIFF
--- a/server/storage/memory.go
+++ b/server/storage/memory.go
@@ -224,7 +224,7 @@ func (st *MemStorage) GetChanges(changeID string, records int, filterName string
 	} else {
 		id, err = strconv.ParseInt(changeID, 10, 32)
 		if err != nil {
-			return nil, err
+			return nil, ErrBadQuery{msg: fmt.Sprintf("change ID expected to be integer, provided ID was: %d", changeID)}
 		}
 	}
 	var (
--- a/server/storage/memory.go
+++ b/server/storage/memory.go
@@ -85,7 +85,7 @@ func (st *MemStorage) UpdateCurrent(gun data.GUN, update MetaUpdate) error {
 // the MemStorage. Behaviour is undefined otherwise
 func (st *MemStorage) writeChange(gun data.GUN, version int, checksum string) {
 	c := Change{
-		ID:        uint(len(st.changes) + 1),
+		ID:        strconv.Itoa(len(st.changes) + 1),
 		GUN:       gun.String(),
 		Version:   version,
 		SHA256:    checksum,
@@ -200,7 +200,7 @@ func (st *MemStorage) Delete(gun data.GUN) error {
 	}
 	delete(st.checksums, gun.String())
 	c := Change{
-		ID:        uint(len(st.changes) + 1),
+		ID:        strconv.Itoa(len(st.changes) + 1),
 		GUN:       gun.String(),
 		Category:  changeCategoryDeletion,
 		CreatedAt: time.Now(),
--- a/server/storage/rethinkdb.go
+++ b/server/storage/rethinkdb.go
@@ -4,7 +4,6 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"sort"
 	"time"
@@ -14,6 +13,8 @@ import (
 	"github.com/docker/notary/tuf/data"
 	"gopkg.in/dancannon/gorethink.v3"
 )
+
+var blackoutTime = 60
 
 // RDBTUFFile is a TUF file record
 type RDBTUFFile struct {
@@ -29,7 +30,22 @@ type RDBTUFFile struct {
 
 // TableName returns the table name for the record type
 func (r RDBTUFFile) TableName() string {
-	return "tuf_files"
+	return TUFFileTableName
+}
+
+// Change defines the the fields required for an object in the changefeed
+type Change struct {
+	ID        string    `gorethink:"id,omitempty"`
+	CreatedAt time.Time `gorethink:"created_at"`
+	GUN       string    `gorethink:"gun"`
+	Version   int       `gorethink:"version"`
+	SHA256    string    `gorethink:"sha256"`
+	Category  string    `gorethink:"category"`
+}
+
+// TableName sets a specific table name for Changefeed
+func (rdb Change) TableName() string {
+	return ChangefeedTableName
 }
 
 // gorethink can't handle an UnmarshalJSON function (see https://github.com/gorethink/gorethink/issues/201),
@@ -65,6 +81,14 @@ func rdbTUFFileFromJSON(data []byte) (interface{}, error) {
 	}, nil
 }
 
+func rdbChangeFromJSON(data []byte) (interface{}, error) {
+	res := Change{}
+	if err := json.Unmarshal(data, &res); err != nil {
+		return Change{}, err
+	}
+	return res, nil
+}
+
 // RethinkDB implements a MetaStore against the Rethink Database
 type RethinkDB struct {
 	dbName   string
@@ -87,35 +111,27 @@ func NewRethinkDBStorage(dbName, user, password string, sess *gorethink.Session)
 // if it's a new role, or the version is greater than the current version
 // for the role. Otherwise an error is returned.
 func (rdb RethinkDB) UpdateCurrent(gun data.GUN, update MetaUpdate) error {
-	now := time.Now()
-	checksum := sha256.Sum256(update.Data)
-	file := RDBTUFFile{
-		Timing: rethinkdb.Timing{
-			CreatedAt: now,
-			UpdatedAt: now,
-		},
-		GunRoleVersion: []interface{}{gun, update.Role, update.Version},
-		Gun:            gun.String(),
-		Role:           update.Role.String(),
-		Version:        update.Version,
-		SHA256:         hex.EncodeToString(checksum[:]),
-		Data:           update.Data,
+	// empty string is the zero value for tsChecksum in the RDBTUFFile struct.
+	// Therefore we can just call through to updateCurrentWithTSChecksum passing
+	// "" for the tsChecksum value.
+	if err := rdb.updateCurrentWithTSChecksum(gun.String(), "", update); err != nil {
+		return err
 	}
-	_, err := gorethink.DB(rdb.dbName).Table(file.TableName()).Insert(
-		file,
-		gorethink.InsertOpts{
-			Conflict: "error", // default but explicit for clarity of intent
-		},
-	).RunWrite(rdb.sess)
-	if err != nil && gorethink.IsConflictErr(err) {
-		return ErrOldVersion{}
+	if update.Role == data.CanonicalTimestampRole {
+		tsChecksumBytes := sha256.Sum256(update.Data)
+		return rdb.writeChange(
+			gun.String(),
+			update.Version,
+			hex.EncodeToString(tsChecksumBytes[:]),
+			changeCategoryUpdate,
+		)
 	}
-	return err
+	return nil
 }
 
-// UpdateCurrentWithTSChecksum adds new metadata version for the given GUN with an associated
+// updateCurrentWithTSChecksum adds new metadata version for the given GUN with an associated
 // checksum for the timestamp it belongs to, to afford us transaction-like functionality
-func (rdb RethinkDB) UpdateCurrentWithTSChecksum(gun, tsChecksum string, update MetaUpdate) error {
+func (rdb RethinkDB) updateCurrentWithTSChecksum(gun, tsChecksum string, update MetaUpdate) error {
 	now := time.Now()
 	checksum := sha256.Sum256(update.Data)
 	file := RDBTUFFile{
@@ -162,11 +178,15 @@ func (rdb RethinkDB) UpdateMany(gun data.GUN, updates []MetaUpdate) error {
 	// find the timestamp first and save its checksum
 	// then apply the updates in alphabetic role order with the timestamp last
 	// if there are any failures, we roll back in the same alphabetic order
-	var tsChecksum string
+	var (
+		tsChecksum string
+		tsVersion  int
+	)
 	for _, up := range updates {
 		if up.Role == data.CanonicalTimestampRole {
 			tsChecksumBytes := sha256.Sum256(up.Data)
 			tsChecksum = hex.EncodeToString(tsChecksumBytes[:])
+			tsVersion = up.Version
 			break
 		}
 	}
@@ -175,7 +195,7 @@ func (rdb RethinkDB) UpdateMany(gun data.GUN, updates []MetaUpdate) error {
 	sort.Stable(updateSorter(updates))
 
 	for _, up := range updates {
-		if err := rdb.UpdateCurrentWithTSChecksum(gun.String(), tsChecksum, up); err != nil {
+		if err := rdb.updateCurrentWithTSChecksum(gun.String(), tsChecksum, up); err != nil {
 			// roll back with best-effort deletion, and then error out
 			rollbackErr := rdb.deleteByTSChecksum(tsChecksum)
 			if rollbackErr != nil {
@@ -184,6 +204,11 @@ func (rdb RethinkDB) UpdateMany(gun data.GUN, updates []MetaUpdate) error {
 			}
 			return err
 		}
+	}
+
+	// if the update included a timestamp, write a change object
+	if tsChecksum != "" {
+		return rdb.writeChange(gun.String(), tsVersion, tsChecksum, changeCategoryUpdate)
 	}
 	return nil
 }
@@ -259,7 +284,7 @@ func (rdb RethinkDB) Delete(gun data.GUN) error {
 	if err != nil {
 		return fmt.Errorf("unable to delete %s from database: %s", gun.String(), err.Error())
 	}
-	return nil
+	return rdb.writeChange(gun.String(), 0, "", changeCategoryDeletion)
 }
 
 // deleteByTSChecksum removes all metadata by a timestamp checksum, used for rolling back a "transaction"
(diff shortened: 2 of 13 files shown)


server/storage/memory.go
```diff
@@ -200,7 +200,7 @@ func (st *MemStorage) Delete(gun data.GUN) error {
 	}
 	delete(st.checksums, gun.String())
 	c := Change{
-		ID:        uint(len(st.changes) + 1),
+		ID:        strconv.Itoa(len(st.changes) + 1),
```
riyazdf: why is this change necessary?  Is this intended to be signed?  Same as above.

server/storage/rethinkdb.go
```diff
@@ -15,6 +14,8 @@ import (
 	"gopkg.in/dancannon/gorethink.v3"
 )
 
+var blackoutTime = 60
```
riyazdf: can you please add a comment for what this intended for?

server/storage/rethinkdb.go
```diff
@@ -29,7 +30,22 @@ type RDBTUFFile struct {
 
 // TableName returns the table name for the record type
 func (r RDBTUFFile) TableName() string {
-	return "tuf_files"
+	return TUFFileTableName
+}
+
+// Change defines the the fields required for an object in the changefeed
+type Change struct {
+	ID        string    `gorethink:"id,omitempty"`
```
riyazdf: I'm no database expert, but is there a reason for this ID being a `string` vs. the mysql version being a `uint`?

server/storage/rethinkdb.go
```diff
@@ -87,35 +111,27 @@ func NewRethinkDBStorage(dbName, user, password string, sess *gorethink.Session)
...
-		GunRoleVersion: []interface{}{gun, update.Role, update.Version},
-		Gun:            gun.String(),
-		Role:           update.Role.String(),
-		Version:        update.Version,
-		SHA256:         hex.EncodeToString(checksum[:]),
-		Data:           update.Data,
+	// empty string is the zero value for tsChecksum in the RDBTUFFile struct.
+	// Therefore we can just call through to updateCurrentWithTSChecksum passing
+	// "" for the tsChecksum value.
+	if err := rdb.updateCurrentWithTSChecksum(gun.String(), "", update); err != nil {
```
riyazdf: Why wouldn't we want to keep around the TS checksum for other lookup functionality?

server/storage/rethinkdb.go
```diff
@@ -294,7 +324,100 @@ func (rdb RethinkDB) CheckHealth() error {
 	return nil
 }
 
+func (rdb RethinkDB) writeChange(gun string, version int, sha256, category string) error {
+	now := time.Now()
+	ch := Change{
```
riyazdf: do we need to specify ID?

server/storage/rethinkdb.go
```diff
@@ -294,7 +324,100 @@ func (rdb RethinkDB) CheckHealth() error {
...
+	_, err := gorethink.DB(rdb.dbName).Table(ch.TableName()).Insert(
+		ch,
+		gorethink.InsertOpts{
+			Conflict: "error", // default but explicit for clarity of intent
+		},
+	).RunWrite(rdb.sess)
+	return err
+}
+
 // GetChanges is not implemented for RethinkDB
```
riyazdf: update function docstring

server/storage/rethinkdb.go
```diff
@@ -294,7 +324,100 @@ func (rdb RethinkDB) CheckHealth() error {
...
 func (rdb RethinkDB) GetChanges(changeID string, pageSize int, filterName string) ([]Change, error) {
-	return nil, errors.New("Not Implemented")
+	var (
+		lower, upper, bound []interface{}
+		idx                 = "rdb_created_at_id"
+		max                 = []interface{}{gorethink.Now().Sub(blackoutTime), gorethink.MaxVal}
```
riyazdf: is it worth introducing `blackoutTime` to rethinkdb when it doesn't seem to be implemented in mysql?

server/storage/storage_test.go
```diff
@@ -355,48 +365,43 @@ func testGetChanges(t *testing.T, s MetaStore) {
...
 
-	c, err = s.GetChanges("10", -4, "")
-	require.NoError(t, err)
-	require.Len(t, c, 4)
-	for i := 0; i < 4; i++ {
-		require.Equal(t, uint(i+5), c[i].ID)
-		require.Equal(t, "busybox", c[i].GUN)
-		require.Equal(t, i+1, c[i].Version)
-	}
+	//c, err = s.GetChanges(full[7].ID, -4, "")
```
riyazdf: either bring back this test or remove?

server/storage/rethinkdb.go
endophage: RethinkDB IDs are UUID strings

server/storage/rethinkdb.go
endophage: This just DRYs out code, I don't think it changed functionality. Did I miss something?

server/storage/memory.go
riyazdf: got it - thanks for the clarification!

server/storage/rethinkdb.go
endophage: No, autoassigned UUID from rethinkdb

server/storage/rethinkdb.go
endophage: Yes, it deals with the eventual consistency of RethinkDB. MySQL doesn't have an eventual consistency issue.

server/storage/storage_test.go
endophage: was planning to delete 👍

server/storage/rethinkdb.go
riyazdf: 👍

server/storage/rethinkdb.go
riyazdf: ah, yes you're right that no functionality has changed.  `TSchecksum` was omitted from this write

server/storage/rethinkdb.go
```diff
@@ -294,7 +324,100 @@ func (rdb RethinkDB) CheckHealth() error {
...
+		order               gorethink.OrderByOpts
+		reversed            bool
+	)
+	if filterName != "" {
+		idx = "rdb_gun_created_at_id"
+		max = append([]interface{}{filterName}, max...)
+		min = append([]interface{}{filterName}, min...)
+	}
+
+	switch changeID {
```
riyazdf: would it be helpful to refactor the MySQL `GetChanges` to factor out a helper that handles the `changeID` and `pageSize` logic?  It might be helpful to consolidate that as well, so that the functionality between the two backends is as identical as possible.

Here's the code I had in mind: https://github.com/docker/notary/blob/master/server/storage/sqldb.go#L279-L294

server/storage/rethinkdb.go
```diff
@@ -294,7 +324,100 @@ func (rdb RethinkDB) CheckHealth() error {
...
+			for i, j := 0, len(changes)-1; i < j; i, j = i+1, j-1 {
+				changes[i], changes[j] = changes[j], changes[i]
+			}
+		}
+	}()
+
+	return changes, res.All(&changes)
+}
+
+func (rdb RethinkDB) bound(changeID, filterName string) ([]interface{}, string) {
```
riyazdf: could you please add a function docstring here?

server/storage/rethinkdb.go
```diff
@@ -294,7 +324,100 @@ func (rdb RethinkDB) CheckHealth() error {
...
+		reversed            bool
+	)
+	if filterName != "" {
+		idx = "rdb_gun_created_at_id"
+		max = append([]interface{}{filterName}, max...)
+		min = append([]interface{}{filterName}, min...)
+	}
+
+	switch changeID {
+	case "0", "-1":
```
riyazdf: is there a semantic difference between `changeID == 0` and `changeID == -1` (and `changeID == -N`)?  It would be nice to collapse this check into `if changeID > 0` and an accompanying `else`

server/storage/rethinkdb.go
riyazdf: from IRL discussion: would be helpful to leave a comment about the compound min/max corresponding to the behavior of `Between` operating on an index

server/storage/rethinkdb.go
```diff
@@ -294,7 +324,100 @@ func (rdb RethinkDB) CheckHealth() error {
...
+				changes[i], changes[j] = changes[j], changes[i]
+			}
+		}
+	}()
+
+	return changes, res.All(&changes)
+}
+
+func (rdb RethinkDB) bound(changeID, filterName string) ([]interface{}, string) {
+	term := gorethink.DB(rdb.dbName).Table(Change{}.TableName()).Get(changeID).Field("created_at")
```
riyazdf: nit: rename to `createdAt` or similar?

server/storage/sql_models.go
```diff
@@ -32,9 +32,9 @@ func (g TUFFile) TableName() string {
 	return TUFFileTableName
 }
 
-// Change defines the the fields required for an object in the changefeed
-type Change struct {
-	ID        uint `gorm:"primary_key" sql:"not null" json:",string"`
+// SQLChange defines the the fields required for an object in the changefeed
+type SQLChange struct {
+	ID        uint `gorm:"primary_key" sql:";not null" json:",string"`
```
riyazdf: is the added `;` intended?

endophage: good catch. I initially tried some fiddling here and must have left this in by accident. It will still work as this is a valid tag format for sql but I'll remove it for consistency.

server/storage/rethinkdb.go
endophage: Going to leave this for a future PR due to time constraints.

server/storage/rethinkdb.go
endophage: `changeID >= 0` and `changeID == -1` both use the min/max bounds but differ in `Asc` vs `Desc` ordering.

server/storage/rethinkdb.go
```diff
@@ -15,6 +14,13 @@ import (
 	"gopkg.in/dancannon/gorethink.v3"
 )
 
+// RethinkDB has eventual consistency. This represents a 60 second blackout
+// period of the most recent changes in the changefeed which will not be
+// returned while the eventual consistency works itself out.
+// It's a var not a const so that the tests can turn it down to zero rather
+// than have to include a sleep.
+var blackoutTime = 60
```
cyli: Why a blackout instead of a quorum read, since we use `gorethink.TableOpts{ReadMode: "majority"}` for `GetCurrent` calls?

server/storage/rethinkdb.go
```diff
@@ -294,7 +329,110 @@ func (rdb RethinkDB) CheckHealth() error {
...
+		}
+	}()
+
+	return changes, res.All(&changes)
+}
+
+// bound creates the correct boundary based in the index that should be used for
+// querying the changefeed.
+func (rdb RethinkDB) bound(changeID, filterName string) ([]interface{}, string) {
+	createdAtTerm := gorethink.DB(rdb.dbName).Table(Change{}.TableName()).Get(changeID).Field("created_at")
```
cyli: What happens if the change ID doesn't exist in the table?  (and can we have a test for that case?)

server/storage/rethinkdb.go
```diff
@@ -294,7 +329,110 @@ func (rdb RethinkDB) CheckHealth() error {
...
+
+	changes := make([]Change, 0, pageSize)
+
+	// Between returns a slice of results from the rethinkdb table.
+	// The results are ordered using BetweenOpts.Index, which will
+	// default to the index of the immediately preceding OrderBy.
+	// The lower and upper are the start and end points for the slice
+	// and the Left/RightBound values determine whether the lower and
+	// upper values are included in the result per normal set semantics
+	// of "open" and "closed"
```
cyli: "open" means including the values, right?

server/storage/rethinkdb.go
endophage: This gives some additional protection in the case of a partition. Also, even with majority reads and writes, we could still end up with a little inconsistency on the most recent few records due to clock skews etc...

server/storage/rethinkdb.go
```diff
@@ -29,7 +35,22 @@ type RDBTUFFile struct {
 
 // TableName returns the table name for the record type
 func (r RDBTUFFile) TableName() string {
-	return "tuf_files"
+	return TUFFileTableName
+}
+
+// Change defines the the fields required for an object in the changefeed
+type Change struct {
```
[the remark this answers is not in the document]
endophage: We still write using a `SQLChange` for the gorm side. The SQL databases -> Go will coerce fine on read, but there were some interesting issues on the insert side where `""` in the ID was generating an error at the database level as it was being interpreted as an explicit `NULL` rather than an empty value that should be populated with an auto-increment ID.

server/storage/rethinkdb.go
endophage: "open" means excluding values.

server/storage/rethinkdb.go
endophage: Good question. Adding the test case. I think this is where the behaviours may unfortunately diverge between gorm and rethinkdb. RethinkDB will likely error due to the `bound` function looking up an explicit change ID, while MySQL will happily just use the ID as a bound.

endophage: Ok, test pushed. As far as the API spec, it probably makes sense for us to update it to specify the behavior is "undefined" (gives us a lot of wiggle room) in the cases where a changeID other than `0`, `-1`, or a known valid ID are provided.

server/storage/rethinkdb.go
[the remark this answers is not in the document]
cyli: It's interesting that even though `SQLChange` specifies a custom column name for `GUN` that it reads fine (since in our tests we check the GUN value).  For safety sake, since we seem to be relying on some underlying behavior of gorm or SQL database -> go, could we also add assertions that the `SHA256` value is what is expected (since that field also specifies a custom SQL column name?)

Also, would it make sense to make `SQLChange` unexpected, since it's an internal implementation and we are actually returning the `Change` struct?  (And add a comment stating why?)

server/storage/rethinkdb.go
endophage: I added 2 variations of the test to simulate multiple expected behaviours https://github.com/docker/notary/pull/1214/commits/51bcded0bfd67a43bc67c78d78759ebbfb234c0d#diff-523c4d973493ddd421ebe1b81b5ec81eR432

server/storage/rethinkdb.go
cyli: > This gives some additional protection in the case of a partition.

Is in the goal is to make change feeds always readable so long as even one node in an n-node cluster is up?

> we could still end up with a little inconsistency on the most recent few records due to clock skews etc

Could you explain how that would happen?

server/storage/rethinkdb.go
cyli: Thanks for the update!  Wondering if we should catch these specifically (where the ID can't be coerced to an int in SQL, or that the ID doesn't exist in rethink, and return an `InvalidChangeID` type error?  Otherwise I think in the handler this gets translated into a 500, whereas it should probably be a 400?  Otherwise we might see some spurious reports/pages/etc.

server/storage/rethinkdb.go
endophage: As I understand it, bearing in mind much of RethinkDB is a mystery to me, during a partition (I think during leader election too) we may have records being written to DBs that aren't synchronized while the partition resolves. After resolution those records end up interleaved. That means you'll get an inconsistent feed during vs after the event.

As far as clock skew, I'm using the Go server time but I don't think the rethinkdb instances are guaranteed to have perfectly synchronized clocks either. Server A inserts a record at time 12345, Server B inserts a record at 12344 according to the servers clock, but in the real world the even happens after Server A's insert.

We don't have to worry about clock skew in the MySQL version because we use the database's time and there is only 1 replica.

server/storage/rethinkdb.go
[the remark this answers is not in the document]
endophage: I can also just add the `sql` tags to the `Change` object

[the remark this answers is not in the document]
endophage: SQLChange and Change are both concrete types and the various storage types are implementing an interface we use at a higher level so type safety means we can never accidentally return a SQLChange

server/storage/rethinkdb.go
cyli: I'm also somewhat hazy on what RethinkDB does, especially since they don't use consensus for the data, just table metadata.  But going by their consistency guarantees here:  https://rethinkdb.com/docs/consistency/, we may want to specify majority reads for change feeds anyway, even with the blackout for better consistency?

According to https://aphyr.com/posts/329-jepsen-rethinkdb-2-1-5, which is a bit old but I don't think the model is outdated, only the primary replica can make changes (guaranteeing serializability of data), and the primary is selected via raft, so I'm not sure if for a single table results can be interleaved as a result of a partition (since only the primary can cause writes).

I can see what you mean about the clock skew though, although if I'm reading this correctly, we order by `CreatedAt`, which is determined by the notary server's clock, as opposed to the rethink server's clock?  Do we perform any checks on the notary server that would guarantee that the clock skew does not exceed 60 seconds?

server/storage/rethinkdb.go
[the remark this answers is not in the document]
cyli: Agree that we can't accidentally return a `SQLChange`, I was just suggesting unexporting it and the comment for readability and future maintainership purposes.  

+1 on adding the sql tags to the `Change` object too.

server/storage/rethinkdb.go
endophage: No, we're relying on NTP or some custom mechanism to keep all the notary servers reasonably in sync.

endophage:
@cyli filed an issue to track that https://github.com/docker/notary/issues/1216

PR #1215, opened 2017-08-26
Keep TUFClient private as tufClient

tiborvass:
Signed-off-by: Tibor Vass <teabee89@gmail.com>

Unless I'm mistaken, this was not meant to be exported.


HuKeping:
I supposed all the abbreviation is upper case?

tiborvass:
@HuKeping I could change it to `_TUFClient` if you prefer, but I was suggested to use `tufClient`.

HuKeping:
@tiborvass I'm OK with it

PR #1218, opened 2017-08-31
jenkinsfile: skeleton to test github integration

RobbKistler:
Signed-off-by: Robb Kistler <robb.kistler@docker.com>


RobbKistler:
@endophage Since I don't have write access to this repo, Jenkins ignores changes I've made to the Jenkinsfile for security reasons.  The failure is because the original Jenkinsfile on the merge target branch is used.

PR #1219, opened 2017-08-31
Add more details to a few doc strings

cpuguy83:
This makes the client package a little more approachable and give sa
basic example of how to get started.

ping @riyazdf

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -1,3 +1,88 @@
+/*
+Package client implements everything required for interacting with a Notary repository.
+
+Usage
+
+Use this package by creating a new repository object and calling methods on it.
+
+	package main
+
+	import (
+		"encoding/hex"
+		"fmt"
+		"net/http"
+		"os"
+		"time"
+
+		"github.com/docker/distribution/registry/client/auth"
+		"github.com/docker/distribution/registry/client/auth/challenge"
+		"github.com/docker/distribution/registry/client/transport"
+		notary "github.com/docker/notary/client"
+		"github.com/docker/notary/trustpinning"
+		"github.com/docker/notary/tuf/data"
+	)
+
+	func main() {
+		rootDir := ".trust"
+		if err := os.MkdirAll(rootDir, 0700); err != nil {
+			panic(err)
+		}
+
+		server := "https://notary.docker.io"
+		image := "docker.io/library/alpine"
+		repo, err := notary.NewFileCachedNotaryRepository(
+			rootDir,
+			data.GUN(image),
+			server,
+			makeHubTransport(server, image),
+			nil,
+			trustpinning.TrustPinConfig{},
+		)
+
+		targets, err := repo.ListTargets()
+		if err != nil {
+			panic(err)
+		}
+
+		for _, tgt := range targets {
+			fmt.Printf("%s\t%s\n", tgt.Name, hex.EncodeToString(tgt.Hashes["sha256"]))
+		}
+	}
+
+	func makeHubTransport(server, image string) http.RoundTripper {
+		base := http.DefaultTransport
+		modifiers := []transport.RequestModifier{
+			transport.NewHeaderRequestModifier(http.Header{
+				"User-Agent": []string{"my-client"},
+			}),
+		}
+
+		authTransport := transport.NewTransport(base, modifiers...)
+		pingClient := &http.Client{
+			Transport: authTransport,
+			Timeout:   5 * time.Second,
+		}
+		req, err := http.NewRequest("GET", server+"/v2/", nil)
+		if err != nil {
+			panic(err)
+		}
+
+		challengeManager := challenge.NewSimpleManager()
+		resp, err := pingClient.Do(req)
+		if err != nil {
+			panic(err)
+		}
+		defer resp.Body.Close()
+		if err := challengeManager.AddResponse(resp); err != nil {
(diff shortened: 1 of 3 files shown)


trustpinning/trustpin.go
```diff
@@ -12,9 +12,24 @@ import (
...
+// 3. TOFUS (TOFU over HTTPS)
+//
+// Only one trust pinning option will be used to validate a particular GUN.
 type TrustPinConfig struct {
-	CA          map[string]string
-	Certs       map[string][]string
+	// CA maps a GUN prefix to file paths containing the root CA.
+	// This file can contain multiple root certificates, bundled in separate PEM blocks.
+	CA map[string]string
+	// CERTS maps a GUN to a list of certificate IDs
```
riyazdf: Let's case to `Certs`

client/client.go
```diff
@@ -1,3 +1,65 @@
...
+		notary "github.com/docker/notary/client"
+		"github.com/docker/notary/cryptoservice"
+		"github.com/docker/notary/storage"
+		"github.com/docker/notary/trustmanager"
+		"github.com/docker/notary/trustpinning"
+	)
+
+	func main() {
+		rootDir := ".trust"
+		if err := os.MkdirAll(trustDir, 0700); err != nil {
```
riyazdf: s/`trustDir`/`rootDir`

client/client.go
```diff
@@ -1,3 +1,65 @@
...
+		}
+
+		remote, err := storage.NewHTTPStore(server.URL, "metadata", "json", "key", http.DefaultTransport)
+		if err != nil {
+			panic(err)
+		}
+
+		cryptoService := cryptoservice.NewCryptoService(trustmanager.NewKeyMemoryStore(demoPassRetriever))
+
+		repo, err := notary.NewNotaryRepository(
```
riyazdf: If you prefer, this can also be changed to:
```
notary.NewFileCachedNotaryRepository(
  rootDir,
  "docker.io/library/alpine",
  "https://notary.docker.io",
  roundTripper, // note this is different from remote
  demoPassRetrieverType, // note this needs to be wrapped
  trustpinning.TrustPinConfig{},
)
```

and then there's no need for the `metadata`, changelist, cryptoservice, or other `filepath.Join`.  That said, it will require some changes for the roundtripper and passretriever

client/client.go
```diff
@@ -1,3 +1,65 @@
...
+			filepath.Join(trustDir, "repo-data"),
```
riyazdf: s/`trustDir`/`rootDir`

client/client.go
```diff
@@ -1,3 +1,65 @@
...
+			metadata,
+			trustpinning.TrustPinConfig{},
+			cryptoService,
+			NewMemChangelist(),
+		)
+		if err != nil {
+			panic(err)
+		}
+
+		repo.Update(true)
```
riyazdf: it might be better to show `repo.ListTargets()`, since it returns output similar to `notary list`?

client/client.go
```diff
@@ -1,3 +1,65 @@
+/*
+Package client implements everything required for interacting with a notary repository.
```
riyazdf: nit: capitalize Notary for consistency

client/client.go
cpuguy83: This can't use the (vanilla) default http transport for the round tripper?

What does it mean that the pass retriever needs to be wrapped?

riyazdf: > This can't use the (vanilla) default http transport for the round tripper?

It likely won't because of Hub's token auth, but it's worth a try to confirm?  @tiborvass and I rigged up a challenge-response round tripper to get tags listing, as Hub will require an anonymous token for public repros.

> What does it mean that the pass retriever needs to be wrapped?

You'll need to pass a passretriever type, but actually it seems you already have satisfied the interface :+1:

cpuguy83: Ok, the example is updated and added a new function to generate the right round tripper for this to work correctly.

cpuguy83:
```
go run notary.go
2.7	9f08005dff552038f0ad2f46b8e65ff3d25641747d3912e3ea8da6785046561a
3.3	8fd4b76819e1e5baac82bd0a3d03abfe3906e034cc5ee32100d12aaaf3956dc7
2.6	9ace551613070689a12857d62c30ef0daa9a376107ec0fff0e34786cedb3399b
latest	1072e499f3f655a032e88542330cf75b02e7bdf673278f701d7ba61629ee3ebe
3.1	d9477888b78e8c6392e0be8b2e73f8c67e2894ff9d4b8e467d1488fcceec21c8
3.2	19826d59171c2eb7e90ce52bfd822993bef6a6fe3ae6bb4a49f8c1d0a01e99c7
3.5	af2a5bd2f8de8fc1ecabf1c76611cdc6a5f1ada1a2bdd7d3816e121b70300308
edge	79d50d15bd7ea48ea00cf3dd343b0e740c1afaa8e899bee475236ef338e1b53b
3.4	833ad81ace8277324f3ca8c91c02bdcf1d13988d8ecf8a3f97ecdd69d0390ce9
3.6	1072e499f3f655a032e88542330cf75b02e7bdf673278f701d7ba61629ee3ebe
integ-test-base	3952dc48dcc4136ccdde37fbef7e250346538a55a0366e3fccc683336377e372
```

Issue #1221, opened 2017-09-01
NewNotaryRepository(gun string, opts func(*repoOpts)...)


riyazdf:
As suggested by @cpuguy83 in a separate conversation: 
> What do you think about changing the notary client initializer to something like `NewNotaryRepository(gun string, opts func(*repoOpts)...)` where things like storage, server, round tripper and set like `WithTheThing(...)`, anything not set is just some default value

I was looking over both `NewNotaryRepository` and `NewFileCachedNotaryRepository` and this would seem to work nicely.  @endophage @cyli @ecordell @HuKeping WDYT?

PR #1223, opened 2017-09-01
Use functional options to configure notary repo

cpuguy83:
Introduces a new type `RepositoryOpt` which is used in
`NewNotaryRepository`.  This replaces the growing list of options to
this function to just what's required making everything else optional.
It uses default values for anything not set.

The usage is like this:

```go
repo := NewNotaryRepository(gun, url, WithMetadataStore(cache),
WithRemoteStore(remote))
```

In this case I've only set the metadata store and the remote store, so
other things like changelist use a default in-memory store.

This can be paired down further for easier examples like so:

```go
  repo := NewNotaryRepository(gun, url)
```

With this, the user should have a functioning repository client using
all default values.

In addition this extracts out the file store and http store setup used
in `NewFileCachedNotaryRepository` to an exported set of options that
any user can call. Note the function signature for is unchanged since it
is used all over.

```go
  fsOpts, _ := FileStorageOpts(baseDir, gun, nil)
  httpStoreOpt, _ := HTTPStoreOpt(url, gun, roundTripper)
  repo := NewNotaryRepository(gun, url, append(fsOpts, httpStoreOpt))
```

This approach allows adding new options in the future without breaking
the package level API, and helps keep the API cleaner.

addresses #1221

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -18,6 +18,7 @@ import (
 	"github.com/docker/notary/client/changelist"
 	"github.com/docker/notary/cryptoservice"
 	store "github.com/docker/notary/storage"
+	"github.com/docker/notary/trustmanager"
 	"github.com/docker/notary/trustpinning"
 	"github.com/docker/notary/tuf"
 	"github.com/docker/notary/tuf/data"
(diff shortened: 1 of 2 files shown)


endophage:
I'm very much against this. It's the new version of passing a map instead of named, typed parameters. It greatly obscures what the expected arguments are and increases the complexity of using a library.

cpuguy83:
>  It's the new version of passing a map instead of named, typed parameters

How does this compare to a map? These are strongly typed options.
We could go a step further and make sure only the client package can define these options.

> It greatly obscures what the expected arguments

I'm not sure how it obscures arguments. The function takes all the arguments that it requires, and follows a pattern for arguments that are optional (within the confines of having a set of usable defaults).

>  increases the complexity of using a library

Can you explain why you think this makes things more complex?

I can say I just spent the last day coming up with an example for how to use the client and trying to tease apart what is actually required by the initializer and it's not at all simple and I had to look at docker/cli for a working example and strip it down from there.


Some reading from Dave Cheney on the usage of this: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

Also note this is the same pattern used by grpc, containerd, and many other packages.

HuKeping:
Haven't take it deeply, personally I think the functional options looks quite cool and I like it , but it's not very clear as the present one.

endophage:
@cpuguy83 given your new function definition: `func NewNotaryRepository(gun, notaryURL string, opts ...RepositoryOpt) *NotaryRepository` how do I know what the `RepositoryOpt` options are? I have to go and find all the functions that return a `RepositoryOpt` type. That's an extra step of indirection that increases the complexity of using the library. 

This manifests itself clearly in that the comment we had on `NewNotaryRepository` explicitly calling out that it expects an initialized cache has been lost. I'm not going to go so far as to say we've done a good job of explaining the valid combinations of arguments, but where does such a comment belong in your suggested update? Does `NewNotaryRepository` need a comment saying "your RepositoryOpts must include at least one call to `WithMetadataStore`."

Additionally, you changed the functionality of the `NotaryRepository`. Defaulting to an in memory key store for the CryptoService is straight up dangerous. It will allow users to not set KeyStores explicitly, have everything work, and risk deploying only to find they rapidly lose the ability to update repositories as keys are lost whenever the process ends.

cpuguy83:
@endophage

> how do I know what the RepositoryOpt options are?

I'll have to check the godoc, but the options should be grouped together.
We could do this better and not pass in a `Repository` object directly and these would be easier to find. Documentation can also fix this.

> Defaulting to an in memory key store for the CryptoService is straight up dangerous

We could error out instead of setting a default crypto service to make this more explicit rather than setting the user up for failure.
Either way we are stuck with runtime checking here since you can pass a `nil` for the interface.
Likewise add docs on the in-memory store for the crypto-service so people don't use it unless they really want it.

PR #1211, opened 2017-08-14
Update vendoring and import for sirupsen/logrus

vdemeester:
Migrate to `github.com/sirupsen/logrus` 👼

- [x] What to do about rethinkdb (i.e. have a "fix branch" for now and update to a recent, rethinkdb fixed version or update it right away) ?

Takes over @simonfrequel PR #1203

Signed-off-by: Vincent Demeester <vincent@sbr.pm>

PR #1224, opened 2017-09-07
logrus: update dependency and Sirupsen -> sirupsen

tiborvass:
Signed-off-by: Tibor Vass <teabee89@gmail.com>


tiborvass:
Seems like dupe of #1211

PR #1227, opened 2017-09-07
Changefeed tests

endophage:
Adds a testchangefeed.py and a testclient.Dockerfile.

### DIFF
--- a/buildscripts/testchangefeed.py
+++ b/buildscripts/testchangefeed.py
@@ -198,9 +198,7 @@ class Tester(object):
             targets1)
 
         self.client.run(
-            ["verify", self.repo_name, "basic_repo_test", "-i", tempfile, "-q"], self.dir,
-            # skip username/password since this is an offline operation
-            username_passwd=())
+            ["verify", self.repo_name, "basic_repo_test", "-i", tempfile, "-q"], self.dir)
 
         changes = self.client.changefeed(gun=self.repo_name)
         assert changes["count"] == 1
--- a/buildscripts/testchangefeed.py
+++ b/buildscripts/testchangefeed.py
@@ -1,0 +1,448 @@
+#!/usr/bin/env python
+
+"""
+Run basic notary client tests against a server
+"""
+
+from __future__ import print_function
+
+import argparse
+from getpass import getpass
+import inspect
+import json
+import os
+import ssl
+import tempfile
+import platform
+import requests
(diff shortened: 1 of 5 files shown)


passphrase/passphrase_test.go
```diff
@@ -165,5 +165,5 @@ func TestGetPassphrase(t *testing.T) {
 	stdin := bufio.NewReader(&in)
 	passphrase, err := GetPassphrase(stdin)
 	require.NoError(t, err)
-	require.Equal(t, string(passphrase), "passphrase\n")
+	require.Equal(t, "passphrase", string(passphrase))
```
endophage: @riyazdf this was necessary due to the change in passphrase.go just above. I'm second guessing this change now though. I think it's probably just Windows being Windows, but I'm wondering if there's any reason it would break on Linux. I think this is _only_ used for logging in to notary, not any key passphrases.

endophage: Oh, and I swapped the order around because it was wrong :P `require.Equal` wants expected then actual. The test failure message was confusing me for a minute.

buildscripts/testchangefeed.py
```diff
@@ -0,0 +1,448 @@
+#!/usr/bin/env python
+
+"""
+Run basic notary client tests against a server
```
riyazdf: this should reflect `changefeed`.  Should this supersede the existing `testclient.py`?

buildscripts/testchangefeed.py
```diff
@@ -0,0 +1,448 @@
...
+
+            When running against Docker Hub, suggest usage like:
+
+                python buildscripts/testclient.py \\
+                    -s https://notary.docker.io \\
+                    -r docker.io/username/reponame \\
+                    -u username
+
+            Note that especially for Docker Hub, the repo has to have been already
+
```
riyazdf: nit whitespace

passphrase/passphrase_test.go
riyazdf: `GetPassphrase` is also used in `requestPassphrase`, which means that it's also used for key passphrases

buildscripts/testchangefeed.py
endophage: Long term yes but in the immediate future I'm not pushing on that.

endophage:
superseded by #1293

PR #1228, opened 2017-09-11
interface: add Update to Repository interface

riyazdf:
We could also take this opportunity to rename what it's exposed as in the interface, if we'd like to move away from `Update`


client/interface.go
```diff
@@ -12,6 +12,7 @@ type Repository interface {
 	Initialize(rootKeyIDs []string, serverManagedRoles ...data.RoleName) error
 	InitializeWithCertificate(rootKeyIDs []string, rootCerts []data.PublicKey, serverManagedRoles ...data.RoleName) error
 	Publish() error
+	Update(foWrite bool) error
```
endophage: `foWrite` -> `forWrite`

riyazdf: oops :)  will fix.

PR #1231, opened 2017-09-25
provide an option to Auto-confirm key rotation with root role

cyc115:
Provide an option to auto confirm key rotation with root role. 
This would make it easier to test key rotation related features in the future. 

Signed-off-by: Chen Yuechuan <Yuechuan.Chen@motorolasolutions.com>


ecordell:
Can you add a test for this?

cyc115:
@ecordell Thanks for the quick response. I have added tests but circle is not passing for some reason.

endophage:
@cyc115 it's failing on some linting
```
./server/handlers/validation.go:231:2: redundant if ...; err != nil check, just return error instead.
./storage/filestore.go:209:2: redundant if ...; err != nil check, just return error instead.```

**Comment by riyazdf** [comment]
@endophage @cyc115: there was an update to lint and I have a fix for those issues here: https://github.com/docker/notary/pull/1232

**Comment by cyc115** [comment]
Thanks for pointing it out and also providing the solution :). I've rebased this commit on the fix provided by riyazdf. @riyazdf @endophage

File: cmd/notary/integration_test.go [hunk 1]
```diff
@@ -1527,6 +1527,49 @@ func TestClientKeyGenerationRotation(t *testing.T) {
...
+	// rotate without -y command
+	_, err = runCommand(t, tempDir, "-s", server.URL, "key", "rotate", "gun", data.CanonicalRootRole.String())
+	require.NoError(t, err)
+	// no rotation
+	assertNumKeys(t, tempDir, 1, 2, true)
+
+	// rotate with auto confirm
+	_, err = runCommand(t, tempDir, "-s", server.URL, "key", "rotate", "gun", data.CanonicalRootRole.String(), "-y")
+	require.NoError(t, err)
+	// no rotation happened due to lack of confirmation
```
endophage: Is this comment correct?

endophage: @cyc115 I think this comment is the only thing holding up merging this. It reads incorrectly to me but maybe I'm misunderstanding.

PR #1234, opened 2017-09-29
[vendor] /x/crypto to master to fix terminal recognition

### LINKED ISSUE #1233: passphrase showing on power
We may have had a regression in notary CLI passphrase library between 0.4.x and 0.5.x: https://github.com/docker/cli/pull/572#pullrequestreview-66197715

cc @tophj-ibm

tophj-ibm:
Fixes #1233

/x/crypto vendored to master, which contains a fix to use /x/unix TCGETS and TCSETS instead of relying on a hardcoded amd64 value. This should fix isTerminal() not being recognized on other platforms, resulting in the password being shown.

ping @riyazdf 

This seems to have also pulled in /x/sys/windows which wasn't being used before, but I can remove it if you want.

Signed-off-by: Christopher Jones <tophj@linux.vnet.ibm.com>


tophj-ibm:
@endophage no, I don't think so. The main change I was tracking is for linux  https://github.com/docker/notary/pull/1234/files#diff-a4ed619ce231357ef8a6aeac9ecfd8f7L10 (those values are amd64)

endophage:
@tophj-ibm we've fixed the lint issues that are causing circleci to fail, if you could rebase it should make CI pass and we'll get this merged.

PR #1246, opened 2017-10-13
Introduce env. variable NOTARY_AUTH for setting username & password

stefanberger:
In case of automation it may not be possible to prompt the user
for username and password. To provide a way to pass username and
password we introduce the NOTARY_AUTH environment variable through
which username and password can be passed. Username and password
must be separated by ':' and base64 encoded.

The following example shows how to set username and password:

> export NOTARY_AUTH="$(echo -n "me:mysecretpassword" | base64)"

Decoding:

> echo $NOTARY_AUTH | base64 -d
me:mysecretpassword

Signed-off-by: Stefan Berger <stefanb@linux.vnet.ibm.com>

### DIFF
--- a/cmd/notary/tuf.go
+++ b/cmd/notary/tuf.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bufio"
 	"crypto/x509"
+	"encoding/base64"
 	"encoding/hex"
 	"encoding/pem"
 	"fmt"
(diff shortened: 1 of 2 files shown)


cmd/notary/tuf.go
```diff
@@ -785,6 +786,30 @@ func (ps passwordStore) Basic(u *url.URL) (string, string) {
...
+			logrus.Error("Authentication string with zero-legnth username")
+			return "", ""
+		} else if i > -1 {
+			username := plain[:i]
+			password := plain[i+1:]
+			password = strings.TrimSpace(password)
+			return username, password
+		}
+
+		logrus.Error("Malformatted authentication string; format must be <username>:<password>")
```
endophage: I'm debating whether it makes sense for this to be a fatal error here. This is only used in the CLI so it seems like it would be desirable to exit ASAP if we've made the decision the inputs are bad. Thoughts?

stefanberger: Would be fine by me. I followed the existing pattern where we don't exit with fatal error on timeout from keyboard input, either. Maybe that should then be changed as well in either a preceding patch or one on top of this here...

stefanberger: The test cases  would not be able to test false inputs, though. Running into failures would exit the test case...

endophage: hmmm yeah, definitely prefer having the tests work properly. We'll leave this as is 👍

PR #1259, opened 2017-11-02
update MAINTAINERS file with updated repo name, update github cdn urls

riyazdf:
Updates a few last places to the proper repo name.

For the CDN images: GH will work even with `docker/notary` in the URL but figured we should update anyway - I've tested the images and it works with the new naming.

### DIFF
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -1,6 +1,6 @@
 # Notary maintainers file
 #
-# This file describes who runs the docker/notary project and how.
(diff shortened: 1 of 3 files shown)


riyazdf:
I deliberately didn't sign off on the commit to check that DCO is up - seems like so far it isn't running, I am investigating.

riyazdf:
@caniszczyk DCO is back!  See message above.

PR #1261, opened 2017-11-07
WIP: A new command to diff two repo versions

endophage:
Using two timestamp checksums found from a changefeed (or known through some other means), this PR creates a new command:

`diff <gun> <ts checksum> <ts checksum>`

It takes all the typical arguments of the `notary` command including `-s` for the server, `--tlscacert`. It does not need a `-d` as it operates entirely in memory, outputting the a diff as a JSON blob with the following 6 fields:

```
{
    "TargetsAdded": <null of []client.TargetSignedStruct>,
    "TargetsUpdated": <null of []client.TargetSignedStruct>,
    "TargetsRemoved": <null of []client.TargetSignedStruct>,
    "RolesAdded": <null or []data.Role>,
    "RolesUpdated": <null or []data.Role>,
    "RolesRemoved": <null or []data.Role>
}
```

TODOs:
-[ ] Cleanup CLI help text, it's inconsistent with things like the command being referenced as `notarydiff` (I couldn't decide on the best name).
-[ ] Verify checksums in the findRoot function.
-[ ] Provide a `-o` option to output the diff to a file. Because of the interactivity with registry login, piping and directing stdout don't work.

cc @clemenko

### DIFF
--- a/client/diff.go
+++ b/client/diff.go
@@ -1,13 +1,15 @@
 package client
 
 import (
-	"errors"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"net/http"
+	"reflect"
 
 	"github.com/sirupsen/logrus"
 
+	"github.com/theupdateframework/notary"
 	store "github.com/theupdateframework/notary/storage"
 	"github.com/theupdateframework/notary/trustpinning"
 	"github.com/theupdateframework/notary/tuf"
@@ -16,7 +18,14 @@ import (
 )
 
 // Diff represents the difference between two versions of the same repo
-type Diff struct {}
+type Diff struct {
+	TargetsAdded   []TargetSignedStruct
+	TargetsRemoved []TargetSignedStruct
+	TargetsUpdated []TargetSignedStruct
+	RolesAdded     []data.Role
+	RolesRemoved   []data.Role
+	RolesUpdated   []data.Role
+}
 
 // NewDiff returns the different between two versions of the same TUF repo
 func NewDiff(gun data.GUN, baseURL string, rt http.RoundTripper, hash1, hash2 string) (Diff, error) {
@@ -25,7 +34,7 @@ func NewDiff(gun data.GUN, baseURL string, rt http.RoundTripper, hash1, hash2 st
 		// baseURL is syntactically invalid
 		return Diff{}, err
 	}
-	
+
 	first, err := setupClient(gun, hash1, remoteStore)
 	if err != nil {
 		return Diff{}, err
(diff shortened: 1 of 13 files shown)


client/diff.go
```diff
@@ -0,0 +1,273 @@
...
+	oldBuilder := tuf.NewRepoBuilder(gun, nil, trustpinning.TrustPinConfig{})
+	if err := oldBuilder.Load(data.CanonicalRootRole, root, 0, true); err != nil {
+		return nil, err
+	}
+	newBuilder := tuf.NewRepoBuilder(gun, nil, trustpinning.TrustPinConfig{})
+	if err := newBuilder.Load(data.CanonicalRootRole, root, 0, true); err != nil {
+		return nil, err
+	}
+
+	tClient := newTufClient(oldBuilder, newBuilder, remote, store.NewMemoryStore(nil))
```
riyazdf: I wonder if we need the two separate builders for newTufClient

client/diff.go
```diff
@@ -0,0 +1,273 @@
...
+			res.RolesUpdated = append(res.RolesUpdated, *role)
+		}
+	}
+	for _, role := range lookupTable {
+		res.RolesRemoved = append(res.RolesRemoved, *role)
+	}
+	return nil
+}
+
+func equivalentRoles(first, second *data.Role) bool {
```
riyazdf: Could we add a comment that we check for paths, thresholds, key ids, and _not_ versions because that isn't a meaningful update in the role?

client/diff.go
```diff
@@ -0,0 +1,273 @@
...
+	err := diffTargets(&res, first, second)
+	if err != nil {
+		return res, err
+	}
+	err = diffRoles(&res, first, second)
+	return res, err
+}
+
+func diffRoles(res *Diff, first, second *tuf.Repo) error {
+	firstRoles := first.GetAllLoadedRoles()
```
riyazdf: Did you have a reason for using this function instead of `ListRoles`?  This should be fine given the ordering of operations (we have an `Update` when getting targets), though I'm wondering if we should add a comment to ensure that the code doesn't get shifted around

cmd/diff/diff.go
```diff
@@ -0,0 +1,337 @@
+package main
```
riyazdf: let's move this to `cmd/notary/diff.go` for consistency, and that way we'll pick up all of the duplicated helper functions as well.

cmd/diff/main.go
```diff
@@ -0,0 +1,203 @@
+package main
```
riyazdf: I don't think this is needed if we move the CLI `diff.go` to `cmd/notary/diff.go`?

cmd/diff/repo_factory.go
```diff
@@ -0,0 +1,45 @@
+package main
```
riyazdf: same here: this will go away with the file move

client/diff.go
endophage: I think ListRoles only returned delegations, or at least appeared to when I was tired last night

riyazdf: ah - it should return all roles.  This code is fine as-is but `ListRoles` may be a safer improvement

PR #1263, opened 2017-11-14
remove warning when checking if delegation certs have expired

eiais:

### DIFF
--- a/tuf/tuf.go
+++ b/tuf/tuf.go
@@ -248,17 +248,14 @@ func (tr *Repo) GetDelegationRole(name data.RoleName) (data.DelegationRole, erro
 				}
 				// Check all public key certificates in the role for expiry
 				// Currently we do not reject expired delegation keys but warn if they might expire soon or have already
-				for keyID, pubKey := range delgRole.Keys {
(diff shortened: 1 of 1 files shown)


eiais:
CC @riyazdf @endophage

tuf/tuf.go
```diff
@@ -248,17 +248,13 @@ func (tr *Repo) GetDelegationRole(name data.RoleName) (data.DelegationRole, erro
...
 					if err != nil {
 						continue
 					}
-					if err := utils.ValidateCertificate(certFromKey, true); err != nil {
-						if _, ok := err.(data.ErrCertExpired); !ok {
-							// do not allow other invalid cert errors
-							return err
-						}
-						logrus.Warnf("error with delegation %s key ID %d: %s", delgRole.Name, keyID, err)
+					if err := utils.ValidateCertificate(certFromKey, false); err != nil {
```
riyazdf: can we add a comment about why the `false` value is ok here?

Can be something like: `do not check delegation certificate expiry once accepted, use the TUF role expiry instead`

endophage:
LGTM

PR #1264, opened 2017-11-14
changefeed: error with 400 for nonexistent change_id

riyazdf:
Check early that the `change_id` parameter exists for rethinkdb instead of returning a 500 on a query error.

Closes https://github.com/theupdateframework/notary/issues/1257


endophage:
I always figured the fix for this would be to default the change ID at the handler/input verification level, rather than erroring when not provided.

riyazdf:
@endophage: that works too.  Do you think `"0"` makes most sense, even in the context of rethinkdb?

endophage:
Yeah, and I also have a strong preference for making Go's "zero" values the defaults whenever possible.

PR #1300, opened 2018-02-14
Fix usage example for client

justincormack:
This fixes the example so it works again, and adds a missing error handler.

This really should be converted to a Go test example so it gets compile tested, but it is a bit long, ie two functions for good reason. Will think about how best to do this, but lets fix the error first.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -30,7 +30,7 @@ Use this package by creating a new repository object and calling methods on it.
 
 		server := "https://notary.docker.io"
 		image := "docker.io/library/alpine"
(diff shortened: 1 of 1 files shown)


eiais:
accidentally duplicated this in #1301

cyli:
Thanks for the example fix @justincormack - I think you're right it should be a go test example.  Since @eiais has also duplicated the fix and moved the code to a `*_test.go` file, perhaps they could also convert it to a go test example?

justincormack:
Closed in favour of #1301

PR #1301, opened 2018-02-16
Fix example client

eiais:
Also move it to a test file.
CC @endophage @cyli


cyli:
Thanks for adding this @eiais - @justincormack is correct and it might do better being a go test example.  For instance, see https://blog.golang.org/examples (larger examples), which shows up in https://golang.org/pkg/sort/ as entire examples in the doc.  If you have time, would you mind converting the main function to an example?

eiais:
@cyli Do you know how to make govet happy with the function name? ExampleClient and Example_client result in ExampleClient refers to unknown identifier: Client. and Example_Client has malformed example suffix: Client.

justincormack:
It needs to be just called `Example()` if it documents the package as a whole, wlse it needs to be named after the function it documents. See https://blog.golang.org/examples

eiais:
@justincormack Totally missed that. Thank you!

cyli:
jenkins, test this please

PR #1307, opened 2018-02-28
Bump the go-connections vendor so that we disable insecure RSA key exchange ciphers and require TLS >= 1.2

cyli:
If this is approved before https://github.com/theupdateframework/notary/pull/1306, ideally it would be included in the 0.6.0 release as well and I'll update the changelog and relevant branches.


cyli:
Thanks @ecordell!

endophage:
LGTM!

PR #1312, opened 2018-03-02
Update minimal dockerfiles to use multi-stage builds to achieve minimalness

cyli:
cc @endophage @justincormack

### DIFF
--- a/server.minimal.Dockerfile
+++ b/server.minimal.Dockerfile
@@ -1,3 +1,21 @@
+FROM golang:1.9.4-alpine AS build-env
+RUN apk add --update git gcc libc-dev
+# Pin to the specific v3.0.0 version
+RUN go get -tags 'mysql postgres file' github.com/mattes/migrate/cli && mv /go/bin/cli /go/bin/migrate
+
+ENV NOTARYPKG github.com/theupdateframework/notary
+
+# Copy the local repo to the expected go path
+COPY . /go/src/${NOTARYPKG}
+WORKDIR /go/src/${NOTARYPKG}
+
+# Build notary-server
+RUN go install \
+    -tags pkcs11 \
+    -ldflags "-w -X ${NOTARYPKG}/version.GitCommit=`git rev-parse --short HEAD` -X ${NOTARYPKG}/version.NotaryVersion=`cat NOTARY_VERSION`" \
+    ${NOTARYPKG}/cmd/notary-server
+
+
 FROM busybox:latest
 MAINTAINER David Lawrence "david.lawrence@docker.com"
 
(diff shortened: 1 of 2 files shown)


server.minimal.Dockerfile
```diff
@@ -1,19 +1,38 @@
...
+    -tags pkcs11 \
+    -ldflags "-w -X ${NOTARYPKG}/version.GitCommit=`git rev-parse --short HEAD` -X ${NOTARYPKG}/version.NotaryVersion=`cat NOTARY_VERSION`" \
+    ${NOTARYPKG}/cmd/notary-server
+
+
 FROM busybox:latest
 MAINTAINER David Lawrence "david.lawrence@docker.com"
 
-# the ln is for compatibility with the docker-compose.yml, making these
+# the line is for compatibility with the docker-compose.yml, making these
```
justincormack: that meant `ln` ie the symlink

signer.minimal.Dockerfile
```diff
@@ -1,20 +1,40 @@
...
+    -tags pkcs11 \
+    -ldflags "-w -X ${NOTARYPKG}/version.GitCommit=`git rev-parse --short HEAD` -X ${NOTARYPKG}/version.NotaryVersion=`cat NOTARY_VERSION`" \
+    ${NOTARYPKG}/cmd/notary-signer
+
+
 FROM busybox:latest
 MAINTAINER David Lawrence "david.lawrence@docker.com"
 
-# the ln is for compatibility with the docker-compose.yml, making these
+# the line is for compatibility with the docker-compose.yml, making these
```
justincormack: and here

justincormack:
looks fine bar comment about comment but needs a rebase.

cyli:
Thanks @justincormack!  Fixed.

server.minimal.Dockerfile
```diff
@@ -1,19 +1,38 @@
+FROM golang:1.9.4-alpine AS build-env
+RUN apk add --update git gcc libc-dev
```
thaJeztah: If "tiny-ness" is important; use `--no-cache` instead;

```bash
docker build -t before -<<EOF
FROM golang:1.9.4-alpine AS build-env
RUN apk add --update git gcc libc-dev
EOF

docker build -t after -<<EOF
FROM golang:1.9.4-alpine AS build-env
RUN apk add --no-cache git gcc libc-dev
EOF
```

```
REPOSITORY      CREATED                  SIZE
after           Less than a second ago   386MB
before          Less than a second ago   387MB
```

cyli: This is using a build image and a final image, though - does the no-cache affect the final image?  This is the before and after image built with the `server.minimal.Dockerfile`, without the `--no-cache` option and with the `--no-cache` option:

```
REPOSITORY                                       TAG                    IMAGE ID            CREATED              SIZE
<none>                                           <none>                 fa5c316b8562        17 seconds ago       540MB
after                                            latest                 f3f0621261c3        About a minute ago   21.1MB
before                                           latest                 f3f0621261c3        About a minute ago   21.1MB
<none>                                           <none>                 82a30458273b        About a minute ago   540MB
```

thaJeztah: No it doesn't affect the final image (but does affect your local cache 😅); it was mainly a nit, and I wanted to mention `--no-cache` because it's awesome: it takes care of everything (updating the index, and removing it afterwards)

cyli: Oh nice!  It seems to only be a MB or less, so in this case it seems ok, but I will keep that in mind for all my other images!

PR #1318, opened 2018-03-21
Don't error out in case home cannot be found

vdemeester:
There is a few case where home could not be found (and would not be
needed). Especially as this piece of code is only there to set a
default value for a flag.

cc @cyli @n4ss @endophage 

Signed-off-by: Vincent Demeester <vincent@sbr.pm>

### DIFF
--- a/cmd/notary/main.go
+++ b/cmd/notary/main.go
@@ -77,10 +77,9 @@ func (n *notaryCommander) parseConfig() (*viper.Viper, error) {
 	// Get home directory for current user
 	homeDir, err := homedir.Dir()
 	if err != nil {
-		return nil, fmt.Errorf("cannot get current user home directory: %v", err)
-	}
-	if homeDir == "" {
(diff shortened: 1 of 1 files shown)


n4ss:
LGTM!

cmd/notary/main.go
```diff
@@ -77,10 +77,7 @@ func (n *notaryCommander) parseConfig() (*viper.Viper, error) {
 	// Get home directory for current user
 	homeDir, err := homedir.Dir()
 	if err != nil {
-		return nil, fmt.Errorf("cannot get current user home directory: %v", err)
```
cyli: nit:  we should specify that the default notary directory instead will be `$(pwd)/.notary` maybe, as part of the warning, since we are still setting a default, it just may be in an unexpected place.

vdemeester: Oh I see what you mean, can definitely change that :wink:

HuKeping:
LGTM

PR #1319, opened 2018-03-21
Remove generation of AUTHORS

justincormack:
This is unused, and currently breaks the spellcheck rule. I was
going to just skip spell checking it, but it seems easier to remove
the file generation instead.


HuKeping:
Thanks @justincormack !

LGTM!

PR #1320, opened 2018-03-22
Rework the FIPS mode detection

justincormack:
The best test we currently have for whether Notary is running in a FIPS
compliant environment is whether the MD5 hash function is registered when
crypto/md5 is linked in to the program. This function is not available
in FIPS mode as it is not an allowed hash function.

### DIFF
--- a/fips.go
+++ b/fips.go
@@ -1,13 +1,14 @@
 package notary
 
-import "os"
+import (
+	"crypto"
+	// Need to import md5 so can test availability.
+	_ "crypto/md5"
+)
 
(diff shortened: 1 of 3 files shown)


HuKeping:
The const `FIPSEnvVar` in the testcode fail the CI

justincormack:
Ah ok, we still have a force override, will adjust.

justincormack:
Ok fixed up the way tests work cc @cyli

HuKeping:
Thanks @justincormack 

LGTM!

PR #1333, opened 2018-04-07
Remove go-homedir dependency

justincormack:
This is overcomplicated for a very simple requirement. It can call
`exec`, it has a cache protected by a mutex because it is so complex.
Just use the per platform environment variables which will always be
set, if the user has a weird environment they can specify full paths.


cmd/notary/util.go
```diff
@@ -52,3 +53,12 @@ func feedback(t *tufCommander, payload []byte) error {
 	os.Stdout.Write(payload)
 	return nil
 }
+
+// homeExpand will expand an initial ~ to the user home directory. This is supported for
+// config files where the shell will not have expanded paths.
+func homeExpand(homeDir, path string) string {
+	if path == "" || path[1] != '~' || (len(path) > 1 && path[1] != os.PathSeparator) {
```
cyli: Should `path[1] != '~'` be `path[0] ~= '~'`?  Since we only want to expand if the first char is `~`?  (e.g. `~/.docker`).

Also, how would we expand for something like `~cyli` or ~justincormack`?  That is not the homedir of the current user, necessarily.

justincormack: ah yes, typo, fixed now.

This has never worked with `~cyli` etc so there is no change here. If you put that on the command line your shell will expand it before `notary` sees it anyway, so it is only for use in the config file, where it would be quite odd to put files in another user's home.

justincormack:
I should add some tests to clarify this.

cyli:
> I should add some tests to clarify this.

That would be awesome, thanks.  I was wondering if we had tests to make sure `~/` had been expanded correctly, but hadn't had a chance to look them up yet.

justincormack:
Added a unit test for `homeExpand`.

PR #1315, opened 2018-03-05
Add permission to ConfigureRepo function

molepigeon:
Fixes https://github.com/theupdateframework/notary/issues/1308

Signed-off-by: Michael Hough <michaelh@uk.ibm.com>

PR #1338, opened 2018-04-09
Bump notary version to 0.6.1 and add changelog.

cyli:
This is a point release for updating some of our dependencies, including the JWT library which includes a security fix, and a bugfix for requesting too many permissions when listing a repo.  #1315.


cyli:
Adding #1333 to various internal fixes, then merging.

PR #1258, opened 2017-11-01
[WIP] re-enable GoASTScanner

eiais:
Signed-off-by: Kyle Spiers <kyle@spiers.me>

PR #1340, opened 2018-04-11
[WIP] Restore Go ast scanner

justincormack:
replace #1258 there are a bunch of changes in the cli syntax.

There are still a lot of errors, looking at what needs doing to fix them and will add commits either here or upstream and rebase.


justincormack:
Ok this is now passing with two tests disabled, so at least it will catch other issues rather than nothing.

PR #1344, opened 2018-04-15
Use golang/x/crypto for ed25519

justincormack:
This is now available in the official upstream packages, so stop
using independent implementation. Small changes in types required.


cyli:
Actually scratch that, I can check manually first.

PR #1348, opened 2018-04-24
Fix gorethink

dmp42:
Moved from docker fork (based on v3.0.0 + sirupsen cap patch), back to upstream (on v3.0.5).
- no code change in notary itself, except `dancannon` -> `gorethink` on imports
Vendor changes:
- updated from `cenk` to `cenkalti`
- removed dep on gorethink.v2 (no longer needed)

Signed-off-by: Olivier Gambier <olivier.gambier@docker.com>


dmp42:
CI failure seems unrelated (fails on master as well). @endophage do you now why it's failing?

Issue #1289, opened 2018-01-17
Enabling notary for other PKCS#11 users than Yubikey


florianeichin:
The only hardware that notary supports as trustmanager is Yubikey. We would like to extend notary to support other devices that work according to PKCS#11 -- and for our case add support for devices that are used through the opencryptoki library. Our suggestion is to split this up into two steps:
1. Create an abstraction for pkcs11 (and move Yubikey in there).
2. add opencryptoki support

So essentially this results in a file structure like the following:
   ```bash
├─trustmanager 
│     ├─pkcs11
│     │   ├─yubikey
 │     │   └─ opencryptoki
```

Other PKCS#11 exploiters can hook in at the level below PKCS#11. To keep enablement of other PKCS#11 trustmanagers simple in general, we think it would make sense, to abstract some of the yubikey files and code pieces to reuse them.
The abstraction could go that far, that just the following yubikeystore functions have to stay hardware specific, everything else could be reused by opencryptoki and other pkcs11 implementations:
- getECDSAKey(), 
- addECDSAKey(), 
- yubiRemoveKey(), 
- Sign(), 
- hardwareListKey(), 
- listObjects(), 
- SetupHSMEnv() 

An (IMHO bad) alternative is to clone the entire trustmanager, which would create a lot of redundant code.
Prior to a PR, we wanted to discuss this structure. If the above approach sounds good to everyone, a PR with code will follow.

utzb:
@endophage @riyazdf could you PTAL whether this approach is good to go for a PR?

endophage:
This is very closely aligned with what I've been wanting to do with our PKCS#11 support. As part of this it would also be great to address the fact that we only support root keys in hardware today. We should eventually support all keys in hardware.

florianeichin:
That sounds great. We would suggest to setup a PR, containing two commits. 
1.  Abstraction and refactoring of the existing yubikey #PKCS11 implementation
2. Implementation of opencryptoki as a second trustmanager besides yubikey

If you're fine with that, we would setup the PR.

nfewx:
I am also interested in improving HSM support in Notary. The approach I had in mind was a generic pkcs11 key store rather than one tied to a particular provider, described at https://gist.github.com/nfewx/839327740b1355f6fd2c9d36aa790190.

jschintag:
To further seperate the specific HSM from notary i thought about 2 possible ways:

1. Using Go-Plugin

[...]

PR #1352, opened 2018-05-08
abstracting pkcs11 package and adding another vendor besides yubikey

florianeichin:
This PR addresses #1289

First commit abstracts the yubikey package into a hardwareindependent pcks11 package and a hardwarespecific yubikey package.

Second commit adds opencryptoki as another hardware implementation.


trustmanager/pkcs11/common/hardwareprivatekey.go
```diff
@@ -0,0 +1,97 @@
...
+
+	return publicKey
+}
+
+// SetLibLoader sets up the pkcs library for further usage
+func (hpk *HardwarePrivateKey) SetLibLoader(loader Pkcs11LibLoader) {
+	hpk.libLoader = loader
+}
+
+// CryptoSigner returns a crypto.Signer tha wraps the HardwarePrivateKey. Needed for
```
justincormack: type "tha"

florianeichin: fixed that and pushed the fix

trustmanager/pkcs11/common/hardwareprivatekey.go
```diff
@@ -0,0 +1,97 @@
...
+// HardwarePrivateKey represents a private key inside of a Hardwarestore
+type HardwarePrivateKey struct {
+	data.ECDSAPublicKey
+	passRetriever notary.PassRetriever
+	slot          HardwareSlot
+	libLoader     Pkcs11LibLoader
+}
+
+// hardwareSigner wraps a HardwarePrivateKey and implements the crypto.Signer interface
+type hardwareSigner struct {
```
endophage: What's the need for the extra wrapper? `crypto.Signer` is only implemented by virtue of the embedded `HardwarePrivateKey`?

trustmanager/pkcs11/common/hardwareprivatekey.go
```diff
@@ -0,0 +1,97 @@
...
+	"fmt"
+	"io"
+
+	"github.com/theupdateframework/notary"
+	"github.com/theupdateframework/notary/tuf/data"
+	"github.com/theupdateframework/notary/tuf/signed"
+)
+
+const (
+	// SigAttempts defines maximum attempts to sign an image before aborting
```
endophage: Notary is not Docker/container specific, shouldn't reference "image"

trustmanager/pkcs11/common/hardwarestore.go
```diff
@@ -0,0 +1,195 @@
...
+		logrus.Debugf("Failed to get an empty %s slot: %s", hardwareName, err.Error())
+		return false, err
+	}
+	logrus.Debugf("Attempting to store key using %s slot %v", hardwareName, slot)
+	key := HardwareSlot{
+		Role:   role,
+		SlotID: slot,
+		KeyID:  keyID,
+	}
+	//err = hardwareKeyStore.AddECDSAKey(ctx, session, privKey, slot, s.PassRetriever, role)
```
endophage: remove commented out line.

trustmanager/pkcs11/common/universal.go
```diff
@@ -0,0 +1,184 @@
...
+// for testing
+const ecdsaPrivateKeySize = 32
+
+var (
+	hardwareName     string
+	hardwareKeyStore HardwareSpecificStore
+)
+
+// SetKeyStore sets up the to be used keystore
+func SetKeyStore(ks HardwareSpecificStore) {
```
endophage: This appears to be restricting us to only being able to access a single hardware device at a time. Is that correct? If yes, how can we not incur that limitation.

trustmanager/pkcs11/common/hardwareprivatekey.go
florianeichin: changed "image" to "artifact"

trustmanager/pkcs11/common/hardwarestore.go
florianeichin: removed the line

trustmanager/pkcs11/common/universal.go
florianeichin: yes this is right, depending on which libs are found (opencryptoki  and yubikey right now) a hardware device is chosen. Yet it would choose one of the libs and performing every crypto operation with it. What would be your suggestion for breaking down this limitation? And what do you mean with "at a time"? You would like to have a selection, which hardware should bes used? Didn't think about it (and it's design implications) too much yet, but with further development there could be a mechanism that chooses one if just one is present and asks which to use if more than one is found.

client/client.go
```diff
@@ -69,7 +70,7 @@ func NewFileCachedRepository(baseDir string, gun data.GUN, baseURL string, rt ht
 	if err != nil {
 		return nil, err
 	}
-
+	pkcs11.Setup()
```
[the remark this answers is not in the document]
florianeichin: You are absolutely right. A non_pkcs11 package should not contain any relation to a pkcs11 package it. I added an init() function in the pkcs11 package.  
The init() just calls Setup(), because Docker for example uses another entrypoint into notary while building/pushing. It uses client/repo(_pkcs11).go as entry. If not calling Setup() there, the whole pkcs11 package would not be imported just pkcs11/common. 
So init of pkcs11 would  not be called. 
Therefore in repo_pcks11.go the call for Setup() remains.

trustmanager/pkcs11/common/hardwareprivatekey.go
florianeichin: I actualy don't know in that case. Took this from the original yubikey implementation. Seems to be implemented like this since the first yubikey enhanced release. So I don't know what was the intention to do it like that. Didn't want to touch things like this, to keep this restructuring as simple as possible.

trustmanager/pkcs11/common/universal.go
endophage: Ideally we would instantiate a key store for every recognized hardware device attached to the system. You might run into cases where say, 2 yubikeys were attached, one having the root or targets key, one having your delegation key.

florianeichin:
As I'm leaving the team, developing the code leading to this PR, https://github.com/jschintag will continue development

PR #1354, opened 2018-05-16
Update test certificates

justincormack:
These had expired!


seb-bah:
@justincormack could you update these again? It looks like they just expired on 3/31/2021.

PR #1361, opened 2018-05-29
Fix small typo

justincormack:


endophage:
@justincormack looks like circleCI is building on your fork? Not sure if that's causing it not to show as fixed?

justincormack:
Ah that would explain it. I mean it doesn't explain why my CircleCI build is failing but yes it could be a weird interference, trying with it disabled.

PR #1362, opened 2018-05-29
Add libpython-dev to CirclCI build

justincormack:
Build is failing not being able to find python.h on CircleCI in #1361


justincormack:
Something weird going on, I think we are getting a really old version of compose...

PR #1375, opened 2018-07-17
[WIP] Update to CircleCI 2.0

### LINKED ISSUE #1316: Update CircleCI to V2
CircleCI is deprecating V1 syntax: https://circleci.com/blog/sunsetting-1-0/

We should update to V2 to avoid running into issues at the deprecation date.

justincormack:
The old CircleCI config is going away at the end of August.

This splits the old parallel script into the new native parallel jobs
format.

fix #1316 

Largely working, except that CircleCI Docker images won't populate volumes as the daemon is remote. This means the database jobs are failing, and I need to fix this still...


justincormack:
#1395 was merged instead, but still need to improve parellelism here.

PR #1377, opened 2018-08-01
Notary client refactor - simpler reader interface

cyli:
Refactor the notary client library to have a separate reader and loader that may be easier to use for clients that do not need any publishing capabilities.

Originally proposed something similar in https://github.com/theupdateframework/notary/issues/968 - this just makes it easier to set up a downloading client that can read to an in-memory cache, or just something that will load from cache without having to provide a CryptoService, etc.

The plan is, if this PR passes design and code review, for a new PR converting all the `client_update_tests` to just use the `LoadTUFRepo` function and an in-memory cache.

cc @justincormack

### DIFF
--- a/client/interface.go
+++ b/client/interface.go
@@ -6,42 +6,145 @@ import (
 	"github.com/theupdateframework/notary/tuf/signed"
 )
 
-// Repository represents the set of options that must be supported over a TUF repo.
-type Repository interface {
-	// General management operations
-	Initialize(rootKeyIDs []string, serverManagedRoles ...data.RoleName) error
-	InitializeWithCertificate(rootKeyIDs []string, rootCerts []data.PublicKey, serverManagedRoles ...data.RoleName) error
-	Publish() error
-
-	// Target Operations
-	AddTarget(target *Target, roles ...data.RoleName) error
-	RemoveTarget(targetName string, roles ...data.RoleName) error
+// ReadOnly represents the set of options that must be supported over a TUF repo for
+// reading
+type ReadOnly interface {
+	// ListTargets lists all targets for the current repository. The list of
+	// roles should be passed in order from highest to lowest priority.
+	//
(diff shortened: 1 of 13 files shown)


cyli:
(Test issues are due to GAS being renamed - I’ll fix that in a separate PR)

client/interface.go
```diff
@@ -6,42 +6,145 @@ import (
 	"github.com/theupdateframework/notary/tuf/signed"
 )
 
-// Repository represents the set of options that must be supported over a TUF repo.
+// Reader represents the set of options that must be supported over a TUF repo for
+// reading
+type Reader interface {
```
justincormack: Wondering if there is a better name than `Reader` as that usually means an `io.Reader`...

cyli: I am happy to name it something else.  Alternately we could just make those methods that take a TUF repo instead of having a wrapper

cyli:
Thanks for the feedback @justincormack and @endophage!  Renamed to `ReadOnly`.

cyli:
I think there were some gosec ci issues in this branch, and I’ll rebase but ci will probably still fail until https://github.com/theupdateframework/notary/pull/1382 is merged

cc @riyazdf @ecordell if you could have a look as well, since we had discussed this possible refactor way way earlier?

client/interface.go
```diff
@@ -6,42 +6,145 @@ import (
 	"github.com/theupdateframework/notary/tuf/signed"
 )
 
-// Repository represents the set of options that must be supported over a TUF repo.
+// ReadOnly represents the set of options that must be supported over a TUF repo for
+// reading
+type ReadOnly interface {
```
riyazdf: nice 👍

thaJeztah:
🎉 

/cc @tiborvass @anusha-ragunathan this one's merged

PR #1382, opened 2018-08-14
Add import of sha1 to gosec whitelist

justincormack:
The code checker, gosec, has started flagging the import of crypto/sha1
as potentially weak crypto. It is being used as the hash function for
PBKDF2 when keys are password protected.

In the short term I think removing the warning is fine as it restores
the status quo; need to look further if we really need this code.


justincormack:
Opened an issue to remind me to look into this in more detail later #1383

HuKeping:
:) have a nice trip

PR #1391, opened 2018-09-08
Lame Circle 2.0 conversion

cyli:
Sorry about the delay, @justincormack.  This converts the circle config file to 2.0 format.  This uses the machine executor, and still uses 4 parallel machines, which is pretty bad.

@justincormack's #1316 would be better, but this is a quick and dirty way to try to keep CI running in the meantime, and should be superceded by #1316 when that is merged.


cyli:
Ok I'm not sure how to get CircleCI to build using this particular circle config file.  I ran a build manually on my fork using https://circleci.com/docs/2.0/examples/#section=configuration, and it seemed to pass.

justincormack:
Hmm, it should just use the new config thats odd, it seemed to work on my config.

(also not sure why docker-jenkins has appeared here)

justincormack:
replaced by #1395

Issue #1388, opened 2018-08-29
Generalized Notary Client External Signer/HSM Interface


dbilling:
I'm new to TUF and notary but I'm familiar with the corporate software signing operations at the company where I work. We use a network-based HSM solution for signing that handles our large organization problems - problems like corporate access control and authorization of signing requests,  ability to do signing requests from data center VMs that don't have accessible usb ports, etc. etc.

As we investigated signing for docker images we've found the most important private keys are stored on the notary client machine or a local yubikey.  Neither of these solutions satisfies the signing needs of a large organization like ours.  I looked through other notary issues and I am excited to see several folks talking about extending the keystore/HSM support on the client. However, unless I'm mistaken (which is entirely possible), it appears the writers of these issues propose to modify notary client to work with their specific HSM and/or their specific PKCS#11 interface package.   This is all good, but takes notary down a (never ending?) path of supporting all sort of different variants of HSMs and HSM interfaces inside the client.  Might it be better to have one external interface for all this? 

So, what I'm proposing here is a generalized "external keystore" plugin solution -- some sort of proper interface/API where any interested party could write a plugin to teach the notary client to interact with their specific HSM/keystore needs -- such that the code that interfaces with that keystore/HSM would be outside notary.   The interface wouldn't be PKCS11 or KMIP or MS-CAPI -- instead it would be a notary defined plugin interface anyone could write a plugin to for whatever local or remote HSM/keystore they need accessible from the notary client.   With such an API in place, the notary client would instantly have a compatibility story for interacting with any local or remote keystore/HSM/HSM system without having to muck with notary itself.

Thoughts?  After writing all this I found the trustmanager remoteks grpc interface, but that seems to be solving a different problem.   Could it be used/extended/adapted for this purpose?

justincormack:
Hmm, I think the hope was that PKCS11 was standard enough to be supported by the devices people use, and that the current work on restructuring that would make it easier to use with other devices. What APIs are you interested in supporting? I have had recent interest in better HSM integration just yesterday, so interested in this area.

dbilling:
[...]

PR #1393, opened 2018-10-10
Proposed interface for GRPC remote keystore

dbilling:
Submitting this PR for initial review and discussion (not for merging)  of a grpc .proto service definition for a remote keystore as discussed in #1388

Most of this .proto facilitates a remote implementation of the current notary Keystore interface.  NewKey() however is designed to be used by new proposed notary functionality as discussed in #1388 , which would allow a HSM to generate keypairs and never expose the private key.

related issues: #1388
Signed-off-by: Dave Billing <dave@davebilling.com>


trustmanager/grpckeystore/grpckeystore.proto
```diff
@@ -0,0 +1,123 @@
...
+}
+
+message NewKeyRsp {
+    string KeyId = 1;       // for NewKey, KeyId is generated by remote store
+    string PublicKey = 2;   // Public Key in PEM format, Private Key is never shared
+    string DebugMsg = 3;    // Optional informational string returned from remote
+}
+
+message AddKeyReq {
+    string KeyId = 1;
```
justincormack: With AddKey, I think there may be interfaces where KeyId cannot be specified?

trustmanager/grpckeystore/grpckeystore.proto
```diff
@@ -0,0 +1,123 @@
...
+    string DebugMsg = 1;
+}
+
+message SignReq {
+    string KeyId = 1;
+    bytes Message = 2;
+}
+
+message SignRsp {
+    string Signature = 1;
```
justincormack: should this be `bytes`?

trustmanager/grpckeystore/grpckeystore.proto
dbilling: I might need some help understanding the case you are concerned about.  From what I can gather from the current yubi and generic keystore implementations, they seem to take for granted that AddKey() provides the keyID via privKey.ID().

trustmanager/grpckeystore/grpckeystore.proto
dbilling: I tried to match the format of how notary's json stores these same elements so it would be easier for an implementer/debugger to follow the trail of information... poking around the TUF metadata it looked like keys and signatures are encoded PEM style/base64 but hashes are shown as hex strings, so my thought was to use the same model here (keys and signatures as  encoded strings, message (hash) as binary bytes.  However, I certainly don't have a strong opinion here... just let me know what you'd like to see...

trustmanager/grpckeystore/grpckeystore.proto
endophage: As the key ID is the TUF key ID, we'd probably always want the client to specify it, otherwise the server would have to understand how to generate a TUF key ID. 

That being said, the ID would need to be generated by the GRPC client wrapper as the `KeyStore.AddKey` method doesn't take an ID. That's how all the other concrete `KeyStore` implementers work though so nothing strange there (https://github.com/theupdateframework/notary/blob/master/trustmanager/keystore.go#L105).

trustmanager/grpckeystore/grpckeystore.proto
endophage: The signatures are `[]byte` types (https://github.com/theupdateframework/notary/blob/master/tuf/data/types.go#L164). Go automatically encodes byte slices to base64 when marshaling them to JSON. I don't think we require signatures to be PEM. It's up to the individual signer/verifier for a given signature algorithm to be consistent with each other.

trustmanager/grpckeystore/grpckeystore.proto
dbilling: OK sounds like we want to keep KeyID in the AddKeyReq message, which will be set by the client wrapper e.g. keyID := privKey.ID() .  Sounds like we should remove keyID from NewKeyResp, since that would have required that the HSM side to generate a keyID, which we want to avoid.

trustmanager/grpckeystore/grpckeystore.proto
dbilling: OK I'll switch Signature to bytes.   Based on your same reasoning, I'm thinking I should also switch PublicKey and PrivateKey to bytes too.

trustmanager/grpckeystore/grpckeystore.proto
dbilling: this will mean AddKey will be required even when NewKey was used (so that the KeyID can be communicated).   Also means that when NewKey was used,  AddKey will not supply a PrivateKey.

endophage: The key ID is generated from the public key so we don't need to use AddKey with NewKey. The fact NewKey returns the public key is sufficient.

PrivateKey.ID() is actually inherited from the PublicKey interface being embedded in PrivateKey (https://github.com/theupdateframework/notary/blob/master/tuf/data/keys.go#L29)

dbilling: If we go with notary (grpc client) always generating the KeyID (even in the NewKey case), the KeyID still needs to get over to the HSM (grpc server) somehow.  Was thinking AddKey would be they way to send it over.  Otherwise how would it get there?

endophage: Ahhh, I see what you're saying. It's an interesting problem. What would be included in the AddKey message to enable the HSM to associate the KeyID with the correct key? We wouldn't have the private key if it had been generated on the HSM via NewKey.

justincormack: Looking at the YubiHSM API for example in https://developers.yubico.com/YubiHSM2/Commands/Generate_Asymmetric_Key.html it needs a key ID thats specific to the HSM (ie 2 bytes), which is a spare ID, so you would not want to specify it (you want the GRPC endpoint to pick a spare one), but it does need to be returned so you can pass it back to use the key. Unless the API endpoint keeps an association between Notary IDs and HSM key ids.

dbilling: I think we have several choices here... I can see any of the following ideas working, but they all have pluses and minuses:
1) idea 1: have the GRPC server generate the KeyID and return it in the NewKeyRsp.  That was my original thought with this .proto.  However, that comes with the downside of having to tach the GRPC server how to generate KeyIDs, as you noted.  In this model, an AddKey would not follow a NewKey, since only one is needed.
2) idea 2: If we want to make sure the notary client generated the KeyID, we could require an AddKey sequence even when NewKey was used, so that the KeyID can be sent over.  AddKey would contain the public key that was returned in NewKey so the server side would know what key we are talking about, but the private key would be empty in this case.
3) idea 3: To avoid overloading AddKey as shown idea 2, but still want have the notary client generate the KeyID, we could add an additional exchange -- something like like UseKey.  UseKey would always follow NewKey, and would basically indicate that the notary client accepted the NewKey and would provide the KeyID in UseKey.

endophage: Given we're implementing a GRPC service, I think it's reasonable to expect there is some type of server wrapping the HSM (let me know if you disagree).

In that case, we don't need to be bound to the ID scheme of the HSM, the wrapping server can implement a mapping from TUF IDs to the HSM's ID scheme.

With those assumptions, I'd prefer option 3 rather than overloading `AddKey`. Something like `Associate(id, publicKey)`

dbilling: @justincormack You bring up a really good point.  I like the idea of adding a server side correlator such as  "RemoteID"  that the server provides on NewKeyRsp, AddKeyRsp, and ListKeysRsp, and that the the client will echo back on GetKeyReq, GetKeyInfoReq, RemoteKeyReq, and SignReq.  That way, the remote store can correlate to keys using whatever it's preferred method happens to be.  However, that's really solves a different issue than the one were talking about here.  To make ListKeys work across a GRPC interface it's looking like the remote server still must store the KeyID, GUN and role alongside every key pair it holds.

endophage:
I'm sure we'll find tweaks that need to be made as the client and server are concretely implemented but I'm good with merging this as an alpha version of the interface.

PR #1395, opened 2018-10-15
Fix circleci

justincormack:
Make an even simpler version of #1391 to fix CI. Remove parallelism, as not working on new machine executor. Can improve later.

Also fix up golint as import path has changed.


justincormack:
cc @cyli @riyazdf @endophage ok this fixes CircleCI so we can move forward, can fix better later

buildscripts/circle_parallelism.sh
```diff
@@ -1,19 +1,20 @@
...
-   docker run --rm -e NOTARY_BUILDTAGS=pkcs11 notary_client make lint
-   ;;
-3) SKIPENVCHECK=1 make TESTDB=rethink testdb
-   SKIPENVCHECK=1 make TESTDB=rethink integration
-   SKIPENVCHECK=1 make TESTDB=postgresql testdb
-   SKIPENVCHECK=1 make TESTDB=postgresql integration
-   ;;
-esac
+
+# parallelism disabled here as not working in circleci 2.0
```
endophage: Parallelism should be working in 2.0 but it's now defined in the config.yaml at the same level as `machine` and `steps`.

PR #1403, opened 2018-11-14
Hot reload of server certificate

krockpot:
For some use cases, the certificate configured on the host may be short lived and rotated regularly. Rather than restart notary, it makes sense for it to regularly update its certificate information based on what is at the configured path.


utils/configuration.go
```diff
@@ -76,7 +79,52 @@ func ParseServerTLS(configuration *viper.Viper, tlsRequired bool) (*tls.Config,
...
+	conf, err := tlsconfig.Server(tlsOpts)
+	if err != nil {
+		return nil, err
+	}
+
+	// Reload certificate hourly
+	var reloadCert *tls.Certificate
+	var m sync.RWMutex
+	var resetCert func()
+	resetCert = func() {
```
endophage: no need for line 91. `resetCert := func() {` is sufficient

utils/configuration.go
```diff
@@ -76,7 +79,52 @@ func ParseServerTLS(configuration *viper.Viper, tlsRequired bool) (*tls.Config,
...
+		m.Lock()
+		defer m.Unlock()
+		reloadCert, err = loadCertificate(tlsOpts.CertFile, tlsOpts.KeyFile)
+		if err != nil {
+			logrus.Warnf("failed to (re)load certificate %s", err)
+		}
+		// Re-run this function hourly
+		time.AfterFunc(time.Hour, resetCert)
```
endophage: The cadence seems like it should be configurable.

utils/configuration.go
krockpot: It's necessary in this case because we call the time.AfterFunc bit inside the declaration

utils/configuration.go
endophage: Default 0 == no rotation works for me. It would be unreasonable to interpret it as "always be rotating" so there shouldn't be any future conflict :)

endophage:
Noticed the dco/no tag is still one here. Please make sure _all_ commits in this PR have the Signed off by line. This will block merging.

krockpot:
ah yeah I think that's because I used the update branch button in the UI. rebased manually.

krockpot:
Unclear why CircleCI was waiting, so I'll just rebase from master again 🤷‍♂️

krockpot:
@endophage should be good to merge (after a few rebases),, don't think I have perms for that though.

krockpot:
sorry to spam but @cyli @endophage  @justincormack any other actions before this can be merged? Thanks!

krockpot:
Just wanted to follow up again and see if there is any official process to have this PR merged now that it's been accepted. Thanks! (@endophage)

marcofranssen:
> Please sign your commits following these rules:
> https://github.com/moby/moby/blob/master/CONTRIBUTING.md#sign-your-work
> The easiest way to do this is to amend the last commit:
> 
> ```
> $ git clone -b "certificate_rotation" git@github.com:krockpot/notary.git somewhere
> $ cd somewhere
> $ git commit --amend -s --no-edit
> $ git push -f
> ```
> 
> Amending updates the existing PR. You **DO NOT** need to open a new one.

https://marcofranssen.nl/secure-2fa-ssh-and-pgp-using-krypton/#PGP-setup Here is also a guide which eases howto setup your GPG key

PR #1409, opened 2018-12-16
Update to Go 1.10.8

justincormack:
Includes denial of service fix on TLS connections, and other security fixes.

1.10.7 images are not out yet, and 1.10.6 has a bug so that breaks build, so need to wait for 1.10.7 images.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.10.3
+FROM golang:1.10.8
 
 RUN apt-get update && apt-get install -y \
 	curl \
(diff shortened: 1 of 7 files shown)


justincormack:
Ok, the base images are up now. We should probably do a release with this update  as it is a DoS issue cc @theupdateframework/notary-maintainers

justincormack:
Updated to Go 1.8. For the Go 1.78 fixes, we ended up adjusting the Docker builds to use the updated Go, as those do not use the Dockerfiles here. But we really should update Go here too now. (I will fix the 1.11 issues but its more complicated due to the OSX cross builds breaking).

cc @cyli @riyazdf @endophage

PR #1416, opened 2019-01-30
Bump golang 1.11, and switch to dockercore/golang-cross for cross-compiling

thaJeztah:
Alternative to https://github.com/theupdateframework/notary/pull/1396, but let me know if it's acceptable to switch to a different base-image (instead of the official Go image)

Added an extra commit to switch to the golang-cross image, and rebased @justincormack's PR on that, then bumped Go to 1.11.5

Note that the version built on Go 1.11 requires `libtool` to be installed on macOS in order to run the binary;

On Go 1.10:

```
otool -L ./cross/darwin/amd64/notary
./cross/darwin/amd64/notary:
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1253.0.0)
	/System/Library/Frameworks/Security.framework/Versions/A/Security (compatibility version 1.0.0, current version 57336.1.9)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1225.1.1)
```

On Go 1.11:

```
otool -L ./cross/darwin/amd64/notary
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1151.16.0)
	/System/Library/Frameworks/Security.framework/Versions/A/Security (compatibility version 1.0.0, current version 57031.1.34)
	@@HOMEBREW_PREFIX@@/lib/libltdl.7.dylib (compatibility version 11.0.0, current version 11.1.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1213.0.0)
```

```
./cross/darwin/amd64/notary version
dyld: Library not loaded: @@HOMEBREW_PREFIX@@/lib/libltdl.7.dylib
  Referenced from: /Users/sebastiaan/go/src/github.com/theupdateframework/notary/./cross/darwin/amd64/notary
  Reason: image not found
```

Installing libtool resolves that;

```
brew install libtool

./cross/darwin/amd64/notary version
notary
 Version:    0.6.1
 Git commit: 1d28d8ae
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.10.8
+FROM golang:1.11.2
 
 RUN apt-get update && apt-get install -y \
 	curl \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.11.2
+FROM golang:1.11.5
 
 RUN apt-get update && apt-get install -y \
 	curl \
--- a/cmd/notary-server/main_test.go
+++ b/cmd/notary-server/main_test.go
@@ -380,9 +380,9 @@ func TestGetCacheConfig(t *testing.T) {
 
 func TestGetGUNPRefixes(t *testing.T) {
 	valids := map[string][]string{
-		`{}`: nil,
-		`{"repositories": {"gun_prefixes": []}}`:         nil,
-		`{"repositories": {}}`:                           nil,
+		`{}`:                                     nil,
+		`{"repositories": {"gun_prefixes": []}}`: nil,
+		`{"repositories": {}}`:                   nil,
 		`{"repositories": {"gun_prefixes": ["hello/"]}}`: {"hello/"},
 	}
 	invalids := []string{
--- a/cross.Dockerfile
+++ b/cross.Dockerfile
@@ -1,10 +1,9 @@
-FROM golang:1.10.8
+FROM dockercore/golang-cross:1.10.8
 
 RUN apt-get update && apt-get install -y \
 	curl \
 	clang \
 	file \
-	libltdl-dev \
 	libsqlite3-dev \
 	patch \
 	tar \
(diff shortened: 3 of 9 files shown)


thaJeztah:
oh, FWIW; this is the repository that that image is built from (it's an auto build); https://github.com/docker/golang-cross

justincormack:
Hmm, why does the libtool dependency change? That seems weird.

thaJeztah:
@justincormack this was the description from the change in that image; https://github.com/docker/golang-cross/commit/41b2acab7ded9c3ffd7de94ef8947b5d2d29313d

possibly @kolyshkin is able to provide more insight though

justincormack:
Doesn't explain why the runtime dependency has changed. This is potentially problematic for consumers of this, like Docker Desktop which is shipping OSX Notary binary.

thaJeztah:
There's some mention here https://developer.apple.com/library/archive/documentation/Porting/Conceptual/PortingUnix/compiling/compiling.html (and found some [here](https://fmttools.ewi.utwente.nl/redmine//projects/jtorx/wiki/Additional_dependency_for_Mac_OS_X_108_(Mountain_Lion)_for_the_anidot_automaton_visualization) (and saw some others that it was needed in recent MacOS versions), but I'm quite a n00b in that area 😅

kolyshkin:
libtool is not a new dependency per se; it is (or, rather, its header, `ltdl.h`) needed to compile miekg/pkcs11 package (which is vendored by docker/cli). It looks like a compile-time (rather than runtime) dependency to me.

justincormack:
@kolyshkin no it is a runtime dependency, and the current versions we ship do not have it. It is a bit weird though, as I would expect them to have it unless we statically link at present.

cyli:
The current notary binary has libtool itself statically linked in:  https://github.com/theupdateframework/notary/blob/master/buildscripts/cross.sh#L7.

If you remove the `libltdl.dylib` and just leave the `libltdl.a` file, it will use the static library.  Luckily the mac `libltdl.a` library at the time didn't have any other dependencies, but the linux one does (`libtool.so`), so this trick worked on mac os but not linux.

If you untar the current `MacOSX10.11.sdk` tarball:
```
$ ls MacOSX10.11.sdk/usr/lib/libltdl*
MacOSX10.11.sdk/usr/lib/libltdl.a
```

And yes the dependencies were injected into that tarball a while ago from somewhere.  :| . This is probably not the ideal way to do things, but you're not supposed to statically link against system libraries in OS X so we can't just build a single static binary.

justincormack:
Ah that makes sense. As the libtool license is GPL except:

As a special exception to the GNU General Public License,
if you distribute this file as part of a program or library that
is built using GNU Libtool, you may include this file under the
same distribution terms that you use for the rest of that program.

and we are not built with that, statically linking seems problematic (and dynamic linking is not ideal). So I think this is better. (I wan't to rewrite all the PKCS#11 stuff in native Go soon).

justincormack:
Going to remove libtool requirement in another PR have opened PR upstream https://github.com/miekg/pkcs11/pull/102

PR #1422, opened 2019-02-18
Replacing 'http' by 'https' for securing kubernetes links

longkb:
This commit aims to replace "http://kubernetes.io/" by
"https://kubernetes.io/" for security.

Co-Authored-By: Nguyen Phuong An <AnNP@vn.fujitsu.com>
Signed-off-by: Kim Bao Long <longkb@vn.fujitsu.com>


eiais:
It looks like both of those changes are in the vendor folder. It might be better to send the changes to those projects :)

justincormack:
Yes, we can't update vendored dependencies.

PR #1427, opened 2019-02-20
Adding `/usr/bin/env bash`

truongnh1992:
This commit aims to add `/usr/bin/env bash` as a shebang line
to indicates scripts use bash shell for interpreting.

Signed-off-by: Nguyen Hai Truong <truongnh@vn.fujitsu.com>

### DIFF
--- a/migrations/rethink_migrate.sh
+++ b/migrations/rethink_migrate.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env sh
 
 # When run in the docker containers, the working directory
 # is the root of the repo.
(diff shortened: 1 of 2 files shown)


justincormack:
I would rather we could just use `/bin/sh` to be honest. Is there anything non Posix in those files?

truongnh1992:
> I would rather we could just use `/bin/sh` to be honest. Is there anything non Posix in those files?

It uses `bash` instead of `sh` shell.

justincormack:
I know it uses `bash` now, but does it need to? Does it use any bash-only features?

marcofranssen:
`/usr/bin/env sh` or `/usr/bin/env bash` is the recommended best practice when writing bash scripts.

https://unix.stackexchange.com/questions/29608/why-is-it-better-to-use-usr-bin-env-name-instead-of-path-to-name-as-my

PR #1434, opened 2019-02-25
Update pkcs#11 library

justincormack:
No longer needs libtool, which simplifies build.


kolyshkin:
LGTM

PR #1440, opened 2019-03-05
Updating warn message in the import flow

### LINKED ISSUE #1064: Improve key import error messaging
Per #1057 it's apparent that the messaging when import fails to parse a PEM block is confusing and should be clearer that the block is being skipped and import will continue attempting to read the remaining PEM blocks in the file.

If it's not already done, we should also consider outputting some kind of success message at the end of the import.

Avni-Sharma:
Fixes #1064


trustmanager/keys.go
```diff
@@ -210,7 +210,7 @@ func checkValidity(block *pem.Block) (string, error) {
 		// check if the key is missing a gun header or has an empty gun and error out since we don't know what gun it belongs to
 		if block.Headers["gun"] == "" {
 			logrus.Warnf("failed to import key (%s) to store: Cannot have canonical role key without a gun, don't know what gun it belongs to", block.Headers["path"])
-			return "", errors.New("invalid key pem block")
+			return "", errors.New("invalid key pem block.")
```
endophage: We're not perfect in execution but generally errors shouldn't end in punctuation c.f. https://github.com/golang/go/wiki/CodeReviewComments#error-strings

PR #1456, opened 2019-05-13
Add http.Request to doAuth to use with challenge.SetHeaders

eclipseo:
Newer challenge.SetHeaders takes two parameters, r *http.Request and
w http.ResponseWriter.
See commit https://github.com/docker/distribution/commit/f730f3ab77d0098f5172a825d9feecc6265f3c48

Signed-off-by: Robert-André Mauchin <zebob.m@gmail.com>

You will need to bump github.com/docker/distribution vendoring if you wanna use this. In Fedora we unbundle all the dependencies, so we need to work with the latest version.


justincormack:
You need to bump the vendoring here as we use vendoring in this repo, regardless of whether you use the vendoring.

eclipseo:
Might need some help with that, running `~/go/bin/vndr github.com/docker/distribution` removes tons of vendor files,  including `github.com/docker/distribution` and I don't know why. I've never used that tool before so maybe I am doing something wrong.

eclipseo:
vndr only works correctly if the package is located in the GOPATH, a caveat not mentioned on the README.

justincormack:
Oh yes, sorry, `vndr` assumes you do that. We should switch to new style modules.

Looks like the new version of `distribution` needs `github.com/opencontainers/image-spec` adding to the vendoring too, from the CircleCI error.

justincormack:
cc @cyli @endophage

PR #1472, opened 2019-07-19
Fix algorithm case handling (fix #1471)

### LINKED ISSUE #1471: key generate algorithm param is not really case-insensitive
The intent is apparently for the algorithm parameter to be case-insensitive:

https://github.com/theupdateframework/notary/blob/ce99e3cc568549b10ced3eb49cbb7e2941924eb4/cmd/notary/keys.go#L206

But then, the algorithm param is passed unchanged to further methods, and if not lowercase will fail during private key generation...

```
$ notary key generate ECDSA --output foo

* fatal: private key type not supported for key generation: ECDSA
```

dubo-dubon-duponey:
Fix #1471

### DIFF
--- a/cmd/notary/keys.go
+++ b/cmd/notary/keys.go
@@ -196,14 +196,14 @@ func (k *keyCommander) keysGenerate(cmd *cobra.Command, args []string) error {
 
 	// If we were provided an argument lets attempt to use it as an algorithm
 	if len(args) > 0 {
-		algorithm = args[0]
+		algorithm = strings.ToLower(args[0])
 	}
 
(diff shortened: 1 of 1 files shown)


HuKeping:
He is right, but I also think whether we should update the switch case for `utils.GenerateKey` that change it like:
```
...
        switch algorithm {
	case data.ECDSAKey:
        case String.ToUpper(data.ECDSAKey):
		return GenerateECDSAKey(rand.Reader)
	case data.ED25519Key:
		return GenerateED25519Key(rand.Reader)
	}
...

```

PR #1474, opened 2019-07-19
Fix #1473

### LINKED ISSUE #1473: `key generate --output` results in keys that can't be imported
`notary key generate --role delegation --output foo`

will generate a PEM *without* a `path` header.

Trying to `import` such a key will then fail.

dubo-dubon-duponey:
Fix #1473 

Note that `ConvertPrivateKeyToPKCS8` is called in a number of (other) places (not only `generate`).

I can't think of a reason why the additional PEM header would hurt, but then, somebody with a better understanding of the codebase should check.

### DIFF
--- a/tuf/utils/x509.go
+++ b/tuf/utils/x509.go
@@ -485,6 +485,7 @@ func ConvertPrivateKeyToPKCS8(key data.PrivateKey, role data.RoleName, gun data.
 		headers["gun"] = gun.String()
 	}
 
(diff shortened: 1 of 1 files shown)


dubo-dubon-duponey:
Not sure why this is breaking - tests seem to pass locally and circle output is hard to decipher.
Ideas?

dubo-dubon-duponey:
Closing for lack of interest.

PR #1486, opened 2019-08-14
CircleCI: update image, and use BuildKit

thaJeztah:
YOLO; https://discuss.circleci.com/t/default-machine-executor-image-update/29308 mentioned CircleCI now has Docker 18.09, so let's see if this works

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,11 +2,22 @@ version: 2
 jobs:
   build:
     machine:
-      image: circleci/classic:201808-01
+      image: ubuntu-1604:201903-01
+      # not supported in the free plan
+      # docker_layer_caching: true
     working_directory: ~/go/src/github.com/theupdateframework/notary
+    environment:
+      DOCKER_BUILDKIT: 1
     steps:
       - add_ssh_keys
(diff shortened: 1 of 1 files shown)


HuKeping:
strange image name, why they give up using circleci at all?

thaJeztah:
yes, for this image they don't use the `circleci` prefix; I was confused as well, but 🤷‍♂️

HuKeping:
As per https://discuss.circleci.com/t/default-machine-executor-image-update/29308, I'd like to merge this PR, otherwise we should change the image to ` circleci/classic:201710-01` which is tooooooooooooooooooooo old.

I don't like the image name without the prefix `circleci ` , but it seems we have no better choice.

thoughts @justincormack ?

thaJeztah:
Also didn't try yet, but I think we can bring back parallelism by putting the steps from the circle-ci script into the circle-ci YAML (as it does allow stages to be executed in parallel).

This would also allow parameterising the `docker build`, and defining a matrix to test against multiple Go versions (adding a `ARG GO_VERSION` to the Dockerfiles, and overriding the default version in CI

Issue #1502, opened 2019-10-04
Helm chart for Notary


patoarvizu:
I started exploring Notary a couple of days ago and I personally find it easier to run things locally on Kubernetes (with minikube or k3s) rather than docker-compose, but I couldn't find a 
Helm chart for Notary anywhere so I started writing one.

I plan on submitting a PR with a more polished chart in a couple of days, I just wanted to open the issue first to get visibility.

HuKeping:
Do we really want introduce helm to this git repo(for notary source code).

Actually I don't like the docker-compose.yml to be here either,  but since it could be a guide  to show people how to deploy notary-server and notary-signer and is simple enough, so I haven't drop it. 

I'm thinking if we should create another separated git repo to hold those templates-such as `notary-template-helm`, `notary-template-docker-compose` rather than tie it up with this git repo.

thoughts @theupdateframework/notary-maintainers ?

patoarvizu:
One of the benefits I can see from splitting things into different repos is that we would be relying less on shared things like manifests, config files with very specific options, documentation, etc. that needs to work for both. If we go that way, we could even abstract things out a little more and instead of providing these things, add automation on how to create them (go templates, `openssl`, or [`step`](https://smallstep.com/docs/cli/) commands, or Terraform running locally with the [TLS provider](https://www.terraform.io/docs/providers/tls/index.html), etc. This way, any changes on one solution won't risk breaking the other (and requiring more specific knowledge to fix them).

On the other hand, I think part of the value of adding this here, is that it shows activity and signals interest in the project (in the context of [this proposal](https://github.com/cncf/toc/issues/272)). I feel that keeping things separate will decrease visibility into the activity of the project. Furthermore, having the examples co-exist with the core code can (in my opinion) remove the cognitive wall of having two separate repos, and potentially encourage more community contributions. In other words, I think the docker-compose and Helm examples can be a form of "gateway code" 🙂

HuKeping:
> Furthermore, having the examples co-exist with the core code can (in my opinion) remove the cognitive wall of having two separate repos, and potentially encourage more community contributions. 

I agree it may increase the activity of this project if we put these two together, but in the meantime I also think it doesn't make much sense if the target is to **bring more contributor to this community.**

I've noticed the proposal at CNCF TOC, I think they're just doing their job, supervise the projects under CNCF and to see if there's any thing that they could help.

HuKeping:
[...]

PR #1503, opened 2019-10-10
Add helm chart

patoarvizu:
Addressing #1502 

I started exploring Notary a couple of weeks ago and while docker-compose was helpful, I was trying to run it with something more natural to me and I was surprised I couldn't find a Helm chart for it, so here's my attempt at that.

First of all, this chart is **not** meant to be used for production, but rather as a way to quickly spin up and tear down a Notary service (I'm adding a section to the README to make that clear). It doesn't mean that it can't evolve into that, but my immediate goal is to have something that can be iterated on very quickly, in a way that's more Kubernetes-friendly.

One thing I want to point out is that, to minimize duplication, I'm symlink-ing from `helm/fixtures`, `helm/migrations`, and `helm/notarysql` to their respective directories one level above. I'm doing that because of a restriction in Helm where charts can't access any files that are outside of its directory. With the symlink, the chart can use the same SQL files and TLS certificates used by the docker-compose files, without having to duplicate them.

This first version of the chart doesn't currently support 100% of the Notary settings. I decided to leave out `auth`, `reporting`, and `caching`, since I considered those to be less important for the initial goal.

I'm not sure what tests to run since this is all new code, but let me know if there's anything else I need to do! I'd welcome any suggestions to the new README section, specially if you think the section about it not being meant for prod should go **before** the main Helm section.

Lastly, I think I'm supposed to merge against `master`, but let me know if I should do it against a release branch!


patoarvizu:
Friendly `ping` here! Should I directly assign maintainers and/or contributors as reviewers?

patoarvizu:
@justincormack @HuKeping Any chance you can take a look? Or do you have any feedback?

HuKeping:
Apologise @patoarvizu ! i've noticed this PR a few days ago, but haven't find a time slot getting to it since it's quite big for this project.

justincormack:
Sorry, been travelling, will take a look soon.

helm/values.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - localhost
+  path: /
+
+loadBalancer:
+  enabled: true
+
+storage:
+  type: mysql
+  pvcSize: 100Mi
+  migrateVersion: v4.6.2
```
blhagadorn: Could you also allow a `storageClass` option here?

helm/values.yaml
```diff
@@ -0,0 +1,35 @@
...
+  enabled: true
+  tls:
+    enabled: true
+  hosts:
+  - localhost
+  path: /
+
+loadBalancer:
+  enabled: true
+
```
blhagadorn: Perhaps it would be good to put in the comments the options for the `type`, mysql, postgres, and file.

helm/values.yaml
patoarvizu: Sure, although the "default" value here would be somewhat tricky because we can set a default for the chart, but the chart doesn't know what the default `StorageClass` is on the target cluster. I guess the way to go here is to **not** set a chart default, and only conditionally set `storageClassName` [here](https://github.com/theupdateframework/notary/pull/1503/files#diff-a6372b10b9b4f207ef15397bd89d20b7R4) if `storage.className` is set?

helm/values.yaml
patoarvizu: You mean inline comments here in the file? I added [this](https://github.com/theupdateframework/notary/pull/1503/files#diff-04c6e90faac2675aa89e2176d2eec7d8R124) on the `README.md`, but I can add that here too if you think that adds value.

blhagadorn: I think in this case it would be helpful (three options that might take a while for someone to find out). I'm a fan of more commented Helm charts when it's not entirely clear, and I think this Consul chart does a great job: https://github.com/hashicorp/consul-helm/blob/master/values.yaml

patoarvizu: It definitely makes sense, I'll add that!

helm/values.yaml
blhagadorn: I think maybe that could be something further down the road, chart looks good to me (and seems to work)! I just had a problem with `pod has unbound immediate PersistentVolumeClaims` on my cloud provider which requires me to set the storage class to something different (this happens with lots of Helm charts so it's not unique), particularly I have to set my `volumeBindingMode` so something custom within the storage class.

helm/templates/_helpers.tpl
```diff
@@ -0,0 +1,52 @@
...
+{{- else -}}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{- define "notary.serverdburl" -}}
+{{- if eq .Values.storage.flavor "mysql" -}}
+{{- if .Values.storage.enabled -}}
+{{ .Values.server.storageCredentials.user }}:%% .Env.PASSWORD %%@tcp(notary-db:3306)/notaryserver
```
ralgozino: According to the documentation the connection string should include: `?parseTime=true`.

Eventhough if it's not present it would be added by default, it should be better if it is included in the chart.

There are 3 more ocurrences of the connection string, maybe it's a good idea to use a variable.

helm/templates/secrets.yaml
```diff
@@ -0,0 +1,110 @@
...
+{{- end }}
+{{ if and .Values.tls.database.signer.cert .Values.tls.database.signer.key }}
+  notary-signer.pem: |
+{{ .Values.tls.database.signer.cert | indent 4 }}
+  notary-signer-key.pem: |
+{{ .Values.tls.database.signer.key | indent 4 }}
+{{- end }}
+{{- else }}
+{{- $rootCA := genCA "Notary Root CA" (int .Values.tls.generated.validityDays) }}
+{{- $serverCert := genSignedCert "notary-server" (list) .Values.tls.generated.server.dns (int .Values.tls.generated.validityDays) $rootCA }}
```
ralgozino: Maybe it could be useful to add here also the fqdn of the service like `.Values.tls.generated.server.dns` +  `.Values.namespace` + `.svc.cluster.local`?

helm/values.yaml
```diff
@@ -0,0 +1,265 @@
...
+  custom: false
+
+  # The CA that signed the server/signer certificates. It's assumed that the same CA is used for bother server
+  # and signer certificates.
+  # It doesn't need to strictly be a root CA, it can be an intermediate one.
+  rootCACert: null
+
+  # The public certificate and corresponding private key for the server.
+  server:
+    cert: null
```
ralgozino: If I want to use my own certificates, in what format should I put the certificate here? base64? string? I would be nice if specified in the comments. The same for the rest of the certificates.

helm/templates/secrets.yaml
patoarvizu: Ah, good idea... I'll do that

helm/values.yaml
patoarvizu: It's a string but yes, it should be clarified in the comments. Thanks!

patoarvizu:
Gentle bump here...

Is there any interest in merging this? It was suggested that this can live in a separate repo, so I'll gladly do that too if that's the preference.

helm/templates/deployments.yaml
```diff
@@ -0,0 +1,371 @@
...
+
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  labels:
+    app.kubernetes.io/name: notary
+    app.kubernetes.io/component: notary-server
+  name: notary-server
+spec:
+  replicas: {{ .Values.signer.replicas }}
```
rubroboletus: In notary-server component, you are assigning signer.replicas instead of server.replicas

helm/templates/ingresses.yaml
```diff
@@ -0,0 +1,27 @@
...
+{{ .Values.ingress.annotations | toYaml | trim | indent 4 }}
+  {{- end }}
+spec:
+  rules:
+  {{- range $host := .Values.ingress.hosts }}
+  - host: {{ $host }}
+    http:
+      paths:
+      - backend:
+          serviceName: notaryserver
```
rubroboletus: Typo, right one service name in your chart is notary-server

patoarvizu:
Thanks for the comments @rubroboletus! I just pushed the changes to correct them.

jeffb4:
Is there any further feedback necessary to merge this PR?

patoarvizu:
I haven't received any further feedback on this PR despite multiple previous requests. Is there any interest in merging it here? Maybe on a different repo?

cc @justincormack @HuKeping @thaJeztah @cquon (As the most recent committers on `master`)

jeffb4:
There was some feedback about `?parseTime=true` but I don't see anything else

patoarvizu:
Ping!

Would it be possible to at least get some acknowledgement from the maintainers to know if it's still worth keeping this PR alive? It's the most commented open PR in the last 2 years and we haven't heard anything from any maintainer in over 6 months.

If you have any guidance on whether you'd like to maintain this in a different repo instead of this one, I'd be happy to follow that, but I would really appreciate if you can let me (and the people following the PR) if we should still expect this to get some attention at any point. Thanks!

DrissiReda:
Has there been any news on this?

patoarvizu:
No, and I don't think the maintainers are very motivated to move forward, since Notary V2 is in the works.

Here's what I'll do: I'm going to tag the maintainers one more time at the bottom of this comment to see if I can get their attention or acknowledgement. If we don't hear back in a reasonable amount of time, I will close this PR and migrate this to a repo on its own, and probably publish it to [Artifact Hub](https://artifacthub.io/).

@cyli @diogomonica @endophage @ecordell @HuKeping @justincormack @NathanMcCauley @riyazdf

PR #1504, opened 2019-10-16
Fix Dockerfile permissions and add .dockerignore

thaJeztah:
### Dockerfile: fix permissions, use "notary" as default user

Before this, running the circle-ci integration tests showed some failures:

```
pkcs11 import was not found anywhere without a build tag, yay
find: './cmd/notary/tuf/gun': Permission denied
find: './cmd/notary/tuf/gun': Permission denied
find: './cmd/notary/tuf/gun': Permission denied
find: './cmd/notary/tuf/gun': Permission denied
find: './cmd/notary/tuf/gun': Permission denied
find: './cmd/notary/tuf/gun': Permission denied
lstat cmd/notary/tuf/gun: permission denied
lstat cmd/notary/tuf/gun: permission denied
```

With this patch applied, the tests run without issues.

This patch modifies the circle-ci script accordingly, and makes a small
convenience change when running the tests locally; adding `-it` allows
for the containers to be terminated (CTRL-C), which was not possible
previously.

### Add .dockerignore to ignore build and test artefacts

Ignore files that are unlikely needed in the dockerfile


.dockerignore
```diff
@@ -0,0 +1,16 @@
+.cover
+.dockerignore
```
thaJeztah: Removed `.git`, as it's used by the `server.Dockerfile` to generate the `version` output. We could consider doing that outside of the container, but that may complicate building the Dockerfile (e.g. through docker hub autobuild)

PR #1506, opened 2019-10-16
gosec: ignore  G108: Profiling endpoint automatically exposed

thaJeztah:
This was a false positive, as no server is started unless `-debug` is enabled.

Also see https://github.com/theupdateframework/notary/pull/1505, which fixes CI to print the actual failure

### DIFF
--- a/cmd/notary-server/main.go
+++ b/cmd/notary-server/main.go
@@ -5,7 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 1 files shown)


thaJeztah:
yay! this is green. I see it's not fully up-to-date with master though 😞, so I'll rebase once (perhaps that check should be disabled)

thaJeztah:
ping @hukeping PTAL

PR #1508, opened 2019-10-19
Private keys and passwords present.

the4rchangel:


tuf/utils/pkcs8_test.go
```diff
@@ -11,33 +11,7 @@ import (
 )
 
 func getRSAKey() (data.PrivateKey, error) {
-	raw := []byte(`-----BEGIN RSA PRIVATE KEY-----
```
thaJeztah: These keys and passwords are specifically created for these tests, so it's ok to have them in here

HuKeping: it seems no risk here

thaJeztah: yes, I think this should be ok to close, unless we want to add a comment in files to explain this

HuKeping:
closing.

Please feel free to reopen it if there is any further work need to do.

PR #1510, opened 2019-10-22
Makefile: disable debug to make CI more readable

thaJeztah:
CI was quite noisy, and probably could do without the `-debug` option (I think it can still be enabled by setting a `DEBUG` env-var if needed)


HuKeping:
merging.

PR #1511, opened 2019-10-22
CircleCI: run jobs in parallel

thaJeztah:
Splitting the jobs the same as they were in the old situation;
https://github.com/theupdateframework/notary/blob/fb795b0bc868746ed2efa2cd7109346bc7ddf0a4/buildscripts/circle_parallelism.sh

follow-up to https://github.com/theupdateframework/notary/pull/1395


thaJeztah:
yup 👍

thaJeztah:
@justincormack @HuKeping PTAL

names for the jobs are a bit lame, but I couldn't find a better name for them
also wasn't sure if the "teardown" step was needed for all of them (let's see if it completes successfully)

.circleci/config.yml
```diff
@@ -16,8 +18,119 @@ jobs:
...
+          command: |
+            docker version
+            docker info
+            docker-compose version
+      - run:
+          name: "Build image"
+          command: docker build --progress=plain -t notary_client .
+      - run:
+          name: "ci"
+          command: docker run --rm -e NOTARY_BUILDTAGS --env-file buildscripts/env.list --user notary notary_client bash -c "make ci && codecov"
```
HuKeping: `NOTARY_BUILDTAGS=none`?

EDIT: ah ,ok, you define it over there.

thaJeztah: Ah, yes, I initially did that as I expected it to show up in the CI UI (but I think I mixed up CircleCI with TravisCI - Travis shows those in the UI 😂)

Still thought it was nice to have all the env-vars defined in a single place

PR #1520, opened 2020-01-09
Bump mariadb from 10.1.28 to 10.4

marcofranssen:
Bumped the mariadb container used in docker-compose to 10.4 docker image.

PR #1522, opened 2020-01-10
Upgrade docker images to Go 1.13.6

marcofranssen:
Folllowing PR should make the image for the cross.Dockerfile available

Running the ./build-image.sh from that PR should create it locally on your machine for testing purposes.

https://github.com/docker/golang-cross/pull/49

Depending on the other PR we might also take out that change for the cross.Dockerfile so this PR doesn't remain open for ages.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.12.12
+FROM golang:1.13.6
 
 RUN apt-get update && apt-get install -y \
 	curl \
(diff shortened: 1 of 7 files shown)


marcofranssen:
Rebased the branch on top of latest master since merging #1520

HuKeping:
i'm a little worried about the update of golang , it cost quite a lot of time to fix CI when we update it last time.

marcofranssen:
I do think it is worth to invest. Especially for better error handling using errorwrapping and also to have babysteps to be able to upgrade more easily to 1.14 in the near future as there we will benefit from huge performance improvements when it comes to `defer`.

With ther Go 1.13 error wrapping error handling can be simplified and improved.

justincormack:
upgraded to 1.14.1 in #1543

PR #1451, opened 2019-04-18
RFC: use go modules

SantiagoTorres:
I just did this for the sake of avoiding GOPATH insanity and quality of life.

What do you people think?

PR #1523, opened 2020-01-10
Feature/go modules

marcofranssen:
I managed to upgrade the codebase to use Go Modules.

In this PR I kept the Go version at 1.12 to not interfere with other PRs I created and to limit the risk of breaking things.


marcofranssen:
Rebased the branch on top of latest master since merging #1520

HuKeping:
love to see go modules reduce tons of code base , seems duplicated with #1451 ?

marcofranssen:
@HuKeping correct it is duplicate, however the other PR does not clean the vendor folder and introduces some other changes. It also fails the CI. This one works with CI and solely stick to switching to go modules. I also updated all the dockerfiles to work with Go modules.

Personally I'm using this now for a bunch of days to work on a docker-compose sandbox.

https://github.com/marcofranssen/notary/compare/feature/go-modules...marcofranssen:feature/sandbox

Once this sandbox works I also will make a PR for that, but it requires the go modules PR. With that sandbox merged also the following PR with Docker documentation can be simplified and improved.

https://github.com/docker/docker.github.io/pull/10078

Currently I'm using the sandbox in an internal project to further finetune and test.

HuKeping:
Would you mind if not changing the golang version for those docker image or any reason we must upgrade it? 

The rest LGTM!

Any suggestions please @thaJeztah @justincormack

marcofranssen:
@HuKeping the reason is to have the patches applied to Go 1.12 in there.

https://golang.org/doc/devel/release.html#go1.12.minor

I consider it a best practice to at least update patches regularly. 

Major and Minor releases is a different story though.

thaJeztah:
Could we still use vendoring? (we can use `go mod` to do the vendoring, just keep using the `vendor` directory).

marcofranssen:
@thaJeztah why would you keep that directory? What would be the advantage of that?

thaJeztah:
Vendoring has been a lifesaver on quite some occasions where either dependencies went AWOL, or in air gapped situations. In addition, the vendor directory allows for easier reviewing of the actual code being used. I understand a proxy could be used to address the "dependency going AWOL" case, but such a proxy would have to be hosted (and maintained).

I'm not a maintainer for this repository, so just my 0.02c

HuKeping:
>  the vendor directory allows for easier reviewing of the actual code being used.

agreed, go modules makes deps more like a  black-box.

marcofranssen:
Did some more research on the question to vendor or not to vendor.

https://github.com/golang/go/wiki/Modules#how-do-i-use-vendoring-with-modules-is-vendoring-going-away

It looks like it complicates development.

- What if you forget to re-vendor after updating modules?
- More complex PRs with lots of vendor updates? Why should we want to review that code?

Anyhow let me know what you guys prefer, then I will adjust the PR to move this forward.

endophage:
Using the `--mod=vendor` flag during testing/build in CI ensures the vendor directory matches the go.sum requirements. I've been using it  that way for the last year or so to resolve the update issue.

As far as more complex PRs, vendoring updates should be done either in a separate PR, or in a separate commit.

In the last 18 months I've had CI fail frequently due to dependency providers (in particular google of all places) not being available 24/7. I'm also not sure any of the maintainers are willing and/or able at this point to maintain a dependency cache. We'd be doing it of our own volition rather than as a company supported exercise.

marcofranssen:
@endophage shall I reintroduce the vendor folder then for the time being? I can do that and then also rebase my other PR #1530 on top.

thaJeztah:
> What if you forget to re-vendor after updating modules?

Yes, the repository should have a vendor check to validate that vendoring is done correctly (looks like there's no check currently? there should be one! 😅). Something like this; https://github.com/containerd/project/blob/b8808e568e74c98eb568f54135fcf6c9c89a3f8e/script/validate/vendor

marcofranssen:
@thaJeztah added the vendor folder and a fixed version of the script you shared.

@endophage @HuKeping are we ready to merge with these changes?

Also contributed the fix to the script here https://github.com/containerd/project/pull/44

buildscripts/validate-vendor.sh
```diff
@@ -0,0 +1,47 @@
...
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+
+set -eu -o pipefail
+
+if [ -f vendor.conf ]; then
+  rm -rf vendor/
+  vndr |& grep -v -i clone
```
justincormack: you can remove all the parts related to `vndr` as we will only be using this with `go mod`

thaJeztah:
Thanks for adding that validation!

I noticed the diff in vendoring is quite big; some of that can be attributed to `go mod` vendoring a slightly different set of files, but it also looks like some dependencies were updated (e.g. I noticed logrus going from 1.4.1 to 1.4.2).

Was this intentional? (Were there modules that enforced the updated version?) Basically; wondering if it's possible to keep this PR narrowed to _just_ the switch to go mod (same version of dependencies, only different tool for dependency management), and keep the updating of dependencies separate (could be either a PR that's merged _before_ or _after_ this one)

marcofranssen:
The script to validate vendor is also tested in this PR where I updated the viper depdency #1530

marcofranssen:
@thaJeztah there where some initial issues with prometheus, which I manually fixed by downgrading it after the switch. Might be some deps have been updated due to that fix.

marcofranssen:
@justincormack @endophage what is required to get this merged?

justincormack:
I would like there not to be any version changes; if there are issues can we deal with them in different PRs please.

marcofranssen:
I redid the whole modularization. See commit history.
There are no version changes except in the last commit which is required for CI to pass. 

`go mod tidy` keeps to update those few dependencies.

**github.com/stretchr/testify** goes from `1.2.2` to `1.4.0`
**golang.org/x/crypto** goes from `v0.0.0-20180411161317-d6449816ce06` to `v0.0.0-20190325154230-a5d413f7728c`
**golang.org/x/sys** goes from `v0.0.0-20180905080454-ebe1bf3edb33` to `v0.0.0-20190215142949-d0b11bdaac8a`

That is required to keep proper track  if all modules are properly vendored.

I have also added following commit 8dd890ce which adds GOFLAGS environment variable in Dockerfiles and Makefile to ensure the vendored modules are used.

I think we should merge the branch as is. The updated dependencies are because newer golang versions have the crypto and sys deps integrated now (assumption) and the testify lib is not very impacting (tests still succeed).

marcofranssen:
For review purposes I didn't delete the vendor.conf yet.

You can use following on the various commits to check the versions haven't changed in commits.

```bash
git diff --no-index --word-diff=color -- vendor.conf go.mod
```

To simplify this I have sorted the `vendor.conf`. If all good I can add one more commit to remove vendor.conf.

marcofranssen:
@HuKeping @justincormack @thaJeztah I found one bug with the dependencies when consuming notary as a lib in another project. I think in my original PR I resolved that by manually downgrading the vendored lib if I remember correctly. #1533 

Shall I try to fix the issue on this PR? Or shall I add another PR on top of this one trying to resolve it? Would be good if this can be merged as I would love to add a few more contributions on top of this.

justincormack:
Can you explain what the issue actually is?

marcofranssen:
@justincormack I have added a steps to reproduce 

https://github.com/theupdateframework/notary/issues/1533#issuecomment-584522088

justincormack:
ok LGTM I think we should get this  merged. cc @HuKeping

HuKeping:
Happy to see this get to be merge ready, thanks @marcofranssen @justincormack !

PR #1532, opened 2020-02-04
Change NewBasicKeyRequest to NewKeyRequest to use cfssl 1.4.1

eclipseo:
Signed-off-by: Robert-André Mauchin <zebob.m@gmail.com>


HuKeping:
Hi @eclipseo seems like we should update `cloudflare/cfssl` first?

PR #1555, opened 2020-05-09
Add expires date to the warning from nearExpiry

directionless:
When notary warns me of an impending expiration, the _first_ thing I want to know is when. This commit adds the date to the error message.


directionless:
Hi? Is there any interest in this?

PR #1559, opened 2020-05-25
Feature/sandbox

marcofranssen:
This PR adds a docker-compose setup that can be utilized to have a sandbox available for playing with notary and docker content trust.

resolves #1561 which was included by rebasing on top of #1560

PR #1562, opened 2020-06-03
linting: fix "Implicit memory aliasing in RangeStmt"

thaJeztah:
wasn't able to get this linting failiure locally for some reason, but CI was failing on this. This is probably goes `G601: Implicit memory aliasing in RangeStmt`, but the linting output is a bit confusing, so not sure which linter is producing it;

```
/go/src/github.com/theupdateframework/notary/tuf/signed/sign.go,103,Implicit memory aliasing in for loop.,MEDIUM,MEDIUM,&sig,CWE-118
make: *** [Makefile:109: lint] Error 1
```

I _think_ this should fix it 🤞


justincormack:
Hmm, you just confused the pattern matching algorithm for this, but technically it is correct about the aliasing, although it seems to be safe here, maybe there is a better fix.

marcofranssen:
What is the state on this one? I would like to rebase #1559 so we could also handle that one.

justincormack:
Looking at this again this is a good fix.

justincormack:
@marcofranssen sorry about delay please go ahead and rebase...

PR #1605, opened 2021-07-27
Convert ECDSA signing code to produce ECDSA signatures in ASN.1 form

stefan-zh:
The Elliptic Curve signatures that Notary produces are not in [ASN.1](https://en.wikipedia.org/wiki/ASN.1) form, which leads to incompatibilities with some popular external libraries and tools such as OpenSSL (https://github.com/theupdateframework/notary/issues/1544) or AWS KMS. The proposed change here will generate ECDSA signatures in ASN.1 form going forward, but also be backwards-compatible in its Verify logic, i.e. all signatures produced so far for users of Notary will continue to be verified successfully as the original verification logic will remain in the code. All that users will need to do, should they wish to upgrade, is rebuild their client, server and signer.

If there are any reasons why Notary should produce non-ASN.1 signatures, please discuss it in this thread.


tuf/signed/verifiers.go
```diff
@@ -242,6 +248,19 @@ func (v ECDSAVerifier) Verify(key data.PublicKey, sig []byte, msg []byte) error
...
+
+	// if the signature is in ASN.1 form, we can verify it directly
+	var signature ecdsaSig
+	if _, err := asn1.Unmarshal(sig, &signature); err == nil {
+		if !ecdsa.Verify(ecdsaPubKey, digest[:], signature.R, signature.S) {
+			logrus.Debugf("failed ECDSA signature validation")
+			return ErrInvalid
+		}
+		return nil
+	}
```
stefan-zh, line 261: If the signature cannot be parsed as ASN.1 signature, we continue below trying to parse it in the old Notary signature style.

justincormack:
It might be good to reference links to the docs for the old and new forms for reference.

stefan-zh:
@justincormack I updated the documentation. See if it looks good to you

stefan-zh:
Hey @diogomonica @endophage  I think you guys have worked on implementing the ECDSA keys functionality in Notary and go-tuf before and I'm curious to hear your thoughts.

PR #1607, opened 2021-08-04
BUGFIX: Resolve build error with 'make cross'

marcofranssen:
The build was failing due to a breaking change to `goclyclo`.

Recently (18 days ago at the time of this writing) `gocyclo` updated to `go1.16` and started using it's `io/fs` package. Due to this `make cross` breaks as it depends on Go `1.12.5`, I took liberty to update it to the latest go `1.13.5` release.

I also had to move the enabling of Go modules to be able to install gocyclo at a specific version/commit.

Also worked a bit on the cache-ability of layers in the image by splitting some RUN statements.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,14 +14,21 @@ RUN apt-get update && apt-get install -y \
 	&& rm -rf /var/lib/apt/lists/*
 
 RUN useradd -ms /bin/bash notary \
-	&& pip install codecov \
-	&& go get golang.org/x/lint/golint github.com/fzipp/gocyclo github.com/client9/misspell/cmd/misspell github.com/gordonklaus/ineffassign github.com/securego/gosec/cmd/gosec/...
+    && pip install codecov
 
-ENV NOTARYDIR /go/src/github.com/theupdateframework/notary
+ENV GO111MODULE=on
+
+# Locked go cyclo on this commit as newer commits depend on Golang 1.16 io/fs
+RUN go get golang.org/x/lint/golint \
+    github.com/client9/misspell/cmd/misspell \
+    github.com/gordonklaus/ineffassign \
+    github.com/securego/gosec/cmd/gosec/... \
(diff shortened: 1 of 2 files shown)


marcofranssen:
For some reason the CI does not seem to be running against the changes I made in this PR. It seems to be failing on the original broken code.

marcofranssen:
Have applied same fix for the Dockerfile. Personally wasn't using that so missed out on this initially.

Dockerfile
```diff
@@ -14,14 +14,21 @@ RUN apt-get update && apt-get install -y \
...
 
-ENV NOTARYDIR /go/src/github.com/theupdateframework/notary
+ENV GO111MODULE=on
+
+# Locked go cyclo on this commit as newer commits depend on Golang 1.16 io/fs
+RUN go get golang.org/x/lint/golint \
+    github.com/client9/misspell/cmd/misspell \
+    github.com/gordonklaus/ineffassign \
+    github.com/securego/gosec/cmd/gosec/... \
+    github.com/fzipp/gocyclo@0c41b8201a7b6d814898607c4e0cd097c0f2ae46
```
stefan-zh, line 26: What about using this commit [`ffe36aa317dcbb421a536de071660261136174dd`](https://github.com/fzipp/gocyclo/tree/ffe36aa317dcbb421a536de071660261136174dd)? It's from April 8th, 2021, which  is newer than `0c41b8201a7b6d814898607c4e0cd097c0f2ae46` which is from October 16th, 2020, and the version in `go.mod` is still `1.15`.

marcofranssen: Good point. No clue how I ended up on that commit. ffe36aa317dcbb421a536de071660261136174dd this is much better.

stefan-zh (approved):

marcofranssen:
@HuKeping @thaJeztah can you help out to get this fix merged?

HuKeping (approved):

PR #1635, opened 2022-04-08
Fix role parsing in golang 1.17+

kylewuolle:
In golang 1.17 [a change was made](https://go.dev/doc/go1.17#minor_library_changes) to remove anything preceding a slash from the filename field of the Content-Disposition header. This causes a problem with notary since it's using the filename field for the role value. This fix changes the parsing of the header value so it grabs the raw filename value instead of using the method that removes the base part of the filename.

This is also outlined in this issue:
https://github.com/notaryproject/notary/issues/1630

### DIFF
--- a/server/handlers/default.go
+++ b/server/handlers/default.go
@@ -6,6 +6,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"io"
+	"mime"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


kylewuolle:
Ping @thaJeztah @marcofranssen

thaJeztah (approved):
LGTM (but not a maintainer 😅)

thaJeztah:
ping @justincormack @HuKeping PTAL (also looks like CI is no longer running, including `codecov`, which was marked as "required"; probably need to rewrite CI to use GitHub actions, or re-enable CircleCI

jonnystoten (approved):

PR #1640, opened 2022-05-18
Fix static binary build

jonnystoten:
Static builds are broken on `master` because of a change introduced in #1636 which added a dependency on the sqlite library which requires cgo. We don't actually support sqlite outside of tests so this PR removes the dependency. I've also added a CI step to build the static binaries to make sure this doesn't happen again.

We discovered this at Docker because [we use the static binaries to build](https://github.com/docker/notary-official-images/blob/master/notary-server/Dockerfile#L15) the [Docker Official Image](https://hub.docker.com/_/notary) for Notary.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,5 @@
 version: 2
 jobs:
-
   job_01:
     machine:
       image: ubuntu-1604:201903-01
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -53,7 +52,7 @@ jobs:
           command: docker build --progress=plain -t notary_client .
       - run:
           name: "ci"
-          command: docker run --rm -e NOTARY_BUILDTAGS --env-file buildscripts/env.list --user notary notary_client bash -c "make ci && codecov"
+          command: docker run --rm -e NOTARY_BUILDTAGS --env-file buildscripts/env.list --user notary notary_client bash -c "make static && make ci && codecov"
```
justincormack, line 55: why do we want to build the static version for ci? If you want to test static build maybe add another target

jonnystoten: Yes it is just to test the build, I can do it in a separate step :+1:

justincormack:
FYI I don't think we should necessarily use static builds for the Official image, but this fix looks fine anyway and testing static builds is good.

PR #1613, opened 2021-09-29
Fix a concurrency bug in `TUFMetaStorage` by using sync.Map

stefan-zh:
My team at work encountered a concurrency issue in 2 places of the [`tuf_store.go::GetChecksum`](https://github.com/notaryproject/notary/blob/master/server/storage/tuf_store.go#L110) function.

This one is related to [`tuf_store.go:119`](https://github.com/notaryproject/notary/blob/master/server/storage/tuf_store.go#L119):
```
fatal error: concurrent map writes

goroutine 18240 [running]:
runtime.throw(0xc2a5e4, 0x15)
                /usr/local/go/src/runtime/panic.go:1114 +0x72 fp=0xc000161430 sp=0xc000161400 pc=0x436782
runtime.mapassign_faststr(0xb435c0, 0xc0000a24e0, 0xc000f39051, 0x40, 0x8)
                /usr/local/go/src/runtime/map_faststr.go:211 +0x3f7 fp=0xc000161498 sp=0xc000161430 pc=0x4153a7
github.com/theupdateframework/notary/server/storage.TUFMetaStorage.GetChecksum(0xd4b140, 0xc0000b4120, 0xc0000a24e0, 0xc000f38fc8, 0x74, 0xc000f39048, 0x8, 0xc000f39051, 0x40, 0x0, ...)
                /go/src/notary/server/storage/tuf_store.go:119 +0x1a6 fp=0xc000161538 sp=0xc000161498 pc=0x9dcb06
```
And this one is related to [`tuf_store.go:111`](https://github.com/notaryproject/notary/blob/master/server/storage/tuf_store.go#L111)
```
fatal error: concurrent map read and map write

goroutine 11789 [running]:
runtime.throw(0xc352e4, 0x21)
	/usr/local/go/src/runtime/panic.go:1114 +0x72 fp=0xc002d60f50 sp=0xc002d60f20 pc=0x436782
runtime.mapaccess2_faststr(0xb435c0, 0xc00007e5a0, 0xc0024dc8c0, 0x40, 0x3, 0xa)
	/usr/local/go/src/runtime/map_faststr.go:116 +0x47c fp=0xc002d60fc0 sp=0xc002d60f50 pc=0x414f9c
[...]

PR #1651, opened 2022-09-12
Remove unbounded cache

jonnystoten:
`TUFMetaStorage` uses an in-memory cache to store all retrieved metadata files. This cache has a few problems:

1. It doesn't have any sort of limit on its size so it will grow indefinitely as more content is retrieved
2. Entries in the cache are not deleted when `Delete` is calls on a `TUFMetaStorage` instance. This means the cache can hold entries that have been deleted from the underlying storage (there seem to be tests for this behavior so could this be for a reason?)
3. It isn't safe for concurrent use, see #1613.

I've taken the scorched-earth approach of simply removing this cache. After #1639 it should be fast to get these files from the database each time. I'll test this on Docker's production instance before merging it to check the performance is still OK.

### DIFF
--- a/server/storage/tuf_store.go
+++ b/server/storage/tuf_store.go
@@ -15,21 +15,13 @@ import (
 // by always starting from a current timestamp and then looking up other data by hash
 type TUFMetaStorage struct {
 	MetaStore
-	// cached metadata by checksum
-	cachedMeta map[string]*storedMeta
 }
 
 // NewTUFMetaStorage instantiates a TUFMetaStorage instance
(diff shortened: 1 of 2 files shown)


jonnystoten:
Looks good on Docker's notary server after a few days. The performance seems fine and more consistent - the average is a little bit slower (~3ms -> ~5ms) but the worst case is much better (>100ms -> ~8ms).

I'm assuming the tests that were testing that deleting from the TUF store doesn't delete from the cache were just wrong - anything relying on this will have other issues because of course if the server is restarted the cache will be cleared and if more than one server is running each one will have different cache contents etc.

This can also close #1613.

PR #1659, opened 2022-11-07
server: use docker/go-metrics utilities for prometheus, add "no_metrics" build-tag

thaJeztah:
The old code was no longer compatible with current versions of prometheus.
This switches the code to use docker/go-metrics, which is compatible with
current versions of prometheus, and already in use in other code in the
dependency tree.

I tried to keep the metrics the same as before, but there may be some
differences.

The first commit separates the metrics code from the server file; second commit introduces a `no_metrics` build-tag to allow building without metrics.

### DIFF
--- a/server/metrics.go
+++ b/server/metrics.go
@@ -8,23 +8,31 @@ import (
 
 	"github.com/docker/go-metrics"
 	"github.com/gorilla/mux"
-	"github.com/prometheus/client_golang/prometheus"
 )
 
 // namespacePrefix is the namespace prefix used for prometheus metrics.
 const namespacePrefix = "notary_server"
 
-func prometheusOpts(operation string) prometheus.SummaryOpts {
-	return prometheus.SummaryOpts{
-		Namespace:   namespacePrefix,
-		Subsystem:   "http",
-		ConstLabels: prometheus.Labels{"operation": operation},
-	}
-}
+// Server uses handlers.Changefeed for two separate routes. It's not allowed
+// to register twice ("duplicate metrics collector registration attempted"),
+// so checking if it's already instrumented, otherwise skip.
+var instrumented = map[string]struct{}{}
 
 // instrumentedHandler instruments a server handler for monitoring with prometheus.
 func instrumentedHandler(handlerName string, handler http.Handler) http.Handler {
-	return prometheus.InstrumentHandlerFuncWithOpts(prometheusOpts(handlerName), handler.ServeHTTP) //lint:ignore SA1019 TODO update prometheus API
+	if _, registered := instrumented[handlerName]; registered {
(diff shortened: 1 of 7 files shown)


thaJeztah:
@jonnystoten @justincormack PTAL - this should get us out of the dependency hell for prometheus, but ⚠️ I'm not a prometheus user; I tried to keep the metrics the same, but please double check if this looks good.

Once this is merged, we're able to update dependencies to current versions (and keep it compatible with code using this module, which very likely uses a more current version of prometheus).

I added a `no_metrics` build-tag which would allow for building without prometheus support, for situations where it's not needed.

server/metrics.go
```diff
@@ -0,0 +1,41 @@
...
+		// handler for this operation is already registered.
+		return handler
+	}
+	instrumented[handlerName] = struct{}{}
+
+	// Preserve the old situation, which used ConstLabels: "operation: <operation>"
+	// for metrics, but ConstLabels in go-metrics are per-namespace, and use
+	// ConstLabels: "handler: <handlerName>" (we pass operationName as handlerName).
+	namespace := metrics.NewNamespace(namespacePrefix, "http", metrics.Labels{"operation": handlerName})
+	httpMetrics := namespace.NewDefaultHttpMetrics(handlerName)
```
thaJeztah, line 33: New situation will have _two_ labels; "old" (`operation: <handlerName>`) and "new" (`handler: <handlerName>`)

server/metrics.go
```diff
@@ -0,0 +1,41 @@
...
+// to register twice ("duplicate metrics collector registration attempted"),
+// so checking if it's already instrumented, otherwise skip.
+var instrumented = map[string]struct{}{}
+
+// instrumentedHandler instruments a server handler for monitoring with prometheus.
+func instrumentedHandler(handlerName string, handler http.Handler) http.Handler {
+	if _, registered := instrumented[handlerName]; registered {
+		// handler for this operation is already registered.
+		return handler
+	}
```
thaJeztah, line 26: Had to add this to prevent the `handlers.Changefeed` being registered twice; it's used for two endpoints, bot using `Changefeed` as handlerName / operationName. Double check if it still works as expected with this.

thaJeztah:
I see there's an alternative PR already for the metrics in https://github.com/notaryproject/notary/pull/1570. /cc @brackendawson (not sure which approach works best, but more eyes welcome!)

brackendawson:
So both are a breaking change to the metric names. This PR would give notary the same metrics as distribution/distribution right? I could see that being a desirable feature. When I looked at this I tried to find the closest substitute for the discontinued prometheus middleware and went with the examples from the prometheus docs.

Hsaameett57 (approved):

PR #1692, opened 2023-09-14
fix: exceeded typo

testwill:


signer/rpc_and_client_test.go
```diff
@@ -34,7 +34,7 @@ func socketDialer(ctx context.Context, socketAddr string) (net.Conn, error) {
 		return net.DialTimeout("unix", socketAddr, time.Until(deadline))
 	}
 
-	return nil, fmt.Errorf("context deadline excedded")
+	return nil, fmt.Errorf("context deadline exceeded")
```
qba73, line 37: since we use a string literal here please consider using `errors.New()` func instead

